PR #936, opened 2026-04-03
Integrate Codex cloud provider

LuigiElleBalotta:
## Summary
- add a first-class Codex cloud provider to MAGDA AI settings
- implement browser OAuth login, token refresh, provider persistence, and Codex model normalization
- document Codex setup in the README and wire MAGDA to the updated `juce-llm` Codex transport

## Notes
- this PR does not change `.gitmodules`; local fork URLs were used only for development
- this PR depends on the matching `juce-llm` Codex transport PR being merged first or in parallel

## Testing
- built Windows release locally
- verified login flow, provider selection, and Codex request iteration against the backend

PR #948, opened 2026-04-07
Add official Codex App Server integration

LuigiElleBalotta:
﻿## Summary

This PR adds a new **Codex App Server** provider to MAGDA.

The goal is to support Codex in a way that is compatible with MAGDA's existing provider model **without relying on unofficial ChatGPT web backend endpoints** and without treating a ChatGPT subscription as a substitute for the OpenAI API.

This implementation is intentionally different from the previous Codex experiment discussed in #936. That earlier approach was based on an undocumented ChatGPT backend endpoint. This PR does **not** use that mechanism.

Instead, this PR integrates MAGDA with the **official OpenAI Codex App Server** interface documented by OpenAI.

Relevant prior reviewer concern:
- https://github.com/Conceptual-Machines/magda-core/pull/936#issuecomment-4183882387

## Why This Is Different From The Previous Approach

The previous Codex work explored calling an undocumented ChatGPT backend endpoint directly. I understand and agree with the concern raised there:

> MAGDA will only support officially supported APIs and access patterns.

This PR is designed specifically to satisfy that requirement.

What this PR **does not** do:
- does not call `chatgpt.com/backend-api/codex/responses`
- does not scrape bearer tokens from browser sessions
- does not ask the user to paste `id_token`, `access_token`, or refresh tokens into MAGDA
- does not treat ChatGPT subscription billing as if it were the OpenAI API
- does not add support for undocumented web backend behavior

What this PR **does** do:
- uses the **official Codex App Server** integration surface published on `developers.openai.com`
- delegates authentication to the local `codex` runtime instead of reimplementing browser/session/token plumbing inside MAGDA
- connects MAGDA to Codex through the documented App Server transport and protocol

In other words, this PR is proposing support for **an officially documented Codex integration pattern**, not support for an unofficial ChatGPT backend workaround.

## Documentation And References Used

This implementation was based on the official OpenAI Codex App Server documentation and the protocol surface exposed by the Codex CLI itself.

Primary documentation:
- https://developers.openai.com/codex/app-server

The App Server documentation describes Codex App Server as the interface intended for embedding Codex in applications. That is the reason this PR implements Codex as a dedicated MAGDA provider rather than extending the generic OpenAI Responses client.

In addition to the published docs, I validated the exact method names and transport expectations against the locally available Codex App Server tooling exposed by the installed `codex` CLI. In practice, this meant checking:
- `codex app-server --help`
- the generated App Server JSON schema from the CLI

That local schema and CLI help were used only to ensure the code matches the documented protocol surface actually exposed by the runtime, including:
- server startup via `codex app-server`
- transport over `ws://...`
- session initialization
- account/login flow
- thread creation
- turn execution
- streamed agent message deltas
- turn completion notifications

### 1. A separate provider instead of reusing the old Codex backend code

This PR introduces **Codex App Server** as its own provider path.

Reasoning:
- the App Server model is not just another OpenAI-compatible HTTP endpoint
- it is a stateful protocol with its own transport and lifecycle
- it has different concepts: initialization, account state, threads, turns, streamed events

Trying to force that into the existing `OpenAIResponsesClient` abstraction would have made the implementation harder to reason about and more fragile. A dedicated provider keeps the supported paths clearly separated:
- OpenAI API remains OpenAI API
- local inference remains local inference
- Codex App Server is integrated through its own official interface

### 2. MAGDA delegates login to the Codex runtime

This PR does not implement custom browser auth logic or token management in MAGDA.

Reasoning:
- the official App Server flow already includes account/login methods
- re-implementing token/session handling in MAGDA would create unnecessary coupling to auth internals
- keeping auth inside the Codex runtime is cleaner, less invasive, and closer to the supported model

### 3. WebSocket transport instead of shelling text through a pseudo-terminal

The integration talks to the local App Server over WebSocket.

Reasoning:
- this follows the documented App Server model
- it is more structured and portable than trying to drive an interactive CLI process as if it were a terminal assistant
- it maps naturally onto MAGDA's request/response lifecycle for agent messaging

### 4. MAGDA-specific integration instead of changes in `juce-llm`

This PR keeps the App Server implementation in MAGDA rather than pushing it into `juce-llm`.

Reasoning:
- this is not an OpenAI-compatible completion transport
- it is an application-level agent/session protocol
- keeping it in MAGDA avoids mixing a stateful Codex App Server client into a lower-level generic LLM transport library

### New provider

Adds a new cloud provider option:
- `Codex App Server`

### New MAGDA-side client/manager

Introduces a Codex App Server integration layer that:
- starts or connects to a local `codex app-server`
- initializes an App Server session
- triggers the documented login flow through the Codex runtime
- starts a thread
- sends user prompts as turns
- receives streamed assistant text from App Server notifications
- returns the final text back through MAGDA's existing agent interfaces

### AI Settings UI support

Adds UI support in `AI Settings -> Cloud` for:
- selecting `Codex App Server`
- logging in through the runtime-managed flow
- enabling/testing the provider

### Config wiring

Adds preset/config plumbing so that once enabled, the provider can be selected and used by the command/chat agent path instead of falling back to API-key-only OpenAI settings.

### Documentation

Updates the README with setup information for the new provider, including the requirement that the local `codex` CLI is installed and available on `PATH`.

## Implementation Notes

The provider uses the local Codex App Server runtime and defaults to:
- WebSocket URL: `ws://127.0.0.1:8765/`
- Model: `gpt-5-codex`

One small but important compatibility detail: the WebSocket URL includes a trailing `/`. During implementation I found that the client library used here expects a path component in the WebSocket URL; `ws://127.0.0.1:8765/` works reliably, while `ws://127.0.0.1:8765` did not.

## Cross-Platform Considerations

This was implemented with MAGDA's cross-platform structure in mind.

The integration avoids platform-specific browser token hacks and instead relies on:
- the local `codex` executable
- the documented App Server protocol
- MAGDA-side transport/session handling

That keeps the design much closer to something portable across the platforms MAGDA already targets.

## What This PR Intentionally Does Not Attempt

This PR does not try to expose every possible App Server capability.

It intentionally focuses on the minimum useful MAGDA integration:
- provider registration
- login
- prompt/response turn handling
- streamed agent text

It does not attempt to add:
- full approval/tooling UX beyond what MAGDA currently needs
- custom token persistence in MAGDA
- direct support for undocumented Codex backend HTTP calls
- replacement of the official OpenAI API provider

## Validation

Local validation performed for this branch:
- `git diff --check`
- `cmake -S . -B build-app-server`
- `cmake --build build-app-server --config Release --target magda_agents`
- `cmake --build build-app-server --config Release --target magda_daw`
- `cmake --build build-app-server --config Release --target magda_daw_app`

I also verified the end-to-end MAGDA flow locally:
- select `Codex App Server` in AI Settings
- login successfully via the runtime-managed flow
- enable the provider
- use MAGDA chat/command flow against the App Server-backed provider

## Files Of Interest

Key files in this PR:
- `magda/agents/codex_app_server_manager.hpp`
- `magda/agents/codex_app_server_manager.cpp`
- `magda/agents/codex_app_server_client.hpp`
- `magda/agents/codex_app_server_client.cpp`
- `magda/agents/llm_presets.hpp`
- `magda/agents/llm_presets.cpp`
- `magda/agents/llm_config_utils.hpp`
- `magda/agents/llm_client_factory.hpp`
- `magda/agents/daw_agent.cpp`
- `magda/daw/ui/dialogs/AISettingsDialog.cpp`
- `magda/agents/CMakeLists.txt`
- `README.md`

## Reviewer Reassurance

The most important point for review is this:

This PR is **not** asking MAGDA to support ChatGPT subscription access through unofficial or undocumented backend APIs.

It is proposing support for the **officially documented Codex App Server integration path** instead.

If the maintainers' policy is that MAGDA should only support officially supported APIs and official access patterns, then this PR is intended to align with that policy rather than work around it.


lucaromagnoli:
Closing this for the same reasons #936 was closed, plus a process concern and a final warning.

## Process

After #936 I added contributing guidelines that ask contributors to open an issue or discussion before submitting large feature PRs — especially new providers. This PR is 926 lines added with no prior discussion. That is the exact pattern those guidelines exist to prevent, and this is now the second time in a row. Opening unsolicited large PRs after the same topic was already declined and after explicit guidelines were published is not how this project accepts contributions.

## Policy

MAGDA's position is that cloud LLM support is limited to the official OpenAI API (and equivalents from other providers) accessed via API keys with API billing. Codex App Server, regardless of how it is documented on `developers.openai.com`, is an integration path designed around ChatGPT subscription seats. That is exactly what was declined in #936 and the decision has not changed. The technical framing in this PR ("official App Server", "WebSocket transport", "delegates auth to the runtime") does not change the underlying access model, which is what the policy is about.

To be explicit about why App Server specifically does not change the analysis: App Server is the embedding interface for the **Codex agent** — it exists so applications that want to embed Codex's coding agent (with its tools, sandbox execution, approval flow, etc.) can drive it programmatically. MAGDA does not embed the Codex agent. MAGDA has its own agent and only needs LLM text generation. This integration consumes `agent_message_delta` events purely as a text channel and discards everything App Server is actually for. Functionally that makes it a chat completion call billed against a ChatGPT subscription seat, which is the exact pattern the policy excludes. If MAGDA were proposing to embed Codex-the-agent inside MAGDA as a coding assistant, App Server would be the right interface for that. That is not what is happening here.

## Final warning

I appreciate the effort that went into this, but it is not going to be merged, and I am not going to keep having this discussion. Any further PR that:

- proposes ChatGPT subscription access as a substitute for the OpenAI API, by any technical route, or
- is opened as a large unsolicited PR without first agreeing on scope through an issue or discussion,

will result in a ban from this repository. This is not personal — I want to focus on shipping a quality product, not relitigating the same policy decision. Please raise an issue first if you would like to propose something genuinely different.

PR #981, opened 2026-04-11
Fix #970: route MIDI through instrument rack wrappers and chain ends

### LINKED ISSUE #970: FX Chain. Blocking of MIDI signal output when passing through third-party plugins.
### What happened?

Probably the biggest pain among the found bugs :)

The Tracktion engine allows placing different types of plugins on one track: midi 1 -> (midi 2 -> midi 3 -> midi (n)...) -> instrument -> FX, which simplifies, speeds up, and optimizes the workflow. 
Currently, 1: when installing a third-party MIDI plugin, the signal enters the MIDI plugin, but at the output of the MIDI plugin, the signal is blocked and does not reach the instrument; 2: when disabling the plugin, the signal also remains blocked. 
Only removing the plugin from the FX Chain allows the signal to pass to the instrument. This signal behavior is also confirmed in the Rack.

### Steps to reproduce

1 Open Magda. Create a track with a non-empty MIDI clip.
2 Place any MIDI plugin (seq, arp, MIDI generator, etc.) on the track, place an instrument (synth, etc.) after it.
3 Press Play. The track indicator shows (flashes) that the signal is present.
4 BUG01: The instrument does not play notes.
5 Disable the plugin by clicking the on/off button.
6 BUG02: The instrument does not play notes.
7 Delete the MIDI plugin. Sound plays.

### Expected behavior

This time without video, hope I described it correctly. If not, I'll record a video.

### MAGDA version

0.4.3

### Operating system

Windows

### Crash log (if applicable)

```text

```

lucaromagnoli:
InstrumentRackManager dropped MIDI at the rack output pin (only audio pins 1/2 were wired), so any plugin chain ending in a wrapped instrument silently lost its MIDI stream. RackSyncManager had the same omission at the last-plugin->rack-output and empty-rack passthrough boundaries.

InstrumentRackManager now adds one MIDI edge at the rack output, picked by audio output channel count: zero outputs (Cthulhu, chord engines, arps) forward the plugin's transformed MIDI; non-zero outputs (Serum, Surge, 4osc, samplers) pass the raw clip MIDI around the synth so downstream sidechain effects and SerumFX-style LFOs still see note events. Using output count instead of producesMidi() avoids synths that lie about emitting MIDI.

RackSyncManager wires pin 0 unconditionally at the chain-end and empty-rack boundaries so a MIDI plugin at the end of a chain still feeds downstream when the chain is bypassed and an empty rack is fully transparent.

### DIFF
--- a/magda/daw/audio/InstrumentRackManager.cpp
+++ b/magda/daw/audio/InstrumentRackManager.cpp
@@ -43,6 +43,32 @@ te::Plugin::Ptr InstrumentRackManager::wrapInstrument(te::Plugin::Ptr instrument
     // MIDI: rack input pin 0 --> synth pin 0
     rackType->addConnection(rackIOId, 0, synthId, 0);
 
+    // MIDI at the rack output: depends on what kind of plugin we wrap.
+    //
+    // - Zero audio outputs => MIDI-only plugin (Cthulhu, chord engines, arps).
+    //   Forward the plugin's transformed MIDI to the rack output so downstream
+    //   sees its events.
+    //
+    // - Has audio outputs => synth (Serum, Surge, 4osc, samplers...). Pass the
+    //   raw clip MIDI around the synth so downstream audio effects that sync
+    //   from note events (SerumFX LFO, sidechain envelopes) still see the
+    //   clip's MIDI after the instrument. We deliberately ignore the plugin's
+    //   own MIDI output here even if it claims producesMidi() — synths often
+    //   lie about that and emit nothing, which would silently kill MIDI flow.
+    //
+    // Exactly one of these edges is added so MIDI is never duplicated.
+    const bool isMidiOnly = [&] {
+        if (auto* ext = dynamic_cast<te::ExternalPlugin*>(instrument.get()))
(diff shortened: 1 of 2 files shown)


lucaromagnoli:
Thanks @nowgad for the detailed reproduction steps on #970 — having both the FX Chain and Rack scenarios spelled out made it easy to confirm both came from the same wiring bug.

nowgad:
> Thanks @nowgad for the detailed reproduction steps on #970 — having both the FX Chain and Rack scenarios spelled out made it easy to confirm both came from the same wiring bug.
@lucaromagnoli  The issue is partially fixed. At the moment, MIDI from the clip is sent to the instrument bypassing the MIDI plugins :) Arpeggiators receive MIDI at the input, but there is nothing at the output. Sequencers also have empty outputs.
Tested with Bluearp, Eon-arp, HY-MPS3, TATAT, Toneworks Catalyst, Ripchord.

https://github.com/user-attachments/assets/67e67867-a728-4d8a-9ee9-1fecf86575ab

Magda v. 0.4.4, Win11

PR #1196, opened 2026-05-06
Fix/1194 midi signal routing

lucaromagnoli:


lucaromagnoli (commented):
## PR Review — Fix/1194 MIDI Signal Routing

**+580 / -133** across 18 files, 4 commits.

### What it does

Three related fixes bundled into one PR:

1. **MIDI consumption by instrument racks**: Previously, wrapper racks passed MIDI through (input passthrough + plugin MIDI out), meaning a second synth downstream would double-trigger. Now MIDI is consumed by the instrument — only audio passes through the rack. This is the core routing fix.

2. **InstrumentMeterTapPlugin**: New internal plugin placed on the synth-output branch inside the rack, so device meters show synth-generated audio only (excluding the upstream audio passthrough). The tap also applies per-device gain, replacing the need for TE's `LevelMeasuringNode` hook for wrapped instruments.

3. **StepSequencer midiThru default**: Changed from `true` to `false`, so step sequencers replace incoming MIDI by default instead of passing it through. Also fixes `restorePluginStateFromValueTree` which was missing `midiThru` from the `copyPropertiesToCachedValues` call.

Plus a cleanup pass removing ~100 lines of DBG/Logger noise.

### Code quality

Well-structured. The tap plugin is clean — straightforward `applyToBuffer`, correct CAS loop for `storeMax`, proper acquire/release pairs on the pointer atomics. The `retiredTapStorage_` vector to prevent dangling pointers during `bindRealtimeTap` rebinding is a nice detail.

Tests are thorough — rack wiring assertions, gain application, meter publication, survival after `metering.clear()`, and the step sequencer default/restore round-trip.

### Issues and suggestions

**1. `PluginManager::getDeviceIdForPlugin` lost the rack fallback (medium)**

```cpp
// Before:
return instrumentRackManager_.getDeviceIdForRack(plugin);
// After:
return it != pluginToDevice_.end() ? it->second : INVALID_DEVICE_ID;
```

This removes the fallback that allowed `getDeviceIdForPlugin` to resolve rack instance plugins to device IDs. If anything else calls `getDeviceIdForPlugin` with a rack instance (e.g. the existing `LevelMeasuringNode` path for non-instrument racks, or future callers), it will silently return `INVALID_DEVICE_ID`. Was this intentional because the meter tap now handles it, or is there a caller that still relies on this path?

**2. `pendingMeterTapsByRack_` keyed on raw `te::Plugin*` (low)**

The map stores a raw pointer as key. Between `wrapInstrument()` and `recordWrapping()`, if the rack instance were somehow destroyed, the pointer would dangle as a map key. In practice this window is tiny and controlled by the same caller, but using the `te::Plugin::Ptr` (or `EditItemID`) as the key would be strictly safer. Not a real bug given the current call flow, but worth noting.

**3. `getOrCreateMeasurer` now conditionally re-registers the client (minor)**

```cpp
if (!it->second->clientRegistered) {
    it->second->measurer.addClient(it->second->client);
    it->second->clientRegistered = true;
}
```

This re-registers a client that was previously unregistered. Is there a path where the measurer could have been cleared but the entry still exists with `clientRegistered = false`? If so, does `addClient` on a measurer that was previously cleared work correctly? Looks fine based on TE's implementation, just flagging.

**4. StepSequencer default change is a behavior change for existing projects**

`midiThru` defaults to `false` now. Existing serialized projects that have a StepSequencer without an explicit `seqMidiThru` property will silently switch from thru to replace on next load. The `restorePluginStateFromValueTree` fix ensures that projects which *do* have the property saved will restore correctly, but projects saved before this property existed will get the new default. Worth confirming this is the desired behavior for existing sessions.

### Verdict

The core MIDI routing fix is correct and well-tested. The meter tap is a solid architectural addition — avoids the hacky `getDeviceIdForRack` lookup during graph building and instead makes metering explicit. Item #1 (the removed fallback) is the only thing I'd want a clear answer on before merging.

PR #1321, opened 2026-05-22
feat: CLAP zero-shot tagging in the media DB indexer (#1319)

### LINKED ISSUE #1319: Media DB indexer never runs CLAP zero-shot tagging
The Python prototype's zero-shot tagging step (`prototypes/media_db/src/media_db/tags.py` + `derive.py`) was never ported to the C++ indexer. Tags currently written to `media_tag` come exclusively from path/filename tokenization. The CLAP audio encoder runs, but its output only ever reaches `media_embedding` for similarity search. The CLAP text encoder runs only at query time, to embed the user's search string.

### Evidence

- `MediaDbIndexer.cpp:638-657` is the only tag-writing path in the audio pipeline. It builds `tags` from `pathTags(f.path)` + `scanTagsFor(...)` (both filename tokenizers) and writes them with `replaceTags(sqlDb, fileId, tags, "path")`. No CLAP involvement.
- `MediaDbIndexer.cpp:176-180` has an explicit acknowledging comment:
  > Mirrors derive.py in the Python prototype. The TAG_FAMILY map there converts CLAP zero-shot tag scores into a family; in C++ we currently don't run zero-shot tagging during indexing, so family comes entirely from pathFamilyHint.
- The only `ClapTextEncoder::embedTokens` call site in the whole tree is `MediaDbQuery.cpp:599` (query side).
- `deriveFamily(path)` in `MediaDbIndexer.cpp:197` only takes a path. Its prototype counterpart `family(top_tags, path)` in `derive.py:232` takes a CLAP top-tag list. The C++ port silently dropped the CLAP input.

### Impact

- Samples with uninformative filenames ("Sample 01.wav", "Untitled.wav", numeric IDs) get no semantic tags at all.
- The `media_fts` tag-text index is starved. Search recall depends on either exact filename tokens or the embedding cosine path.
- `media_file.family` is path-only, so anything outside a recognised folder hint comes back as "unknown" even when CLAP would have classified it confidently.

### Outline of fix

1. Add a curated prompt list to the indexer, mirroring `tags.py` as the starting taxonomy (around 40 prompts covering drums, bass/lead, acoustic, FX, texture descriptors). Map each to a coarse family for the family-derivation pass.
2. At indexer init, embed every prompt once via `ClapTextEncoder` + `RobertaTokenizer` (both already lazy-loaded by `MediaDbContext`), cache as a `[K, 512]` row-major matrix. Roughly 80 KB of RAM, one-time cost.
3. In the per-file pipeline, after `encoder_->embed(...)` produces the audio embedding, dot-product against the prompt matrix (L2-normalized on both sides so it's straight cosine), take the top-K above a threshold (the prototype uses around 0.25-0.3).
4. Insert those as a second `replaceTags(sqlDb, fileId, clapTags, "clap-zeroshot")` call with `confidence = cosine_score`. Schema already supports it.
5. Extend `deriveFamily` to consume the CLAP top-tag list as a fallback after the path hint, matching `derive.py:232` precedence (path > CLAP > "unknown"). Skip prompts whose family is "texture" so descriptors like "warm sound" don't override real instrument tags.

### Out of scope

- Few-shot anchor tagging (using hand-labelled reference samples as audio-domain centroids). Worth a separate issue once zero-shot is in and the recall gaps are visible.
- Threshold tuning per prompt. Start with a single global threshold from the prototype; per-tag calibration is a follow-up.

lucaromagnoli:
## Summary

- The C++ indexer wrote tags only from path/filename tokens; samples with uninformative names got no semantic tags at all, and `media_file.family` defaulted to `unknown` outside recognised folder structures. The CLAP audio encoder ran for similarity search but was never projected against text prompts during indexing.
- Adds a curated 40-prompt taxonomy mirroring `prototypes/media_db/src/media_db/tags.py`, embeds the prompts once on first use via the existing `ClapTextEncoder` + `RobertaTokenizer`, and scores each audio embedding against the cached `[K, 512]` prompt matrix during the post-scan embedding pass. Top hits above 0.28 are written to `media_tag` with `source_model='clap-zeroshot'`; `media_fts` is refreshed in the same transaction so tag-text search picks them up.
- `media_file.family` is overridden from the top non-texture CLAP tag only when the path-derived family was `unknown` — preserves the existing precedence rule (path > CLAP > unknown). Texture descriptors (`"a warm sound"` etc.) can still emit a tag but never set the family.
- Tagger is lazily initialised; when the text encoder / tokenizer aren't loaded the indexer keeps producing audio embeddings and skips tagging without per-file noise. Init failures are reported once via the existing failure callback.

Closes #1319.

## Test plan

- [x] `make test-build` — debug test binary links cleanly
- [x] `./tests/magda_tests "[media_db][zero-shot]"` — 3 hermetic cases pass, 2 model-gated cases skip with `MAGDA_MEDIA_DB_CLAP_TEXT_MODEL` / `MAGDA_MEDIA_DB_TOKENIZER_JSON` instructions
- [x] `./tests/magda_tests "[media_db]"` — 74 cases pass, no regressions in existing scan / indexer / query suites
- [x] `make debug` — main app links cleanly with the new indexer DI
- [ ] **Needs a manual rescan with the Sample Tagger models installed** to confirm `clap-zeroshot` rows appear in `media_tag` and that tag-text search surfaces semantically-named samples that filename-only tagging missed
- [ ] **Spot-check a folder of uninformatively-named samples** (`Sample01.wav`, numeric IDs) — should now carry CLAP tags and a non-`unknown` family

## Notes

- Branch is based on `dev/0.9.0`, not `main` — the media DB code only exists there.
- Thresholds (`kZeroShotTagThreshold = 0.28F`, `kZeroShotFamilyFloor = 0.20F`) are `constexpr` in `MediaDbZeroShotTags.hpp` so per-tag calibration is a follow-up without ABI risk.
- Out of scope (matches the issue): few-shot anchor tagging, per-prompt threshold calibration.

### DIFF
--- a/magda/daw/media_db/CMakeLists.txt
+++ b/magda/daw/media_db/CMakeLists.txt
@@ -10,6 +10,7 @@ target_sources(magda_daw PRIVATE
     MediaDbIndexer.cpp
     MediaDbMetadata.cpp
     MediaDbQuery.cpp
+    MediaDbZeroShotTags.cpp
     MelSpectrogram.cpp
     PathRules.cpp
     RobertaTokenizer.cpp
--- a/magda/daw/media_db/MediaDbIndexer.cpp
+++ b/magda/daw/media_db/MediaDbIndexer.cpp
@@ -656,7 +656,7 @@ void applyZeroShotTagging(sqlite3* db, std::int64_t fileId,
         // ambiguous samples (a 0.5s vocal hit sounds drum-like to CLAP, but
         // a producer who put it in /Vocals/ knew what they made).
         if (fileRow->family.empty() || fileRow->family == "unknown") {
-            const auto clapFamily = familyFromTopTags(hits);
+            const auto clapFamily = familyFromTopLabels(hits);
             if (!clapFamily.empty()) {
                 updateFamily(db, fileId, clapFamily);
             }
@@ -758,6 +758,13 @@ void processOneFile(sqlite3* sqlDb, const ScannedFile& f, MediaDbIndexer::Stats&
         replaceTags(sqlDb, fileId, tags, "path");
         if (existing && f.kind == "audio") {
             deleteEmbeddings(sqlDb, fileId);
+            // Stale CLAP zero-shot tags described the previous audio
+            // content; clear them in the same transaction so the file can't
+            // be left with semantic tags that no longer describe the bytes
+            // on disk. The next embedding pass re-derives them when a
+            // tagger is available; when it isn't, the file simply has no
+            // CLAP tags rather than misleading ones (issue #1319).
+            replaceTags(sqlDb, fileId, {}, "clap-zeroshot");
         }
 
         upsertFts(sqlDb, fileId, buildPathText(f.path), buildTagText(tags));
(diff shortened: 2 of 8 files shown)


lucaromagnoli:
Review comment 1/3

In `magda/daw/media_db/MediaDbZeroShotTags.cpp`, the prompt list is used both as the CLAP text prompts and as the strings returned by `scoreEmbedding()`, which means `applyZeroShotTagging()` writes values like `"the sound of a kick drum"` directly into `media_tag`. The prototype keeps terse labels (`"a kick drum"`) and only applies `PROMPT_TEMPLATE` for embedding. Storing the full prompt will pollute displayed tags and FTS with repeated `the sound of` terms. Can we split stored labels from prompt strings, embed the prompts, and return/write the labels?

lucaromagnoli:
Review comment 2/3

In `magda/daw/media_db/MediaDbIndexer.cpp`, when changed audio invalidates its embedding, the old `clap-zeroshot` tags are left behind unless the later embedding pass has a live tagger and replaces them. If the text model/tokenizer is missing or tagger init fails, the new embedding can be written while stale semantic tags from the previous audio content remain in `media_tag` and FTS. We should clear `source_model = clap-zeroshot` whenever audio embeddings are invalidated, or explicitly clear them in the embedding pass when zero-shot tagging is unavailable.

lucaromagnoli:
Review comment 3/3

In `magda/daw/ui/panels/content/MediaDbBrowserContent.cpp`, `ctx.textEncoder()` / `ctx.tokenizer()` lazy-load the large text model during Analyze, even though the constructor parameters are described as nullable and there may be no rows that need embedding. The scan path does the same. Consider using non-loading accessors or deferring this load until after `pendingEmbeddings` / selected audio rows are known, so ordinary indexing does not pay the text-model load cost just because the bundle is installed.

PR #1348, opened 2026-05-29
Fix arrange zoom paint crash

### LINKED ISSUE #1345: Segmentation fault when zooming in completely on waveform in Arrange view
**MAGDA version:** 0.9.1 (Linux x86_64 AppImage)
**OS:** Arch Linux (kernel 7.0.3-arch1-2)
**CPU:** Intel Core i5-4258U (Haswell, AVX2/FMA supported)
**RAM:** 3.7 GiB

## Description

MAGDA crashes with a segmentation fault when zooming into a waveform completely in the Arrange view. The crash is 100% reproducible by scrolling the horizontal zoom to its maximum level on an audio clip.

## Steps to reproduce

1. Open any project with an audio track
2. In Arrange view, zoom in horizontally on a waveform (scroll up or use zoom control)
3. Continue zooming until fully zoomed in
4. App crashes

## Terminal output at time of crash

```
sched_reserve: fused Gated Delta Net (chunked) enabled
sched_reserve:        CPU compute buffer size =   311.00 MiB
sched_reserve: graph nodes  = 959
sched_reserve: graph splits = 1
sched_reserve: reserve took 10.57 ms, sched copies = 1
Segmentation fault (core dumped) ./MAGDA-0.9.1-Linux-x86_64.AppImage
```

## Notes

The AI model loads and the llama context initialises successfully. The crash happens at maximum zoom, suggesting the waveform renderer may be encountering a divide-by-zero or out-of-bounds access when `samplesPerPixel` approaches zero (or the zoom ratio overflows). The project file's saved `horizontalZoom` value remains at a normal level after the crash, so the bug is only triggered interactively.

lucaromagnoli:


magda/daw/ui/components/clips/ClipComponent.cpp
```diff
@@ -56,6 +56,77 @@ void showExternalEditorFailedAlert(const juce::String& message) {
 constexpr int MIDI_PREVIEW_MIN_NOTE = 21;   // A0
 constexpr int MIDI_PREVIEW_MAX_NOTE = 108;  // C8
 
+juce::Path makeClippedRoundedRectPath(juce::Rectangle<int> bounds,
```
lucaromagnoli, line 59: Nit (non-blocking): JUCE's `Path::addRoundedRectangle` already has a per-corner-curve overload that does exactly this. Because `visibleBounds.getX() == bounds.getX()` whenever the left edge is visible (and symmetric on the right), the whole helper collapses to ~10 lines:

```cpp
juce::Path makeClippedRoundedRectPath(juce::Rectangle<int> bounds,
                                      juce::Rectangle<int> visibleBounds, float radius) {
    juce::Path path;
    if (bounds.isEmpty() || visibleBounds.isEmpty()) return path;

    const bool leftVisible  = visibleBounds.getX()     <= bounds.getX();
    const bool rightVisible = visibleBounds.getRight() >= bounds.getRight();
    const auto vb = visibleBounds.toFloat();
    const float r = juce::jlimit(0.0f,
                                 juce::jmin(vb.getWidth(), vb.getHeight()) * 0.5f,
                                 radius);

    path.addRoundedRectangle(vb.getX(), vb.getY(), vb.getWidth(), vb.getHeight(),
                             r, r,
                             leftVisible, rightVisible, leftVisible, rightVisible);
    return path;
}
```

Same behavior, less custom path code to maintain. Not a blocker — the explicit version is fine too.

PR #1353, opened 2026-05-29
feat: add Ollama (and OpenAI-compat) local-server provider

sPROFFEs:
## Summary

Adds **Ollama** as a first-class LLM provider, routed through the existing `juce-llm` OpenAI-Chat wire format. Same code path also works for OpenAI-compatible local servers like **GPUStack** (optional bearer token is forwarded). The DAW gains a "Local (Ollama)" preset alongside the existing cloud/embedded presets — no new dependencies, no changes to `juce-llm`.

UX summary, from the AI Settings dialog:

1. **Cloud tab → Provider: Ollama (Local)** — the *API Key* row becomes *Base URL* (default `http://localhost:11434/v1`, password-mask off). A second optional *API Key* row appears for servers like GPUStack that require auth (vanilla Ollama ignores it). A *Model* row appears with a combo + **Refresh** button that probes `GET /v1/models` and lists pulled models.
2. **Test** does the same `/v1/models` probe — no model required, validates URL + key together. Reports HTTP status / connection failure with a human-readable detail.
3. **Add** (or just **OK**) persists URL, key and picked model. Ollama remains selectable in the combo so the model/URL/key can be edited later; selecting it re-prefills the fields.
4. **Config tab → Mode: Cloud → Provider: Ollama (Local)** writes the `LOCAL_OLLAMA` preset, overriding each agent role's model with the user's pick. The chat status bar reflects the real model in use.

At request time, `toLLMProviderConfig`:
- Resolves the base URL from per-agent → credential → default.
- Normalises the URL so both `http://host:port` and `http://host:port/v1` route to `.../v1/chat/completions`.
- Injects the user-supplied bearer token, or a harmless placeholder if blank (Ollama ignores it).
- Substitutes `Config::ollamaModel` when set, regardless of what the preset stored.

## Files changed

| File | What |
|---|---|
| `magda/agents/llm_presets.hpp` | `provider::OLLAMA`, `preset::LOCAL_OLLAMA`, default model IDs, `DEFAULT_OLLAMA_BASE_URL` |
| `magda/agents/llm_presets.cpp` | `LOCAL_OLLAMA` preset (Llama 3.1 for router/command/controller, Qwen 2.5 for music) |
| `magda/agents/llm_config_utils.hpp` | Map `ollama` → `OpenAIChat`; base-URL override from credential slot; bearer-token injection; URL normalisation; model override |
| `magda/daw/core/Config.{hpp,cpp}` | `ollamaModel` and `ollamaApiKey` fields with get/set + JSON save/load |
| `magda/daw/ui/dialogs/AISettingsDialog.cpp` | Ollama entry in `getKnownProviders`, dynamic field labels, model picker with `/v1/models` autodiscover, persistence on OK, re-editable after Add |
| `magda/daw/CMakeLists.txt` | Registers the new icon |
| `assets/icons/cloud_providers/ollama.svg` | New |

8 files changed, 354 insertions(+), 11 deletions(-). No changes to `tracktion_engine`, `juce-llm`, or any submodule.

## Design notes / trade-offs

- **Why Cloud tab, not Local tab.** The existing Local tab is wired specifically to the in-process `llama.cpp` (`LlamaModelManager` / GGUF file picker). Ollama is HTTP and architecturally identical to other cloud providers — just a different base URL. Routing it through the Cloud-tab credentials UI was the minimal-surface choice. The Local tab is untouched.
- **Credential slot reused as base URL.** Per-provider credentials are a `map<string, string>`. Reusing the Ollama slot for the URL avoided introducing a new schema field while keeping the same "registered providers" list semantics. The optional API key lives in a separate `ollamaApiKey` field.
- **CFG / structured output.** CFG-constrained output stays gated to OpenAI Responses + GPT-5 (`supportsOpenAICFG` unchanged). Other providers — including Ollama — use free-form output with the grammar shown as instruction text. Smaller open-weight models will produce more DSL parse errors; the interpreter surfaces them as chat errors rather than executing badly.

## Test plan

- [x] Builds clean on Linux (Debian 13, GCC 14, Ninja, Release).
- [x] **Test** button reports "N models available" against a running Ollama; reports HTTP 401 against a GPUStack server when the key is blank; reports HTTP 401 detail when the key is wrong.
- [x] **Refresh** populates the Model combo from `/v1/models`.
- [x] Picking a model + OK persists it; reopening Settings re-shows it; chat status bar reflects it.
- [x] Editing the model after Add and clicking OK saves the new pick (no need to re-Add).
- [x] Chat sends to `http://host:port/v1/chat/completions` regardless of whether the user stored `host:port` or `host:port/v1`.
- [ ] macOS / Windows builds — not validated locally; should be free of platform-specific code.
- [ ] Tested against GPUStack — wire format is identical; reporter has confirmed `/v1/models` probe and the Bearer header path.


lucaromagnoli:
Thanks for putting this together. I appreciate the work that went into it.

Before I go too far with the review, could you please confirm the Contributor License Agreement?

https://github.com/Conceptual-Machines/magda-core/blob/main/CLA.md

A reply here with the following is enough for now:

> I have read and agree to the MAGDA Contributor License Agreement.

One small process note: for future feature work, please open an issue first so we can discuss scope before you spend time implementing. Provider integrations can affect the roadmap and ongoing maintenance, so it helps to align upfront.

I’ll take a closer look at the implementation and leave technical feedback separately.

sPROFFEs:
Hi! Nice to read you. Yes I check the CLA and "I have read and agree to the MAGDA Contributor License Agreement"

About implementation... it was only tested under Linux amdx64 and worked almost perfect but the backend model does really matter. Maybe this week I could test it under w11 and MacOS. 

Also tested remote endpoint was a self-hosted ollama (no api key needed) server running qwen3.6 and qwen2.5-coder. 
Also tested with gemma4 and llama3 but non coding oriented models does not work very well. 

They're some issues as the models do not really fully interact the samples database, etc.. I do not test it with paid LLMs such as GPT/Opus but with local models it did not work. 

I would still testing it anyway. This is such a nice project. Keep on it !

lucaromagnoli:
> Hi! Nice to read you. Yes I check the CLA and "I have read and agree to the MAGDA Contributor License Agreement"
> 
> About implementation... it was only tested under Linux amdx64 and worked almost perfect but the backend model does really matter. Maybe this week I could test it under w11 and MacOS.
> 
> Also tested remote endpoint was a self-hosted ollama (no api key needed) server running qwen3.6 and qwen2.5-coder. Also tested with gemma4 and llama3 but non coding oriented models does not work very well.
> 
> They're some issues as the models do not really fully interact the samples database, etc.. I do not test it with paid LLMs such as GPT/Opus but with local models it did not work.
> 
> I would still testing it anyway. This is such a nice project. Keep on it !

Hey, thanks for this. I’ll get back to you properly next week; I’m quite slammed this week.
When you mention the models not fully interacting with the sample database, can you clarify what you mean? The media/sample DB does not go through the LLM agent flow; it uses a separate ML model running through ONNX for embeddings/search. So I want to make sure I understand what behavior you expected there.

Thanks again for testing this on Linux and for the model notes. Windows/macOS testing would be useful if and when you get a chance.

magda/daw/ui/dialogs/AISettingsDialog.cpp
```diff
@@ -66,10 +66,20 @@ const std::vector<ProviderInfo>& getKnownProviders() {
          magda::model::DEEPSEEK_CHAT, BinaryData::deepseek_svg, BinaryData::deepseek_svgSize},
         {magda::provider::OPENROUTER, "OpenRouter", magda::provider::OPENROUTER, "",
          magda::model::LLAMA_70B, BinaryData::openrouter_svg, BinaryData::openrouter_svgSize},
+        // Ollama: a local HTTP server (or remote). The "credential" stores the
+        // base URL, not an API key. testBaseUrl falls back to DEFAULT_OLLAMA_BASE_URL
+        // when the field is left blank.
+        {magda::provider::OLLAMA, "Ollama (Local)", magda::provider::OLLAMA,
```
lucaromagnoli, line 72: The main thing for me: Ollama is a local provider, so surfacing it under Cloud here (and routing the `local_ollama` preset into Cloud mode) is the wrong mental model for users. I understand the Cloud tab was the pragmatic choice since Ollama is HTTP while the Local tab is currently GGUF / in-process only. But a local Ollama server belongs on the local side. I'd rather it live under the Local section, even if that means the Local tab grows a "local HTTP server" option next to the in-process llama path, than ship Ollama presented as cloud.

lucaromagnoli:
Thanks for this, and for the provider wiring in particular. Routing it through the OpenAI-Chat path is exactly right, and the `toLLMProviderConfig` / preset plumbing is the genuinely fiddly part.

I'm going to close it though, for reasons that aren't about code quality:

**1. Placement.** Ollama runs locally, so surfacing it under Cloud (and routing the preset into Cloud mode) is the wrong model for users. Its correct home is the Local section, which today only knows about in-process GGUF / llama.cpp. Giving it a proper home means redesigning that Local side first, so this can't land correctly until that happens.

**2. I'd rather not have anything Ollama-specific.** The better shape is one generic "local / OpenAI-compatible server" provider, configured by base URL, that covers LM Studio, Ollama, GPUStack and friends, instead of a per-tool entry with `ollama*` config fields and an Ollama icon.

I'll fold the wiring from here into that redesign when I do the Local rework, so there's nothing more you need to do, and you'll be credited. For next time, opening an issue to discuss a feature like this before building it saves exactly this kind of mismatch.

Thanks again.

PR #1575, opened 2026-06-25
Feat/fm synth

lucaromagnoli:


lucaromagnoli:
Merging directly into feat/faust-instruments instead of via PR.

PR #1596, opened 2026-06-26
Make DAWproject schema validation advisory, not a hard gate

lucaromagnoli:
Importing a DAWproject exported from Cubase 15 aborted with schema errors (out-of-order Devices / automation points vs the bundled XSD's strict <xs:sequence>), so the project never loaded. Our parser is order- independent (it locates children by name), so these files are readable even when they fail strict validation.

Validate for diagnostics and log any schema problems, but continue to the parser instead of returning false. The parser is now the real gate: it fails only if the project genuinely cannot be read. error stays free of schema noise so a real parse failure still surfaces a clear message.


aloomens:
I was able to import 2 different DAWproject exports from Cubase 15 without a problem after this fix.

PR #1628, opened 2026-06-28
Fix arrangement clips dropped from the engine on project load

lucaromagnoli:
## Problem

On project load, arrangement clips (MIDI and audio) were silently dropped from Tracktion Engine, so reopening a saved project played back silent even though the clip data was intact in the model and visible in the UI. Instruments only responded to live keyboard input, not their clips.

## Root cause

`ProjectSerializer::commitStagedData` restored everything inside three coalesced `BatchScope`s. TE `AudioTrack`s are created lazily by `AudioBridge::syncAll` when the track batch's `notifyTracksChanged()` flushes. Because `BatchScope` destructors run in reverse declaration order (`trackBatch` declared first, so it flushes last), the clip notification flushed before the track notification. `ClipSynchronizer` then bailed with "no TE track" and dropped each arrangement clip.

This regressed in #1596, which introduced the batching to fix O(N^2) panel rebuilds on load. Before that, `restoreTrack` notified immediately, so TE tracks existed before clips were restored. It shipped in v0.12.2.

## Fix

Commit the restored tracks in their own batch that closes before clips are restored, so the TE `AudioTrack`s exist by the time clips sync into them.

## Tests

New `test_project_load_clip_ordering_juce.cpp` drives the real `saveToFile` -> `loadAndStage` -> `commitStaged` path with the engine wired, asserting clips land on their TE tracks after load:

- MIDI arrangement clip survives load with all notes
- Audio arrangement clip survives load
- Clips on every track survive (ordering holds beyond the first track)
- Instrument track plus its MIDI clip both survive (the reported scenario)
- Render check: the instrument is actually driven by its restored MIDI clip (RMS > 0)

All five fail without the fix (verified by reverting the ordering) and pass with it.

### DIFF
--- a/magda/daw/project/serialization/ProjectSerializer.cpp
+++ b/magda/daw/project/serialization/ProjectSerializer.cpp
@@ -607,25 +607,36 @@ void ProjectSerializer::commitStagedData(std::vector<TrackInfo>& stagedTracks,
     // the UI for seconds on a large project. The batch fires one notification when
     // the scope closes.
     {
-        TrackManager::BatchScope trackBatch;
         ClipManager::BatchScope clipBatch;
         AutomationManager::BatchScope automationBatch;
 
         // Clear all existing data from managers
-        trackManager.clearAllTracks();
         clipManager.clearAllClips();
         automationManager.clearAll();
 
-        // Restore tracks
-        for (auto& track : stagedTracks) {
-            trackManager.restoreTrack(track);
-        }
+        // Restore tracks in their own batch that closes BEFORE clips are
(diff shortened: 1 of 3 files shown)


lucaromagnoli:
In `tests/CMakeLists.txt`, this adds `test_project_load_clip_ordering_juce.cpp` to the `magda_juce_tests` binary, but I do not see a matching `add_test(NAME ... COMMAND magda_juce_tests "Project Load Clip Ordering")` entry below. The surrounding JUCE suites are registered individually, so this regression test will compile but likely will not run under `ctest`/CI unless it gets its own CTest registration.

Issue #1643, opened 2026-06-29
Headless magda-cli for e2e testing


lucaromagnoli:
## Goal

A headless `magda-cli` console binary that can load a `.mgd` project, apply a command or sequence of commands, render/dump results, and save — primarily to enable scriptable end-to-end testing without driving the GUI.

## Why it's feasible now

- The engine already boots headless in tests (`tests/test_export_no_assertion_juce.cpp` constructs `TracktionEngineWrapper` + Edit + renderer with no UI).
- `magda_daw` is a static lib already linked by `magda_tests`, `magda_juce_tests`, and the `magda_plugin_scanner` console app — the CMake/entry-point template already exists (`magda/daw/engine/plugin_scanner_main.cpp`).
- Load/save is fully UI-decoupled: `ProjectManager` is a singleton; `loadProject()`/`saveProjectAs()` work standalone (`.mgd` = gzip JSON).

## Design decisions

- The CLI is a plain C++ binary linked against `magda_daw`/`magda_scripting`. It calls the engine's C++ API (`MagdaApi`, `ProjectManager`, `UndoManager` commands) **directly** — no scripting runtime required.
- "Apply a command" = a `name → C++ call` dispatch table over `MagdaApi`. A "sequence" = a flat command file parsed and dispatched line-by-line (argument parsing, not a DSL).
- **Lua is deferred.** `magda_scripting` already exists, so hosting Lua is cheap, but it's only needed for control flow inside test scripts (loops/conditionals/inline asserts). Not on the critical path.

## Proposed surface

```
magda-cli run    project.mgd  --cmds cmds.txt  [--out out.mgd]
magda-cli exec   project.mgd  add-track audio  set-tempo 120  [--out out.mgd]
magda-cli render project.mgd  --wav out.wav    [--from 0 --to 32bars]
magda-cli dump   project.mgd  --json
```

## Phases (sub-issues)

- P0 — CLI target + headless boot + `load → save` round-trip
- P1 — command dispatch over `MagdaApi` + `dump --json`
- P2 — offline `render --wav` + wire into CTest as e2e golden tests
- P3 — fill missing `MagdaApi` verbs needed by tests

## Risks / gotchas

- **macOS is not auto-headless** — the existing `isHeadless` check is Linux/`DISPLAY`-only; needs a forced-headless flag/env so the wrapper skips window/timer components on mac.
[...]

PR #1648, opened 2026-06-29
Headless magda CLI epic

lucaromagnoli:
## Summary

Implements the headless magda-cli epic from #1643 across P0-P3:

- Adds a headless CLI bootstrap target and engine initialization path.
- Adds project command dispatch for headless project mutation and JSON state dumps.
- Adds offline WAV rendering and a CTest headless e2e covering init, mutation, dump, save, and render.
- Exposes MIDI note verbs through MagdaApi, delegates to existing undoable command implementations, and wires the verbs through CLI and Lua bindings.

## Validation

- cmake -S . -B build-cli -DMAGDA_BUILD_TESTS=OFF -DMAGDA_HAVE_CLAP=OFF -DMAGDA_BUILD_EXAMPLES=OFF
- cmake --build build-cli --target magda_cli -j 8
- magda_cli boot
- cmake -S . -B build-cli-p2 -DMAGDA_BUILD_TESTS=ON -DMAGDA_HAVE_CLAP=OFF -DMAGDA_BUILD_EXAMPLES=OFF
- cmake --build build-cli-p2 --target magda_cli -j 8
- ctest --test-dir build-cli-p2 -R magda_cli_headless_e2e --output-on-failure
- cmake -S . -B build-cli-p3 -DMAGDA_BUILD_TESTS=ON -DMAGDA_HAVE_CLAP=OFF -DMAGDA_BUILD_EXAMPLES=OFF
- cmake --build build-cli-p3 --target magda_cli -j 8
- cmake --build build-cli-p3 --target magda_scripting -j 8
- ctest --test-dir build-cli-p3 -R magda_cli_headless_e2e --output-on-failure

Draft until CI confirms the new branch state.

### DIFF
--- a/magda/daw/cli/magda_cli_main.cpp
+++ b/magda/daw/cli/magda_cli_main.cpp
@@ -2,6 +2,11 @@
 #include <juce_events/juce_events.h>
 #include <juce_gui_basics/juce_gui_basics.h>
 
+// clang-format off
+#include <tracktion_engine/tracktion_engine.h>
+// clang-format on
+
+#include <atomic>
 #include <functional>
 #include <iostream>
 #include <memory>
@@ -14,6 +19,7 @@
 #include "api/magda_api.hpp"
 #include "api/project_api.hpp"
 #include "api/track_api.hpp"
+#include "audio/AudioBridge.hpp"
 #include "engine/TracktionEngineWrapper.hpp"
 #include "project/ProjectInfo.hpp"
 #include "project/ProjectManager.hpp"
@@ -21,14 +27,18 @@
 
 namespace {
 
+namespace te = tracktion;
+
 void printUsage(std::ostream& out) {
     out << "magda-cli " << MAGDA_VERSION << "\n"
         << "\n"
         << "Usage:\n"
         << "  magda-cli boot\n"
+        << "  magda-cli init <out.mgd>\n"
         << "  magda-cli run <project.mgd> [--out <out.mgd>]\n"
         << "  magda-cli run <project.mgd> --cmds <cmds.txt> [--out <out.mgd>] [--dump-json]\n"
         << "  magda-cli exec <project.mgd> <commands...> [--out <out.mgd>] [--dump-json]\n"
+        << "  magda-cli render <project.mgd> --wav <out.wav> [--from <time>] [--to <time>]\n"
         << "\n"
         << "Commands:\n"
         << "  set-tempo <bpm>\n"
@@ -93,6 +103,13 @@ std::optional<double> parseDouble(const juce::String& text) {
 std::optional<int> parseInt(const juce::String& text) {
     int value = 0;
     if (std::istringstream{text.toStdString()} >> value)
+        return value;
+    return std::nullopt;
+}
+
+std::optional<int> parsePositiveInt(const juce::String& text) {
+    auto value = parseInt(text);
+    if (value && *value > 0)
         return value;
     return std::nullopt;
 }
@@ -340,6 +357,15 @@ struct RunOptions {
     juce::StringArray execTokens;
 };
 
+struct RenderOptions {
+    juce::File input;
+    juce::File wavOutput;
+    std::optional<double> fromSeconds;
+    std::optional<double> toSeconds;
+    std::optional<double> sampleRate;
+    std::optional<int> bitDepth;
+};
+
 bool loadProjectForCli(const juce::File& input, HeadlessEngineSession& session) {
     auto& projectManager = magda::ProjectManager::getInstance();
     if (!projectManager.loadProject(input, [&session](const magda::ProjectInfo& info) {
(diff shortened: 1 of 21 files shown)


magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,958 @@
...
+
+bool restoreProjectTiming(magda::TracktionEngineWrapper& engine, const magda::ProjectInfo& info) {
+    engine.setTempo(info.tempo);
+    engine.setTimeSignature(info.timeSignatureNumerator, info.timeSignatureDenominator);
+    return true;
+}
+
+std::optional<double> parseDouble(const juce::String& text) {
+    double value = 0.0;
+    if (std::istringstream{text.toStdString()} >> value)
```
lucaromagnoli: `parseDouble`/`parseInt` use `istringstream >> value`, which accepts non-finite and partial input. Two consequences:

- `set-tempo nan` / `inf`: the guard at line 254 (`!bpm || *bpm <= 0.0`) is false for NaN and +Inf, so a non-finite tempo is written to the engine and ProjectManager and saved, breaking every beats/seconds conversion downstream.
- Trailing garbage passes silently: `delete-track 3x` deletes track 3, `add-midi-clip 1 0 4junk` uses length 4, `transpose 12.9` drops the `.9`.

Reject input that isn't fully consumed, and reject non-finite doubles with `std::isfinite`.

magda/daw/engine/TracktionEngineWrapperInit.cpp
```diff
@@ -479,20 +494,24 @@ bool TracktionEngineWrapper::initialize() {
...
-        // Configure audio devices if user has preferences
-        juce::Logger::writeToLog("[Init] configureAudioDevices()...");
-        configureAudioDevices();
-        juce::Logger::writeToLog("[Init] configureAudioDevices() done");
-
-        // Setup MIDI devices
-        juce::Logger::writeToLog("[Init] setupMidiDevices()...");
-        setupMidiDevices();
-        juce::Logger::writeToLog("[Init] setupMidiDevices() done");
+        if (!isHeadlessRuntime()) {
```
lucaromagnoli, line 495: Device/MIDI init was unconditional before; gating it behind `isHeadlessRuntime()` changes behavior for existing callers, not just the CLI. The heuristic returns true on any Linux box with `DISPLAY` unset, and the Catch2 `magda_tests` CI job runs without `xvfb` (unlike `magda_juce_tests`, which uses `xvfb-run`). Catch2 tests that call `engine.initialize()` (e.g. `test_export_transport_stop.cpp`) now take the skip-init path where they previously got a configured device manager.

Safer to have the tests and the CLI opt in explicitly via `setForceHeadless`, so the path doesn't depend on ambient `DISPLAY`.

tests/magda_cli_headless_e2e.cmake
```diff
@@ -0,0 +1,89 @@
...
+)
+if(NOT render_result EQUAL 0)
+    message(FATAL_ERROR "magda-cli render failed: ${render_stderr}\n${render_stdout}")
+endif()
+if(NOT EXISTS "${wav_file}")
+    message(FATAL_ERROR "magda-cli render did not create ${wav_file}")
+endif()
+
+file(SIZE "${wav_file}" wav_size)
+if(wav_size LESS 44)
```
lucaromagnoli: This only checks the file is >= 44 bytes with a RIFF/WAVE header. The fixture is a MIDI clip on an audio track with no instrument, so the render is silence by construction; nothing asserts sample count, duration, or non-silent audio. Combined with the headless device-init skip, the render path is effectively unverified: an empty, zero-length, or wrong-duration output would still pass. Add an instrument to the fixture and assert a non-trivial sample count / duration.

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,958 @@
...
+  private:
+    CommandResult setTempo(const juce::StringArray& tokens, size_t& index) {
+        if (index >= static_cast<size_t>(tokens.size()))
+            return fail("set-tempo requires <bpm>");
+        auto bpm = parseDouble(tokens[static_cast<int>(index++)]);
+        if (!bpm || *bpm <= 0.0)
+            return fail("set-tempo requires a positive numeric bpm");
+
+        engine_.setTempo(*bpm);
+        magda::ProjectManager::getInstance().setTempo(*bpm);
```
lucaromagnoli: `set-tempo` is the only mutating verb that bypasses `getMagdaApi()` and the undoable commands: it writes `engine_.setTempo()` and `ProjectManager::setTempo()` directly. That can diverge from how a GUI tempo edit propagates (tempo-lane sync / timecode remap), and the two writes have to stay in lockstep by hand. Route it through MagdaApi like the other verbs.

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,958 @@
...
+    CommandResult dump(const juce::StringArray& tokens, size_t& index) {
+        if (index >= static_cast<size_t>(tokens.size()) ||
+            tokens[static_cast<int>(index)] != "--json")
+            return fail("dump requires --json");
+        ++index;
+        dumpJson();
+        return {};
+    }
+
+    static bool isCommand(const juce::String& token) {
```
lucaromagnoli, line 446: `isCommand()` duplicates the verb list in `execute()` (~216) and `printUsage()` (~45), and it's load-bearing: variadic verbs use it to find where one command's args stop. Add a verb to `execute()` but forget it here and the next command's tokens get silently swallowed as arguments, with no error. A single command table (name + handler) consumed by all three sites removes the desync.

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,958 @@
...
+
+    if (file.exists() && !file.deleteFile()) {
+        std::cerr << "Failed to replace output file: " << file.getFullPathName() << "\n";
+        return false;
+    }
+
+    return true;
+}
+
+bool renderWav(magda::TracktionEngineWrapper& engine, const RenderOptions& options) {
```
lucaromagnoli, line 634: `renderWav` re-implements `MainWindowExport.cpp`'s render path: same `Renderer::Parameters` fields, `freePlaybackContextIfNotRecording`, the force-enable-plugins loop, and prepare/restore. Worth extracting a shared headless render helper so a future fix to one path doesn't miss the other. (The force-enable-without-restore at line 608 matches the GUI export, so that part is consistent, not a regression.)

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,958 @@
...
+        if (!length)
+            return fail(lastParseError_);
+
+        int velocity = 100;
+        if (index < static_cast<size_t>(tokens.size()) &&
+            !isCommand(tokens[static_cast<int>(index)])) {
+            auto parsedVelocity = parseInt(tokens[static_cast<int>(index)]);
+            if (!parsedVelocity)
+                return fail("add-midi-note velocity must be an integer");
+            velocity = *parsedVelocity;
```
lucaromagnoli: Velocity isn't clamped to 0..127 here; an out-of-range value (e.g. `add-midi-note 1 0 60 1 999`) is stored and serialized straight through. Clamp at the CLI boundary.

lucaromagnoli (commented):
Self-review of the headless CLI. Overall solid. Main things: a few input-validation gaps in the CLI parsers (NaN/Inf tempo is the worst), the device-init gating quietly changes behavior for the non-xvfb Catch2 test job, and the render e2e renders silence so it doesn't actually verify audio. Details inline.

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,1013 @@
...
+  private:
+    CommandResult setTempo(const juce::StringArray& tokens, size_t& index) {
+        if (index >= static_cast<size_t>(tokens.size()))
+            return fail("set-tempo requires <bpm>");
+        auto bpm = parseDouble(tokens[static_cast<int>(index++)]);
+        if (!bpm || *bpm <= 0.0)
+            return fail("set-tempo requires a positive numeric bpm");
+
+        engine_.setTempo(*bpm);
+        engine_.getMagdaApi().project().setTempo(*bpm);
```
lucaromagnoli, line 266: Better, the ProjectManager write now goes through `ProjectApi::setTempo()`. Only partly resolved though: this is still two manual writes (`engine_.setTempo()` plus `project().setTempo()`) that have to stay in lockstep, and tempo is still the one verb that isn't a single undoable command like the others. Fine to leave for now, but a single `MagdaApi` tempo op that updates both the engine tempo sequence and the project info would close it.

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -0,0 +1,1013 @@
...
+
+    if (file.exists() && !file.deleteFile()) {
+        std::cerr << "Failed to replace output file: " << file.getFullPathName() << "\n";
+        return false;
+    }
+
+    return true;
+}
+
+bool renderWav(magda::TracktionEngineWrapper& engine, const RenderOptions& options) {
```
lucaromagnoli, line 634: Still open from the last round (lowest priority): `renderWav` duplicates `MainWindowExport.cpp`'s render path (same `Renderer::Parameters` fields, `freePlaybackContextIfNotRecording`, the force-enable loop, prepare/restore). Worth extracting a shared headless render helper at some point so the two can't drift. Not a blocker.

lucaromagnoli (commented):
Re-checked against the feedback commit. The parsers (NaN/Inf + trailing garbage), the headless device-init gating, the velocity clamp, the command-table dedup, and the e2e (real instrument + meaningful size floor) all look properly addressed. I confirmed the CLI sets `setForceHeadless(true)` before init, so dropping the DISPLAY heuristic is safe and other callers get unconditional init back. Two small leftovers noted inline.

PR #1660, opened 2026-06-30
Epic/1630 decouple device uis

lucaromagnoli:


magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -1658,20 +1806,184 @@ bool DeviceCustomUIManager::createDrumGridUI(const magda::DeviceInfo& device,
...
+            initialPath = callbacks.getNodePath();
+        deviceUiContext_ = std::make_shared<magda::BasicDeviceUiContext>(device, initialPath);
+    } else if (auto* basicContext =
+                   dynamic_cast<magda::BasicDeviceUiContext*>(deviceUiContext_.get())) {
+        basicContext->setDevice(device);
+    }
+    if (auto* basicContext = dynamic_cast<magda::BasicDeviceUiContext*>(deviceUiContext_.get())) {
+        basicContext->setParameterController(std::make_shared<CallbackDeviceParameterController>(
+            device, callbacks.onParameterChanged));
+        basicContext->setCommandController(std::make_shared<CallbackDeviceCommandController>(
```
lucaromagnoli, line 1991: This command-controller lambda runs ~145 lines inline, handling sampler + IR + both sequencers as one long `if (command == …)` chain. It centralizes the coupling nicely, but mixes many device concerns into a single function and re-implements the old focused lambdas in one place. Consider extracting per-device dispatch into named private methods (`executeSamplerCommand`, `executeSequencerCommand`, …) or a small dispatch table — it would also make the command surface unit-testable.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -555,6 +658,119 @@ tracktion::engine::Plugin::Ptr DeviceCustomUIManager::getLivePlugin() const {
...
+            dynamic_cast<daw::audio::PolyStepSequencerPlugin*>(plugin.get()), exportButton,
+            dragOwner, event);
+    }
+
+    return handleStepSequencerPatternExternalDrag(
+        dynamic_cast<daw::audio::StepSequencerPlugin*>(plugin.get()), exportButton, dragOwner,
+        event);
+}
+
+bool DeviceCustomUIManager::getSequencerStepRecordingState(bool polyphonic, int& position,
```
lucaromagnoli, line 721: Inconsistent routing through the command surface: `randomizeSequencerPattern` and `toggleSequencerStepRecording` go through `deviceUiContext_->commands()`, but `getSequencerStepRecordingState` (here), `copySequencerPatternToClipboard`, and `handleSequencerPatternExternalDrag` bypass it and call `getLivePlugin()` + `dynamic_cast` directly. Not a bug, but the abstraction seam is only half-applied — worth either routing these consistently or adding a comment on why they intentionally stay direct.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -267,18 +283,103 @@ bool isMidiFxPlugin(const juce::PluginDescription& desc) {
...
+    juce::var executeCommand(const juce::Identifier& command,
+                             const juce::var& arguments = {}) override {
+        return execute_ ? execute_(command, arguments) : juce::var{};
+    }
+
+  private:
+    std::function<juce::var(const juce::Identifier&, const juce::var&)> execute_;
+};
+
+void writeParameterChange(const DeviceCustomUIManager::Callbacks& callbacks, int paramIndex,
```
lucaromagnoli, line 352: Latent sharp edge: this only falls back to `callbacks.onParameterChanged` when `deviceUiContext == nullptr`. If a caller ever passes a custom (non-`BasicDeviceUiContext`) context whose `parameters()` returns null, the manager never sets a parameter controller on it (controllers are only wired for `BasicDeviceUiContext` in `create()`), so parameter writes are silently dropped with no fallback. No current caller hits this, but consider asserting a controller exists, or keeping the `onParameterChanged` fallback when `parameters() == nullptr`.

magda/daw/ui/components/chain/custom_ui/OscilloscopeUI.cpp
```diff
@@ -105,20 +105,20 @@ OscilloscopeUI::~OscilloscopeUI() {
     colourCombo_.setLookAndFeel(nullptr);
 }
 
-void OscilloscopeUI::setPlugin(daw::audio::OscilloscopePlugin* plugin) {
-    if (plugin_ == plugin)
+void OscilloscopeUI::setTelemetrySource(std::shared_ptr<OscilloscopeTelemetrySource> telemetry) {
+    if (telemetry_ == telemetry)
```
lucaromagnoli, line 109: This `telemetry_ == telemetry` early-out is now effectively dead: `bindAnalyzerPlugins()` and `refreshMiniAnalyzers()` allocate a fresh `make_shared<…TelemetrySource>` on every call, so the pointers never compare equal. Effect: every rebind re-allocates an adapter and resets `lastTapWritePosition_ = 0`, forcing a re-pull from the ring (and `refreshMiniAnalyzers` can be called frequently). Harmless for correctness, but the same-plugin optimization is lost. If you want to keep it meaningful, compare the underlying plugin pointer rather than the adapter shared_ptr. (Same pattern in SpectrumAnalyzerUI.)

lucaromagnoli (commented):
A few non-blocking observations from reviewing the decoupling. The core lifetime-safety design is sound — these are mostly readability / consistency / sharp-edge notes.

PR #1693, opened 2026-07-04
Give session clips engine identity

lucaromagnoli:


magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -249,7 +249,24 @@ void ClipSynchronizer::clipsChanged() {
...
     // Sync session clips to ClipSlots
     const auto& sessionClips = clipManager.getSessionClips();
-    bool sessionClipsSynced = removeVacatedSessionSlots(sessionClips);
+    bool sessionClipsSynced = false;
+    std::unordered_set<ClipId> currentSessionClipIds;
+    currentSessionClipIds.reserve(sessionClips.size());
+    for (const auto& clip : sessionClips)
+        currentSessionClipIds.insert(clip.id);
+
+    for (const auto& entry : clipIds_.snapshot()) {
```
lucaromagnoli: Efficiency: this loop iterates the whole clipIds_ map (arrangement + session) and calls getSessionTeClip() for every entry that isn't a current session clip, which includes every arrangement clip. getSessionTeClip() does an O(tracks x slots) scan with an itemID.toString().toStdString() allocation per slot, so each clipsChanged() pass does roughly arrangementClips x tracks x slots string conversions that always miss for arrangement clips. On a large project that is real message-thread latency versus the old O(lastKnownSlots) diff. Consider restricting this to session mappings (or skipping current arrangement clips), or resolving slot itemIDs into a set once.

magda/daw/audio/session/ArrangementClipSyncPlanner.cpp
```diff
@@ -18,17 +18,18 @@ ArrangementClipSyncPlan buildArrangementClipSyncPlan(tracktion::Edit& edit,
...
-
     std::unordered_map<std::string, tracktion::AudioTrack*> engineIdToParentTrack;
     for (auto* track : tracktion::getAudioTracks(edit)) {
         for (auto* teClip : track->getClips())
             engineIdToParentTrack[teClip->itemID.toString().toStdString()] = track;
     }
 
+    for (const auto& [clipId, engineId] : clipIds.snapshot()) {
+        if (currentClipIds.find(clipId) == currentClipIds.end() &&
+            engineIdToParentTrack.find(engineId) != engineIdToParentTrack.end())
```
lucaromagnoli: Behaviour change worth a look: with this guard, a deleted clip whose TE clip is already off the timeline is skipped, so it never reaches removeClipFromEngine() and its clipIds_ entry is never erased. For example a track deletion that removes its timeline clips before this runs would leave those clips' mappings lingering in clipIds_ for the rest of the session. It is a memory-only leak and harmless to sync, but flagging it since the previous path always cleaned the mapping.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -248,8 +259,20 @@ void ClipSynchronizer::clipsChanged() {
     }
 
     // Sync session clips to ClipSlots
-    const auto& sessionClips = clipManager.getSessionClips();
-    bool sessionClipsSynced = removeVacatedSessionSlots(sessionClips);
+    bool sessionClipsSynced = false;
+    for (const auto& entry : clipIds_.snapshot()) {
```
lucaromagnoli: Thanks, this addresses both earlier notes. One follow-up: this session-removal loop now looks like dead code. The planner's clipsToRemove already covers any clip that is in neither the current arrangement nor the current session set, and removeClipFromEngine -> removeTeClipByEngineId searches slots too, so deleted session clips are removed (and their clipIds_ mapping erased) in the clipsToRemove pass above. By the time this loop takes its snapshot, the neither-current set is already empty, so it never does anything. It could be deleted for clarity.

PR #1703, opened 2026-07-05
Add macOS plugin format preference

lucaromagnoli:


lucaromagnoli (commented):
Reviewed the format-preference mechanism end to end (each finding below was adversarially verified against the code, and two candidate issues were refuted along the way). The core idea is sound and the pass-through on non-mac platforms is right; it also incidentally fixes a pre-existing dangling bestMatch pointer in the DSL lookup. Four confirmed issues and two robustness notes, worst first. The favorites/aliases data loss is the one I would fix before merging.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -482,7 +482,8 @@ void PluginBrowserContent::loadExternalPlugins() {
     }
 
     auto& knownPlugins = engine_->getKnownPluginList();
-    auto pluginTypes = knownPlugins.getTypes();
+    auto pluginTypes =
+        magda::PluginPreferences::getInstance().preferredExternalPlugins(knownPlugins.getTypes());
```
lucaromagnoli: **Confirmed: filtering permanently deletes favorites/aliases saved on the hidden format.**

Favorites and custom aliases are keyed by format-specific uniqueId, and both files are rewritten from the visible `plugins_` list only (`saveFavorites`/`saveAliases` iterate `plugins_`, no merge with on-disk state). With prefer-VST3, a favorite/alias on the AU build first disappears from view, then the next toggleFavorite or alias edit rewrites the XML without it. Switching the preference back does not restore it; the data is gone.

Either key favorites/aliases by the format-agnostic group identity, or make the save functions merge with entries whose keys are not in the current `plugins_` list.

magda/agents/dsl_interpreter.cpp
```diff
@@ -1180,10 +1181,12 @@ bool Interpreter::executeAddFx(const Params& params) {
     }
 
     const auto& knownPlugins = teWrapper->getKnownPluginList();
+    const auto pluginTypes =
+        PluginPreferences::getInstance().preferredExternalPlugins(knownPlugins.getTypes());
```
lucaromagnoli: **Confirmed: the DSL's `format=` hint can no longer reach the non-preferred format.**

`executeAddFx` reads `params.get("format")` and filters on it inside the loop (line 1194), but the loop now iterates the pre-collapsed list. With the default prefer-VST3, `.fx.add(name="Pro-Q 3", format="AU")` errors "FX plugin not found" even though the AU is scanned and loadable. dsl_grammar.hpp:128 explicitly advertises the hint.

Suggest: only apply `preferredExternalPlugins` when no format hint was given (fall back to the raw list when `formatHint.isNotEmpty()`).

magda/daw/core/PluginPreferences.cpp
```diff
@@ -97,6 +124,64 @@ void PluginPreferences::setBrowserCategoryOverride(const juce::String& pluginIde
...
+    std::lock_guard<std::mutex> lock(mutex_);
+    return externalFormatPreference_;
+}
+
+void PluginPreferences::setExternalPluginFormatPreference(
+    ExternalPluginFormatPreference preference) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (externalFormatPreference_ == preference)
+        return;
+    externalFormatPreference_ = preference;
```
lucaromagnoli: **Confirmed: changing the preference does not take effect until a rescan or restart.**

This saves but notifies nothing (the Listener class has no format callback), and `PluginSettingsDialog::applySettings` fires no KnownPluginList broadcast. An open plugin browser keeps its stale `plugins_` cache, so the new dropdown looks broken, while a freshly opened add-device menu already shows the new filtering, so two surfaces disagree within one session.

Add a listener callback (or reuse the existing ListenerList) and have the plugin browser refresh on it.

magda/daw/core/PluginPreferences.cpp
```diff
@@ -18,6 +20,31 @@ constexpr const char* kMidiFxCategoryOverride = "MIDI FX";
 bool hasGlobalKitDefault(const juce::String& pluginIdentifier) {
     return pluginIdentifier != kDrumGridBuiltinId;
 }
+
+bool isVST3Format(const juce::String& format) {
```
lucaromagnoli: **Reuse: this is roughly the 7th copy of pluginFormatName parsing, plus a second format enum.**

An identical predicate already exists in DeviceCustomUIManager.cpp:388 (`pluginFormatFromDescription`), with hand-rolled copies in ChainPanel, TrackChainContent, PostFxPanelContent, instruction_executor and dsl_interpreter. And `ExternalPluginFormatPreference` parallels `magda::PluginFormat` (DeviceInfo.hpp:16), which this header already includes, with its own to/from-string round-trip.

Suggest hoisting one shared format classifier next to `PluginFormat` and storing the preference as a `PluginFormat` restricted to VST3/AU.

magda/daw/core/PluginPreferences.cpp
```diff
@@ -18,6 +20,31 @@ constexpr const char* kMidiFxCategoryOverride = "MIDI FX";
...
+}
+
+PluginPreferences::ExternalPluginFormatPreference formatPreferenceFromString(
+    const juce::String& value) {
+    if (value.equalsIgnoreCase(kFormatAU) || value.containsIgnoreCase("AudioUnit"))
+        return PluginPreferences::ExternalPluginFormatPreference::AU;
+    return PluginPreferences::ExternalPluginFormatPreference::VST3;
+}
+
+juce::String duplicateGroupKey(const juce::PluginDescription& desc) {
```
lucaromagnoli, line 32: **Robustness note (verified as a no-op on my current scan, but worth knowing):** name+manufacturer is a heuristic identity. JUCE fills manufacturerName from unrelated per-format metadata (AU component display string vs VST3 factory vendor), so vendor-string drift makes a pair never collapse (benign: both stay listed). The reverse case, two genuinely different products sharing name and vendor across formats, would get one dropped from every menu and lookup. I parsed the actual PluginList.xml on this machine (868 entries, 8 real AU/VST3 twins): all collapse correctly today and no false twin exists, so nothing is live; just do not tighten or trust this key further without a real identity check.

magda/daw/core/PluginPreferences.cpp
```diff
@@ -97,6 +124,64 @@ void PluginPreferences::setBrowserCategoryOverride(const juce::String& pluginIde
...
+
+void PluginPreferences::setExternalPluginFormatPreference(
+    ExternalPluginFormatPreference preference) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (externalFormatPreference_ == preference)
+        return;
+    externalFormatPreference_ = preference;
+    saveUnlocked();
+}
+
```
lucaromagnoli, line 133: **Altitude: the filter is opt-in by convention at 8 call sites.**

Every remaining raw `getTypes()` site today is a legitimate exact-identifier lookup (ParameterConfigDialog, drag-drop resolution, project restore), so nothing is currently wrong. But nothing marks the filtered/unfiltered split, and the next plugin-listing surface will most naturally call `getTypes()` and silently bypass the preference. A `TracktionEngineWrapper::getPreferredPluginTypes()` accessor would make the right call the easy one and document the split in one place.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
lucaromagnoli: Fixed in the update: saveFavorites/saveAliases now merge on-disk entries whose keys are not in the visible list before rewriting, so hidden-format curation survives preference flips. Side effect worth knowing: favorites of uninstalled plugins also persist now instead of being pruned, which seems fine.

magda/agents/dsl_interpreter.cpp
lucaromagnoli: Fixed: an explicit format hint now searches the raw KnownPluginList (formatHint.isNotEmpty() ? knownPlugins.getTypes() : teWrapper->getPreferredPluginTypes()), so the hint can reach the non-preferred format again.

magda/daw/core/PluginPreferences.cpp
lucaromagnoli: Fixed: Listener gained externalPluginFormatPreferenceChanged, the setter notifies outside the mutex (matching the drum-grid notifier), and the plugin browser subscribes and refreshes. One residue: AIChatConsoleContent's alias cache does not subscribe, so a preference change while the console panel is active keeps the old aliases until the panel is re-activated. Minor, but a one-line listener there would close it.

magda/daw/core/PluginPreferences.cpp
lucaromagnoli: Fixed: the parallel enum is gone (preference is a magda::PluginFormat), pluginFormatFromName() is hoisted to DeviceInfo.hpp, and DeviceCustomUIManager delegates to it. One note: pluginFormatFromName defaults unknown names to VST3, which makes the dedup's old neither-format pass-through arm unreachable. Harmless today (macOS external formats are only VST3/AudioUnit/VST), but worth revisiting if CLAP support ever lands.

magda/daw/core/PluginPreferences.cpp
lucaromagnoli: Unchanged in the update, which is fine. This was a robustness note, not a live bug; all 8 real AU/VST3 twins on my scan collapse correctly.

magda/daw/core/PluginPreferences.cpp
lucaromagnoli: Fixed: TracktionEngineWrapper::getPreferredPluginTypes() is now the chokepoint, all call sites use it, and the doc comment distinguishes presentation from exact-identifier lookup.

PR #1732, opened 2026-07-09
Feat/clip automation UI

lucaromagnoli:


magda/daw/core/AutomationCommands.cpp
```diff
@@ -271,6 +273,190 @@ bool DuplicateAutomationTimeSelectionCommand::shouldDuplicateLane(
...
+}
+
+// ============================================================================
+// BakeModulationCommand
+// ============================================================================
+
+void BakeModulationCommand::execute() {
+    auto& autoMgr = AutomationManager::getInstance();
+    const auto* lane = autoMgr.getLane(laneId_);
+    if (!lane || !lane->isAbsolute())
```
lucaromagnoli: `getOrCreateLane` returns the existing singleton regardless of the requested type. If this target already has a clip-based lane, this path still opens the bake action, but `BakeModulationCommand::execute()` returns at the `!lane->isAbsolute()` guard before replacing points or disabling the links. The user gets a no-op bake. Either convert/handle clip-based lanes before creating the command, or disable the bake action for those targets; please add coverage for that existing clip-lane case.

PR #1734, opened 2026-07-10
Support sidechain on master track

lucaromagnoli:


magda/daw/audio/plugins/SidechainPlugin.cpp
```diff
@@ -137,18 +149,28 @@ void SidechainPlugin::applyToBuffer(const te::PluginRenderContext& fc) {
         // Attack when ducking (gain falling), release when recovering.
         const float coeff = rampValue_ < gain ? attackCoeff : releaseCoeff;
         gain += coeff * (rampValue_ - gain);
-        for (int ch = 0; ch < usedChannels; ++ch)
-            channels[ch][i] *= gain;
+        if (sidesOnly) {
```
lucaromagnoli, line 152: Sides mode only touches channels[0] and [1]. With usedChannels > 2, the channels beyond stereo skip the else branch entirely and pass through un-ducked (the per-channel gain never gets applied to them). It is theoretical for a stereo-focused app, but a >2-channel buffer in Sides mode leaves the higher channels unprocessed. A one-line comment noting Sides is stereo-only, or ducking channels >= 2 with the normal gain, would close the gap.

magda/daw/audio/plugin_manager/PluginManagerModifiers.cpp
```diff
@@ -1017,9 +1038,35 @@ std::pair<int, int> PluginManager::computeModLinkFingerprint(TrackId trackId,
...
+        macroList = &teTrack->getMacroParameterListForWriting();
+        visitHostPlugins = [teTrack](const std::function<void(te::Plugin*)>& visit) {
+            for (int pi = 0; pi < teTrack->pluginList.size(); ++pi) {
+                if (auto* plugin = teTrack->pluginList[pi])
+                    visit(plugin);
+            }
+        };
+    } else if (trackId == MASTER_TRACK_ID) {
+        if (auto* masterTrack = edit_.getMasterTrack()) {
+            defaultModifierList = masterTrack->getModifierList();
```
lucaromagnoli, line 1058: macroList stays null on the master path (only the teTrack branch assigns it), so master devices never get TE macro parameters and macro->parameter modulation silently will not apply on master. I checked: this is correct, not an oversight, because TE's Track/MasterTrack is not a MacroParameterElement and so has no macro list to give, and every downstream use guards the null. Worth a short comment here saying exactly that, so a future reader does not mistake it for a missing assignment.

lucaromagnoli (commented):
Reviewed the master-sidechain threading and the new Sides mode. Overall this looks solid and safe to merge: the refactor of syncDeviceModifiers to take a ModifierList/MacroParameterList/visitHostPlugins is clean, the master's lack of a macro list is handled null-safely (the walker and both teardown helpers guard it), the device-removal teardown mirrors the regular-track path so there is no new lifetime hazard, and the M/S math is correct and directly tested. Two small notes inline, neither blocking.

One coverage question: testMasterModifierAndSourceMonitor asserts the structural wiring (master modifier created, source monitor retained) but does not assert that the master device actually ducks in response to the source signal. If there is no render-level test that covers that end to end, worth a manual check.

PR #1738, opened 2026-07-10
Feat/ghost clips

lucaromagnoli:


magda/daw/core/ClipManager.cpp
```diff
@@ -1112,6 +1213,13 @@ ClipId ClipManager::splitClipAtBeat(ClipId clipId, double splitBeat, double temp
 
             clip->midiNotes = leftNotes;
             rightClip.midiNotes = rightNotes;
+
+            // Partitioning rewrites shared content, so ghost halves cannot
+            // stay in their link group: the next propagation would truncate
+            // every sibling. Looped-MIDI and audio splits are pure windowing
+            // (midiOffset / offset) and keep their group.
+            clip->linkGroupId = 0;
```
lucaromagnoli, line 1235: This only detaches non-looped MIDI. For linked audio and looped-MIDI clips, execution falls through with linkGroupId intact, then appends an L suffix to the left clip and calls notifyClipPropertyChanged below. Since the name is now a shared field, that notification renames every existing group member to the L-suffixed name; the new right half still has the R-suffixed name until its next property change, at which point it can rename the whole group again. A linked split therefore leaves the group internally inconsistent. Either detach both halves for every split, or avoid the L/R name mutation and make the split members shared-content semantics explicit. Please add coverage for a linked audio or looped-MIDI split with a third sibling.

PR #1740, opened 2026-07-11
Feat/grid quantization picker

lucaromagnoli:


magda/daw/ui/components/common/GridDivisionMenu.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+        slash_.setText("/", juce::dontSendNotification);
+        slash_.setJustificationType(juce::Justification::centred);
+        slash_.setColour(juce::Label::textColourId, DarkTheme::getSecondaryTextColour());
+        addAndMakeVisible(slash_);
+        hint_.setJustificationType(juce::Justification::centred);
+        hint_.setFont(FontManager::getInstance().getUIFont(11.0f));
+        hint_.setColour(juce::Label::textColourId, DarkTheme::getSecondaryTextColour());
+        addAndMakeVisible(hint_);
+        changed();
```
lucaromagnoli: The constructor ends with `changed()`, which fires `onChanged_` before the user has touched anything. Opening Custom... on the automation snap path immediately calls `setClipSnapX(clipId, true, <same values>)` and emits a spurious `notifyClipsChanged`; on BottomPanel it dispatches an event that only stays harmless because of the no-change guard. Set the hint text here without invoking the callback - this shape will bite as soon as a callee stops being idempotent.

magda/daw/ui/components/common/GridDivisionMenu.hpp
```diff
@@ -0,0 +1,287 @@
...
+    {"1/16T", 1, 24},
+    {"1/32T", 1, 48},
+    {"1/2.", 3, 4},
+    {"1/4.", 3, 8},
+    {"1/8.", 3, 16},
+    {"1/16.", 3, 32},
+    {"1/32.", 3, 64},
+}};
+
+inline std::pair<int, int> normaliseGridDivision(int numerator, int denominator) {
```
lucaromagnoli: This clamp+gcd block now exists in four places: here, ClipManager::setClipGridSettings, AutomationManager::setClipSnapX/Y, and the SetGridQuantize handler - with the 999/64 magic numbers in each. The core managers rightly can't include a daw/ui header, but this helper only needs juce::jlimit and <numeric>: move it to a core-level header and share it everywhere. Would also let one test cover all call sites; right now the tests only exercise this copy.

magda/daw/ui/components/common/GridDivisionMenu.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+class GridDivisionCustomEditor final : public juce::Component {
+  public:
+    GridDivisionCustomEditor(int numerator, int denominator,
+                             std::function<void(int, int)> onChanged)
+        : onChanged_(std::move(onChanged)) {
+        numerator_ = std::make_unique<juce::TextEditor>();
+        denominator_ = std::make_unique<juce::TextEditor>();
+        for (auto* control : {numerator_.get(), denominator_.get()}) {
+            control->setInputRestrictions(3, "");
```
lucaromagnoli: `setInputRestrictions(3, "")` limits length but allows any characters; letters silently fall into the restore-last-valid path. Passing "0123456789" makes the field self-explanatory.

magda/daw/ui/panels/BottomPanel.cpp
```diff
@@ -1405,12 +1427,11 @@ void BottomPanel::layoutMidiHeaderControls(juce::Rectangle<int> headerBounds) {
...
     x -= 4;
-    x -= 24;
-    gridDenominatorLabel_->setBounds(x, y + vPad, 24, h - vPad * 2);
-    x -= 8;
-    gridSlashLabel_->setBounds(x, y, 8, h);
-    x -= 24;
-    gridNumeratorLabel_->setBounds(x, y + vPad, 24, h - vPad * 2);
+    x -= 64;
+    gridDivisionButton_->setBounds(x, y + vPad, 64, h - vPad * 2);
+    gridNumeratorLabel_->setVisible(false);
```
lucaromagnoli: Visibility fighting: `addMidiControlsToHeader()` still sets these labels visible, then this layout pass force-hides them every time - with a potential one-frame flash on content switch. Cleaner to never show them in `addMidiControlsToHeader()` and drop these lines from layout.

magda/daw/ui/panels/TransportPanel.cpp
```diff
@@ -429,13 +430,13 @@ void TransportPanel::resized() {
...
     if (gridVisible_) {
-        const int numDenWidth = 30;
         const int gridGap = 4;
         const int btnWidth = 44;
         int gridX = x + 6;
-        gridNumeratorLabel->setBounds(gridX, rowY1, numDenWidth, rowHeight);
-        gridDenominatorLabel->setBounds(gridX, rowY2, numDenWidth, rowHeight);
-        const int gridBtnX = gridX + numDenWidth + gridGap;
+        gridDivisionButton->setBounds(gridX, rowY1, 56, rowHeight * 2 + 2);
+        gridNumeratorLabel->setBounds(0, 0, 0, 0);
```
lucaromagnoli: These labels stay `setVisible(gridVisible_)` in resized() above but get zero bounds here - visible zero-sized components. Prefer `setVisible(false)` at the visibility site and drop the zero-bounds placement.

magda/daw/ui/panels/content/AutomationClipEditorContent.cpp
```diff
@@ -263,16 +274,20 @@ void AutomationClipEditorContent::syncSnapControls() {
...
+    snapXDivision_->setDivision(clip->snapXNumerator, clip->snapXDenominator);
+    snapYDivision_->setDivision(clip->snapYNumerator, clip->snapYDenominator);
     const auto dim = [](juce::Component& c, bool on) {
         c.setEnabled(on);
         c.setAlpha(on ? 1.0f : 0.6f);
     };
     dim(*snapXNum_, clip->snapXEnabled);
     dim(*snapXDen_, clip->snapXEnabled);
     dim(snapXSlash_, clip->snapXEnabled);
+    dim(*snapXDivision_, clip->snapXEnabled);
```
lucaromagnoli, line 209: Unlike BottomPanel (where the old labels genuinely back the picker as state-holders), `snapXNum_`/`snapXDen_`/slash and the Y counterparts are now fully unreachable here - no longer populated into the header, yet still created, synced, and dimmed. They can be deleted outright.

lucaromagnoli (commented):
Musical-notation picker reads well: division table is musically correct (1/4T = 1/6, dotted = 3/2x), menu id-to-index dispatch checks out against the 6/4/5 array layout, gcd reduction preserves grid spacing, SafePointer guards the popup target, and the tests are model-level. Findings inline. One manual check worth doing before merge: the custom editor now admits odd denominators (e.g. 7) end to end - grid math is pure fraction arithmetic so nothing should care, but a quick pass over the piano roll ruler with an odd denominator would confirm it.

PR #1741, opened 2026-07-11
Feat/render looped MIDI clips

lucaromagnoli:


magda/daw/core/ClipOperations.hpp
```diff
@@ -991,10 +991,12 @@ class ClipOperations {
 
         if (clip.loopEnabled && clip.loopLengthBeats > 0.0) {
             double loopLen = clip.loopLengthBeats;
-            double phase = clip.midiOffset;
+            double phase = std::fmod(clip.midiOffset, loopLen);
```
lucaromagnoli, line 994: Consider the existing wrapPhase(clip.midiOffset, loopLen) helper (ClipInfo.hpp:21) instead of raw std::fmod. It exists exactly for loop phase calculations and is already used on this same field at line 252. The difference is negative input: model setters clamp midiOffset with jmax(0.0, ...), but ClipSerializer.cpp:416 reads it from the project file unclamped. A negative value makes fmod return a negative phase, which shifts cycle 0 past beat 0 and silently drops the head of the clip. wrapPhase wraps into [0, period) and closes that hole for free.

tests/test_clip_commands.cpp
```diff
@@ -593,6 +593,122 @@ TEST_CASE("JoinClipsCommand - basic MIDI join", "[clip][command][join]") {
...
+        note.velocity = 100;
+        clip->midiNotes.push_back(note);
+    }
+
+    // A two-beat pattern with one beat phase offset renders at beats 0..9,
+    // including controller data and pitch bends.
+    clip->loopEnabled = true;
+    clip->loopLengthBeats = 2.0;
+    clip->loopLength = 1.0;
+    clip->midiOffset = 1.0;
```
lucaromagnoli, line 623: The fmod normalization is one of the two changed lines but is untested: all cases here use midiOffset < loopLen. One case with midiOffset = 3.0 and loopLen = 2.0 would pin the wrap-around behavior (and matches what ClipManager::setClipMidiOffset can actually produce, since it only clamps to >= 0).

lucaromagnoli (commented):
Core fix verified against playback semantics: ClipSynchronizer sets the TE loop range to [0, loopLen) with TE offset = midiOffset, so the old cycle count left [numCycles*loopLen - phase, clipLen) silent whenever phase > 0. The new count covers the tail, the existing trim guards prevent overshoot, and the exact-multiple case adds no spurious repetition. It also fixes the same undercount in the split-clip flatten path (ClipCommands.cpp:996).

One pre-existing edge this PR makes user-reachable: the looped flatten branch resets midiOffset and the loop fields but leaves midiTrimOffset untouched. midiTrimOffset accumulates on left-resize of non-looped clips and nothing resets it when looping is enabled. So: left-resize a non-looped clip, enable loop, flatten. The result is a non-looped clip with a stale nonzero midiTrimOffset, and consumers switch back to reading it once loopEnabled is false (ClipOperations.hpp:87, ClipComponent.cpp:678), shifting the freshly baked notes. Worth resetting midiTrimOffset = 0.0 in the looped branch while in this function.

PR #1742, opened 2026-07-11
Add delta solo for effects and racks

### LINKED ISSUE #1675: Delta solo feature for effects and racks
### What would you like?

Simple button in effects bar, I think. Instead of Reaper-like mouse right button use.

### Use case

This is one of the most helpful features in Reaper, so making it even easier would be great for my workflow.

lucaromagnoli:
## Summary

- add a delta-solo button to effect and rack headers, including collapsed and nested rack views
- persist delta-solo state and keep it mutually exclusive with bypass
- synchronize device and rack state to latency-aligned wet-minus-dry processing
- update the Tracktion Engine submodule to Conceptual-Machines/tracktion_engine#26

## Why

Delta solo makes it possible to hear only what an effect or rack changes. This provides a direct control in the device bar instead of requiring a mouse-button convention.

For plugins that report latency, the dry input is delayed before subtraction so the result remains phase-aligned. Racks reuse Tracktion's existing compensated dry-return path and retain their wet/dry automation values.

Closes #1675.

## Validation

- `cmake --build cmake-build-debug --target magda_tests -j4`
- `cmake-build-debug/tests/magda_tests "[delta_solo]"` - 33 assertions in 5 test cases passed
- full Catch2 suite - 1,413 test cases and 159,703 assertions passed; 6 optional model tests skipped
- `cmake --build cmake-build-debug --target magda_daw_app -j4`

### DIFF
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -578,8 +578,10 @@ void AudioBridge::devicePropertyChanged(const ChainNodePath& devicePath) {
     // Apply enablement directly on the TE plugin: syncFromDeviceInfo only
     // knows the device's own bypassed flag, not the chain gate; and plugins
     // without a processor (e.g. Chord Engine) have no other sync path.
-    if (auto tePlugin = pluginManager_.getPlugin(devicePath))
+    if (auto tePlugin = pluginManager_.getPlugin(devicePath)) {
         tePlugin->setEnabled(effectiveEnabled);
+        tePlugin->setDeltaSoloEnabled(device->deltaSolo);
+    }
 
     if (auto tePlugin = pluginManager_.getPlugin(devicePath))
         daw::audio::syncPluginMidiInThru(tePlugin.get(), device->midiInThru);
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -581,10 +581,8 @@ void AudioBridge::devicePropertyChanged(const ChainNodePath& devicePath) {
     if (auto tePlugin = pluginManager_.getPlugin(devicePath)) {
         tePlugin->setEnabled(effectiveEnabled);
         tePlugin->setDeltaSoloEnabled(device->deltaSolo);
-    }
-
-    if (auto tePlugin = pluginManager_.getPlugin(devicePath))
         daw::audio::syncPluginMidiInThru(tePlugin.get(), device->midiInThru);
+    }
 
     // Wrapped instruments consume MIDI while active. Only top-level devices own
     // instrument wrapper racks; post-fx/mixer-analysis ids are section-local and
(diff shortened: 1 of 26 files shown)


magda/daw/audio/AudioBridge.cpp
```diff
@@ -578,8 +578,10 @@ void AudioBridge::devicePropertyChanged(const ChainNodePath& devicePath) {
     // Apply enablement directly on the TE plugin: syncFromDeviceInfo only
     // knows the device's own bypassed flag, not the chain gate; and plugins
     // without a processor (e.g. Chord Engine) have no other sync path.
-    if (auto tePlugin = pluginManager_.getPlugin(devicePath))
+    if (auto tePlugin = pluginManager_.getPlugin(devicePath)) {
         tePlugin->setEnabled(effectiveEnabled);
+        tePlugin->setDeltaSoloEnabled(device->deltaSolo);
```
lucaromagnoli, line 583: Now that this lookup has braces, the second consecutive getPlugin(devicePath) lookup right below (the midiInThru sync) can move into the same block and drop the duplicate lookup.

lucaromagnoli (commented):
Reviewed together with the TE PR (Conceptual-Machines/tracktion_engine#26); the substantive findings are all on that side (per-block cost of the always-on dry capture, delta FIFO continuity across graph rebuilds, raw pointer capture in the rack lambdas). The MAGDA side is clean.

What I verified:

- Bypass/delta mutual exclusivity is enforced in both directions at every model setter, including the bare-id variants, and mirrored in the UI without notification loops (dontSendNotification throughout).
- Every sync path pushes deltaSolo to TE: createPluginOnly, AudioBridge::devicePropertyChanged, DeviceProcessor/ExternalPluginProcessor::syncFromDeviceInfo, rack loadRackContents, updateElementPropertiesRecursive, and applyRackInstanceState. Enabling delta on a bypassed device also reaches TE correctly: setDeviceDeltaSoloByPath clears bypassed, and devicePropertyChanged recomputes effectiveEnabled from the updated model.
- RackComponent placing deltaButton_ from the right edge in resizedHeaderExtra is safe: NodeComponent::resized removes [delete][power][preset] before calling it, so the button lands left of the reserved cluster.
- Delta button visibility limited to DeviceType::Effect keeps the flag impossible to set on MIDI utilities, where a stale TE-side flag would subtract the audio passthrough to silence.
- Serialization is guarded for pre-feature projects and matches the surrounding hasProperty style; TE state stays runtime-only, MAGDA model authoritative.

Minor test note: the mutual-exclusion test covers the path-based setters only; setDeviceBypassed/setRackBypassed (bare-id) got the same clearing logic but are untested. Not blocking.

magda/daw/audio/AudioBridge.cpp
lucaromagnoli: Still open at the current head; the follow-up commits did not touch this. Trivial, not blocking.

PR #1743, opened 2026-07-11
Fix Faust momentary button controls

lucaromagnoli:


lucaromagnoli (commented):
Reviewed the diff plus surrounding code. Core `MomentaryParamButton` logic is sound: the mouse-gesture and `onStateChange` paths are correctly gated by `mouseGestureActive_`, `release()` covers every lifecycle exit (destructor, rebuildTabs, setOverlayOnly, cancelGesture, setParameterInfo) so I found no stuck-on path, and the `pressed_` dedup prevents double-emits. Test coverage is solid and serialization is backward-safe (hasProperty guard). One functional gap and two minor notes inline. Low risk overall.

magda/daw/audio/plugins/FaustPlugin.cpp
```diff
@@ -138,6 +138,7 @@ float denormalizeForBinding(const FaustParamPool::ActiveBindingDescriptor& b, fl
     const float n = juce::jlimit(0.0f, 1.0f, normalized);
     switch (b.kind) {
         case FaustParamSlot::Kind::Boolean:
+        case FaustParamSlot::Kind::Trigger:
```
lucaromagnoli, line 141: These new `Kind::Trigger` cases are unreachable in the FX path: `UIHarvester::addButton` (around line 96) still emits `Kind::Boolean`, so no FX slot is ever a Trigger. A Faust *effect* with a `button()` therefore still renders as a latching toggle, i.e. the same bug this PR fixes for instruments. Change the FX `addButton` to emit `Kind::Trigger` to match `VoiceHarvester::addButton`, or drop these cases if FX triggers are intentionally out of scope.

magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
```diff
@@ -289,7 +291,7 @@ struct VoiceHarvester : public ::UI {
     }
 
     void addButton(const char* label, FAUSTFLOAT* zone) override {
-        emitControl(FaustParamSlot::Kind::Boolean, label, zone, 0, 0, 1, 1);
+        emitControl(FaustParamSlot::Kind::Trigger, label, zone, 0, 0, 1, 1);
```
lucaromagnoli, line 294: Modeling a trigger as a per-block-sampled parameter means a press+release shorter than one audio block (or two host writes coalesced within a block) can be dropped. Fine for human press-and-hold; worth a note if any DSP relies on catching every edge.

magda/daw/ui/components/chain/custom_ui/FaustInstrumentTabbedUI.hpp
```diff
@@ -47,8 +48,8 @@ class FaustInstrumentTabbedUI : public juce::Component {
     /// refresh by DeviceCustomUIManager.
     void updateFromParameters(const std::vector<magda::ParameterInfo>& params);
 
-    /// Flat list of every tab's sliders (pool index carried via setParamIndex).
-    /// Consumed by DeviceSlotComponent::setupCustomUILinking().
+    /// Flat list of every tab's linkable sliders (pool index carried via
+    /// setParamIndex). Momentary buttons are intentionally excluded.
```
lucaromagnoli, line 52: Excluding momentary buttons here also removes them from MIDI-learn / drag-linking, so a Faust trigger can't be mapped to a hardware pad. Reasonable given modulatable=false; flagging as a deliberate design call, not a defect.

PR #1747, opened 2026-07-12
Add Signalsmith as the default time-stretch engine

lucaromagnoli:


magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -994,6 +1009,26 @@ void ClipSynchronizer::launchSessionClip(ClipId clipId, bool forceImmediate) {
                           ? clip_launch::computeQuantizedBeat(edit_, clip->launchQuantize)
                           : std::optional<te::MonotonicBeat>{};
 
+    if (clip) {
```
lucaromagnoli: Leftover debug scaffolding: this `[StretchTrace][LaunchRequest]` block builds a `juce::String` and writes to the logger on every session clip launch. Looks like it was for diagnosing this change. Worth removing before merge. It is message-thread and `launchHandle` is null-checked above, so no crash or RT-safety issue, just log spam plus an allocation on a hot user action.

magda/daw/core/TimeStretchModes.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace magda::time_stretch_mode {
+
+// These values are persisted in project files and must stay aligned with
+// tracktion::engine::TimeStretcher::Mode.
+inline constexpr int kDisabled = 0;
+inline constexpr int kSoundTouchNormal = 3;
+inline constexpr int kSoundTouchBetter = 4;
+inline constexpr int kSignalsmith = 15;
```
lucaromagnoli, line 10: The comment says all values must stay aligned with `TimeStretcher::Mode`, but only `kSignalsmith` is actually checked (the `STATIC_REQUIRE` in test_audio_clip_stretch.cpp). Consider extending those static checks to `kDisabled` / `kSoundTouchNormal` / `kSoundTouchBetter` too, so a future TE enum reshuffle cannot silently corrupt persisted project values.

magda/daw/core/ClipManager.cpp
```diff
@@ -1540,8 +1540,8 @@ void ClipManager::setAutoTempo(ClipId clipId, bool enabled, double bpm) {
             ClipOperations::setAutoTempo(*clip, enabled, bpm);
 
             // Ensure time-stretching is enabled when beat mode is on
-            if (enabled && clip->timeStretchMode == 0)
-                clip->timeStretchMode = 4;  // soundtouchBetter
+            if (enabled && clip->timeStretchMode == time_stretch_mode::kDisabled)
+                clip->timeStretchMode = time_stretch_mode::kSignalsmith;
```
lucaromagnoli, line 1545: IWYU nit: this file now uses `time_stretch_mode::` but only gets `TimeStretchModes.hpp` transitively via `ClipInfo.hpp` (same for the two stretch-mode combo files). An explicit `#include "core/TimeStretchModes.hpp"` keeps it robust against include-chain changes. The tests already include it directly.

magda/daw/ui/panels/content/AudioClipPropertiesContent.cpp
```diff
@@ -261,9 +261,10 @@ void AudioClipPropertiesContent::createControls() {
     stretchModeCombo_->setColour(juce::ComboBox::textColourId, DarkTheme::getTextColour());
     stretchModeCombo_->setColour(juce::ComboBox::outlineColourId,
                                  DarkTheme::getColour(DarkTheme::BORDER));
-    stretchModeCombo_->addItem("Off", 1);
-    stretchModeCombo_->addItem("SoundTouch", 4);
-    stretchModeCombo_->addItem("SoundTouch HQ", 5);
+    stretchModeCombo_->addItem("Off", time_stretch_mode::kDisabled + 1);
+    stretchModeCombo_->addItem("Signalsmith", time_stretch_mode::kSignalsmith + 1);
```
lucaromagnoli, line 267: Minor: `ClipInspectorSections.cpp` carries a `combo ID = mode + 1` comment explaining this scheme. Worth mirroring it here, since Signalsmith's ID is now 16 sitting between 1 and 4 and reads oddly without the note.

magda/daw/core/ClipOperations.hpp
```diff
@@ -759,9 +759,10 @@ class ClipOperations {
         if (enabled) {
             clip.analogPitch = false;  // Analog pitch is incompatible with autoTempo
 
-            // Auto-tempo requires time-stretching — enable SoundTouch if disabled
-            if (clip.timeStretchMode == 0 || clip.timeStretchMode == 3)
-                clip.timeStretchMode = 4;  // soundtouchBetter
+            // Auto-tempo requires time-stretching. Preserve any explicitly selected
+            // engine, otherwise enable the default quality engine.
+            if (clip.timeStretchMode == time_stretch_mode::kDisabled)
+                clip.timeStretchMode = time_stretch_mode::kSignalsmith;
```
lucaromagnoli, line 766: Intended behavior change worth flagging: previously an explicit SoundTouch normal (mode 3) selection was upgraded to HQ (4) when enabling auto-tempo; now an explicit choice is preserved. This is the fix you want and it is tested (`Explicit SoundTouch remains selected`), but it is a change from prior releases for existing projects.

magda/daw/core/ClipInfo.hpp
```diff
@@ -415,14 +416,14 @@ struct ClipInfo {
...
+    // default quality engine. UI readouts must show this effective mode,
     // not the raw field, so the inspector and the audio editor agree — e.g.
-    // after a session drop auto-enables beat mode. (4 = soundtouchBetter.)
+    // after a session drop auto-enables beat mode.
     int getEffectiveTimeStretchMode() const {
         if (timeStretchMode == 0 && !isAnalogPitchActive() &&
             (autoTempo || warpEnabled || std::abs(speedRatio - 1.0) > 0.001 ||
              std::abs(pitchChange) > 0.001f)) {
-            return 4;  // soundtouchBetter (TE's defaultMode)
+            return time_stretch_mode::kSignalsmith;
```
lucaromagnoli, line 426: Behavior note: existing projects saved with mode Off plus auto-tempo / warp / speed (the silent-default path) will now render through Signalsmith instead of SoundTouch HQ. Quality upgrade, and the field was never explicitly stored, so this is reasonable. Just flagging that saved projects will sound slightly different after this ships.

lucaromagnoli (commented):
Review of the Signalsmith default engine change. One should-fix (leftover debug trace) plus a few nits and behavior notes, all inline. The direction and test coverage look solid.

PR #1752, opened 2026-07-13
Fix session and arrangement clip playback

lucaromagnoli:
## What changed

- preserve warp mode and marker maps when arrangement clips are copied into session slots
- synchronize restored warp markers into Tracktion clips
- reconnect Back to Arrangement activation to launcher playback ownership
- add regression coverage for arrangement/session switching, warp-state copying, Signalsmith startup transients, and session launch de-clicking
- update the Tracktion Engine revision for the playback ownership and transient fix: https://github.com/Conceptual-Machines/tracktion_engine/pull/28

## Why

Several arrangement/session paths had diverged. Arrangement clips could continue processing behind a launched session clip, copied session clips lost their warp-marker state, and the UI was not notified when launcher playback took ownership. The session launcher also attenuated the first 256 samples, making the first warped transient sound softer than subsequent loop transients.

## Impact

Session loops are seamless even when arrangement clips exist, pasted warped clips retain visible and functional warp markers, Back to Arrangement activates correctly, and the first launched transient retains its attack.

## Validation

- transient-focused tests: 627 assertions across 2 cases
- Signalsmith tests: 156 assertions across 4 cases
- clip synchronization and arrangement/launcher regression coverage
- full `magda_daw_app` Debug build
- repository pre-commit hooks

### DIFF
--- a/magda/daw/audio/session/ClipSynchronizer.cpp
+++ b/magda/daw/audio/session/ClipSynchronizer.cpp
@@ -142,21 +142,46 @@ bool syncFollowActionToTracktionClip(te::Clip& teClip, const ClipInfo& clip, dou
     return changed;
 }
 
-void restoreWarpMarkers(te::WarpTimeManager& warpManager,
-                        const std::vector<ClipInfo::WarpMarker>& markers) {
-    if (markers.empty())
-        return;
+constexpr double warpMarkerSyncEpsilonSeconds = 1.0e-6;
+
+bool warpMarkerMapsMatch(const juce::Array<te::WarpMarker*>& engineMarkers,
+                         const std::vector<ClipInfo::WarpMarker>& savedMarkers) {
+    if (engineMarkers.size() != static_cast<int>(savedMarkers.size()))
+        return false;
+
+    for (size_t i = 0; i < savedMarkers.size(); ++i) {
+        const auto* engineMarker = engineMarkers[static_cast<int>(i)];
+        const auto& savedMarker = savedMarkers[i];
+        if (engineMarker == nullptr ||
+            std::abs(engineMarker->sourceTime.inSeconds() - savedMarker.sourceTime) >
+                warpMarkerSyncEpsilonSeconds ||
+            std::abs(engineMarker->warpTime.inSeconds() - savedMarker.warpTime) >
+                warpMarkerSyncEpsilonSeconds) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool restoreWarpMarkersIfNeeded(te::WarpTimeManager& warpManager,
+                                const std::vector<ClipInfo::WarpMarker>& markers) {
+    // A valid map needs both boundary markers. In particular, do not insert a
+    // lone saved marker on top of TE's recreated boundaries: that can create a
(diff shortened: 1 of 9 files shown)


magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -142,6 +142,66 @@ bool syncFollowActionToTracktionClip(te::Clip& teClip, const ClipInfo& clip, dou
...
+    }
+
+    if (!clip.warpMarkers.empty()) {
+        auto& warpManager = audioClip.getWarpTimeManager();
+        const auto& existingMarkers = warpManager.getMarkers();
+
+        // A newly-created WarpTimeManager contains only its two default
+        // boundary markers. Replace those with the marker map copied from the
+        // source clip, but never overwrite an already-live edited map.
+        if (existingMarkers.size() <= 2) {
```
lucaromagnoli: Restore loop: since enableWarp now eagerly stores markers, every warp-enabled clip with no user markers carries a saved map of exactly 2 (the boundaries). Restoring a 2-marker map leaves TE at 2 markers, so this guard re-fires on every subsequent syncClipPropertyToEngine call:

- restoreWarpMarkers runs removeAllMarkers + 2x moveMarker through the edit UndoManager on every gain/pan/transpose tweak: spurious undo steps plus TreeWatcher restarts
- `changed = true` unconditionally, so reallocateAndNotify() rebuilds the graph per property change, including per drag tick on a gain drag

The pre-existing arrangement-path block (line 1971) has the same `<= 2` pattern; it was dormant because warpMarkers was only populated at save time, but the eager store makes it live too.

Suggest comparing saved vs existing markers (size + epsilon on source/warp times) and skipping the restore when identical. That kills the loop at both call sites.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -142,6 +142,66 @@ bool syncFollowActionToTracktionClip(te::Clip& teClip, const ClipInfo& clip, dou
...
+void restoreWarpMarkers(te::WarpTimeManager& warpManager,
+                        const std::vector<ClipInfo::WarpMarker>& markers) {
+    if (markers.empty())
+        return;
+
+    // removeAllMarkers deliberately recreates TE's start/end boundaries, so
+    // inserting the complete saved list would duplicate both boundaries.
+    warpManager.removeAllMarkers();
+
+    if (markers.size() == 1) {
```
lucaromagnoli: Minor: removeAllMarkers just recreated both boundary markers, so this branch inserts the saved marker on top of them, giving 3 markers. If the saved marker duplicates a boundary you get a zero-length segment, and moveMarker's stretch-ratio math then divides by srcLen == 0. Unreachable with the current save code (always stores >= 2), so an early return or jassert would do.

lucaromagnoli (commented):
Two findings, one real and one minor. Everything else checked out: the boundary move-then-insert-middles restore is safe against TE's stretch-ratio clamping for any valid saved map, the paste condition preserves prior cross-view behavior for non-warped clips, and the listener wiring in MainWindow is guarded.

PR #1754, opened 2026-07-14
Fix/audio warp bugs

lucaromagnoli:


magda/daw/core/ParameterUtils.cpp
```diff
@@ -230,6 +230,15 @@ ParameterNormalizedValue modelToNormalizedValue(ParameterModelValue model,
     return ParameterNormalizedValue::clamped(model.value);
 }
 
+float modelToTeValue(ParameterModelValue model, const ParameterInfo& info) {
+    const float teSpan = info.teMaxValue - info.teMinValue;
+    if (teSpan <= 0.0f)
+        return model.value;
+
+    const auto normalized = modelToNormalizedValue(model, info);
+    return info.teMinValue + normalized.value * teSpan;
```
lucaromagnoli, line 242: modelToNormalizedValue honors ParameterScale::Logarithmic and anchor skew, but this line projects that normalized position linearly into the TE range. For a param whose model range equals its TE range but is log-scaled (say 20-20000 Hz), 440 would come out near 9 kHz; the old raw write was correct for that shape. Current pad devices only route 0..1 host slots through here, so nothing breaks today, but the helper reads as general-purpose. Suggest an early return of model.value when the model and TE ranges match: matching ranges mean the TE native value is already in display units, and it makes the pass-through test exact by construction.

magda/daw/core/AutomationCommands.cpp
```diff
@@ -454,32 +521,89 @@ void enableModLinks(const std::vector<BakeModulationCommand::ModLinkRef>& refs)
...
 
+    const auto* lane = autoMgr.getLane(laneId_);
+
+    if (!capturedLaneAuthority_) {
+        previousLaneDisabled_ = isAutomationPersistentlyDisabled(lane->authorityState);
+        capturedLaneAuthority_ = true;
+    }
+
     createdClipId_ = autoMgr.createClip(laneId_, startBeat_, endBeat_ - startBeat_);
     if (createdClipId_ == INVALID_AUTOMATION_CLIP_ID)
```
lucaromagnoli: If createClip fails, this early return leaves the lane preparation applied (retype, visibility) and skips exitAllLinkModes, and the inert command stays in undo history. Undo still restores previousLaneState_, so it is recoverable, but rolling back via restoreBakeAutomationLaneState on this path would match the other failure exits in performModulationBake.

magda/daw/audio/WarpMarkerManager.cpp
```diff
@@ -71,16 +72,27 @@ void WarpMarkerManager::setTransientSensitivity(
...
+    pending.engineId = (it != clipIdToEngineId.end()) ? it->second : std::string{};
+    pending.edit = &edit;
 
-    // Restart immediately so the latest sensitivity doesn't wait behind an old
-    // detection job. WarpTimeManager::detectTransients detaches/cancels the
-    // previous job before submitting the replacement.
-    applySensitivityNow(edit, det.engineId, clipId, sensitivity);
+    // A drag can generate dozens of values per second. Restart the quiet-period
+    // timer so only the final value clears the cache and launches detection.
+    startTimer(150);
```
lucaromagnoli, line 83: Nit: this is one shared timer, so every call restarts the 150 ms window for all clips. A continuing drag on clip A postpones clip B's pending detection indefinitely. One slider at a time makes this academic; a line in the header comment would cover it.

magda/daw/ui/components/chain/custom_ui/PolySynthUI.cpp
```diff
@@ -256,6 +258,22 @@ PolySynthUI::~PolySynthUI() {
             btn->setLookAndFeel(nullptr);
 }
 
+void PolySynthUI::setLivePlugin(magda::daw::audio::compiled::MagdaPolySynthCompiledPlugin* plugin) {
+    if (livePlugin_ == plugin)
+        return;
+
+    livePlugin_ = plugin;
```
lucaromagnoli, line 265: Nit: the raw livePlugin_ pointer plus a 60 Hz timer relies on detachFromLivePlugin always running before the plugin dies. Same pattern as struckUI and nimbusUI, so it is consistent; it just inherits the same lifetime assumption.

lucaromagnoli (commented):
Review findings: one real edge case in modelToTeValue worth fixing before merge, one recoverable early-return gap in the clip bake, two nits. Verified separately: cancelEditing re-entry is safe (editor cleared before onEditCancel), the debounced detection still refreshes the waveform via TransientCacheListener, the deferred Edit* cannot dangle (WarpMarkerManager dies with AudioBridge and stops its timer), and the MSVC /O2 genex only triggers a D9025 warning since Debug flags omit /RTC1.

Issue #1746, opened 2026-07-12
Unify arrangement/session playback ownership and loop/retrigger semantics


lucaromagnoli:
## Context

While integrating Signalsmith Stretch in #1590, repeated session-loop and retrigger failures exposed a broader playback architecture problem. The visible artifacts were fades, gaps, doubled attacks, and retriggers whose UI playhead restarted while audio did not reliably restart.

The individual failures have local fixes, but the underlying problem is that playback ownership and discontinuity semantics are distributed across several layers which can disagree.

### 1. Arrangement/session output ownership has two sources of truth

`ArrangerLauncherSwitchingNode` has access to the audio-thread `LaunchHandle` state, but arrangement mixing is also controlled by the message-thread `AudioTrack::playSlotClips` property.

Those states can temporarily disagree. We observed the arranger and launcher paths both reaching the track output while a session clip was active, producing doubled attacks and phase/comb-filter artifacts that sounded like a fade at loop boundaries.

Both paths may need to be processed for readiness or transitions, but exactly one path should own the audible output outside an explicit transition.

### 2. Audio-loop ownership is duplicated

Looping stretched audio currently involves several independent mechanisms:

- `LaunchHandle::setLooping()` splits the session timeline at each cycle.
- `SlotControlNode` processes the split as separate sections and changes local offsets.
- `BeatRangeReader` maps unbounded beats into a loop range.
- `LoopReader` wraps the source samples.
- `TimeStretchReader` owns algorithm history plus input/output FIFOs.

The same musical loop is therefore represented as both a launch-state restart and a circular audio-source operation. Boundary rounding can leave a zero-frame head section, causing one layer to see a natural wrap while another sees a seek.

### 3. Semantic events are inferred from positions

Reader code currently infers `natural loop`, `retrigger`, and `seek` from continuity flags and floating-point position changes. During testing:

- a natural wrap was interpreted as a full stretcher reset;
- a quantized retrigger near zero was interpreted as a natural loop wrap;
- sub-sample boundary error lost the wrap event between adjacent sections;
- generic seek crossfades softened deliberate retrigger transients.

These are semantic transport events and should be carried explicitly through the graph rather than reconstructed from beat/sample positions.

### 4. Configured and instantiated stretch modes can diverge

A session clip could report SoundTouch HQ (`mode=4`) while its newly constructed audio graph actually used Signalsmith (`mode=15`). Mode resolution happened in more than one place and creation did not consistently apply the model selection.

[...]

PR #1755, opened 2026-07-14
Dev/0.15.0

lucaromagnoli:


magda/daw/audio/insert_capture/InsertRenderCaptureService.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+        if (!tap->startCapture(file, startSec, endSec, sampleRate)) {
+            tapPlugin->deleteFromParent();
+            continue;
+        }
+        taps->plugins.push_back(tapPlugin);
+        taps->files.push_back(file);
+    }
+
+    if (taps->plugins.empty())
```
lucaromagnoli: This only rejects the setup when every tap failed. With multiple routed inserts, failure to create/arm one tap is skipped above, so the pass reports success as long as another tap exists; the offline export then silently loses the uncaptured insert return. Capture setup should be all-or-nothing: if any qualifying insert cannot get a tap, remove the taps already created and fail the pass.

magda/daw/ui/windows/MainWindowExport.cpp
```diff
@@ -352,39 +353,119 @@ void MainWindow::performExport(const ExportAudioDialog::Settings& settings,
...
+                    if (success) {
+                        launchRender();
+                    } else {
+                        captureService->cleanupAfterRender();
+                        juce::AlertWindow::showMessageBoxAsync(juce::AlertWindow::InfoIcon,
+                                                               tr("export.alert.cancelled_title"),
+                                                               tr("export.alert.cancelled_body"));
+                    }
+                });
+            if (!started) {
```
lucaromagnoli: `startCapturePass()` can return `false` because tap/plugin or temporary-file setup failed (and this branch is already gated by `exportNeedsCapturePass()`). Continuing with `launchRender()` then performs the exact offline render that cannot produce the hardware return, so the export can silently contain missing or dry audio. Please abort the export and surface a capture-setup error here instead.

lucaromagnoli (commented):
Full review of the 0.15.0 aggregate. Inline comments carry the file-specific findings (5 major, the rest minor/nit). Items that could not be anchored because the lines are outside the diff hunks:

- `AutomationPlaybackEngine.cpp:288` (clearStaleTarget): manual-value restore uses `setParameter`, which TE drops when a modifier is active. The line predates this PR but the new authority design keeps modifiers attached on automated parameters, so it is now reachable: delete a volume lane while an LFO is linked and the fader stays pinned instead of returning to the manual value. Should be `setParameterFromHost`.
- `TrackManager::deleteTrack`: deleting a sidechain SOURCE track never clears `DeviceInfo::sidechain.sourceTrackId` on destination tracks/master. The device header keeps showing an engaged source, the stale id is serialized, and the LFO just never triggers again. Sends and track-input routing are cleaned here; sidechain configs need the same sweep.
- `AutomationCurveEditor.cpp:834-869` (onStepStamped): gatherContext compares clip-local `p.beatPosition` against editor-space `gridStart`/`gridEnd` while the subsequent `onPointAdded` calls correctly subtract `clipOffset_`; in the non-looped clip editor a step stamp can add a duplicate recovery point on an existing right-edge point and flip the wrong segment shape.
- Ghost-group perf: every property notification deep-copies all shared content to every sibling (including per-tick drag changes), and `isGhostClip`/`getLinkGroupIndex` are O(total clips) called 2-3x per clip paint, so full repaints are O(n^2) in clip count. Message-thread only, but worth a cheap map.
- `ClipComponent.cpp:3535` menu item 27: multi-selection Enable/Disable flips each clip individually, so a mixed selection ends up inverted rather than uniformly set (may be intended).
- `AutomationInfo.hpp:169` getEndLocalBeat: exact `local == 0.0` fmod compare; non-representable length/loop ratios can return ~1e-16 and hold the cycle-start value across the following gap.
- Stale comments: `InsertCapturePlugin.hpp:82` references a nonexistent `InsertFreezeService`; `InsertCapturePlugin.cpp:50` claims the service aborts on device-rate change mid-capture but no such abort exists.
- Test gaps: no flattenClipLane overlap test (see the ClipLaneFlattener comment), no Touching/Writing -> Disable transition test, no playback-engine authority-transition test.

TE submodule bump (7cd99dc..b4976d16) verified: the setParameterFromHost patch is intact at tip, `Mode::signalsmith = 15` is appended without renumbering persisted modes, and the #1746 seam instrumentation was removed. One TE-side note: WarpReader::setPosition now ignores any seek within 10 samples of the current position, which also swallows genuine sub-10-sample nudges on warped clips.

Clean: audio-thread safety at every new applyToBuffer/capture/playback site, beats-domain and getCurrentBaseValue/setParameterFromHost conventions at the new sites (exceptions inline), chain power via TrackChain::enabled without touching device bypass flags, master-track guards, momentary-button release paths, GestureRouter gesture pairing, delta solo mutual exclusion, and the per-suite JUCE test isolation. The [MidiRec] writeToLog calls are deliberate message-thread field diagnostics, not leftovers.

The two existing comments on the capture-pass partial-failure path stand; not repeated here.

magda/daw/core/ClipLaneFlattener.cpp
```diff
@@ -0,0 +1,106 @@
...
+              });
+
+    std::vector<AutomationPoint> out;
+
+    // src carries curve metadata (type / tension / handles) into the bake
+    // loop's tessellation; the value is always the lane-level resolved one.
+    const auto emit = [&](double beat, const AutomationPoint* src) {
+        if (beat < 0.0)
+            return;
+        if (!out.empty() && beat <= out.back().beatPosition)
```
lucaromagnoli, line 85: **major:** The monotonic guard refuses any breakpoint at or before `out.back().beatPosition`, so a front-priority clip that starts later than a longer underlying clip contributes zero points in the overlap. BakeModulationToClipCommand creates exactly this shape (baked clip moved to the front of clipIds over an existing longer clip): the UI shows the baked curve via clipIds-order precedence, but the flattened TE curve has no points inside the baked range and Linear segments get no tessellation, so the fresh bake is inaudible at playback. Front-priority overlap needs to trim the underlying clip's points instead of refusing the winner's; there is currently no flattenClipLane overlap test.

magda/daw/core/ClipLaneFlattener.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+        // Anchor the clip start when the first audible point sits later
+        // (interpolatePoints holds the first point's value before it).
+        if (clip->points.front().beatPosition > kEdgeEpsilon)
+            emit(start, nullptr);
+
+        for (double base = 0.0; base < clip->lengthBeats; base += cycle) {
+            // Loop wrap: hold the cycle's outgoing value, then jump into the
+            // next iteration.
+            if (base > 0.0)
```
lucaromagnoli, line 144: **nit:** The unroll loop has no iteration cap: a tiny positive loopLengthBeats (e.g. 0.001) on a long clip emits millions of breakpoints on the message thread. Zero/negative cycle is guarded; cap the iteration count or floor the cycle length.

magda/daw/core/ClipManager.cpp
```diff
@@ -2909,18 +3050,30 @@ void ClipManager::notifyClipPropertyChanged(ClipId clipId) {
     if (auto* clip = getClip(clipId))
         syncActiveMidiTake(*clip);
 
+    // Ghost clips: every content edit funnels through here, so mirror the
+    // shared fields to the link-group siblings and notify them in the same
+    // pass (no recursion — siblings are notified directly below/batched).
+    const auto siblings = propagateLinkGroupContent(clipId);
```
lucaromagnoli, line 3120: **major:** Flattening a looped ghost clip funnels through this propagation and copies the unrolled note list to every sibling, but loopEnabled/loopLengthBeats/midiOffset are per-instance and stay intact, so a still-looping sibling now loops the first cycle of the unrolled list: with nonzero phase its audible pattern rotates, and seam-crossing notes are permanently truncated. splitClipAtBeat's MIDI branch explicitly detaches both halves for exactly this hazard; FlattenMidiClipCommand needs the same detach (or a ghost guard on the flatten menu item).

magda/daw/core/ClipManager.cpp
```diff
@@ -3214,7 +3373,13 @@ std::vector<ClipId> ClipManager::pasteFromClipboardBeats(double pasteBeat, Track
             // Copy properties
             auto* newClip = getClip(newClipId);
             if (newClip) {
-                newClip->name = clipData.name + " (copy)";
+                // Ghost membership survives whole-clip copy/paste. Range
+                // copies cleared it at copy time, and clearAllClips scrubs the
+                // clipboard so cross-project pastes come in unlinked. Ghosts
+                // keep the source name (they are the same content).
+                newClip->linkGroupId = clipData.linkGroupId;
```
lucaromagnoli, line 3444: **major:** Pasting a whole-clip copy of a ghost rejoins the link group carrying the stale clipboard snapshot. Scenario: copy ghost A; edit sibling B (mirrors to A); paste C; move any note on C and propagateLinkGroupContent pushes the pre-copy note list over A and B, reverting the intervening edit. restoreClip guards this exact case by adopting the live group's content on rejoin; the paste path needs the same adoption after setting linkGroupId.

magda/daw/core/ClipManager.cpp
```diff
@@ -3214,7 +3373,13 @@ std::vector<ClipId> ClipManager::pasteFromClipboardBeats(double pasteBeat, Track
             // Copy properties
             auto* newClip = getClip(newClipId);
```
lucaromagnoli, line 3438: **minor:** Paste copies volume/pan/fades/grid/warp/channels but not `ClipInfo::enabled`, so a disabled clip pastes as enabled. Duplicate and split both preserve the flag; paste is the one lossy path.

magda/daw/core/ClipCommands.cpp
```diff
@@ -1930,7 +2185,34 @@ void BounceInPlaceCommand::execute() {
...
+    if (trackNeedsInsertCapture(*teTrack)) {
+        if (!runInsertCapturePass(*engine_, bounceRange.startSeconds,
+                                  bounceRange.endSeconds + bounceTailSeconds)) {
+            restoreTransport();
+            return;
+        }
+        captureScope.service = engine_->getInsertRenderCaptureService();
+
+        // The pass ran the live transport; listener callbacks may have
+        // invalidated the model/engine clip pointers — re-resolve.
```
lucaromagnoli, line 2216: **minor:** This bail-out returns without setting `errorMessage_`, and MainWindow only toasts when getErrorMessage() is non-empty, so the user sits through the real-time capture pass and then nothing happens, with no message. Set errorMessage_ like every other failure branch in this function.

magda/daw/audio/plugins/InsertCapturePlugin.cpp
```diff
@@ -0,0 +1,203 @@
...
+void InsertCapturePlugin::applyToBuffer(const te::PluginRenderContext& fc) {
+    if (fc.destBuffer == nullptr || fc.bufferNumSamples <= 0)
+        return;
+
+    // Playback mode: substitute the captured return during an offline render.
+    // The isRendering gate keeps file reads off the live audio thread.
+    if (auto* reader = activeReader_.load(std::memory_order_acquire); reader != nullptr) {
+        if (!fc.isRendering)
+            return;  // live graph: transparent passthrough
+        const auto m = mapBlockToCaptureWindow(
```
lucaromagnoli, line 147: **major:** The capture wav is written at the live device rate, but the offline render runs at the user-selected export rate (44.1/48/96/192k in ExportAudioDialog). This playback branch maps block times with `reader->sampleRate` and reads file samples 1:1 into render-rate blocks with no resampling and no guard: device 48k + export 44.1k means the per-block length exceeds blockNumSamples, gets clamped, and ~8% of file samples are silently skipped between blocks (the inverse leaves silent tails in every block). The export completes with garbled hardware audio and no error. Either resample here, force the capture pass to the export rate, or refuse the mismatch up front.

magda/daw/audio/plugins/InsertCapturePlugin.cpp
```diff
@@ -0,0 +1,203 @@
...
+    for (int c = 0; c < kNumChannels; ++c)
+        chans[c] = c < srcChannels
+                       ? fc.destBuffer->getReadPointer(c, fc.bufferStartSample + bufferOffset)
+                       : zeroBuf_.data();
+
+    writer->write(chans, numSamples);
+    nextFileSample_ += numSamples;
+    samplesWritten_.store(nextFileSample_, std::memory_order_relaxed);
+
+    if (fc.editTime.getEnd().inSeconds() >= windowEnd_)
```
lucaromagnoli, line 228: **major:** `ThreadedWriter::write` returns false when the FIFO is full (slow or stalled disk) and writes nothing, but this call and the writeSilence path above ignore the return and advance `nextFileSample_` anyway. The sequential-write reconciliation then believes those samples exist, so everything after the drop lands early: the rest of the captured return is time-shifted and the export still reports success. Track refused writes and fail the pass (or at least surface an error).

magda/daw/audio/plugins/InsertCapturePlugin.cpp
```diff
@@ -0,0 +1,203 @@
...
+    samplesWritten_.store(0, std::memory_order_relaxed);
+    complete_.store(false, std::memory_order_relaxed);
+    activeWriter_.store(threadedWriter_.get(), std::memory_order_release);
+    return true;
+}
+
+void InsertCapturePlugin::stopCapture(bool keepFile) {
+    if (threadedWriter_ == nullptr)
+        return;
+
```
lucaromagnoli, line 105: **minor:** The audio-thread handoff rests on a fixed 80 ms sleep after clearing `activeWriter_`: an in-flight applyToBuffer that loaded the pointer just before the store and whose wall time exceeds 80 ms hits a use-after-free in write(). Practically mitigated because the transport is stopped first, but it is a timing assumption rather than a guarantee, and removeTaps() calling this per tap freezes the message thread 80 ms x N during cancel/cleanup.

magda/daw/audio/ExternalInsertDeviceEnablement.hpp
```diff
@@ -0,0 +1,44 @@
...
+  public:
+    explicit ExternalInsertDeviceEnablement(tracktion::engine::Edit& edit);
+
+    /** Derive pass. Returns true when any device's enablement changed (the
+        caller should reallocate the playback graph). */
+    bool refresh();
+
+  private:
+    tracktion::engine::Edit& edit_;
+    std::set<juce::String> autoEnabledInputs_;
```
lucaromagnoli, line 58: **minor:** These sets are in-memory only while TE persists device enablement in engine-global settings, so a port auto-enabled in session A stays enabled after a restart or crash; the fresh instance then sees it as enabled-at-startup and never auto-disables it. The enable/disable round-trip only holds within one process lifetime, and the feature permanently mutates the user's Audio Settings across sessions. Worth persisting the auto-enabled set so restore survives a restart. Also nit: keying by device name aliases identically named ports on duplicate interfaces.

docs/EXTERNAL_INSERT_PLAN.md
```diff
@@ -0,0 +1,205 @@
+# External FX & External Instrument — Integration Plan
+
+Status: design, pre-implementation. Branch: `feat/external-instrument-fx`.
```
lucaromagnoli: **minor:** Committed handoff/planning material with a stale header: this says 'Status: design, pre-implementation' while the phasing section marks everything DONE, the open question about a DeviceProcessor factory is resolved by shipped code, and Phase 5 claims freeze round-trip tests that do not exist (shipped tests are registry + capture-window math). The HANDOFF file likewise references the pre-merge feature branch. I would drop both files from the repo; if they stay, the status header and stale claims need correcting.

lang/ja.json
```diff
@@ -105,6 +105,7 @@
     "mute.tooltip": "ミュート",
     "solo.tooltip": "ソロ",
     "record.tooltip": "録音待機",
+    "enable.tooltip": "トラックの有効/無効",
```
lucaromagnoli, line 108: **major (process):** ja/ru/zh arrive with hand-authored translations for the new `enable.tooltip` and `export.capture.*` keys, and this file also rewords the existing `no_receives` value. Non-English locale files are Crowdin-owned: committed locale edits collide with the next sync and can silently lose translations on diverging keys. Keep only en.json changes in the repo and push these via `crowdin upload translations` instead.

magda/daw/audio/midi/MidiInputRouter.cpp
```diff
@@ -313,6 +313,81 @@ void MidiInputRouter::removeSurfaceOnlyMidiInputTargets() {
...
+                tokens.remove(i);
+        }
+        return tokens.joinIntoString(" ").toLowerCase();
+    };
+    const auto stemA = stripDirection(a);
+    const auto stemB = stripDirection(b);
+    if (stemA.isNotEmpty() && stemA == stemB)
+        return true;
+
+    if (!a.containsAnyOf("0123456789") && !b.containsAnyOf("0123456789")) {
```
lucaromagnoli: **major:** Tier-3 'same first word' matching treats any two same-brand ports as the same hardware. With an External Instrument sending to 'Arturia MicroFreak', a separate 'Arturia BeatStep Pro' controller (no digits, shared first word, length >= 3) is classified as the sendback port and removed from All-Inputs routing below, so the controller silently stops playing that track with no UI indication. Tiers 1-2 are sound; this tier should require more than a shared brand prefix (e.g. also match the model token).

magda/daw/audio/modifiers/ModifierHelpers.hpp
```diff
@@ -167,15 +172,31 @@ inline void applyLFOProperties(te::LFOModifier* lfo, const ModInfo& modInfo,
...
+    // hold (or a free-running loop), and the trigger path resyncs them.
     //
     // Audio-triggered LFOs are still gated by the audio sidechain path so
     // one-shots can continue through release while normal loops close on
     // the audio gate.
-    lfo->setGateOnTriggerSource(modInfo.triggerMode == LFOTriggerMode::MIDI);
-    if (modInfo.triggerMode == LFOTriggerMode::MIDI)
+    lfo->setGateOnTriggerSource(modInfo.triggerMode == LFOTriggerMode::MIDI &&
+                                !crossTrackSidechain);
+    if (crossTrackSidechain)
```
lucaromagnoli: **minor:** The in-place path forces setGated(false) on ALL cross-track LFOs, including Audio-triggered ones, even though the comment says the audio sidechain path owns that gate. createModifier restores setGated(true) for Audio mode after calling this; the in-place caller (ModifierSync::updateDeviceModifierProperties) does not. Tweak any mod property on an audio-triggered LFO that was gated closed and the gate is forced open: the LFO free-runs and modulates the destination until the next source level drop re-gates it. Mirror the Audio-mode restore on the in-place path.

magda/daw/audio/plugins/SidechainPlugin.cpp
```diff
@@ -0,0 +1,181 @@
...
+    // ramp upward (recovery) steps over the measured stair width so the
+    // recovery is continuous at source; ramp downward steps (the duck hit)
+    // within one block so the onset stays punchy. The attack/release poles
+    // then only shape, they no longer have to hide discontinuities.
+    if (target != lastTarget_) {
+        constexpr int kMaxStairSamples = 4096;
+        const bool goingDown = target < rampValue_;
+        const int width =
+            goingDown ? fc.bufferNumSamples
+                      : juce::jlimit(fc.bufferNumSamples, kMaxStairSamples, samplesSinceChange_);
```
lucaromagnoli: **minor:** Three edges around the stair-stepping: (1) `jlimit(fc.bufferNumSamples, kMaxStairSamples, ...)` asserts lower <= upper, so a block size over 4096 (offline render, some hosts) jasserts on every upward gain hop in debug builds; (2) `samplesSinceChange_` increments unboundedly and is signed, which is UB overflow after ~13.5 h at 44.1k if the target never changes; clamp it at kMaxStairSamples; (3) the fixed `channels[8]` array a few lines up leaves channels 8+ completely unducked on wider buses.

magda/daw/ui/panels/content/AutomationClipEditorContent.cpp
```diff
@@ -0,0 +1,708 @@
...
+        if (clip == nullptr || !clip->looping)
+            return;
+        timeRuler_->setLoopRegion(0.0, juce::jmax(0.0, displayEnd), true, true);
+    };
+    timeRuler_->onLoopDragEnded = [this](double displayStart, double displayEnd) {
+        juce::ignoreUnused(displayStart);
+        const auto* clip = getClip();
+        if (clip == nullptr || !clip->looping)
+            return;
+        double bpm = 120.0;
```
lucaromagnoli: **minor:** This round-trips beats through the seconds-based TimeRuler and back (`lengthBeats = displayEnd * bpm / 60.0`), with bpm falling back to 120 when TimelineController::getCurrent() is null; the same pattern repeats at :343 and :593-650. If the controller is unavailable or the tempo changes mid-gesture the committed loop length is scaled by the wrong tempo. Per the beats-authoritative rule the editor should stay in beats end to end instead of converting through ruler seconds.

magda/daw/ui/components/chain/params/ModulationBakeAction.cpp
```diff
@@ -0,0 +1,201 @@
...
+double midiContentEndBeats(const magda::ControlTarget& target, double bpm) {
+    if (target.isEditScoped())
+        return 0.0;
+    auto& clipManager = magda::ClipManager::getInstance();
+    double end = 0.0;
+    for (magda::ClipId clipId : clipManager.getClipsOnTrack(target.devicePath.trackId)) {
+        const auto* clip = clipManager.getClip(clipId);
+        if (clip == nullptr || !clip->isMidi() || clip->view == magda::ClipView::Session)
+            continue;
+        end = juce::jmax(end, clip->getStartBeats(bpm) + clip->getTimelineLength(bpm) * bpm / 60.0);
```
lucaromagnoli: **nit:** `getStartBeats(bpm) + getTimelineLength(bpm) * bpm / 60.0` is a seconds round-trip where `clip->getEndBeats(bpm)` already exists; numerically identical at fixed bpm, but it silently degrades via the stale seconds cache when placement.lengthBeats <= 0.

magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
```diff
@@ -1115,6 +1115,18 @@ void TrackHeadersPanel::trackDevicesChanged(TrackId trackId) {
     }
 }
 
+void TrackHeadersPanel::devicePropertyChanged(const magda::ChainNodePath& devicePath) {
```
lucaromagnoli, line 1118: **minor:** This now triggers updateTrackHeaderLayout() plus a full-panel repaint for ANY device property change on a visible track, but only External Insert routing changes need it. Dragging a device gain slider notifies per mouse tick, so gestures re-layout every header continuously on large projects. Gate it on the device being an External Insert (or on the routing-relevant property).

magda/daw/audio/automation/ModulationBaker.cpp
```diff
@@ -0,0 +1,155 @@
...
+            // sampling density doesn't need to track tempo automation.
+            const double bps = tempoMap.bpmAt(opts.startBeat) / 60.0;
+            const double rate = std::max(1.0e-6, static_cast<double>(source.mod.rate));
+            cycleBeats = bps / rate;
+        }
+        step = std::min(step, cycleBeats / std::max(1, opts.samplesPerCycle));
+    }
+    if (!anyBakeable)
+        return {};
+    step = std::max(step, opts.minStepBeats);
```
lucaromagnoli: **minor:** The flat 1/128-beat step floor overrides samplesPerCycle for fast LFOs: a triplet 32nd cycle (1/12 beat) gets ~10 samples per cycle instead of 64, so fast bakes come out visibly faceted after RDP simplification. Scale the floor by cycle length instead.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -1097,9 +1234,16 @@ void ClipSynchronizer::configureSessionAutoTempo(te::WaveAudioClip* audioClip,
...
     syncAudioSourceInterpretationToLoopInfo(*audioClip, *clip);
 
-    // Ensure valid stretch mode (autoTempo requires time-stretching)
-    if (audioClip->getTimeStretchMode() == te::TimeStretcher::disabled)
-        audioClip->setTimeStretchMode(te::TimeStretcher::defaultMode);
+    // Auto-tempo requires stretching, but it must still honour the user's
+    // explicit SoundTouch/Signalsmith selection. Fresh slot clips otherwise
+    // retain TE's default mode, making the old modes sound identical to
+    // Signalsmith until a later property edit happens to rebuild the graph.
+    auto desiredMode = static_cast<te::TimeStretcher::Mode>(clip->timeStretchMode);
```
lucaromagnoli, line 1253: **minor:** For autoTempo session clips this applies the stretch-mode change before the property block at :519 runs, so that block's mode-changed check never fires and needsGraphReallocation / setUsesProxy(false) are skipped on a mid-playback engine switch, contradicting that block's own comment. It works today only because TE's TreeWatcher restart list happens to include elastiqueMode; if that implicit restart is ever coalesced away, switching SoundTouch -> Signalsmith on a playing session clip goes silent until the next rebuild. Worth making the explicit rebuild request unconditional.

magda/daw/engine/TracktionEngineWrapperTransport.cpp
```diff
@@ -96,6 +96,36 @@ void TracktionEngineWrapper::stop() {
...
+        // note-off cleanup from the graph teardown (resetSynthsOnTrack only
+        // covers internal synths), so a note-on in flight at stop rings
+        // forever. Flush all-notes-off through every insert's MIDI send.
+        sendAllNotesOffToExternalInserts();
+    }
+}
+
+void TracktionEngineWrapper::sendAllNotesOffToExternalInserts() {
+    if (!currentEdit_ || !engine_)
+        return;
```
lucaromagnoli, line 110: **nit:** sendAllNotesOffToExternalInserts skips inserts where `!insert->isEnabled()`, so disabling an External Instrument insert while hardware notes are sounding and then stopping leaves the synth ringing. Also missing a `break` after the device-name match (harmless, just extra scanning).

PR #1829, opened 2026-07-24
Break DAW-agent dependency cycle

### LINKED ISSUE #1828: Break circular dependency between magda_daw and magda_agents
## Problem

`magda_daw` and `magda_agents` form a circular dependency at the static-library level. CMake only tolerates it because both archives are linked together into the final `magda_daw_app` executable, where the undefined symbols resolve. Neither library can build, test, or be reasoned about in isolation.

## The cycle

**agents -> daw** (`magda/agents/CMakeLists.txt`)
- `magda_agents` `PUBLIC`-links `magda_daw`.
- 35 of 36 agent sources `#include "../daw/..."`, reaching deep into `core/`, `api/`, and `engine/` (e.g. `TracktionEngineWrapper`, `TrackManager`, `AliasRegistry`, `magda_api`).

**daw -> agents** (closes the loop)
- `magda/daw/core/MixAnalysisService.hpp:10` includes `agents/mixing_agent.hpp` for `MixAnalysisAgent::Input`.
- `magda/daw/audio/analysis/OfflineMixAnalysis.hpp` and `MixAnalysisInput.hpp` do the same.
- Those `.cpp`s are compiled into the `magda_daw` archive via `target_sources(magda_daw PRIVATE ...)`, so the DAW library genuinely depends on an agent header.

## Why it matters

- Breaks layering: the AI layer and the DAW core are mutually entangled instead of AI sitting cleanly on top.
- `magda_agents` cannot be built or unit-tested standalone.
- Blocks ever extracting the agent layer (own target boundary, optional/separately-licensed build, or eventually its own repo).
- Makes incremental builds and dependency reasoning worse than they need to be.

## Fix direction

The DAW should never depend on an agent type. The only thing pulling agents back into the DAW is the `MixAnalysisAgent::Input` measured-data struct.

- Move the shared measured-input data type out of `mixing_agent.hpp` into a neutral location the DAW already owns (DAW core or a small shared header). The mixing agent then consumes the DAW-owned type.
- After that, drop `magda_agents` from `magda_daw`'s effective dependencies and make `magda_agents` a strict upper layer: agents depend on `magda_daw` / `magda_api`, DAW depends on neither.
- Add a check (or at least a note) so no future DAW source reintroduces an `agents/` include.

## Acceptance

- No source compiled into the `magda_daw` archive includes any `agents/` header.
- `magda_agents` links on top of `magda_daw` with no back-reference.
- Build stays green (`magda_daw_app`, tests).

lucaromagnoli:


lucaromagnoli (commented):
LGTM. This breaks the magda_daw <-> magda_agents cycle the right way: the measurement structs move into a DAW-owned `MixAnalysisData`, the DAW stops including any agent header, and the boundary is now enforced at configure time.

I checked that `OfflineMixAnalysis::start` has a single caller (`MixAnalysisService`), which already ran measure-only, so dropping the in-driver agent path is dead-code removal rather than a behavior change. Nothing still references the old `MixAnalysisAgent::TrackMix/Segment/MaskingPair`, `skipAgent`, or `Request::question`, and `tonalBandLabels()` is used only inside `mixing_agent.cpp`, so no DAW code pulls the agent header back in.

A few minor, non-blocking notes inline. The only one worth acting on is tightening the CMake include regex.

magda/daw/CMakeLists.txt
```diff
@@ -370,6 +370,34 @@ add_subdirectory(media_db)
...
+    if(IS_ABSOLUTE "${_magda_daw_source}")
+        set(_magda_daw_source_path "${_magda_daw_source}")
+    else()
+        set(_magda_daw_source_path "${CMAKE_CURRENT_SOURCE_DIR}/${_magda_daw_source}")
+    endif()
+    if(EXISTS "${_magda_daw_source_path}"
+       AND _magda_daw_source_path MATCHES "\\.(c|cc|cpp|cxx|h|hh|hpp|hxx)$")
+        file(READ "${_magda_daw_source_path}" _magda_daw_source_text)
+        if(_magda_daw_source_text
+           MATCHES "#[ \t]*include[ \t]*[<\"][^>\"]*agents/")
```
lucaromagnoli: This also matches paths like `.../subagents/...`, and any `#include "...agents/..."` sitting in a comment or string literal, either of which would raise a spurious FATAL_ERROR. No such path exists today, but anchoring to `[<"]agents/` would make it precise.

Worth a one-line note in the comment above that this is a best-effort text scan over the listed sources, not a real dependency-graph check, so a header pulled in transitively but not listed in SOURCES would not be caught.

magda/agents/mixing_agent.hpp
```diff
@@ -18,84 +19,23 @@ namespace magda {
...
-        // Continuity across analyses of the same mix (#886 memory): a compact
-        // block with the previous verdict + measured changes since, so the model
-        // builds on prior advice instead of repeating it. Empty on a first run.
+        MixAnalysisData measurements;
+        std::string question;
         std::string priorContext;
     };
 
+    // Labels for the macro bands in MixAnalysisData::Track::tonalDb, in order.
+    static const std::vector<std::string>& tonalBandLabels();
```
lucaromagnoli: Minor: these labels describe `MixAnalysisData::Track::tonalDb`, so conceptually they belong next to the data type rather than on the agent. It is only used inside `mixing_agent.cpp` today so nothing leaks the agent header, just flagging for consistency.

magda/daw/audio/analysis/OfflineMixAnalysis.cpp
```diff
@@ -253,13 +253,13 @@ class AnalysisJob : public juce::Thread {
             return err;
         }
 
-        MixAnalysisAgent::Input input;
+        MixAnalysisData input;
```
lucaromagnoli: Nit: `input` reads oddly now that this holds measured data rather than agent input. The audio test renamed its local to `measurements`; matching that here would be tidier.

magda/daw/audio/analysis/MixAnalysisInput.cpp
```diff
@@ -207,18 +207,18 @@ std::vector<SectionBound> autoSections(int total, int n) {
...
-MixAnalysisAgent::Input MixAnalysisInput::build(double sr, const std::vector<Source>& tracks,
-                                                const juce::AudioBuffer<float>* masterIn,
-                                                const std::vector<Source>& references,
-                                                const Options& opts) {
-    MixAnalysisAgent::Input input;
+MixAnalysisData MixAnalysisInput::build(double sr, const std::vector<Source>& tracks,
+                                        const juce::AudioBuffer<float>* masterIn,
+                                        const std::vector<Source>& references,
+                                        const Options& opts) {
+    MixAnalysisData input;
```
lucaromagnoli: Same nit as OfflineMixAnalysis: this is now a `MixAnalysisData`, so a name like `data` or `measurements` would read better than `input`.

PR #1879, opened 2026-07-26
Dev/0.17.0

lucaromagnoli:


magda/agents/command_model_onnx.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<int64_t> ids = enc.inputIds;
+    std::vector<int64_t> mask = enc.attentionMask;
+    std::array<Ort::Value, 2> inputs{
+        Ort::Value::CreateTensor<int64_t>(mem, ids.data(), ids.size(), shape.data(), shape.size()),
+        Ort::Value::CreateTensor<int64_t>(mem, mask.data(), mask.size(), shape.data(),
+                                          shape.size())};
+
+    static const char* kInputNames[] = {"input_ids", "attention_mask"};
+    static const char* kOutputNames[] = {"intent_logits", "slot_logits"};
+    auto outputs = impl_->session.Run(Ort::RunOptions{nullptr}, kInputNames, inputs.data(),
```
lucaromagnoli, line 249: **Should fix before release.** `Ort::Session::Run` throws `Ort::Exception` on any inference failure (corrupt but loadable model, shape mismatch, allocator failure) and nothing up the stack catches it: `CommandAgent::generateLocal` guards construction only, and neither the orchestrator workflow lambda nor `RequestThread::run` has a handler. JUCE's thread entry point swallows the exception in release builds, so no Completed/Failed event fires and the console sticks on Thinking... with the turn lost. Wrapping inference in try/catch and surfacing a Failed event fixes the whole path.

magda/agents/command_model_onnx.cpp
```diff
@@ -0,0 +1,334 @@
...
+        // maps.json stores name -> id; invert into dense id-indexed tables.
+        auto invert = [](const juce::var& obj, std::vector<std::string>& out) {
+            auto* dyn = obj.getDynamicObject();
+            if (dyn == nullptr)
+                return;
+            for (const auto& prop : dyn->getProperties()) {
+                const int id = static_cast<int>(prop.value);
+                if (id >= static_cast<int>(out.size()))
+                    out.resize(static_cast<size_t>(id) + 1);
+                out[static_cast<size_t>(id)] = prop.name.toString().toStdString();
```
lucaromagnoli: **Should fix before release.** The id parsed from maps.json is unvalidated. A negative id skips the resize guard and executes `out[static_cast<size_t>(-1)]`, an out-of-bounds write that the `catch (std::exception)` fallback in command_agent.cpp cannot recover from; a huge positive id instead resizes the vector to gigabytes. The model dir is user configurable (`MAGDA_COMMAND_MODEL_DIR` / settings) and these files carry no hash, so a corrupt or tampered maps.json is realistic input. Reject ids outside a sane [0, N) range while parsing.

magda/agents/command_model.cpp
```diff
@@ -0,0 +1,1001 @@
...
+            ++i;
+            continue;
+        }
+        if (j < n && text[j] == '.' && j + 1 < n &&
+            std::isdigit(static_cast<unsigned char>(text[j + 1]))) {
+            ++j;
+            while (j < n && std::isdigit(static_cast<unsigned char>(text[j])))
+                ++j;
+        }
+        return std::stod(text.substr(i, j - i));
```
lucaromagnoli: **Should fix before release.** `std::stod` throws `std::out_of_range` for a numeric literal exceeding double range, and `valueOrFirstNumber` falls back to `parseNumber` over arbitrary user text, so typing a 309+ digit number into the console reaches this line. Nothing catches it (same uncaught path as the ONNX inference finding), so the console turn dies silently. `renderPrediction` is shared, so both backends inherit this.

magda/agents/command_model.cpp
```diff
@@ -0,0 +1,1001 @@
...
+    auto toks = tokenize(expandAliasBrackets(text));
+    if (static_cast<int>(toks.size()) > d::kMaxLen)
+        toks.resize(d::kMaxLen);
+    const int L = static_cast<int>(toks.size());
+    out.tokens = toks;
+    if (L == 0)
+        return out;
+
+    // Encode to ids, padded to kMaxLen (pad id 0).
+    std::array<int, 64> ids{};  // kMaxLen == 24 <= 64
```
lucaromagnoli, line 641: **Minor.** The guarantee in this comment deserves to be `static_assert(d::kMaxLen <= 64)`; regenerating command_model_data.cpp with a larger kMaxLen would silently overrun the array.

magda/daw/stem_separation/StemSeparationService.cpp
```diff
@@ -0,0 +1,387 @@
...
+                    undo.executeCommand(std::make_unique<magda::DeleteTrackCommand>(trackId));
+            }
+            if (groupId != INVALID_TRACK_ID &&
+                magda::TrackManager::getInstance().getTrack(groupId) != nullptr)
+                undo.executeCommand(std::make_unique<magda::DeleteTrackCommand>(groupId));
+            report(INVALID_TRACK_ID, error);
+        });
+    };
+
+    pool_->addJob([this, sourceClipId, engine, filePath, baseName, sourceTrackId, startBeat,
```
lucaromagnoli, line 286: **Should fix before release.** This job lambda has no try/catch and `juce::ThreadPool::runNextJob` swallows exceptions, so on any throw neither `report` nor `failAndCleanup` runs: `pendingJobs_` never decrements, `isBusy()` stays true forever, the activity banner sticks at Splitting stems N%, the completion callback never fires, and the phase-1 empty stem tracks linger. `std::bad_alloc` is a realistic trigger given the Spleeter memory profile (see comment there). Wrap the job body and route failures through the normal cleanup.

lucaromagnoli (commented):
Review of the 0.17.0 release aggregation. The constituent PRs were reviewed individually at merge time, so this pass covers release-level checks plus fresh spot reviews of the three biggest new subsystems: stem separation, the remote API contract, and the agent runtime / command model.

Release-level checks all pass:
- main is fully contained in dev/0.17.0; nothing on main is lost by merging
- version derives from git tags, no hardcoded bump needed
- no credentials in the diff; no network listeners ship (the remote API is contract-only)
- translation diffs are tiny, no Crowdin key trampling
- CI/packaging additions are ASCII-clean; the new VC++ redistributable staging verifies the Authenticode signature before bundling, nicely done
- MuseHub bundle removal from release.yml is deliberate (packaging/windows/musehub went with it in the Windows packaging fix)
- the FM0 anti-click work in magda_fm.dsp is well reasoned, and the export power fix ships with a dedicated JUCE test

Repo hygiene, not anchorable inline: prototypes/command-model-poc commits about 39k lines of training JSONL plus a model.pt binary (0.2 MB). Precedent exists with prototypes/media_db, but this is two thirds of the PR's line count and sits in every clone forever; consider LFS or an external bucket for future datasets.

Verdict: mergeable as a release aggregation; nothing here regresses shipped behavior, and the clean areas (bounded agent runtime, tokenizer, cancellation plumbing, beats-domain discipline in the DTOs, error envelopes) are genuinely solid. The two clusters I would fix before tagging 0.17.0 are the uncaught command-model inference exceptions and the stem-separation wedge/OOM group, since both turn ordinary user actions (typing a big number, splitting a long song) into stuck-forever UI. Details inline.

magda/daw/stem_separation/StemSeparationService.cpp
```diff
@@ -0,0 +1,387 @@
...
+        double sampleRate = 0.0;
+        if (!decodeFile(filePath, input, sampleRate)) {
+            failAndCleanup("Could not read audio file");
+            return;
+        }
+
+        // Throttle progress to whole-percent steps on the message thread,
+        // feeding the activity banner.
+        auto lastPercent = std::make_shared<int>(-1);
+        auto progress = [this, lastPercent](float frac) {
```
lucaromagnoli: **Should fix before release.** This progress lambda unconditionally returns true and no cancel API exists, so separation cancellation is dead code from the user's perspective even though all three separators implement progress-cancel. A mis-clicked Demucs split on a long song can only be waited out, and it blocks the single-slot queue plus app quit (see destructor comment).

magda/daw/stem_separation/StemSeparationService.cpp
```diff
@@ -0,0 +1,387 @@
...
+}  // namespace
+
+StemSeparationService& StemSeparationService::getInstance() {
+    static StemSeparationService instance;
+    return instance;
+}
+
+StemSeparationService::StemSeparationService() : pool_(std::make_unique<juce::ThreadPool>(1)) {}
+
+StemSeparationService::~StemSeparationService() {
```
lucaromagnoli, line 159: **Should fix before release.** The worker is an `addJob(std::function)` lambda, so the interrupt signal from `removeAllJobs(true, ...)` is never polled and the progress callback never returns false. Quitting mid-split blocks static destruction for the remainder of ORT inference, which on a long song is minutes of apparent freeze before a force kill. Polling a cancellation flag in the progress callback (and honouring it) fixes quit and user cancel together.

magda/daw/stem_separation/StemSeparationService.cpp
```diff
@@ -0,0 +1,387 @@
...
+    if (!jpath.existsAsFile())
+        return false;
+
+    std::unique_ptr<juce::AudioFormatReader> reader(fm.createReaderFor(jpath));
+    if (!reader || reader->lengthInSamples <= 0 || reader->sampleRate <= 0 ||
+        reader->numChannels < 1)
+        return false;
+
+    const int channels = static_cast<int>(reader->numChannels);
+    const int len = static_cast<int>(reader->lengthInSamples);
```
lucaromagnoli, line 73: **Follow-up.** `lengthInSamples` is truncated to `int` with no guard; a file over 2^31 samples yields a negative or wrong length and `setSize` UB. Even legal long files are fully decoded with no cap. Worth a bounds check with a clean error.

magda/daw/stem_separation/SpleeterSeparator.cpp
```diff
@@ -0,0 +1,286 @@
...
+    #include <cmath>
+    #include <filesystem>
+
+namespace magda::stems {
+
+namespace {
+
+// Same ratio convention as the other backends (input samples consumed per
+// output sample).
+std::vector<float> resampleChannel(const float* in, int numSamples, double srcRate,
```
lucaromagnoli, line 23: **Should fix before release.** This uses the length-less `LagrangeInterpolator::process` overload, whose contract requires at least `speedRatio * numOutputSamples` input samples; with outLen = ceil(numSamples / ratio) that exceeds numSamples whenever the length is not ratio aligned, so it reads past the exactly sized input buffer (heap over-read; ASan fires on any odd-length 88.2 kHz clip). The overload taking `numInputSamplesAvailable` exists and is the fix. All three separator test suites feed 44.1 kHz only, so this path is never exercised; a non-44.1 kHz test would lock the fix in.

magda/daw/stem_separation/SpleeterSeparator.cpp
```diff
@@ -0,0 +1,286 @@
...
+    const int numFrames = 1 + (paddedLen - sp::kFftSize) / sp::kHop;
+    const int paddedFrames = sp::paddedFrameCount(numFrames);
+    const int numSplits = paddedFrames / sp::kFramesPerSplit;
+
+    const std::vector<float> window = periodicHann();
+    juce::dsp::FFT fft(12);  // 4096
+
+    // Full complex spectra per channel (for mask application) and the
+    // magnitude tensor both models consume.
+    std::array<std::vector<float>, 2> spectra;  // numFrames x kNumBins x (re,im)
```
lucaromagnoli: **Should fix before release.** Whole-file materialization: full complex spectra for both channels plus the input tensor and two full-size outputs all go through a single `session.Run`. That is roughly 60+ bytes per sample per channel, so an hour-long stereo file needs several GB and OOMs on typical machines, which then triggers the wedged-service failure (no try/catch around the pool job). Chunking the ORT runs the way DemucsSeparator does bounds the peak.

magda/daw/stem_separation/SpleeterSeparator.cpp
```diff
@@ -0,0 +1,286 @@
...
+                                              spleeter::kKeptBins};
+        auto inputTensor = Ort::Value::CreateTensor<float>(memoryInfo, const_cast<float*>(x.data()),
+                                                           x.size(), shape.data(), shape.size());
+
+        const char* inputNames[] = {"x"};
+        const char* outputNames[] = {"y"};
+        auto outputs =
+            session.Run(Ort::RunOptions{nullptr}, inputNames, &inputTensor, 1, outputNames, 1);
+
+        const float* data = outputs[0].GetTensorData<float>();
```
lucaromagnoli, line 94: **Follow-up.** The output tensor is read for a hard-coded element count without checking `GetTensorTypeAndShapeInfo()`. Since `isInstalled` is size-only and `makeSeparator` loads whatever file sits at the path, a swapped or corrupt on-disk model that still parses becomes a heap over-read instead of a clean failure.

magda/daw/stem_separation/DemucsSeparator.cpp
```diff
@@ -0,0 +1,222 @@
...
+    #include <filesystem>
+
+namespace magda::stems {
+
+namespace {
+
+// Resample one channel. ratio-style matches BasicPitchTranscriber: the
+// LagrangeInterpolator's speedRatio is input samples consumed per output
+// sample.
+std::vector<float> resampleChannel(const float* in, int numSamples, double srcRate,
```
lucaromagnoli, line 39: **Should fix before release.** Same Lagrange over-read as SpleeterSeparator: the length-less overload reads ratio * ceil(N / ratio) input samples, past the exactly sized buffer. The back-resample from `acc` (sized exactly `total`) has the same issue; Spleeter's back-resample is only accidentally safe because its buffer carries FFT padding.

magda/daw/stem_separation/DemucsSeparator.cpp
```diff
@@ -0,0 +1,222 @@
...
+        const std::array<int64_t, 3> inShape = {1, 2, demucs::kSegmentSamples};
+        auto inputTensor = Ort::Value::CreateTensor<float>(
+            memoryInfo, const_cast<float*>(mix.data()), mix.size(), inShape.data(), inShape.size());
+
+        const char* inputNames[] = {"mix"};
+        const char* outputNames[] = {"stems"};
+        auto outputs =
+            session.Run(Ort::RunOptions{nullptr}, inputNames, &inputTensor, 1, outputNames, 1);
+
+        const float* data = outputs[0].GetTensorData<float>();
```
lucaromagnoli: **Follow-up.** Same missing output shape check as SpleeterSeparator: the tensor is consumed for a hard-coded element count without consulting `GetTensorTypeAndShapeInfo()`.

magda/daw/stem_separation/StemModelDownloader.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  private:
+    // Stream `entry.url` into place via a temp file, verifying the SHA-256
+    // before the atomic rename; mirrors SampleTaggerDownloader.
+    bool downloadOne(const ManifestEntry& entry, Progress& p) {
+        const auto dest = StemModelDownloader::modelsDir().getChildFile(entry.filename);
+
+        juce::URL url(entry.url);
+        int statusCode = 0;
+        const auto options = juce::URL::InputStreamOptions(juce::URL::ParameterHandling::inAddress)
```
lucaromagnoli: **Follow-up.** `withConnectionTimeoutMs` is connect-only; there is no read timeout, so the read loop can block indefinitely on a half-dead connection (Wi-Fi drop without RST mid-download). The destructor's `stopThread(10000)` then blocks the message thread for 10 seconds and falls through to JUCE's killThread. A read deadline avoids both.

magda/daw/stem_separation/StemModelDownloader.cpp
```diff
@@ -0,0 +1,291 @@
...
+        juce::TemporaryFile tmp(dest);  // atomic rename on commit
+        auto out = std::make_unique<juce::FileOutputStream>(tmp.getFile());
+        if (!out->openedOk()) {
+            p.errorMessage = juce::String("Could not write to ") + tmp.getFile().getFullPathName();
+            return false;
+        }
+
+        constexpr int kChunk = 1 << 16;  // 64 KiB
+        juce::MemoryBlock buf(kChunk);
+        for (;;) {
```
lucaromagnoli, line 234: **Follow-up.** No size ceiling in the read loop: bytes append until EOF, so a misbehaving or compromised server fills the disk before the SHA-256 check ever rejects the file. Cheap fix: abort once written exceeds entry.size.

magda/daw/stem_separation/StemModelDownloader.cpp
```diff
@@ -0,0 +1,291 @@
...
+                p.errorMessage = juce::String("Read error on ") + entry.filename;
+                return false;
+            }
+            if (read == 0)
+                break;
+            if (!out->write(buf.getData(), static_cast<size_t>(read))) {
+                p.errorMessage = juce::String("Write error on ") + dest.getFullPathName();
+                return false;
+            }
+            p.bytesDone += read;
```
lucaromagnoli, line 261: **Minor.** One `callAsync` per 64 KiB chunk is roughly 4,800 message-thread posts per model, hundreds per second on fast links; throttling to whole-percent changes is enough. For the record, the rest of this downloader is the strongest in the PR: HTTPS, SHA-256 over the temp file, atomic rename, and truncated downloads fail the hash.

magda/daw/api/remote_api.cpp
```diff
@@ -0,0 +1,1281 @@
...
+juce::var nullableId(const std::optional<int>& id) {
+    return id ? juce::var(*id) : juce::var();
+}
+
+template <typename Id>
+std::optional<Id> readNullableId(const juce::var& object, const char* property) {
+    const auto value = object[property];
+    if (value.isVoid())
+        return std::nullopt;
+    return static_cast<Id>(static_cast<int>(value));
```
lucaromagnoli: **Should fix before the 1.0 contract freezes.** Id decode casts a JUCE int64 to `int` and no id schema declares a maximum, so a clips.get with clipId 4294967336 (2^32 + 40) validates and silently aliases to clip 40. Same pattern in `readIntegerArray` (childIds, clipIds, deviceIds) and `readNullableId`. Either add maxima to the id schemas or range-check in decode and return validation_failed.

magda/daw/api/remote_api.cpp
```diff
@@ -0,0 +1,1281 @@
...
+        "additionalProperties":false
+    })json");
+    return value;
+}
+
+const juce::var& trackSchema() {
+    static const auto value = parseSchema(R"json({
+        "type":"object",
+        "properties":{
+            "id":{"type":"integer","minimum":-2},
```
lucaromagnoli: **Should fix before the 1.0 contract freezes.** minimum -2 is meant to admit MASTER_TRACK_ID but also admits -1, which is INVALID_TRACK_ID; a selection.set with trackId -1 passes validation and hands the reserved sentinel straight to `SelectionApi::selectTrack`. A flat minimum cannot express {-2} union [0, inf); this needs anyOf/enum or an explicit -1 rejection. Same domain appears in selectionSchema and the tracks.get input.

magda/daw/api/remote_api.cpp
```diff
@@ -0,0 +1,1281 @@
...
+            "type":{"type":"string","enum":["audio","group","aux","master","multi_out","chord"]},
+            "name":{"type":"string"},
```
lucaromagnoli, line 152: **Should fix before the 1.0 contract freezes.** No string in any schema has maxLength and no array has maxItems, so a 500 MB track name or a million-entry clipIds array validates and lands in the model unbounded (memory/UI DoS from one request). Blanket caps at the validation layer are cheap now and a breaking change later.

magda/daw/api/remote_api.hpp
```diff
@@ -0,0 +1,312 @@
...
+struct AutomationLaneInfo;
+struct ClipInfo;
+struct DeviceInfo;
+struct ProjectInfo;
+struct TrackInfo;
+
+namespace remote {
+
+inline constexpr int API_VERSION_MAJOR = 1;
+inline constexpr int API_VERSION_MINOR = 0;
```
lucaromagnoli, line 25: **Should fix before the 1.0 contract freezes.** Nothing in these three files documents or asserts which thread must service requests, yet the projections read live message-thread state by reference (`makeSessionDto` iterates TrackManager's UI-mutated track vector). The first transport that calls this off the message thread races container mutation with no guard. One doc line here plus JUCE_ASSERT_MESSAGE_THREAD in the handlers pins the contract while it is still cheap.

magda/daw/api/remote_api_projection.cpp
```diff
@@ -0,0 +1,352 @@
...
+    return dto;
+}
+
+AutomationLaneDto makeAutomationLaneDto(const AutomationLaneInfo& lane) {
+    AutomationLaneDto dto;
+    dto.id = lane.id;
+    dto.type = lane.type == AutomationLaneType::Absolute ? "absolute" : "clip_based";
+    dto.name = lane.getDisplayName();
+    dto.target.kind = juce::String(toString(lane.target.kind));
+    if (!lane.target.isEditScoped() && lane.target.devicePath.trackId != INVALID_TRACK_ID)
```
lucaromagnoli, line 370: **Should fix before the 1.0 contract freezes.** Master-track lanes emit trackId -2 here, but automationTargetSchema declares minimum 0, so automation.getLane responses violate the published output schema. The contract is internally inconsistent about whether master is addressable (trackSchema and selectionSchema say minimum -2).

magda/agents/agent_runtime.cpp
```diff
@@ -0,0 +1,431 @@
...
+                                             "Tool call ID must not be empty");
+            } else if (!seenToolCallIds.insert(call.id).second) {
+                toolResult = makeErrorResult(call, result.finalRevision, "duplicate_tool_call_id",
+                                             "Tool call ID was already used in this run");
+            } else if (tool == nullptr) {
+                toolResult = makeErrorResult(call, result.finalRevision, "tool_not_allowed",
+                                             "Tool is not allowed for this agent surface");
+            } else {
+                const bool isMutation = tool->access == ToolAccess::Mutation;
+                if (isMutation && result.mutations >= effectiveDefinition.budget.maxMutations) {
```
lucaromagnoli, line 336: **Follow-up.** MutationLimit, RepeatedToolCall, and cancel/timeout mid-batch all return after the assistant turn with toolCalls was pushed but before Tool-role turns exist for the remaining calls. `RunResult.conversation` exists precisely for resuming, and real provider adapters reject a tool_use without a matching tool_result, so such transcripts are unreplayable. Synthesizing cancelled tool results before returning keeps the invariant.

magda/daw/audio/plugins/DeviceServices.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+void unregisterDeviceServices(tracktion::engine::Edit& edit) {
+    std::scoped_lock lock(servicesMutex);
+    servicesByEdit.erase(&edit);
+}
+
+DeviceServices getDeviceServices(tracktion::engine::Edit& edit) {
+    std::scoped_lock lock(servicesMutex);
+    return resolve(servicesByEdit[&edit]);
```
lucaromagnoli: **Minor.** `servicesByEdit[&edit]` default-inserts on lookup, so a query after `unregisterDeviceServices` resurrects an entry keyed by a soon-dangling Edit pointer and the global map grows. `find()` plus a static empty fallback avoids it.

magda/agents/command_model_downloader.cpp
```diff
@@ -0,0 +1,261 @@
...
+        out.reset();
+#endif
+
+        p.phase = Phase::Verifying;
+        postProgress(p);
+
+        // Size is the completeness check — a truncated download is the failure
+        // that actually happens. Integrity comes from the pinned revision:
+        // /resolve/<sha>/ cannot serve different bytes later.
+        if (tmp.getFile().getSize() != entry.size) {
```
lucaromagnoli, line 163: **Note, accepted risk.** Completeness here is a byte-size check only; integrity rests on TLS plus the pinned HF revision, which the header comment documents and which is defensible. Just flagging it as the one download path in this release without a content hash, in contrast to the stem model downloader's SHA-256.

prototypes/command-model-poc/HANDOFF_ONNX.md
```diff
@@ -0,0 +1,249 @@
+# Handoff: ONNX export + C++ backend for the encoder command model
```
lucaromagnoli: **Repo hygiene.** Handoff docs stay local by repo convention; this one reads as session-to-session working state rather than durable documentation. Suggest dropping it from the release and folding anything durable into findings.md.

PR #1905, opened 2026-07-27
feat: consolidate plugin metadata storage

lucaromagnoli:


magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,390 @@
...
+    if (auto xml = juce::parseXML(aliasesFile)) {
+        for (auto* child : xml->getChildIterator()) {
+            const auto key = child->getStringAttribute("key");
+            if (key.isNotEmpty())
+                setAlias(key, child->getStringAttribute("alias"));
+        }
+    }
+
+    saveExclusions(loadExclusionList(legacyDirectory.getChildFile("plugin_exclusions.txt")));
+    exec(db_, "INSERT INTO plugin_store_meta (key, value) "
```
lucaromagnoli: The import is not atomic and the completion stamp can throw.

Four separate transactions, then a bare `INSERT` on a PRIMARY KEY, gated by a check-then-insert. Two failure modes:

1. Interrupted mid-import: the flag is never stamped, so the next open re-imports the legacy files and clobbers anything the user changed in between.
2. Two connections opening for the first time both pass the `SELECT` at line 157, and the second one's `INSERT` hits a UNIQUE constraint. That throws out of the constructor, so `openDefault()` fails and the caller's `catch` clears its state. `PluginScanner` and `PluginScanCoordinator` each open independently, so this is reachable.

Wrapping the whole of `importLegacyFilesOnce()` in a single `Transaction` and using `INSERT OR IGNORE` fixes both.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -972,115 +966,58 @@ void PluginBrowserContent::toggleFavorite(const PluginBrowserInfo& plugin) {
...
-        juce::String key = p.uniqueId.isNotEmpty() ? p.uniqueId : p.name;
-        p.isFavorite = favoriteKeys.contains(key);
+    try {
+        const auto favoriteKeys = magda::PluginMetadataStore::openDefault().favoriteKeys();
+        for (auto& plugin : plugins_) {
+            const auto key = plugin.uniqueId.isNotEmpty() ? plugin.uniqueId : plugin.name;
+            plugin.isFavorite = favoriteKeys.contains(key);
+        }
+    } catch (const std::exception& e) {
+        DBG("Failed to load plugin favorites: " << e.what());
```
lucaromagnoli, line 1010: A transient open failure here silently destroys the user's favorites.

The catch leaves every `isFavorite` at its default `false`, and the next `saveFavorites()` writes `favorite=0` for every visible plugin. Aliases take the same path at line 1020 (every alias written back as NULL), and `loadExclusions` in the scanner clears the vector so the next `saveExclusions` does DELETE-all plus zero inserts.

The old XML code had the same shape, but an unparseable XML file is a much rarer event than a locked SQLite file, and all four datasets now sit behind one lock. A `loaded_` flag that makes save a no-op when the last load failed would close it.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+PluginMetadataStore::~PluginMetadataStore() {
+    sqlite3_close(db_);
+}
+
+PluginMetadataStore PluginMetadataStore::openDefault() {
+    return PluginMetadataStore(paths::pluginMetadataFile());
+}
+
+void PluginMetadataStore::createSchema() {
```
lucaromagnoli, line 107: No schema versioning. `MediaDatabase.hpp` next door has `kSchemaVersion`, `PRAGMA user_version`, and an ordered migration ladder; this store relies only on `CREATE TABLE IF NOT EXISTS`, so adding a column later has no upgrade path on installs that already have the file.

Unlike the media DB, this one is not a rebuildable cache: favorites and aliases only exist here once the legacy files stop being written. Stamping `user_version=1` now costs one line and leaves the door open.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+#include <stdexcept>
+#include <string>
+
+#include "core/AppPaths.hpp"
+
+namespace magda {
+namespace {
+
+class Statement {
```
lucaromagnoli: `Statement`, `Transaction`, `exec`, and the error type all already exist in `magda/daw/media_db/MediaDatabase.hpp`, with a typed `MediaDatabaseError` and move semantics this one lacks. Two divergent SQLite wrappers in one binary will drift, and the next store makes three. Either reuse `MediaDatabase` directly or lift the shared pieces into a small sqlite util header both can sit on.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,390 @@
...
+        db_ = nullptr;
+        throw;
+    }
+}
+
+PluginMetadataStore::~PluginMetadataStore() {
+    sqlite3_close(db_);
+}
+
+PluginMetadataStore PluginMetadataStore::openDefault() {
```
lucaromagnoli, line 93: Every call site does `openDefault()` per operation, so each one opens the file, writes `journal_mode=WAL`, runs all the DDL, and probes the migration flag before doing any actual work. `toggleFavorite` pays that twice (save then reload), on the message thread.

Holding one connection per owner would match the `MediaDatabase` comment about one instance per thread and drop the per-click cost to a single statement.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,390 @@
...
+        importLegacyFilesOnce();
+    } catch (...) {
+        sqlite3_close(db_);
+        db_ = nullptr;
+        throw;
+    }
+}
+
+PluginMetadataStore::~PluginMetadataStore() {
+    sqlite3_close(db_);
```
lucaromagnoli: `sqlite3_close_v2` would be safer here: plain `sqlite3_close` returns SQLITE_BUSY and leaks the handle if any statement is still open, and the return value is discarded either way.

lucaromagnoli (commented):
Consolidating the four metadata files into one SQLite store is the right call, and the sparse-row scheme that keeps favorites and aliases alive across a rescan is a genuine improvement over the old merge-on-write XML dance. Two things I checked and they hold up: the alias key match in `AIChatConsoleContent` now lines up exactly with what `PluginBrowserContent` writes (`createIdentifierString()` for external, `pluginId` for internal), replacing the old fuzzy name match; and `sqlite3` is already linked `PUBLIC` into `magda_daw` via `media_db/CMakeLists.txt`, so the new TU builds without a CMake change.

Comments inline. The only one I would call a blocker is the legacy import atomicity, which is a data-loss path on first run. The rest are design and hygiene.

One pre-existing thing worth its own issue, not this PR: `PluginScanner` and `PluginScanCoordinator` each keep their own `excludedPlugins_` and each `saveExclusions()` is a destructive full-table replace, so the last writer wins. That was true of the text file too, but a shared DB makes it easier to fix properly.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,390 @@
...
+}
+
+void PluginMetadataStore::importLegacyFilesOnce() {
+    Statement migrated(
+        db_, "SELECT 1 FROM plugin_store_meta WHERE key='legacy_import_complete' LIMIT 1");
+    if (sqlite3_step(migrated.get()) == SQLITE_ROW)
+        return;
+
+    const auto legacyDirectory = file_.getParentDirectory();
+    const auto pluginListFile = legacyDirectory.getChildFile("PluginList.xml");
```
lucaromagnoli: The four legacy filenames are hardcoded here (lines 162, 171, 180, 189) while `paths::pluginListFile()`, `pluginFavoritesFile()`, `pluginAliasesFile()`, and `pluginExclusionsFile()` still exist in `AppPaths` and are now unreferenced everywhere else. Two owners of the same names.

I would route through `paths::` when the store is opened at the default location, or delete the accessors and let this be the one place that knows them. Related: `saveExclusionList` in `PluginExclusions.cpp` has no callers left.

magda/daw/engine/PluginMetadataStore.hpp
```diff
@@ -0,0 +1,88 @@
...
+    [[nodiscard]] juce::StringArray favoriteKeys() const;
+
+    void setAlias(const juce::String& key, const juce::String& alias);
+    void saveAliases(const std::map<juce::String, juce::String>& aliases);
+    [[nodiscard]] std::map<juce::String, juce::String> aliases() const;
+
+    void saveExclusions(const std::vector<ExcludedPlugin>& entries);
+    [[nodiscard]] std::vector<ExcludedPlugin> loadExclusions() const;
+
+    [[nodiscard]] std::vector<PluginMetadataRecord> query(
```
lucaromagnoli, line 77: `query()` and `PluginMetadataQuery` have no production callers, only tests. The browser still pulls everything out of `KnownPluginList` and filters in C++, which is the thing this API looks designed to replace. Either wire the browser to it in this PR or drop it until the consumer lands, so the surface does not sit here untested against real data.

tests/test_plugin_metadata_store.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "magda/daw/engine/PluginMetadataStore.hpp"
+
+namespace {
+
+struct TempDirectory {
+    TempDirectory()
+        : directory(juce::File::getCurrentWorkingDirectory().getChildFile(
```
lucaromagnoli: `getCurrentWorkingDirectory()` deviates from the convention in the rest of `tests/`, which uses `juce::File::getSpecialLocation(juce::File::tempDirectory)`. It also means a crashed test leaves a directory in the build tree.

tests/test_plugin_metadata_store.cpp
```diff
@@ -0,0 +1,138 @@
...
+    result.fileOrIdentifier = path;
+    result.uniqueId = uniqueId;
+    result.deprecatedUid = uniqueId;
+    result.isInstrument = instrument;
+    return result;
+}
+
+}  // namespace
+
+TEST_CASE("PluginMetadataStore round-trips descriptions and user metadata",
```
lucaromagnoli, line 79: The highest-value path is untested: a favorite or alias surviving a rescan that no longer reports that plugin. That is the `description_xml IS NOT NULL` guard on the DELETE in `saveKnownPlugins`, and it is the entire justification for the sparse-row design, but nothing exercises it.

Concretely: save two plugins, favorite one and alias the other, then `saveKnownPlugins` with a list containing neither, and assert both the favorite and the alias are still there while `query()` returns nothing.

lucaromagnoli:
Closes #943

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,388 @@
...
+            value TEXT NOT NULL
+        );
+        CREATE INDEX IF NOT EXISTS idx_plugin_format_instrument_favorite
+            ON plugin (format, is_instrument, favorite);
+        CREATE INDEX IF NOT EXISTS idx_plugin_file_or_identifier
+            ON plugin (file_or_identifier);
+        CREATE INDEX IF NOT EXISTS idx_plugin_alias ON plugin (alias);
+    )SQL");
+    if (version < kPluginMetadataSchemaVersion)
+        sqlite::exec(db_.get(), "PRAGMA user_version = 1");
```
lucaromagnoli: The stamp hardcodes `1` rather than `kPluginMetadataSchemaVersion`, so bumping the constant leaves this string behind and the file gets marked with the wrong version.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+PluginMetadataStore& PluginMetadataStore::defaultForCurrentThread() {
+    thread_local std::unique_ptr<PluginMetadataStore> store;
+    const auto expectedFile = paths::pluginMetadataFile();
+    if (!store || store->file() != expectedFile)
+        store = std::make_unique<PluginMetadataStore>(openDefault());
+    return *store;
+}
+
+void PluginMetadataStore::createSchema() {
```
lucaromagnoli, line 107: The version read and the newer-than-build guard are the right shape, but there is no migration ladder and the stamp below is applied unconditionally.

At v1 that is correct, because the `CREATE TABLE IF NOT EXISTS` block below *is* the v1 schema, so anything that reads as 0 genuinely is v1 after it runs. At v2 it turns into a landmine: `CREATE TABLE IF NOT EXISTS` will not add a column to a table that already exists, so a v1 file gets stamped v2 having never been migrated, and the guard will then happily accept it forever.

`migrateMediaDatabase` next door already has the ladder shape to copy. Even just a comment here saying the next schema change has to add a migration step before the stamp would stop this from being discovered the hard way.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -558,13 +552,43 @@ void PluginBrowserContent::loadExternalPlugins() {
...
+            plugin.isExternal = true;
+            plugin.uniqueId = record.key;
+            plugin.fileOrIdentifier = record.fileOrIdentifier;
+            plugins_.push_back(std::move(plugin));
+        }
 
-    for (const auto& desc : pluginTypes) {
-        plugins_.push_back(PluginBrowserInfo::fromPluginDescription(desc));
+        DBG("Loaded " << records.size() << " external plugin metadata records");
+        return;
```
lucaromagnoli, line 564: The fallback below only fires on an exception, never on an empty result. If the query succeeds and returns zero rows while `KnownPluginList` is populated, the browser shows internal plugins only, with nothing in the log beyond `Loaded 0 external plugin metadata records`.

That is reachable through the new guard rather than being hypothetical: `loadPluginList()` failing sets `pluginMetadataLoaded_ = false`, which makes `savePluginList()` skip for the rest of the session, so a completed scan never reaches the DB that this now reads from. It does self-heal on the next rescan because `clearPluginList()` sets the flag back to true, but between those two points the user has a populated `KnownPluginList` and an empty browser.

Falling through when `records` is empty and the engine reports types would cover it.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -558,13 +552,43 @@ void PluginBrowserContent::loadExternalPlugins() {
...
 
-    auto pluginTypes = engine_->getPreferredPluginTypes();
+    try {
+        juce::StringArray preferredKeys;
+        for (const auto& description : engine_->getPreferredPluginTypes())
+            preferredKeys.add(description.createIdentifierString());
+
+        const auto records = magda::PluginMetadataStore::defaultForCurrentThread().query();
+        for (const auto& record : records) {
+            if (!preferredKeys.contains(record.key))
```
lucaromagnoli: `juce::StringArray::contains` is a linear scan, and this runs once per record, so the whole loop is O(n*m) string compares on the message thread for every `refreshPluginList()`. Fine at a few hundred plugins, less so at a few thousand. An `std::unordered_set<juce::String>` built alongside `preferredKeys` makes it linear.

magda/daw/core/SqliteUtils.hpp
```diff
@@ -0,0 +1,177 @@
+#pragma once
```
lucaromagnoli, line 1: The extraction is the right call and the savepoint nesting is what makes the atomic import possible, so this file is doing real work. The placement is the only problem: the namespace is `magda::sqlite`, deliberately generic, but the file sits in `media_db/` and `engine/PluginMetadataStore.hpp` now reaches across module boundaries to include it. Somewhere neutral (`core/`, or its own `sqlite/`) would match what the namespace already says this is.

lucaromagnoli (commented):
All ten points from the first pass are addressed, and the SQLite extraction went further than I asked: pulling `Error`/`Connection`/`Statement`/`Transaction`/`exec` into `SqliteUtils.hpp` and refactoring `MediaDatabase` onto it leaves one wrapper instead of two, and the savepoint-based nesting in `sqlite::Transaction` is precisely what lets the legacy import become one atomic unit without unpicking the per-dataset transactions underneath it. The two new migration tests (interrupted import via an ABORT trigger, concurrent first opens on two threads) exercise exactly the failure modes I flagged, which is better than asserting the fix shape.

Six follow-ups inline, none of them blockers. The two I would actually act on before merge are the schema-stamp landmine and the browser's empty-result path.

One thing worth confirming rather than assuming: `MediaDatabase::Transaction` used to throw when nested (plain `BEGIN`) and now silently opens a savepoint. That is an improvement, but it changes shipped media DB semantics inside a plugin-metadata PR, so the media_db tests are the ones I would want green here.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,388 @@
...
+    const auto version = readUserVersion(db_.get());
+    if (version < 0)
```
lucaromagnoli: Dead branch: `readUserVersion` throws when the pragma returns no row, and `PRAGMA user_version` never yields a negative value, so this can't fire.

magda/daw/engine/PluginMetadataStore.cpp
```diff
@@ -0,0 +1,388 @@
...
+        SET description_xml=NULL, format=NULL, category=NULL, manufacturer=NULL,
+            file_or_identifier=NULL, is_instrument=0
+        WHERE description_xml IS NOT NULL
+          AND plugin_key NOT IN (SELECT plugin_key FROM plugin_seen)
+    )SQL");
+    transaction.commit();
+}
+
+void PluginMetadataStore::loadKnownPlugins(juce::KnownPluginList& plugins) const {
+    juce::KnownPluginList loaded;
```
lucaromagnoli: Staging into a local list so `plugins` is only cleared once the read has fully succeeded is the right fix for the wipe-on-failure path. Minor cost: `KnownPluginList::addType` does a dedup scan, and this pays it twice per plugin (once here, once in the copy loop below). A `std::vector<juce::PluginDescription>` as the staging buffer keeps the same guarantee without the second scan.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -558,13 +554,53 @@ void PluginBrowserContent::loadExternalPlugins() {
...
+            plugin.isFavorite = record.isFavorite;
+            plugin.isExternal = true;
+            plugin.uniqueId = record.key;
+            plugin.fileOrIdentifier = record.fileOrIdentifier;
+            queriedPlugins.push_back(std::move(plugin));
+        }
 
-    for (const auto& desc : pluginTypes) {
-        plugins_.push_back(PluginBrowserInfo::fromPluginDescription(desc));
+        if (queriedPlugins.size() == static_cast<std::size_t>(pluginTypes.size())) {
```
lucaromagnoli: This closes the empty-result hole, but the all-or-nothing equality can wedge into a permanent fallback, and the way in is this PR's own migration path.

`createIdentifierString()` is not injective over `KnownPluginList` entries. JUCE builds it as `format + name + hex(fileOrIdentifier.hashCode()) + hex(idToUse)` with `idToUse = uniqueId != 0 ? uniqueId : deprecatedUid`, while `KnownPluginList`'s dedupe key is `(fileOrIdentifier, deprecatedUid, uniqueId)` — the distinction your own `test_known_plugin_list_duplicates.cpp` was written to document. So a pair differing only in `uniqueId == 0` against `uniqueId == X`, with `deprecatedUid == X` on both, is two distinct types in the list but a single identifier string. That is the stale-XML-meets-fresh-scan pair, and `importLegacyFilesOnce()` now pulls exactly that XML in.

When it happens: `plugin_key TEXT PRIMARY KEY` collapses both into one row, so the DB cannot represent them separately; `preferredKeys` collapses to one entry while `pluginTypes.size()` still counts two; the equality can never hold, and every refresh for the rest of the session silently takes the `KnownPluginList` path.

The same wedge happens without any collision, just from drift — a scan whose save was skipped by the `pluginMetadataLoaded_` guard, or a row nulled by an earlier rescan. The direction is safe, since the fallback is the old correct behaviour, but the DB path becomes dead weight announced only by a DBG line.

A per-key merge removes the cliff: build an `unordered_map<juce::String, const PluginMetadataRecord*>` from `records`, then iterate `pluginTypes` and use the record when the key is present, `PluginBrowserInfo::fromPluginDescription(description)` otherwise. Still O(n), never drops a plugin, never falls back wholesale, and the collision case degrades to two entries sharing metadata instead of disabling the feature.

lucaromagnoli (commented):
Everything from the last pass is in. The `static_assert` is a stronger guard than the comment I suggested, since bumping the constant now fails the build rather than relying on someone reading a note. Two unprompted improvements worth calling out: the `busy_timeout` bump to 60000 lines up with `MediaDatabase`, and the `queriedPlugins` staging vector incidentally fixed a latent bug from the previous commit, where an exception thrown partway through the record loop would have left entries in `plugins_` and then had the fallback append the whole list again on top.

I was going to flag that `PRAGMA journal_mode=WAL` still sat outside the serializing transaction, but `3d2ff803` got there first. I checked every `PluginMetadataStore` construction site and they are all in the main app (the scan worker is a separate process but never opens the store), so a process-wide mutex is the right scope and the comment describes it accurately.

One finding left, inline.

Issue #1885, opened 2026-07-27
Close the AudioEngine abstraction gap (57 direct TracktionEngineWrapper users)


lucaromagnoli:
Part of #1882 (Phase 0).

The abstract AudioEngine interface (engine/AudioEngine.hpp) exists with two implementors, but 57 files still name TracktionEngineWrapper directly (roughly a 50/50 split with proper AudioEngine usage).

- Migrate callers to AudioEngine; extend the interface where callers genuinely need more (known leak: OfflineRenderHelper returns tracktion::Renderer::Parameters and needs an engine-neutral render-request type).
- Acceptance: TracktionEngineWrapper is named only inside engine/ and app wiring; UI, core, api, agents, cli talk to AudioEngine.

lucaromagnoli:
One additional acceptance criterion: make the abstraction musical-time-first, not merely free of `TracktionEngineWrapper` names.

The current `AudioEngine` boundary still exposes seconds for locate, loop regions, transport position, and session playheads, and it imports UI-owned `TransportStateListener`. That would preserve a seconds-first and UI-coupled contract for the native implementation.

Suggested invariant:

> Musical timeline state crosses the engine boundary in beats. Render-domain calculations use integer samples internally. Persisted seconds and repeated beat↔second round-trips are prohibited.

Acceptance additions:
- Engine-facing transport, loop, and launcher APIs use beat-domain value types.
- The engine interface has no `ui/` includes.
- Seconds/samples are exposed only through explicit hardware/source/render-domain types, rather than untyped `double`.
- TE-specific seconds conversion remains inside the TE adapter until cutover.

Issue #1880, opened 2026-07-26
[Bug]: Export Audio : Powered off devices in device panel become activated when rendering Audio.


mikobuntu:
### What happened?

If I use the power off button either on the device itself of at the master control for the whole track as soon as i Export my song it becomes activated and is included in the export.

After the song export the GUI itself ( the power button ) do not change state ( they still show as powered off ), but now when i play back my song in MAGDA these powered off devices are now fully active.

### Steps to reproduce

[1] Create 1 track and add for example a drumloop
[2] Add an internal delay to this track
[3] turn off the delay device using it's power button at top of UI
[4] Export the loop/song
[5] playback the exported song, you will hear that the delay is also rendered
[6] play your song again in MAGDA and you will hear the delay is now active on your drumloop
[7] You will notice that the power button is still showing the off state, but the delay effect is running

### Expected behavior

Any devices that are powered off, should remain in the same state during and after exporting audio

### MAGDA version

0.16.1

### Operating system

Linux

### Crash log (if applicable)

```text

```

PR #1909, opened 2026-07-28
refactor: close AudioEngine abstraction gap (#1885)

lucaromagnoli:


magda/daw/engine/OfflineRenderHelper.cpp
```diff
@@ -28,23 +30,125 @@ void restorePluginsAfterOfflineRender(TracktionEngineWrapper& engine) {
...
+        params_.checkNodesForAudio = request_.checkNodesForAudio;
+        params_.realTimeRender = request_.realTimeRender;
+        params_.time =
+            tracktion::TimeRange(tracktion::TimePosition::fromSeconds(request_.startSeconds),
+                                 tracktion::TimePosition::fromSeconds(request_.endSeconds));
+        params_.endAllowance = tracktion::TimeDuration::fromSeconds(request_.endAllowanceSeconds);
+
+        const auto allTracks = tracktion::getAllTracks(edit_);
+        auto* bridge = engine_.getAudioBridge();
+        if (!request_.trackIds.empty()) {
```
lucaromagnoli: A non-empty `trackIds` whose entries do not resolve leaves `tracksToDo` empty, and TE reads an empty bitset as *no filter* rather than *no tracks* (`tracktion_Renderer.cpp`: `cnp.allowedTracks = r.tracksToDo.isZero() ? nullptr : &tracksToDo;`). So a request that asks for one specific track and fails to find it renders the entire mix instead.

That case is reachable. Every call site that previously built the bitset by hand guarded it explicitly and skipped:

```cpp
// RenderTimeSelectionCommand, before
auto* teTrack = bridge->getAudioTrack(trackId);
if (!teTrack) continue;
if (trackIndex < 0) continue;
```

Now `RenderTimeSelectionCommand` writes a per-track stem file that is silently the full mix, and `OfflineMixAnalysis` does the same for its stems, which then feed the masking measurements. `bridge == nullptr` hits the same path via the `break` on line 70.

Suggest failing loudly: if `!request.trackIds.empty()` and no bit ends up set, have `createOfflineRenderTask` return `nullptr` so callers take their existing failure path.

magda/daw/audio/analysis/OfflineMixAnalysis.cpp
```diff
@@ -187,18 +167,19 @@ class AnalysisJob : public juce::Thread {
...
         // one master pass plus one pass per track.
-        std::vector<tk::AudioTrack*> teTracks;
+        std::vector<TrackId> trackIds;
         if (deep) {
             if (request_.trackSet.empty()) {
-                for (auto* track : tk::getAudioTracks(edit_))
-                    teTracks.push_back(track);
-            } else if (auto* bridge = engine_.getAudioBridge()) {
+                for (const auto& track : TrackManager::getInstance().getTracks())
+                    if (track.type == TrackType::Audio)
```
lucaromagnoli: This narrows the default track set. The old code used `tk::getAudioTracks(edit_)`, which is every TE `AudioTrack` in the edit. `track.type == TrackType::Audio` is a different set: `Group`, `Aux` and `MultiOut` tracks are all TE `AudioTrack`s carrying a different `TrackType`, so they drop out of a deep analysis entirely.

Aux returns are the ones that matter here. A reverb return no longer produces a stem, so it is invisible to the masking measurements, which is most of the point of the deep pass.

Since `TrackType` is cosmetic metadata rather than an engine distinction, the filter probably wants to be an exclusion of the chord track (which the old code did include, and arguably should not) instead of an inclusion of one type.

magda/daw/audio/analysis/OfflineMixAnalysis.cpp
```diff
@@ -41,74 +36,67 @@ juce::String formatSeconds(double seconds) {
...
+    request.destination = outFile;
+    request.useMasterPlugins = useMasterPlugins;
+    if (trackId)
+        request.trackIds = {*trackId};
+
+    // Creating and destroying a render task changes transport/plugin state, so
+    // keep those lifecycle operations on the message thread. Only the actual
+    // render loop runs on this analysis worker.
+    std::unique_ptr<OfflineRenderTask> task;
+    if (!runOnMessageThreadBlocking(
```
lucaromagnoli: The render task lifecycle moved from once-per-run to once-per-pass, and that is a real change in what happens between stems.

The old `AnalysisJob` ctor held one `ReallocationInhibitor`, called `prepareForRendering()` once, and set `setOfflineRenderActive(true)` once for the whole run: master pass plus every stem. Now each `renderPass` builds a task (stop transport, free playback context, prepare plugins, gate playback) and destroys it (restore plugins, `ensureContextAllocated()`, ungate playback) around a single pass.

On a 20-track deep analysis that is 21 playback-context free/realloc cycles and 21 plugin prepare/restore cycles instead of one. It also reopens the playback gate between every pass, which the removed comment described precisely:

```
// Block the transport for the render's lifetime: starting playback while
// the render owns the edit corrupts the node graph (NodeRenderContext
// asserts). Cleared in run()'s completion.
```

Hitting space between two stem passes now lands in that window.

A run-scoped session object that the per-pass tasks attach to would keep the engine-neutral shape without the churn.

magda/daw/audio/analysis/OfflineMixAnalysis.cpp
```diff
@@ -41,74 +36,67 @@ juce::String formatSeconds(double seconds) {
...
+    }
+
+    juce::WaitableEvent completed;
+    if (!juce::MessageManager::callAsync([&operation, &completed]() {
+            operation();
+            completed.signal();
+        }))
+        return false;
+
+    completed.wait();
```
lucaromagnoli: Unbounded `wait()` on a background thread, blocking on the message thread. `callAsync` returning `true` only means the message was posted, not that the message thread will ever pump it. During shutdown, or while the message thread is blocked on something that is itself waiting on this job, this hangs forever rather than failing.

A timeout plus a bail-out on expiry would keep the analysis worker killable.

magda/daw/audio/ClipCommands.cpp
```diff
@@ -1449,50 +1376,23 @@ void RenderTimeSelectionCommand::execute() {
...
+        OfflineRenderRequest request;
+        request.destination = trackState.renderedFile;
+        request.bitDepth = project.renderBitDepth;
+        request.sampleRate = project.sampleRate;
+        request.usePlugins = false;
+        request.useMasterPlugins = false;
+        request.startSeconds = startTime_;
+        request.endSeconds = endTime_;
+        request.trackIds = {trackId};
+        request.resumePlaybackAfterRender = true;
```
lucaromagnoli: This is inside the per-track loop, so playback restarts after every track and is stopped again for the next one. The old code stopped once before the loop and called `transport.play(false)` once after it.

For a multi-track selection that is an audible start/stop per track, plus a playback-context free/realloc cycle each time (see the related comment on `OfflineMixAnalysis`). Setting `resumePlaybackAfterRender` only on the final iteration, or better, hoisting the transport handling out of the loop, restores the old behavior.

magda/daw/audio/ClipCommands.cpp
```diff
@@ -2153,7 +2048,9 @@ void BounceInPlaceCommand::execute() {
     // project (e.g. on the Desktop) where the project can't track it.
     auto bouncesDir = ProjectManager::getInstance().getBouncesDirectory();
     if (bouncesDir == juce::File())
-        bouncesDir = edit->editFileRetriever().getParentDirectory().getChildFile("bounces");
+        bouncesDir =
```
lucaromagnoli: This fallback only runs when `getBouncesDirectory()` is empty, which is exactly the case where there is no saved project, and therefore where `getCurrentProjectFile()` is an invalid `juce::File`. `juce::File().getParentDirectory().getChildFile("bounces")` is not a usable path.

The old fallback used `edit->editFileRetriever().getParentDirectory()`, which always resolved to a real directory. The writability check below turns this into a clean error rather than a crash, so the effect is that bouncing in an unsaved project stops working instead of landing next to the edit file. Same code in `BounceToNewTrackCommand`.

magda/daw/engine/AudioEngine.hpp
```diff
@@ -1,19 +1,111 @@
...
 #include <string>
 #include <unordered_map>
+#include <utility>
+#include <vector>
 
 #include "../audio/midi/RecordingNoteQueue.hpp"
 #include "../core/ClipTypes.hpp"
+#include "../core/ParameterDetector.hpp"
 #include "../core/TempoMap.hpp"
 #include "../ui/state/TransportStateListener.hpp"
```
lucaromagnoli: This is the `ui/` include the acceptance criterion rules out. `AudioEngine` inherits `AudioEngineListener` from here, so the engine interface still depends on a UI-owned header, and any non-TE implementor inherits that coupling.

Moving `AudioEngineListener` (or just the transport-state part of it) into the engine or core layer would close this without touching any call site.

magda/daw/engine/AudioEngine.hpp
```diff
@@ -105,6 +207,9 @@ class AudioEngine : public AudioEngineListener {
     virtual void setLooping(bool enabled) = 0;
     virtual void setLoopRegion(double startSeconds, double endSeconds) = 0;
     virtual bool isLooping() const = 0;
+    virtual std::pair<double, double> getLoopRegionSeconds() const {
```
lucaromagnoli: New seconds-domain surface on the engine boundary: `getLoopRegionSeconds()` here, `getEditLengthSeconds()` above, and `startSeconds`/`endSeconds`/`endAllowanceSeconds` on `OfflineRenderRequest`.

The acceptance criterion asks the opposite direction: loop and transport APIs in beat-domain value types, with seconds confined to explicit render/hardware/source types. Render range is arguably legitimately render-domain, but it should say so through a typed render-domain value rather than a bare `double`, and the loop region is squarely musical timeline state.

The untyped `std::pair<double, double>` also makes the unit invisible at the call site. `MainWindowExport.cpp` immediately does `loopRange.first * projectTempo / 60.0`, which is a seconds-to-beats round-trip at the UI layer, exactly what the coordinate-system rule prohibits.

magda/daw/engine/AudioEngine.hpp
```diff
@@ -122,14 +227,107 @@ class AudioEngine : public AudioEngineListener {
...
+    virtual const AudioBridge* getAudioBridge() const = 0;
 
     // ===== MIDI Management =====
-    virtual class MidiBridge* getMidiBridge() = 0;
-    virtual const class MidiBridge* getMidiBridge() const = 0;
+    virtual MidiBridge* getMidiBridge() = 0;
+    virtual const MidiBridge* getMidiBridge() const = 0;
+
+    // ===== Application Services =====
+    virtual MagdaApi& getMagdaApi() {
```
lucaromagnoli: Most of the roughly 40 methods added to this interface come with a default body that returns `{}`, `nullptr`, `false` or does nothing, and this one throws.

That inverts what the abstraction is for. The point of extracting `AudioEngine` is that a second implementor cannot be written without confronting every capability the app depends on. With defaults, a native engine compiles clean and silently returns nothing: no plugin list, no groove templates, no render task, no sampler media, no wave channels. The failures surface as empty UI and dead menu items rather than as a build error naming what is missing.

Pure virtual for anything a real engine must provide, defaults reserved for genuinely optional capabilities, would make the remaining gap visible and measurable. Throwing from a getter is also awkward for a caller that just wants to know whether the facility exists.

magda/daw/engine/AudioEngine.hpp
```diff
@@ -122,14 +227,107 @@ class AudioEngine : public AudioEngineListener {
...
+        return nullptr;
+    }
+
+    // ===== Project Media =====
+    virtual std::vector<SamplerMediaReference> getSamplerMediaReferences() {
+        return {};
+    }
+
+    // ===== Edit-Wide Tempo Sequences =====
+    virtual std::unique_ptr<UndoableCommand> createTempoSequenceRippleCommand(
```
lucaromagnoli, line 285: `double startBeat, double endBeat` is the untyped-`double` case the acceptance criterion calls out, just on the beat side rather than the seconds side. This is a new API, so it can take the beat-domain value type directly instead of establishing another bare-`double` boundary that later has to be migrated.

magda/daw/cli/magda_cli_main.cpp
```diff
@@ -49,8 +43,9 @@ juce::File defaultOutputFor(const juce::File& input) {
 class HeadlessEngineSession {
   public:
     bool initialize() {
-        engine_ = std::make_unique<magda::TracktionEngineWrapper>();
-        engine_->setForceHeadless(true);
+        auto engine = std::make_unique<magda::TracktionEngineWrapper>();
```
lucaromagnoli: The acceptance list names `cli` among the layers that should talk to `AudioEngine`, and everything downstream of here now does. Only construction still needs the concrete type, because `setForceHeadless` is not on the interface.

`createDefaultAudioEngine()` already exists for exactly this. Giving it a headless option would let the CLI drop the last `TracktionEngineWrapper` reference outside `engine/`.

magda/daw/engine/TracktionEngineWrapperPlugins.cpp
```diff
@@ -566,4 +580,228 @@ void TracktionEngineWrapper::detectNewPlugins(
     });
 }
 
+bool TracktionEngineWrapper::isPluginScanRunning() const {
+    return isScanning_ || (pluginScanCoordinator_ && pluginScanCoordinator_->isScanning());
+}
+
+std::vector<ExcludedPlugin> TracktionEngineWrapper::getExcludedPlugins() const {
+    auto* self = const_cast<TracktionEngineWrapper*>(this);
```
lucaromagnoli: `const_cast<TracktionEngineWrapper*>(this)` to call a non-const accessor, same again in `getPluginScanReportFile()`. If `getPluginScanCoordinator()` lazily creates the coordinator then these methods are not really const and the interface should say so; if it does not, a const overload is cheaper than the cast.

magda/agents/dsl_interpreter.cpp
```diff
@@ -1371,8 +1357,7 @@ bool Interpreter::executeAddFx(const Params& params) {
         if (nameMatch) {
             // Filter by format hint if provided
             if (formatHint.isNotEmpty()) {
-                auto descFormat = desc.pluginFormatName.toLowerCase();
-                if (!descFormat.contains(formatHint.toLowerCase()))
+                if (!desc.getFormatString().containsIgnoreCase(formatHint))
```
lucaromagnoli: The format hint now matches against `DeviceInfo::getFormatString()`, which returns `"AU"`, where it previously matched `juce::PluginDescription::pluginFormatName`, which is `"AudioUnit"`.

The grammar documents `VST3, AU, VST` so the documented values still work, but `format="AudioUnit"` used to match and now silently filters everything out. Models emit the long form fairly often. Matching on `maybePluginFormatFromName(formatHint)` against `desc.format` would accept both spellings and drop the substring comparison, which also happens to make `format="VST"` match VST3.

magda/daw/ui/views/SessionView.cpp
```diff
@@ -3671,31 +3666,28 @@ void SessionView::timerCallback() {
...
-                double pos = (atPos >= 0.0) ? atPos : transport.getPosition().inSeconds();
-                const double projectBpm = isValidBpm(bpm) ? bpm : DEFAULT_BPM;
-                double beatDuration = 60.0 / projectBpm;
-                double beatPhase = std::fmod(pos, beatDuration) / beatDuration;
-                newBlinkOn = (beatPhase < 0.5);
-                posBeats = pos * projectBpm / 60.0;
-            }
+        if (audioEngine_->isPlaying()) {
+            transportPlaying = true;
+            const double bpm = audioEngine_->getTempo();
```
lucaromagnoli: `getTempo()` returns the BPM at the current playhead (`tempoSequence.getBpmAtBeat(getPositionBeats())`), where the old code read `getBpmAt(TimePosition())`, the BPM at the start of the edit.

For the blink phase the local tempo is arguably the better answer. For `posBeats = pos * projectBpm / 60.0` two lines down it is not: multiplying elapsed seconds by the tempo *at the playhead* is wrong under any tempo map, and now wrong in a position-dependent way. That conversion should go through the engine's beats/seconds facade rather than a bare multiply.

lucaromagnoli (commented):
Reviewed against #1885 and the acceptance comment on it.

The name migration lands well: direct `TracktionEngineWrapper` users drop from 71 files to 27, and what remains is `engine/`, app wiring and tests. `OfflineRenderHelper` no longer leaks `tracktion::Renderer::Parameters`, which was the known leak called out in the issue. The `OfflineRenderTask` shape is the right abstraction.

Two things from the acceptance list are not met yet:

- **No `ui/` includes in the engine interface.** `AudioEngine.hpp` still includes `../ui/state/TransportStateListener.hpp`.
- **Musical-time-first.** The boundary gets wider in seconds rather than narrower: `getEditLengthSeconds()`, `getLoopRegionSeconds()`, and `OfflineRenderRequest::{startSeconds, endSeconds, endAllowanceSeconds}` are all new, alongside the existing `locate(double)` and `setLoopRegion(double, double)`. `createTempoSequenceRippleCommand` passes beats as untyped `double`, which the criterion rules out for both domains.

The substantive review comments are inline. The one I would fix before merging is the track-filter resolution in `OfflineRenderHelper.cpp`: a non-empty `trackIds` that fails to resolve silently degrades to rendering every track, and three call sites that previously guarded against it no longer do.

Two behavior changes that are probably intentional but are not mentioned anywhere, so they read as accidental:

- `OfflineMixAnalysis` no longer force-enables disabled plugins before rendering. That matches the #1880 rule that `prepareEditForOfflineRender` documents, but it changes what a deep analysis measures.
- In both Bounce commands the insert-capture pass and the FX bypass now run before the transport is stopped, since stopping moved into the render-task constructor. Worth confirming the capture pass still behaves the same with a live playback context.

On testing: only `tests/MockMagdaApi.hpp` changed. Nothing covers `createOfflineRenderTask`, and the track-filter case above is exactly the kind of silent degradation a model-level test would catch cheaply.

magda/daw/engine/AudioEngine.hpp
```diff
@@ -30,6 +133,9 @@ class AudioEngine : public AudioEngineListener {
     // ===== Lifecycle =====
     virtual bool initialize() = 0;
     virtual void shutdown() = 0;
+    virtual bool hasActiveEdit() const = 0;
+    virtual BeatDuration getEditLengthBeats() const = 0;
+    virtual juce::File getEditFile() const = 0;
 
     // ===== Transport =====
     virtual void play() = 0;
```
lucaromagnoli, line 141: Loop region, edit length, the ripple command and both render ranges are all beat-domain or explicit render-domain now, so the bulk of the criterion is met.

What is left in this section is the part the issue comment named first: `locate(double positionSeconds)`, `getCurrentPosition()`, `getSessionPlayheadPosition()` and `getActiveClipPlayheadPositions()` are still untyped seconds crossing the boundary. `getAudioThreadTransportSeconds()` is fine as-is since it is explicitly a sampled render-domain value, though it could take the same treatment as `RenderTimePosition` for consistency.

Not asking for it in this PR, it is a wider change than what is already here. Worth carrying forward as a follow-up so Phase 0 does not close with the transport half-migrated.

magda/daw/engine/OfflineRenderHelper.cpp
```diff
@@ -12,12 +12,48 @@ void prepareEditForOfflineRender(tracktion::Edit& edit) {
...
-    // afterwards.
+    // so powered-off devices must stay off through the render and afterwards.
 }
 
+std::optional<juce::BigInteger> resolveOfflineRenderTrackFilter(
+    const OfflineRenderRequest& request, int numTracks,
+    const std::function<int(TrackId)>& indexForTrack) {
+    juce::BigInteger result;
+
+    if (!request.trackIds.empty()) {
```
lucaromagnoli, line 25: The dangerous direction is handled: zero resolved tracks now returns `nullopt` rather than silently becoming Tracktion's all-tracks sentinel.

Partial resolution still passes through quietly, though. `trackIds = {20, 999}` renders track 20 alone and reports success, which the second test case pins as intended behavior. A caller that asked for two tracks and got one has no way to tell.

That is the safe direction and nothing in tree currently passes multi-track sets, so this is fine as it stands. If a multi-track request ever appears, returning the resolved count, or logging what was dropped, would keep it from being invisible.

lucaromagnoli (commented):
Re-reviewed at 9f052c99. Every point from the previous pass is addressed, and several of the fixes go further than what I asked for. Verified individually:

- The track filter now lives in `resolveOfflineRenderTrackFilter` returning `std::optional`, so an explicit filter that resolves to nothing fails instead of broadening to every track. The same guard was added for `clipIds`, which I had missed. Four Catch2 cases cover it, including the exclusion-becomes-sentinel direction.
- `OfflineRenderSession` is the right split. Transport, playback context, plugin preparation and the playback gate are now owned per run rather than per pass, and `RenderTimeSelectionCommand` creates one session outside its loop.
- The analysis track set is `!= TrackType::Chord`, which restores the old `getAudioTracks` membership and drops the chord track, which arguably should never have been in it.
- `runOnMessageThreadBlocking` now times out. The Queued/Running split is a good call: cancelling a queued callback is safe, waiting out a running one keeps its captures alive, and the shutdown deadlock is the queued case.
- `getEditFile()` restores the original bounce fallback path, and `PlaybackResumeScope` puts the transport stop back ahead of the capture pass and the FX bypass.
- `AudioEngineListener` moving into `engine/` clears the `ui/` include, and the pure-virtual pass means the stub engine in `test_project_serialization.cpp` now has to implement everything, which is exactly the signal that was missing.
- `BeatRange`/`BeatPosition`/`BeatDuration` for musical state with `RenderTimeRange` at the render boundary is the shape the acceptance criterion was asking for. MIDI export losing its seconds-to-beats round-trip and the SessionView blink phase going through `tempoMap->timeToBeat()` are both improvements over what was there before this PR.

All checks green including Windows.

Two residual notes inline, neither blocking. Beyond those, `TransportInterface` still carries `setLoopRegion(double, double)` and reaches `ArrangementView`, so there is a second seconds-domain transport boundary the interface work has not touched. Out of scope here, but worth a line in whatever tracks the rest of Phase 0.

PR #1912, opened 2026-07-28
Fix Drum Grid sampler voice mode restore

### LINKED ISSUE #1910: [Bug]: Drum Grid: Voice modes always default to Poly
### What happened?

When trying to change the Drum Grid's voice to Mono or Legato, they automatically get set back to poly as can be seen in my screen recording.

https://github.com/user-attachments/assets/4a4b1138-f29e-4131-bbb7-e36c054b2d17

### Steps to reproduce

[1] Drag a sample to the timeline
[2] Context click the Audio Clip and choose 'Slice at Grid to Drum Grid'
[3] Select the new track that is created which opens the Drum Grid in the bottom panel
[4] For each slice change the default VOICE setting to Mono
[5] click back through all your slices and here you will see that the VOICE mode has been reset to Poly

### Expected behavior

The chosen VOICE mode should not default to Poly

### MAGDA version

0.17.0-rc0

### Operating system

Linux

### Crash log (if applicable)

```text
*
```

lucaromagnoli:


magda/daw/ui/components/chain/drum_grid/PadDeviceSlot.cpp
```diff
@@ -288,7 +288,8 @@ void PadDeviceSlot::setupForSampler(daw::audio::MagdaSamplerPlugin* sampler) {
         sampler->releaseValue.get(), sampler->pitchValue.get(), sampler->fineValue.get(),
         sampler->levelValue.get(), sampler->sampleStartValue.get(), sampler->sampleEndValue.get(),
         sampler->loopEnabledValue.get(), sampler->loopStartValue.get(), sampler->loopEndValue.get(),
-        sampler->velAmountValue.get(), sampleName, sampler->getRootNote());
+        sampler->velAmountValue.get(), sampleName, sampler->getRootNote(),
+        sampler->voiceModeValue.get(), sampler->glideValue.get());
```
lucaromagnoli, line 292: This call site was able to break silently because the three trailing parameters carry defaults: `int rootNote = 60, float voiceMode = 0.0f, float glide = 0.0f` (SamplerUI.hpp:48-50). Substituting plausible values in a restore function means a missed argument compiles clean and the resulting Poly looks correct.

Dropping those three defaults would have made this a compile error, and would also flag the same gap still present at DeviceCustomUIManager.cpp:1883. With 16 positional arguments, 10 of them bare floats, a small params struct is the more durable fix, but that is beyond this PR.

lucaromagnoli (commented):
Traced the fix end to end and it is correct. The write path already persisted the value (SamplerUI::setVoiceMode -> onParameterChanged(12) -> syncCachedValueFromParam -> voiceModeValue), and both properties are in the copyPropertiesToCachedValues list in restorePluginStateFromValueTree, so only the read-back was dropping them. Reading the CachedValue here also matches how the other 13 arguments in the same call are sourced.

Two follow-ups worth folding in while the branch is open. Neither blocks the merge.

1. The same omission is still live in `DeviceCustomUIManager::executeSamplerCommand` (DeviceCustomUIManager.cpp:1883). That call also stops at `sampleName`, so loading a sample into the standalone Sampler resets the displayed root note to 60, voice mode to Poly and glide to 0. Lower severity than #1910: it is display only, since every setValue on that path uses dontSendNotification, and it self-heals on the next `DeviceCustomUIManager::update()`, which does pass all 16 arguments. Still the identical mistake.

2. The default arguments are what allowed this to happen. See the inline note.

Unrelated but inside the function being touched: PadDeviceSlot.cpp:253 uses `setParameter` rather than `setParameterFromHost`. `PadDeviceSlot::getLinkableSliders` wires these sampler sliders for mod/macro linking, so a pad parameter with an active LFO or macro will silently drop UI writes. Pre-existing, not introduced here.

No test added, and I do not think one is available: this is a dropped argument at a UI call site, out of reach of a model-level test. Removing the defaults is the coverage.

lucaromagnoli:
Closes #1910

magda/daw/ui/components/chain/drum_grid/PadDeviceSlot.cpp
```diff
@@ -250,7 +250,7 @@ void PadDeviceSlot::setupForSampler(daw::audio::MagdaSamplerPlugin* sampler) {
     samplerUI_->onParameterChanged = [sampler](int paramIndex, float value) {
         auto params = sampler->getAutomatableParameters();
         if (paramIndex >= 0 && paramIndex < params.size()) {
-            params[paramIndex]->setParameter(value, juce::sendNotification);
+            params[paramIndex]->setParameterFromHost(value, juce::sendNotificationSync);
```
lucaromagnoli, line 253: This change is correct, and it makes the next line worth a look.

`syncCachedValueFromParam` reads `params[paramIndex]->getCurrentValue()` (MagdaSamplerPlugin.cpp:626), which is the modulated value, not the base. `PadDeviceSlot::getLinkableSliders` wires these sampler sliders for mod/macro linking, so with an LFO or macro on a pad parameter the write now lands on the parameter and the very next line persists base + modulation into the CachedValue. The saved base then drifts by whatever the modifier offset happened to be at that instant, and that is what `restorePluginStateFromValueTree` pushes back on reload.

Strictly pre-existing: the old `setParameter` dropped the parameter write while this sync still ran. But switching to `setParameterFromHost` is what makes editing a modulated pad parameter a working path, so the drift goes from mostly unreachable to routine.

Every other read in the codebase uses `getCurrentBaseValue()` (ParameterInfoBuilder.cpp:22, AutomationManager.cpp:103, AutomationPlaybackEngine.cpp:713). Fix is that one word at MagdaSamplerPlugin.cpp:626. Simpler still, this lambda already holds the exact base value the user set, so it could pass `value` in and skip the read-back entirely.

lucaromagnoli (commented):
Re-reviewed at 17a8b296. All three follow-ups landed correctly.

On removing the defaults, I checked repo-wide that `SamplerUI::updateParameters` has exactly three call sites (PadDeviceSlot.cpp:286, DeviceCustomUIManager.cpp:1883 and :2294) and all three now pass 16 arguments, so this compiles. The wrap points at all three sites fit the 100 column ColumnLimit, so the format check should stay green.

On `setParameterFromHost(value, juce::sendNotificationSync)`: consistent with the sibling call site in the same file (`setupForSharedDeviceUi`), and passing the raw `value` is right here since TE parameters take their own range, so no `modelToTeValue` conversion is needed.

One new finding that the setParameterFromHost switch surfaces. See the inline note on PadDeviceSlot.cpp:253. Not introduced by this PR and not a merge blocker.

Issue #1352, opened 2026-05-29
Menu shortcut hints should read from ApplicationCommandManager


lucaromagnoli:
Menu items show shortcut hints as **hardcoded strings** per platform in `magda/daw/ui/windows/MenuManager.cpp` (e.g. `⌘Z` on macOS, `Ctrl+Z` elsewhere). Once users can remap keys (#20), these hints will be wrong.

## Work
- Derive each menu item's hint from the command's current `KeyPress` via `ApplicationCommandManager::getKeyMappings()` / `ApplicationCommandInfo`, rendered with `KeyPress::getTextDescriptionWithIcons()`.
- Drop the hardcoded `\t…` suffixes.
- Ideally drive menu items from `CommandID`s directly so enablement + hint + dispatch share one source of truth (currently `MenuManager` uses a separate `MenuIDs` enum + callbacks that re-dispatch via `invokeDirectly`).

Depends on #20.

Issue #20, opened 2026-01-17
Keyboard binding system


lucaromagnoli:
## Goal
Make keyboard shortcuts **user-remappable and persistent**. The defaults already exist (`getCommandInfo()` calls `addDefaultKeypress`), and `ApplicationCommandManager::getKeyMappings()` is wired as the key listener — but user remaps are never saved or reloaded.

## Scope
- On startup, after `registerAllCommandsForTarget`, restore user overrides via `KeyPressMappingSet::restoreFromXml()` from `Config` (#22); fall back to defaults otherwise.
- On change, `createXml()` → `Config.setKeyboardBindings()` → `save()`.
- Conflict detection when a key is already bound (surface to the editor #862).
- Reset-to-defaults (`resetToDefaultMappings()`).

## Files
- `magda/daw/ui/windows/MainWindow.cpp` (L359-379 setup), `MainWindowCommands.cpp`, `CommandIDs.hpp`.

Depends on #22. Enables #1351 (promote hardcoded fallbacks), #1352 (dynamic menu hints), #862 (editor UI), #25, #23. Part of epic:command-registry.

PR #1914, opened 2026-07-28
Fix File menu keyboard shortcuts

### LINKED ISSUE #1911: [Bug]: Shortcuts -> File menu: Shortcut combos aren't working for any of the File operations
### What happened?

I have noticed that shortcuts don't work for all of the File operations, example Save project , Open project , Export Audio etc.

Here are my shortcuts:- 

Image

### Steps to reproduce

[1] The easiest method to test this is try to use your shortcut combo for `Save Project As ...` which by default is `CTRL + SHIFT + S`

Here you will notice that no save dialogue is opened.

### Expected behavior

All shortcut combos should work for File menu.

### MAGDA version

0.17.0

### Operating system

Linux

### Crash log (if applicable)

```text
*
```

lucaromagnoli:
Fixes #1911. Scoped to the File menu shortcut bugs only — Save Project, Save As and Export Audio.

New Project and Open Project are deliberately **not** here. They had no default keypress, so wiring them up is a new capability rather than a fix, and they are what pulled in the unsaved-changes work. They moved to #1916 (stacked on this branch), along with the other newly bindable commands. The unsaved-changes bug is its own PR off `main`.

## 1. File menu shortcuts never fired

Save Project, Save As and Export Audio were fully registered: `getAllCommands()` returned them and `getCommandInfo()` gave them names, the "File" category and keypresses (`saveProject` = Cmd/Ctrl+S, `saveProjectAs` = Cmd/Ctrl+Shift+S). So they appeared in the shortcuts editor and looked bound. But `perform()` had no case for them, so each fell through to `default: return false` — JUCE reads that as "not handled" and nothing happens.

Menu clicks kept working because `MenuManager::menuItemSelected()` is a separate path off its own `menuItemID` enum.

`perform()` now routes into `MenuManager::invokeApplicationCommand()`, the same callbacks a menu click uses, so both paths converge on one action per command.

## 2. The File menu showed no shortcut hints

It was the only menu making no `keyHint()` calls, so even Save's Cmd/Ctrl+S was invisible next to its item. Combined with the dead keys above, the File shortcuts read as entirely absent.

## 3. Lint targets were unusable off macOS

`lint`, `lint-changed`, `lint-fix` and `lint-file` hardcoded `/opt/homebrew/opt/llvm/bin/clang-tidy`, so every one failed on Linux even with the distro package installed. A `CLANG_TIDY` variable now prefers the Homebrew binary when present and falls back to `PATH`, with install hints for both platforms when neither resolves.

## Testing

`tests/test_menu_manager.cpp` covers the File commands dispatching to their callbacks plus the unconfigured-callback and unhandled-command rejection paths.

Full suite passes: 1549 passed, 13 skipped, 0 failures. Debug app builds clean. clang-tidy reports nothing new on the changed lines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -411,21 +411,22 @@ lint-fix:
 		exit 1; \
 	fi
 	@echo "⚠️  This will modify your source files!"
-	@read -p "Continue? [y/N] " -n 1 -r; \
-	echo; \
-	if [[ $$REPLY =~ ^[Yy]$$ ]]; then \
-		find magda/daw -name "*.cpp" -type f -exec \
-			$(CLANG_TIDY) \
-			{} \
-			--config-file=.clang-tidy \
-			--format-style=file \
-			-p=$(BUILD_DIR) \
-			--fix \
-			--fix-errors \;; \
-		echo "✅ Fixes applied"; \
-	else \
-		echo "❌ Cancelled"; \
-	fi
+	@printf "Continue? [y/N] "; \
+	read REPLY; \
+	case "$$REPLY" in \
+		[Yy]*) \
+			find magda/daw -name "*.cpp" -type f -exec \
+				$(CLANG_TIDY) \
+				{} \
+				--config-file=.clang-tidy \
+				--format-style=file \
+				-p=$(BUILD_DIR) \
+				--fix \
+				--fix-errors \;; \
+			echo "✅ Fixes applied" ;; \
+		*) \
+			echo "❌ Cancelled" ;; \
+	esac
 
 # Lint specific file
 .PHONY: lint-file
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,15 @@ else
     LAUNCH             :=
 endif
 
+# clang-tidy lookup. Homebrew's LLVM is newer than the Apple toolchain's, so
+# prefer it when it is installed; otherwise take whatever is on PATH (the
+# distro package on Linux). Override with `make lint CLANG_TIDY=/path/to/it`.
+CLANG_TIDY ?= $(shell if [ -x /opt/homebrew/opt/llvm/bin/clang-tidy ]; then \
+		echo /opt/homebrew/opt/llvm/bin/clang-tidy; \
+	else \
+		command -v clang-tidy 2>/dev/null; \
+	fi)
+
 # Default target
 .PHONY: all
 all: debug
@@ -345,14 +354,15 @@ lint:
 		echo "❌ compile_commands.json not found. Run 'make debug' first."; \
 		exit 1; \
 	fi
-	@if [ ! -f /opt/homebrew/opt/llvm/bin/clang-tidy ]; then \
-		echo "❌ clang-tidy not found at /opt/homebrew/opt/llvm/bin/clang-tidy"; \
-		echo "Install with: brew install llvm"; \
+	@if [ -z "$(CLANG_TIDY)" ]; then \
+		echo "❌ clang-tidy not found on PATH or at /opt/homebrew/opt/llvm/bin"; \
+		echo "Install with: brew install llvm   (macOS)"; \
+		echo "              apt install clang-tidy   (Debian/Ubuntu)"; \
 		exit 1; \
 	fi
 	@echo "📋 Analyzing magda/daw sources..."
 	@find magda/daw -name "*.cpp" -type f -exec \
-		/opt/homebrew/opt/llvm/bin/clang-tidy \
+		$(CLANG_TIDY) \
 		{} \
 		--config-file=.clang-tidy \
 		--format-style=file \
@@ -368,13 +378,17 @@ lint-changed:
 		echo "❌ compile_commands.json not found. Run 'make debug' first."; \
 		exit 1; \
 	fi
+	@if [ -z "$(CLANG_TIDY)" ]; then \
+		echo "❌ clang-tidy not found on PATH or at /opt/homebrew/opt/llvm/bin"; \
+		exit 1; \
+	fi
 	@CHANGED_FILES=$$(git diff --name-only --diff-filter=d HEAD | grep '\.cpp$$' || true); \
 	if [ -z "$$CHANGED_FILES" ]; then \
 		echo "No modified .cpp files found"; \
 	else \
 		echo "Analyzing: $$CHANGED_FILES"; \
 		for file in $$CHANGED_FILES; do \
-			/opt/homebrew/opt/llvm/bin/clang-tidy \
+			$(CLANG_TIDY) \
 				$$file \
 				--config-file=.clang-tidy \
 				--format-style=file \
@@ -397,7 +411,7 @@ lint-fix:
 	echo; \
 	if [[ $$REPLY =~ ^[Yy]$$ ]]; then \
 		find magda/daw -name "*.cpp" -type f -exec \
-			/opt/homebrew/opt/llvm/bin/clang-tidy \
+			$(CLANG_TIDY) \
 			{} \
 			--config-file=.clang-tidy \
 			--format-style=file \
@@ -417,7 +431,11 @@ lint-file:
 		exit 1; \
 	fi
 	@echo "🔍 Analyzing $(FILE)..."
-	@/opt/homebrew/opt/llvm/bin/clang-tidy \
+	@if [ -z "$(CLANG_TIDY)" ]; then \
+		echo "❌ clang-tidy not found on PATH or at /opt/homebrew/opt/llvm/bin"; \
+		exit 1; \
+	fi
+	@$(CLANG_TIDY) \
 		$(FILE) \
 		--config-file=.clang-tidy \
 		--format-style=file \
--- a/Makefile
+++ b/Makefile
@@ -406,6 +406,10 @@ lint-fix:
 		echo "❌ compile_commands.json not found. Run 'make debug' first."; \
 		exit 1; \
 	fi
+	@if [ -z "$(CLANG_TIDY)" ]; then \
+		echo "❌ clang-tidy not found on PATH or at /opt/homebrew/opt/llvm/bin"; \
+		exit 1; \
+	fi
 	@echo "⚠️  This will modify your source files!"
 	@read -p "Continue? [y/N] " -n 1 -r; \
 	echo; \
--- a/magda/daw/ui/windows/MainWindowCommands.cpp
+++ b/magda/daw/ui/windows/MainWindowCommands.cpp
@@ -563,6 +563,11 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     };
 
     switch (info.commandID) {
+        case saveProject:
+        case saveProjectAs:
+        case exportAudio:
+            return MenuManager::getInstance().invokeApplicationCommand(info.commandID);
+
         case undo:
             UndoManager::getInstance().undo();
             return true;
--- a/magda/daw/ui/windows/MainWindowCommands.cpp
+++ b/magda/daw/ui/windows/MainWindowCommands.cpp
@@ -106,10 +106,14 @@ void MainWindow::MainComponent::getAllCommands(juce::Array<juce::CommandID>& com
         newAudioTrack, newMidiTrack, deleteTrack, duplicateTrackNoContent,
         duplicateTrackContentOnly, toggleMuteSelectedTracks, toggleSoloSelectedTracks,
         // View
-        zoom, toggleArrangeSession, cycleViewForward, cycleViewBackward, uiScaleUp, uiScaleDown,
+        toggleArrangeSession, cycleViewForward, cycleViewBackward, uiScaleUp, uiScaleDown,
         togglePianoRollFullscreen,
         // Help
-        showHelp, about};
+        about};
+
+    // `zoom` describes a group of View-menu controls, not one action, and
+    // `showHelp` is still a placeholder. Do not expose either as a bindable
+    // command until each has a concrete action to perform.
 
     commands.addArray(allCommands, juce::numElementsInArray(allCommands));
 }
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -392,12 +406,16 @@ lint-fix:
...
 		exit 1; \
 	fi
+	@if [ -z "$(CLANG_TIDY)" ]; then \
+		echo "❌ clang-tidy not found on PATH or at /opt/homebrew/opt/llvm/bin"; \
+		exit 1; \
+	fi
 	@echo "⚠️  This will modify your source files!"
 	@read -p "Continue? [y/N] " -n 1 -r; \
 	echo; \
 	if [[ $$REPLY =~ ^[Yy]$$ ]]; then \
```
lucaromagnoli: The clang-tidy path fix is right, but `lint-fix` still fails on Debian and Ubuntu for a second reason.

The Makefile sets no `SHELL`, so recipes run under `/bin/sh`, which is dash on those distros. Both `read -p ... -n 1 -r` and `[[ $REPLY =~ ... ]]` are bash-only, so the target dies right here, before the new `CLANG_TIDY` guard above ever gets a chance to help. It passes locally because macOS `/bin/sh` is bash in POSIX mode.

Either add `SHELL := /bin/bash` near the top of the Makefile, or make the prompt POSIX:

```make
	@printf "Continue? [y/N] "; \
	read REPLY; \
	case "$$REPLY" in \
		[Yy]*) find magda/daw -name "*.cpp" -type f -exec ... ;; \
		*) echo "Cancelled" ;; \
	esac
```

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -563,6 +563,11 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     };
 
     switch (info.commandID) {
+        case saveProject:
+        case saveProjectAs:
+        case exportAudio:
+            return MenuManager::getInstance().invokeApplicationCommand(info.commandID);
```
lucaromagnoli: This fixes the three cases, but nothing stops the bug class from coming back. The defect shape is `getAllCommands()` listing a command that `perform()` has no case for: it shows up bound in the shortcuts editor and silently does nothing.

Seven commands are still in that state after this PR: `about`, `deleteTrack`, `newProject`, `openProject`, `showHelp`, `toggleArrangeSession`, `zoom`. None has a default keypress today, so none is a live bug, but each becomes a dead binding the moment someone assigns a key in the editor, which is exactly the capability #20 and #1352 shipped.

With #1916 stacked on top adding more bindable commands, a test asserting that every ID in `getAllCommands()` is handled by `perform()` is worth more than the three cases added here.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -563,6 +563,11 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     };
 
     switch (info.commandID) {
+        case saveProject:
+        case saveProjectAs:
+        case exportAudio:
```
lucaromagnoli: `exportAudio` has no `addDefaultKeypress` in `getCommandInfo()`. By the rule the description uses to defer New Project and Open Project to #1916 (no default keypress, so wiring them up is a new capability rather than a fix) this one belongs there too. Both this case and the new `keyHint(CommandIDs::exportAudio)` call are inert until someone binds a key.

No objection to including it, the code is right either way. But the stated boundary does not hold, and "Export Audio shortcut never fired" overstates it. Only Save and Save As were live bugs.

magda/daw/ui/windows/MenuManager.cpp
```diff
@@ -38,6 +38,30 @@ void MenuManager::initialize(const MenuCallbacks& callbacks) {
...
+            break;
+        case CommandIDs::exportAudio:
+            callback = &callbacks_.onExportAudio;
+            break;
+        default:
+            return false;
+    }
+
+    if (!*callback)
+        return false;
```
lucaromagnoli: `false` here means "known command, callback never wired". `false` at the `default:` above means "unknown command". `perform()` collapses both into unhandled, so a File callback that nobody connects reproduces the exact symptom this PR fixes, with no signal anywhere.

A `jassertfalse` on this branch would surface that in debug instead of silently doing nothing.

tests/test_menu_manager.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "magda/daw/ui/windows/MenuManager.hpp"
+
+namespace {
+
+class MenuCallbacksGuard {
+  public:
+    explicit MenuCallbacksGuard(magda::MenuManager& menuManager) : menuManager_(menuManager) {}
+
+    ~MenuCallbacksGuard() {
+        menuManager_.initialize({});
```
lucaromagnoli, line 16: This clobbers rather than restores: it wipes whatever the singleton held instead of putting back what was there. The callbacks live in a process-wide singleton shared by every Catch2 case in the binary, so the blast radius is the whole target.

Harmless today because nothing else touches `MenuManager`. Saving and restoring would need a `callbacks()` accessor on `MenuManager`; if that is not worth adding, a one-line comment saying the reset to empty is deliberate would do.

tests/CMakeLists.txt
```diff
@@ -139,6 +139,8 @@ set(TEST_SOURCES
     test_gesture_router.cpp
     # Key mapping persistence tests (epic:command-registry -- #20)
     test_key_mapping_store.cpp
+    test_menu_manager.cpp
+    ../magda/daw/ui/windows/MenuManager.cpp
```
lucaromagnoli, line 144: Worth a comment line here. This is the first production source in `TEST_SOURCES` (every other `../magda/...` entry belongs to `magda_juce_tests` further down), and it sits between two test files that both carry a comment header.

Something like: `MenuManager.cpp` builds into `magda_daw_app`, not `magda_daw`, so `magda_tests` does not pick it up transitively.

lucaromagnoli (commented):
Reviewed at `ad3b31d`. The diagnosis is right and the fix is the correct shape: the menu-click path and the keypress path now converge on one callback per command instead of duplicating the invocation.

I checked the central claim across all 64 entries in `getAllCommands()`. `saveProject` and `saveProjectAs` were the only two with an `addDefaultKeypress` and no `perform()` case, so the live bug is fully fixed.

Six inline notes. The one that matters is the Makefile: `lint-fix` still dies on Debian and Ubuntu for a second reason this commit does not touch.

Nit: em-dashes in both commit bodies and in the description, against the no-em-dash rule. The branch squashes, so it is fixable at merge time.

lucaromagnoli (commented):
Re-reviewed at `040db910`. Five of six are properly fixed. One introduced a new problem.

**Verified fixed.** I ran the new `lint-fix` recipe body under `/bin/dash` directly rather than reading it: it parses clean and both branches behave (`y` runs the find, `n` and empty both cancel). I also swept the rest of the Makefile for `[[`, `read -` and `<<<`, and there is nothing left. That one is genuinely resolved.

**Coverage is better than what I asked for.** Hoisting the dispatch above the local switch beats adding cases by hand. I recomputed the sets against `getAllCommands()`: all 64 commands are now handled by one path or the other, and zero appear in both, so the early return did not silently kill any existing `perform()` case.

**Unprompted bonus:** moving `calls` above `MenuCallbacksGuard` fixed a latent ordering problem. The counters used to be declared after the guard, so they died before the guard's destructor tore down the lambdas capturing them. Harmless in practice since nothing invoked them at that point, but the new order is correct rather than lucky.

One new bug and two smaller notes inline.

magda/daw/ui/windows/MenuManager.cpp
```diff
@@ -38,6 +38,53 @@ void MenuManager::initialize(const MenuCallbacks& callbacks) {
...
+            callback = &callbacks_.onExportAudio;
+            break;
+        case CommandIDs::deleteTrack:
+            callback = &callbacks_.onDeleteTrack;
+            break;
+        case CommandIDs::toggleArrangeSession:
+            callback = &callbacks_.onToggleArrangeSession;
+            break;
+        case CommandIDs::zoom:
+            callback = &callbacks_.onZoom;
```
lucaromagnoli: This is the wrong callback, and it is a new bug rather than a pre-existing one.

`CommandIDs::zoom` is registered in the **View** category as "Zoom" / "Zoom controls" (`MainWindowCommands.cpp:424-426`). `callbacks_.onZoom` is the **Window** menu's maximize entry (`MenuManager.hpp:107`, under `// Window menu`), and its implementation is a stub:

```cpp
// MainWindowMenus.cpp:918
callbacks.onZoom = []() {
    // TODO: Implement window zoom functionality
    ...  "Window zoom functionality not yet implemented."
};
```

So a user who binds "Zoom" from the View category gets an alert about window zoom. Two unrelated concepts that happen to share a name.

My earlier comment listed the five unhandled commands as evidence for wanting a coverage assertion, not as a to-do list, and this one should not have been wired. "Zoom controls" is not a single action, so I would drop `zoom` from `getAllCommands()` entirely rather than point it at an unrelated stub.

magda/daw/ui/windows/MenuManager.cpp
```diff
@@ -38,6 +38,53 @@ void MenuManager::initialize(const MenuCallbacks& callbacks) {
...
+            break;
+        case CommandIDs::showHelp:
+            callback = &callbacks_.onShowHelp;
```
lucaromagnoli: Same shape as `zoom` but defensible: `onShowHelp` is also a "not yet implemented" alert stub (`MainWindowMenus.cpp:928`), but at least it is the *right* stub for the command.

Flagging only so it is a deliberate choice. Binding Help to a key now pops a placeholder dialog, which is honest but not useful.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -476,6 +476,12 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     auto& selectionManager = SelectionManager::getInstance();
     auto selectedClips = selectionManager.getSelectedClips();
 
+    // Command-backed application/menu actions share the same callback path as
+    // menu clicks. Keeping this before the local switch prevents registered
+    // menu commands from silently becoming dead custom key bindings.
+    if (MenuManager::getInstance().invokeApplicationCommand(info.commandID))
```
lucaromagnoli: Good restructure, and the comment explains it well.

One residual gap, low priority given the shape: the invariant is now unguarded in the mirror direction. If a command ends up in both this dispatcher and the local switch below, the early return wins and the switch case dies silently, which is the same class of bug as the original report. There are zero overlaps today, so this is a note rather than a request.

tests/test_menu_manager.cpp
```diff
@@ -0,0 +1,71 @@
...
+        magda::CommandIDs::zoom,
+        magda::CommandIDs::showHelp,
+        magda::CommandIDs::about,
+    };
+
+    for (const auto commandID : commandIDs)
+        REQUIRE(menuManager.invokeApplicationCommand(commandID));
+
+    for (const auto callCount : calls)
+        CHECK(callCount == 1);
```
lucaromagnoli: This reports a bare `callCount == 1` failure with no indication which of the ten commands did not fire, so a regression means bisecting by hand.

Either index the loop and `CAPTURE(i)`, or compare the whole array against an expected one so Catch2 prints both sides.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -106,10 +106,14 @@ void MainWindow::MainComponent::getAllCommands(juce::Array<juce::CommandID>& com
...
-        zoom, toggleArrangeSession, cycleViewForward, cycleViewBackward, uiScaleUp, uiScaleDown,
+        toggleArrangeSession, cycleViewForward, cycleViewBackward, uiScaleUp, uiScaleDown,
         togglePianoRollFullscreen,
         // Help
-        showHelp, about};
+        about};
+
+    // `zoom` describes a group of View-menu controls, not one action, and
+    // `showHelp` is still a placeholder. Do not expose either as a bindable
+    // command until each has a concrete action to perform.
```
lucaromagnoli: The reasoning is exactly right, but the placement reads as a comment on `commands.addArray(...)` below rather than on the two omissions above it.

It belongs above the array, or split beside the `// View` and `// Help` group lines it actually describes, so the next person editing those groups sees it.

tests/test_menu_manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+        magda::CommandIDs::saveProject,          magda::CommandIDs::saveProjectAs,
+        magda::CommandIDs::exportAudio,          magda::CommandIDs::deleteTrack,
+        magda::CommandIDs::toggleArrangeSession, magda::CommandIDs::about,
+    };
+
+    for (const auto commandID : commandIDs)
+        REQUIRE(menuManager.invokeApplicationCommand(commandID));
+
+    for (size_t i = 0; i < calls.size(); ++i) {
+        CAPTURE(i, commandIDs[i]);
```
lucaromagnoli, line 54: `calls` and `commandIDs` now have to stay the same length by hand, and this indexes `commandIDs[i]` bounded by `calls.size()` using `operator[]`. Add a command to one array and not the other and it reads out of bounds rather than failing the test.

```cpp
static_assert(calls.size() == commandIDs.size());
```

closes it. This hazard came in with the `CAPTURE` change I asked for, so it is on me.

lucaromagnoli (commented):
Re-reviewed at `62cadcb9`. All four points addressed, and the `zoom` one went further than I asked.

Dropping `zoom` and `showHelp` from `getAllCommands()`, `getCommandInfo()` and the dispatcher, rather than rewiring them, is the right call. Moving the dispatch into the switch's `default:` also fixes the shadowing note cleanly: a local case now always wins and MenuManager is only the fallback.

Verified on the new HEAD: 62 registered commands, zero unhandled by both paths, zero shadowed. I also grepped for stray `CommandIDs::zoom` and `CommandIDs::showHelp` references outside the enum and found only the two new negative assertions, so nothing dangles.

Two cosmetic leftovers inline. Neither blocks.

PR #1916, opened 2026-07-28
Make the rest of the File menu bindable to shortcuts

### LINKED ISSUE #1915: [Bug]: File menu shortcut coverage gaps — 3 commands unbound by default, 8 menu actions not bindable at all
### What happened?

While reviewing #1914 (which fixes the File menu shortcuts that were registered but never dispatched), two gaps in File menu shortcut coverage remain. Both are visible to users in Preferences → Shortcuts.

**Gap 1 — three File commands have no default keypress**

`CommandIDs` defines five File commands. Only two ship with a default binding:

| Command | Default keypress |
| --- | --- |
| `saveProject` | Cmd/Ctrl+S |
| `saveProjectAs` | Cmd/Ctrl+Shift+S |
| `newProject` | none |
| `openProject` | none |
| `exportAudio` | none |

`newProject`, `openProject` and `exportAudio` appear in the shortcuts editor with an empty binding, so they only work once the user assigns a key. Cmd/Ctrl+N and Cmd/Ctrl+O are the conventional defaults and are unclaimed; Export Audio has no obvious convention and could reasonably stay unbound.

See `getCommandInfo()` in `magda/daw/ui/windows/MainWindowCommands.cpp:311-329`.

**Gap 2 — most of the File menu has no command ID at all**

Only 5 of the 13 File menu actions are backed by a `CommandID`. These 8 have a `MenuManager` menu item and a callback, but no entry in `CommandIDs`, so they never reach `getAllCommands()`, never appear in the shortcuts editor, and cannot be bound to a key at all:

- Close Project
- Project Settings
- Collect Files
- Export MIDI
- Import DAWproject
- Export DAWproject
- Quit
- Open Recent (parameterised — likely out of scope)

Giving these command IDs means adding the enum entries, a `getCommandInfo()` case with the "File" category, and a `perform()` case routing through `MenuManager::invokeApplicationCommand()` — the dispatch helper #1914 adds. The menu-click path already converges on the same callbacks, so no action logic gets duplicated.

### Steps to reproduce

1. Open Preferences → Shortcuts and select the File category.
2. Observe that New Project, Open Project and Export Audio are listed with no binding (gap 1).
3. Observe that Close Project, Project Settings, Collect Files, Export MIDI and the DAWproject import/export actions are absent from the list entirely (gap 2).

### Expected behavior

Every File menu action that takes no arguments is bindable in the shortcuts editor, and the actions with conventional shortcuts (New, Open) ship with those defaults.

### MAGDA version

0.17.0

### Operating system

All

### Notes

Follow-up to #1911 / #1914. #1914 fixes the dispatch path for the five commands that already exist; this issue covers the coverage gaps around it.

lucaromagnoli:
Closes #1915. **Stacked on #1914** — targets `fix/1911-file-menu-shortcuts`, since it builds on the `invokeApplicationCommand()` dispatch helper that PR adds. Merge #1914 first and this retargets to `main` cleanly.

Additions only; no bug fixes here.

## New Project and Open Project

Registered as commands but never dispatched from `perform()`, and shipped with no default keypress — so they sat in the shortcuts editor unbound. They now dispatch through `MenuManager::invokeApplicationCommand()` and default to Cmd/Ctrl+N and Cmd/Ctrl+O; I checked both were unclaimed.

These are here rather than in #1914 because binding them is a new capability, not a fix — and because they are what surfaced the unsaved-changes bug now in its own PR.

## Six more actions become bindable

Close Project, Project Settings, Collect Files, Export MIDI and the DAWproject import/export actions had a menu item and a callback but no `CommandID` at all, so they never reached `getAllCommands()`, never appeared in the shortcuts editor, and could not be bound to a key.

They now have IDs in the File category and dispatch through the same helper, so each command keeps a single action path rather than two parallel implementations. Every command-backed File item renders its binding via `keyHint()`.

## Deliberately left out

- **Quit** — its menu item is `#if !JUCE_MAC` only and the platform already provides Cmd+Q / Alt+F4. A command ID would either double-handle with the macOS app menu or make the shortcuts list vary by platform.
- **Open Recent** — parameterised per path, so it doesn't fit the command model.

## Testing

`tests/test_menu_manager.cpp` extended to all 11 File commands dispatching to their callbacks. Debug app builds clean; menu and project tests pass (86 assertions, 5 cases).

### DIFF
--- a/magda/daw/ui/windows/CommandIDs.hpp
+++ b/magda/daw/ui/windows/CommandIDs.hpp
@@ -10,6 +10,12 @@ enum {
     saveProject = 0x0003,
     saveProjectAs = 0x0004,
     exportAudio = 0x0005,
+    closeProject = 0x0006,
+    projectSettings = 0x0007,
+    collectFiles = 0x0008,
+    exportMidi = 0x0009,
+    importDawProject = 0x000A,
+    exportDawProject = 0x000B,
 
     // Edit menu
     undo = 0x1000,
--- a/magda/daw/ui/windows/MainWindowCommands.cpp
+++ b/magda/daw/ui/windows/MainWindowCommands.cpp
@@ -98,7 +98,8 @@ void MainWindow::MainComponent::getAllCommands(juce::Array<juce::CommandID>& com
         splitAllTracksAtCursor, copyTimeRange, cutTimeRange, deleteTimeRange, copyLoopRange,
         cutLoopRange, deleteLoopRange, pasteRipple,
         // File menu
-        newProject, openProject, saveProject, saveProjectAs, exportAudio,
+        newProject, openProject, saveProject, saveProjectAs, exportAudio, closeProject,
+        projectSettings, collectFiles, exportMidi, importDawProject, exportDawProject,
         // Transport
         play, stop, record, goToStart, goToEnd, addMarker, goToPreviousMarker, goToNextMarker,
         goToLoopStart, goToLoopEnd, goToSelectionStart, goToSelectionEnd,
@@ -311,9 +312,14 @@ void MainWindow::MainComponent::getCommandInfo(juce::CommandID commandID,
         // File menu
         case newProject:
             result.setInfo("New Project", "Create a new project", "File", 0);
+            result.addDefaultKeypress('n', juce::ModifierKeys::commandModifier);
             break;
         case openProject:
             result.setInfo("Open Project", "Open an existing project", "File", 0);
+            result.addDefaultKeypress('o', juce::ModifierKeys::commandModifier);
+            break;
+        case closeProject:
+            result.setInfo("Close Project", "Close the current project", "File", 0);
             break;
         case saveProject:
             result.setInfo("Save Project", "Save the current project", "File", 0);
@@ -326,6 +332,23 @@ void MainWindow::MainComponent::getCommandInfo(juce::CommandID commandID,
             break;
         case exportAudio:
             result.setInfo("Export Audio", "Export project to audio file", "File", 0);
+            break;
+        case exportMidi:
+            result.setInfo("Export MIDI", "Export project to a MIDI file", "File", 0);
+            break;
+        case projectSettings:
+            result.setInfo("Project Settings", "Open the project settings dialog", "File", 0);
+            break;
+        case collectFiles:
+            result.setInfo("Collect Files", "Copy external media into the project folder", "File",
+                           0);
+            break;
+        case importDawProject:
+            result.setInfo("Import DAWproject", "Import a DAWproject exchange file", "File", 0);
+            break;
+        case exportDawProject:
+            result.setInfo("Export DAWproject", "Export the project as a DAWproject exchange file",
+                           "File", 0);
             break;
 
         // Transport
@@ -563,9 +586,17 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     };
 
     switch (info.commandID) {
+        case newProject:
+        case openProject:
         case saveProject:
         case saveProjectAs:
         case exportAudio:
+        case closeProject:
+        case projectSettings:
+        case collectFiles:
+        case exportMidi:
+        case importDawProject:
+        case exportDawProject:
             return MenuManager::getInstance().invokeApplicationCommand(info.commandID);
 
         case undo:
--- a/magda/daw/ui/windows/MainWindowCommands.cpp
+++ b/magda/daw/ui/windows/MainWindowCommands.cpp
@@ -563,6 +563,11 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     };
 
     switch (info.commandID) {
+        case saveProject:
+        case saveProjectAs:
+        case exportAudio:
+            return MenuManager::getInstance().invokeApplicationCommand(info.commandID);
+
         case undo:
             UndoManager::getInstance().undo();
             return true;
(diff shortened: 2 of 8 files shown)


lucaromagnoli (commented):
Reviewed at the head of `feat/file-menu-bindable-commands`.

The mechanical part is clean: six new `CommandID`s, all eleven File commands routed through `invokeApplicationCommand()`, and `menuItemSelected()` fully converted off the old `if (callbacks_.onX)` idiom. I verified Cmd+N and Cmd+O are genuinely unclaimed across all 70 registered commands, so that claim holds.

My concern is sequencing rather than code, and it comes down to two lines. Details inline.

One thing I could not check from the diff: `ProjectManager::showUnsavedChangesDialog()` uses the synchronous `showYesNoCancelBox`, so it spins a modal loop. With this PR that loop now runs inside `perform()` during `ApplicationCommandManager` key dispatch rather than from a menu callback. Different call context, and a classic reentrancy source. Worth confirming which context the #1915 bug actually reproduces in before assuming a fix covers both.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -311,9 +312,14 @@ void MainWindow::MainComponent::getCommandInfo(juce::CommandID commandID,
         // File menu
         case newProject:
             result.setInfo("New Project", "Create a new project", "File", 0);
+            result.addDefaultKeypress('n', juce::ModifierKeys::commandModifier);
             break;
         case openProject:
             result.setInfo("Open Project", "Open an existing project", "File", 0);
+            result.addDefaultKeypress('o', juce::ModifierKeys::commandModifier);
```
lucaromagnoli: These two lines are the only risk in the PR, and I would hold them.

The description says New Project and Open Project "are what surfaced the unsaved-changes bug now in its own PR". So this binds the two most reflexive shortcuts in any application to a destructive path whose guard is known-broken and not yet fixed.

Everything else here is inert by default: none of the other six commands gets a default keypress, so the "make it bindable" capability lands in full either way. Dropping just these two `addDefaultKeypress` calls until the unsaved-changes PR merges (or merging that one first) costs nothing and removes the exposure entirely.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -563,9 +586,17 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
     };
 
     switch (info.commandID) {
+        case newProject:
```
lucaromagnoli: Concrete bug that Cmd+N makes easy to hit, in code outside this diff.

`MainWindowMenus.cpp:176-181`: `onNewProject` clears the selection and dispatches `ClearTimeSelectionEvent` **before** calling `projectManager.newProject()`, which is what runs the unsaved-changes prompt. So choosing Cancel still destroys the selection and the time selection. `onCloseProject` around line 240 has the same shape.

Pre-existing, and reaching it via a menu click is at least deliberate. Cmd+N is one key away from Cmd+B and Cmd+M, both of which are bound, so this goes from "rare" to "routine typo". Fix is to move the two clears into the success branch.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -563,9 +586,17 @@ bool MainWindow::MainComponent::perform(const InvocationInfo& info) {
...
         case saveProject:
         case saveProjectAs:
         case exportAudio:
+        case closeProject:
+        case projectSettings:
+        case collectFiles:
+        case exportMidi:
+        case importDawProject:
+        case exportDawProject:
             return MenuManager::getInstance().invokeApplicationCommand(info.commandID);
```
lucaromagnoli: Same coverage-test point I left on #1914, not repeating it at length. Just noting it got sharper: 70 commands now, 5 still with no `perform()` case (`about`, `deleteTrack`, `showHelp`, `toggleArrangeSession`, `zoom`).

This is exactly the kind of PR that grows that set, which is the argument for asserting `getAllCommands()` is a subset of the handled set rather than adding cases by hand.

magda/daw/ui/windows/CommandIDs.hpp
```diff
@@ -10,6 +10,12 @@ enum {
     saveProject = 0x0003,
     saveProjectAs = 0x0004,
     exportAudio = 0x0005,
+    closeProject = 0x0006,
```
lucaromagnoli, line 13: `closeProject` is destructive and now appears in the shortcuts editor. No default keypress, so the risk is low, but a user can bind it to a bare letter and nothing in the editor distinguishes it from Project Settings sitting right next to it.

magda/daw/ui/windows/MainWindowCommands.cpp
```diff
@@ -327,6 +333,23 @@ void MainWindow::MainComponent::getCommandInfo(juce::CommandID commandID,
         case exportAudio:
             result.setInfo("Export Audio", "Export project to audio file", "File", 0);
             break;
+        case exportMidi:
+            result.setInfo("Export MIDI", "Export project to a MIDI file", "File", 0);
+            break;
+        case projectSettings:
+            result.setInfo("Project Settings", "Open the project settings dialog", "File", 0);
```
lucaromagnoli, line 341: These names are hardcoded English while the matching menu items go through `tr()`, so the shortcuts editor lists them in English regardless of locale.

Pre-existing across all 70 commands rather than anything this PR introduced, so I would not fix it here. Worth its own issue though, and this adds six more entries to the pile.

magda/daw/ui/windows/MenuManager.cpp
```diff
@@ -409,27 +444,22 @@ void MenuManager::menuItemSelected(int menuItemID, int topLevelMenuIndex) {
...
-            if (callbacks_.onImportDawProject)
-                callbacks_.onImportDawProject();
+            invokeApplicationCommand(CommandIDs::importDawProject);
             break;
         case ExportDawProject:
-            if (callbacks_.onExportDawProject)
-                callbacks_.onExportDawProject();
+            invokeApplicationCommand(CommandIDs::exportDawProject);
             break;
         case Quit:
```
lucaromagnoli, line 475: `Quit` is now the only File item left on the old `if (callbacks_.onQuit)` idiom. The reasoning in the description for excluding it is sound, but it is no longer self-evident from the code. A one-line comment here would stop a later cleanup from "fixing" the inconsistency and double-handling with the macOS app menu.

PR #1917, opened 2026-07-28
Prompt to save unsaved changes on New/Open/Close

lucaromagnoli:
Independent of #1914 / #1916 — branches off `main`, touches no File menu code. Found while testing that PR's New/Open shortcuts: they fired correctly, but opening a project silently discarded unsaved work.

## New/Open/Close never prompted to save

All three check `isDirty_` and raise the unsaved-changes prompt, so the guard was already in place everywhere. It just never fired, because the flag was set from only a handful of places — tempo/time-signature/loop changes, marker edits, the project settings dialog and two external-file paths.

Adding tracks, adding or moving clips, recording, editing notes and every other edit left it `false`. The prompt never appeared and the work was discarded without warning.

Every one of those edits runs through `UndoManager::executeCommand()`, so that single choke point now marks the project dirty — covering all 479 call sites rather than chasing them individually. `undo()` and `redo()` mark it too: both move state away from what is on disk.

## Two problems that only surface once the prompt fires

- **The undo history outlived the project.** Its commands reference track and clip ids from the old project, so undoing after a New/Open/Close applied them to unrelated state. Every project boundary now clears it, including both async commit paths (`loadProjectAsync` and `importDawProjectAsync`, which commit separately from the sync `loadProject`).
- **A cancelled prompt reported itself as an error.** `loadProjectAsync` passed `"Cancelled by user"` and `openProjectFile` showed any error in an alert, so declining to discard your work popped a spurious warning. It now passes an empty error — the convention the sync callers and the DAWproject import path already use — and `openProjectFile` stays silent for it.

## Testing

`tests/test_project_dirty_tracking.cpp` covers execute/undo/redo marking dirty, a command inside a compound operation, and the undo history being dropped at a project boundary.

The risk worth flagging for review: ten test files execute undo commands and several call `loadProject`/`newProject`, so a wrongly-dirty singleton could raise the modal prompt and hang CI. The new tests hand the singleton back clean via a temp-file save, and the full suite passes in both random and declaration order — 1549 passed, 13 skipped, 0 failures, no hang. Debug app builds clean.

### DIFF
--- a/magda/daw/core/UndoManager.cpp
+++ b/magda/daw/core/UndoManager.cpp
@@ -22,10 +22,10 @@ void UndoManager::executeCommand(std::unique_ptr<UndoableCommand> command) {
     }
 
     const auto beforeStateId = currentStateId_;
-    auto& projectManager = ProjectManager::getInstance();
-    projectManager.beginUndoableMutation();
-    command->execute();
-    projectManager.endUndoableMutation();
+    {
+        ProjectManager::UndoableMutationScope mutationScope;
+        command->execute();
+    }
     currentStateId_ = nextStateId_++;
 
     // If in compound operation, collect commands instead of pushing to stack
@@ -66,10 +66,10 @@ bool UndoManager::undo() {
 
     DBG("📝 UNDO: Undoing '" << entry.command->getDescription() << "'");
 
-    auto& projectManager = ProjectManager::getInstance();
-    projectManager.beginUndoableMutation();
-    entry.command->undo();
-    projectManager.endUndoableMutation();
+    {
+        ProjectManager::UndoableMutationScope mutationScope;
+        entry.command->undo();
+    }
     currentStateId_ = entry.beforeStateId;
 
     // Push to redo stack
@@ -92,10 +92,10 @@ bool UndoManager::redo() {
 
     DBG("📝 UNDO: Redoing '" << entry.command->getDescription() << "'");
 
-    auto& projectManager = ProjectManager::getInstance();
-    projectManager.beginUndoableMutation();
-    entry.command->execute();
-    projectManager.endUndoableMutation();
+    {
+        ProjectManager::UndoableMutationScope mutationScope;
+        entry.command->execute();
+    }
     currentStateId_ = entry.afterStateId;
 
     // Push to undo stack
@@ -122,6 +122,10 @@ juce::String UndoManager::getRedoDescription() const {
 }
 
 void UndoManager::clearHistory() {
+    // The state ids deliberately survive this. Dropping the stacks removes the
+    // route back to the saved state but not the fact that current state differs
+    // from it, so undoHistoryDirty_ stays correct and a later save still clears
+    // it through markCurrentStateSaved().
     undoStack_.clear();
     redoStack_.clear();
     compoundCommands_.clear();
--- a/magda/daw/core/UndoManager.cpp
+++ b/magda/daw/core/UndoManager.cpp
@@ -1,5 +1,7 @@
 #include "UndoManager.hpp"
 
+#include "../project/ProjectManager.hpp"
+
 namespace magda {
 
 // ============================================================================
@@ -21,6 +23,11 @@ void UndoManager::executeCommand(std::unique_ptr<UndoableCommand> command) {
 
     // Execute the command
     command->execute();
+
+    // Every undoable command mutates project state, so this is the one place
+    // that has to flag unsaved changes — otherwise New/Open/Close discard the
+    // user's work without ever showing the unsaved-changes prompt.
+    ProjectManager::getInstance().markDirty();
 
     // If in compound operation, collect commands instead of pushing to stack
     if (compoundDepth_ > 0) {
@@ -57,6 +64,9 @@ bool UndoManager::undo() {
     // Undo the command
     command->undo();
 
+    // Undoing moves state away from what is on disk just as much as doing.
+    ProjectManager::getInstance().markDirty();
+
     // Push to redo stack
     redoStack_.push_back(std::move(command));
 
@@ -78,6 +88,8 @@ bool UndoManager::redo() {
 
     // Re-execute the command
     command->execute();
+
+    ProjectManager::getInstance().markDirty();
 
     // Push to undo stack
     undoStack_.push_back(std::move(command));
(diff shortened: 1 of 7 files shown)


lucaromagnoli (commented):
Reviewed the full diff. The state-id model is more than the description claims: because `savedStateId_` is a stable identity rather than a boolean, undoing back to the saved point actually clears the flag, and the `beforeStateId != savedStateId_` merge guard keeps the on-disk state reachable by undo instead of letting command merging swallow it. That is a subtle thing done right.

Two things I want to call out as good beyond the main fix:

- The cancel-path ordering bug I raised on #1916 is fixed here, in all four sites, with the clears moved into the success branches.
- `safeThis` in `importDawProjectFile` fixes a real raw-`this` capture across an async boundary that had nothing to do with this PR.

On layering: `core/UndoManager.cpp` including `project/ProjectManager.hpp` looks like an inversion, but `core/` already does exactly this in `ClipManager`, `SelectionManager`, `TrackCommands` and `MixAnalysisService`. Consistent rather than new, so I am not raising it. The `friend class UndoManager` is the genuinely new coupling, and the suggestion on `executeCommand` would remove it.

Three issues inline, plus one nit.

magda/daw/core/UndoManager.cpp
```diff
@@ -91,14 +111,14 @@ juce::String UndoManager::getUndoDescription() const {
...
 
 juce::String UndoManager::getRedoDescription() const {
     if (redoStack_.empty()) {
         return {};
     }
-    return redoStack_.back()->getDescription();
+    return redoStack_.back().command->getDescription();
 }
 
 void UndoManager::clearHistory() {
```
lucaromagnoli, line 124: `clearHistory()` clears the stacks but never touches `currentStateId_` / `savedStateId_` and never calls `updateProjectDirtyState()`.

That is safe today only by convention: all five call sites are in ProjectManager and each is immediately followed by `clearDirty()`, which syncs the ids back up. But this is a public method with nothing signalling that requirement.

Call it on its own while the project is dirty and `undoHistoryDirty_` stays `true` with an empty undo stack. The user then cannot undo back to clean, because there is nothing left to undo, so the project prompts to save on every New/Open/Close for the rest of the session with no way out except saving.

Syncing `savedStateId_ = currentStateId_` and calling `updateProjectDirtyState()` inside `clearHistory()` makes it self-consistent and costs nothing at the existing call sites.

magda/daw/core/UndoManager.cpp
```diff
@@ -19,27 +21,37 @@ void UndoManager::executeCommand(std::unique_ptr<UndoableCommand> command) {
         return;
     }
 
-    // Execute the command
+    const auto beforeStateId = currentStateId_;
+    auto& projectManager = ProjectManager::getInstance();
+    projectManager.beginUndoableMutation();
     command->execute();
+    projectManager.endUndoableMutation();
```
lucaromagnoli: This begin/end pair is not exception safe, and it appears three times (here, `undo()`, `redo()`).

If any command throws out of `execute()`, `undoableMutationDepth_` never unwinds. From then on `markDirty()` sees a non-zero depth and never sets `externalDirty_`, so the project silently stops becoming dirty. That is precisely the bug this PR fixes, restored permanently and invisibly, and nothing would surface it.

An RAII guard is the direct fix. Inverting is better though: expose `UndoManager::isExecutingCommand()` and have `ProjectManager::markDirty()` query it instead of being bracketed. There is no counter to leak, the three call sites collapse to nothing, and `friend class UndoManager` goes away with it.

magda/daw/project/ProjectManager.cpp
```diff
@@ -524,15 +556,38 @@ void ProjectManager::setLoopSettings(bool enabled, double startBeats, double end
...
+    if (undoableMutationDepth_ == 0)
+        externalDirty_ = true;
+    refreshDirtyState();
 }
 
 void ProjectManager::clearDirty() {
-    if (isDirty_) {
-        isDirty_ = false;
+    externalDirty_ = false;
+    UndoManager::getInstance().markCurrentStateSaved();
```
lucaromagnoli, line 567: `clearDirty()` ends up incrementing `mutationRevision_`: `markCurrentStateSaved()` calls `updateProjectDirtyState()`, which calls `setUndoHistoryDirty()`, which does `++mutationRevision_`.

`saveProject()` calls `clearDirty()` at line 207, and Cmd+S stays live while the loading message is showing. So saving during a background load trips the `mutationRevision_ != startingRevision` check in `loadProjectAsync`, the load is abandoned, and the user is told "The current project changed while the new project was loading." Nothing changed. They saved.

A counter meant to detect mutations should not be bumped by the path that clears dirty state. Either increment it only in `markDirty()`, or have `setUndoHistoryDirty()` bump it only when `dirty` is true.

magda/daw/project/ProjectManager.cpp
```diff
@@ -28,6 +29,9 @@ static const char* const kTempPrefix = "UnsavedProject_";
 static constexpr int kStaleTempDays = 7;
 static const char* const kAutosaveExtension = ".autosave";
 static constexpr int kDefaultAutoSaveIntervalMs = 60000;
+static const char* const kProjectChangedWhileLoading =
+    "The current project changed while the new project was loading. "
+    "Open the file again to avoid losing those changes.";
```
lucaromagnoli, line 34: This is a new user-visible string in hardcoded English: it reaches the user through `openProjectFile`'s alert, whose title goes through `tr()`.

Consistent with the other hardcoded strings already in this file ("Unsaved Changes" and friends), so I would not fix it here. Worth noting that the file is accumulating untranslatable user-facing text.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -188,6 +195,11 @@ void MainWindow::setupMenuCallbacks() {
                                                        tr("dialogs.new_project"), message);
             }
         } else {
+            SelectionManager::getInstance().clearSelection();
+            if (mainComponent && mainComponent->mainView)
+                mainComponent->mainView->getTimelineController().dispatch(
+                    ClearTimeSelectionEvent{});
```
lucaromagnoli, line 202: This is the fix for the cancel-path bug I raised on #1916, and it is done in all four sites rather than just this one. Choosing Cancel at the prompt no longer destroys the selection and time selection on the way out. Good.

PR #1921, opened 2026-07-29
Fix repeated levels telemetry resets

### LINKED ISSUE #1919: Levels meter repeatedly resets and shows no signal on Linux
## Summary

The Levels meter does not update reliably during audio playback on Linux. Most readouts remain at `-inf` or show `no signal`, although a screenshot can capture valid readings. The same behaviour has been reported under both Ubuntu Xorg and Wayland.

Original report: https://github.com/Conceptual-Machines/magda-core/discussions/1918

## Actual behaviour

- LUFS and related meter readouts generally remain at `-inf` during playback.
- `no signal` flashes briefly at the bottom of the Levels UI.
- Taking a screenshot can capture valid readings.
- Screen recording does not make the readings update normally.
- Reproduced by the reporter under both Xorg and Wayland.

## Expected behaviour

The Levels meter should update continuously while non-silent audio passes through it, without flashing `no signal` or repeatedly returning to `-inf`.

## Likely root cause

`DeviceSlotComponent::timerCallback()` runs at 30 Hz and calls `refreshInlinePluginBindings()` for analysis devices. That reaches `DeviceCustomUIManager::bindAnalyzerPlugins()`, which unconditionally calls `LevelsUI::setTelemetrySource()` even when the telemetry source has not changed.

Unlike the Oscilloscope and Spectrum UI implementations, `LevelsUI::setTelemetrySource()` does not return early when passed the same source. Each redundant call therefore:

1. Calls `setActive(false)` on the existing source.
2. Reassigns the same source.
3. Calls `setActive(true)`.
4. Sets `LevelsPlugin::pendingReset_`, causing `TrackMeasurer` to reset on the audio thread.

Because both timers run at approximately 30 Hz, the meter is continually reset. Platform-specific timer ordering likely determines whether a valid snapshot is briefly visible, which is consistent with the intermittent readings and `no signal` flashes.

Relevant code:

- `magda/daw/ui/components/chain/DeviceSlotComponent.cpp` (`timerCallback`)
- `magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp` (`bindAnalyzerPlugins`)
- `magda/daw/ui/components/chain/custom_ui/LevelsUI.cpp` (`setTelemetrySource`)
- `magda/daw/audio/plugins/LevelsPlugin.hpp` (`setActive` / `pendingReset_`)

## Proposed fix

Make same-source binding idempotent, matching Oscilloscope and Spectrum:

```cpp
void LevelsUI::setTelemetrySource(std::shared_ptr<LevelsTelemetrySource> telemetry) {
    if (telemetry_ == telemetry)
        return;

    // Existing binding logic...
}
```

Add a regression test using a fake `LevelsTelemetrySource` to verify that rebinding the same source does not toggle its active state or reset measurement.

## Acceptance criteria

- Levels readings update continuously during playback on Linux/Xorg.
- Levels readings update continuously during playback on Linux/Wayland.
- `no signal` does not flash while non-silent audio is passing through the device.
- Rebinding the same telemetry source is a no-op.
- Binding a different source and hiding/showing the UI still activate, deactivate, and reset measurement as intended.

lucaromagnoli:


magda/daw/audio/analysis/TrackMeasurer.hpp
```diff
@@ -108,12 +108,14 @@ class TrackMeasurer {
...
-            return;  // block exceeds prepared size; skip (signal still passes through upstream)
 
         double sumMidSq = 0.0, sumSideSq = 0.0, sumLR = 0.0, sumLL = 0.0, sumRR = 0.0;
         float samplePeak = 0.0f;
-        const bool capture = captureSpectrum_.load(std::memory_order_acquire);
+        // The scratch buffer is only used for optional spectrum capture. Hosts
+        // can deliver blocks larger than the size reported during prepare();
+        // keep measuring loudness and peaks even when that capture must be skipped.
+        const bool capture = captureSpectrum_.load(std::memory_order_acquire) &&
+                             numSamples <= static_cast<int>(scratchL_.size());
```
lucaromagnoli: Correct for the loudness path, but it changes the spectrum path from stopping to splicing.

Before, an oversized block skipped everything, so the ring simply received nothing. Now measurement continues while the ring silently misses exactly those blocks. A host that alternates block sizes therefore hands the masking FFT discontinuous audio with splices in it, which is a worse input for that analysis than receiving nothing at all.

Worth making the choice explicitly rather than inheriting it. Either accept gappy capture and say so in the comment, or drop the size condition entirely and flush `scratchL_` to the ring whenever it fills:

```cpp
if (capture) {
    scratchL_[scratchFill_++] = static_cast<float>(mid);
    if (scratchFill_ == scratchL_.size()) {
        spectrumRing_.write(scratchL_.data(), static_cast<int>(scratchFill_));
        scratchFill_ = 0;
    }
}
```

with a flush of the remainder after the loop. That handles any block size against a fixed scratch, allocates nothing, and removes the gate rather than working around it.

tests/test_track_measurer.cpp
```diff
@@ -90,6 +90,21 @@ TEST_CASE("TrackMeasurer - sample peak reflects amplitude", "[measurer]") {
     REQUIRE(m.read().samplePeakDb == Catch::Approx(-6.02f).margin(0.2f));
 }
 
+TEST_CASE("TrackMeasurer - live blocks larger than prepared scratch are still measured",
+          "[measurer]") {
+    TrackMeasurer m;
+    m.prepare(kSr, 64, false);
```
lucaromagnoli: This misses both paths the guard removal actually newly exposes.

`false` disables true peak, so `oversamplePeak` never runs on an oversized block, and that is the one remaining loop over `numSamples` in `process()`. Capture is never enabled either, so the new `numSamples <= scratchL_.size()` condition is never exercised.

Both are safe today, but only by inspection. A later change that puts another prepared-size buffer inside that loop reintroduces the overrun with nothing failing. Passing `true` here, plus a case with `setSpectrumCaptureEnabled(true)` and oversized blocks, pins the behaviour this PR is deliberately introducing.

magda/daw/ui/components/chain/custom_ui/LevelsUI.cpp
```diff
@@ -44,6 +44,9 @@ LevelsUI::~LevelsUI() {
 }
 
 void LevelsUI::setTelemetrySource(std::shared_ptr<LevelsTelemetrySource> telemetry) {
+    if (telemetry_ == telemetry)
+        return;
```
lucaromagnoli, line 48: Right fix, and the test genuinely discriminates: without this, re-setting the same source produces `setActive(false)` then `setActive(true)` through `updateActiveState()`, so `activationStates` would be 3 entries rather than 1.

lucaromagnoli (commented):
Two independent fixes. The risky half is `TrackMeasurer::process`, since it removes a bounds guard on the audio thread, so I traced that rather than eyeballing it.

**The guard removal is safe.** Every buffer the loop and its tail touch is either gated by the new `capture` flag or not sized from `prepare()`:

- `scratchL_[i]` and `spectrumRing_.write(...)` are both under `capture`
- `oversamplePeak` works on a compile-time `std::array<float, kTapsPerPhase>` delay line, not a prepared-size buffer
- `closeGatingBlock` indexes `blockEnergy_` modulo `kRing` and clamps the histogram bin with `jlimit`

Nothing else is sized from `maxBlockSize`, so an oversized block cannot overrun anything.

**The `LevelsUI` early return is also safe**, for a reason worth writing down since it is not obvious from the diff: it skips `updateActiveState()`, but `visibilityChanged()` and `parentHierarchyChanged()` both call that independently, so the active flag cannot be left stale by the new path.

One off-diff note: `scratchR_` is dead. It is sized to `maxBlockSize` in `prepare()` at line 72 and declared at line 442, and referenced nowhere else. Your new comment correctly says "the scratch buffer", singular. Worth deleting while you are in this file.

Two inline comments, neither blocking.

Issue #1924, opened 2026-07-29
Faust: name device pages from group labels and render a tab strip


lucaromagnoli:
## What

Faust device pages are numbered, not named. `vgroup` / `hgroup` / `tgroup` labels in the DSP should name them, and the page nav should render as a tab strip.

## Why it is cheap

Paging already exists and is purely arithmetic: `FaustDeviceLayout` puts pool slot `N` on page `N / 32` (`magda/daw/ui/components/chain/layout/FaustDeviceLayout.cpp`, `kCellCount = 32`). The only missing piece is the name.

The harvester already walks the group tree. `UIHarvester::pushGroup` (`magda/daw/audio/plugins/FaustPlugin.cpp:45`) parses the group label and keeps only `parsed.metadata`, discarding `parsed.cleanLabel`:

```cpp
void pushGroup(const char* label) {
    auto parsed = parseFaustLabel(juce::String::fromUTF8(label != nullptr ? label : ""));
    groupStack.push_back(parsed.metadata);
}
```

## Work

- Keep the clean group label alongside the metadata on the group stack.
- Record, per group, the range of pool indices its controls occupy.
- Map each group to the page its range falls on, and expose the names to the UI.
- Render the existing page nav as a named tab strip.

## Authoring convention

A group maps to a tab only when its controls sit in one page-aligned `[idx:N]` block: 0 to 31 is the first tab, 32 to 63 the second, and so on. Groups that straddle a page boundary or interleave with another group have no single tab to name, so they fall back to the numbered page.

This needs documenting wherever the `[idx:N]` contract is described, since it is an authoring rule rather than something the host can enforce.

Part of #1375.

lucaromagnoli:
Correction to the authoring convention above. `FaustParamPool::kSize` is 64 and a page is 32 cells, so page-aligned tabs cap at two tabs and burn 32 slots each regardless of occupancy. No runtime DSP currently exceeds 14 controls, so none of them paginate and none could express more than two tabs.

Tabs should therefore be decoupled from pages. Group membership defines the tab, `[idx:N]` continues to mean slot identity only, and `cellFor` consumes a per-tab ordered list of pool indices rather than deriving the index from the page number. This removes the page-alignment authoring rule entirely, at the cost of a real change to `FaustDeviceLayout` rather than just plumbing a label through.

Issue #1913, opened 2026-07-28
Faust runtime instruments ignore [role:projectTempo]


lucaromagnoli:
Runtime Faust instruments silently ignore `[role:projectTempo]`.

`FaustPlugin::applyToBuffer` samples the tempo sequence once per block and fans the BPM out to every binding tagged with the role:

```cpp
// FaustPlugin.cpp:571-585
double cachedBpm = -1.0;
for (const auto& b : active->activeBindings) {
    if (!b.zone || b.role != FaustControlRole::ProjectTempo)
        continue;
    if (cachedBpm < 0.0)
        cachedBpm = edit.tempoSequence.getBpmAt(fc.editTime.getStart());
    *b.zone = static_cast<FAUSTFLOAT>(cachedBpm);
}
```

`FaustInstrumentPlugin::applyToBuffer` has no equivalent. It skips non-User roles when fanning out user parameters (FaustInstrumentPlugin.cpp:625-627) but never writes the tempo zone afterwards, so the control sits at whatever default the `.dsp` declared (120 in the usual idiom) for the lifetime of the instance.

The failure is silent: the parser accepts the tag, the pool allocates the slot, `[hidden:1]` hides it from the grid, and the DSP just reads a constant. Nothing warns the author. This came up in discussion #1845, where the tag is the documented answer for tempo-synced effects.

Compiled devices are unaffected. Their wrappers write the zone by hand, e.g. `MagdaDelayCompiledPlugin.cpp:369`.

## What needs doing

Write the tempo zone in `FaustInstrumentPlugin::applyToBuffer`, fanning out to every voice. The instrument uses `mydsp_poly` with per-voice zones harvested into `voiceZonesBySlot`, so unlike the FX path this has to loop over voices the same way the user-parameter block above it does, rather than writing a single zone.

Worth a look while in there: whether a tempo-tagged slot should be excluded from `poolParams_` entirely on the instrument side, since it is host-driven and not user-driven.

lucaromagnoli:
Fixed in #1939: `FaustInstrumentPlugin::applyToBuffer` now writes the project tempo to every voice zone tagged `[role:projectTempo]`, matching the effect path. Merged into dev/0.18.0, so the closing keyword did not fire automatically.

PR #1938, opened 2026-07-30
Issue/1930

### LINKED ISSUE #1930: Faust: converge the effect and instrument implementations
## What

The Faust effect and the Faust instrument are two parallel implementations of the same device. They harvest the DSP differently and render it differently, and they have already drifted in ways that produce real bugs. Converge them.

## The duplication

**Two harvesters.** `FaustPlugin::UIHarvester` (`magda/daw/audio/plugins/FaustPlugin.cpp:37`) and the equivalent inside `FaustInstrumentPlugin.cpp:276` do the same job against the same Faust `UI` interface. They are not the same code and do not behave the same.

**Two renderers.** The effect is `FaustUI` (a header strip) plus the shared `paramGrid_` / `ParamHostComponent` driven by `FaustDeviceLayout`. The instrument is `FaustInstrumentTabbedUI`, which builds its own rows, its own `LinkableTextSlider`s and its own `MomentaryParamButton`s from the pool directly via `paramInfoFromSlot`, bypassing `DeviceInfo::parameters` entirely.

## Drift this has already caused

- **Tabs exist on one side only.** The instrument harvester keeps the group label (`c.group = topLevelAuthorGroup()`), `FaustParamSlot::group` carries it, and `FaustInstrumentTabbedUI` renders tabs from it. The effect harvester's `pushGroup` keeps only the parsed metadata and throws `cleanLabel` away, and `FaustDeviceLayout` has no notion of groups. `FaustParamSlot::group` is even commented "Unused by the Faust effect."
- **`[hidden:1]` was honoured on one side only.** `FaustInstrumentTabbedUI.cpp:100` has always skipped `slot.hidden`. The effect pushed hidden slots into `DeviceInfo::parameters`, where they rendered as `(slot N)` placeholder cells. Fixed for the effect separately; the two paths needed the same fix in two places.
- **Layout constants leak across the boundary.** `FaustInstrumentTabbedUI` carves `FaustUI::kHeaderHeight` for the shared header, so growing that header silently clipped it on instruments until it was changed to ask for the desired height.

Each of these is the same class of bug: a change lands on one path and the other is forgotten.

## Direction

Share the harvester first, since that is where the behavioural divergence starts and it has no UI risk. One `UIHarvester` producing `HarvestedControl` (group label included) for both devices, so `[idx:N]`, `[hidden:1]`, `[role:...]`, style and group annotations mean exactly one thing regardless of which device loaded the patch.

The renderers are the larger question and can follow separately. Options worth weighing at that point:

- Give the effect the instrument's tab treatment, which is what #1924 asks for. Once the group label reaches the effect's pool this is mostly a `DeviceParamLayout` change.
- Or move the instrument onto the shared `paramGrid_` and give the grid tabs, so there is one renderer and the tab support benefits both.

The second is the smaller long-term surface but the larger single step, since the instrument UI's per-row widget construction and linking would have to move to `ParamSlotComponent`.

## Relationship to other issues

#1924 (tab strip for the effect) partly overlaps: it was written assuming tabs had to be built from scratch, when the instrument already has a working implementation reading `FaustParamSlot::group`. Whichever lands first should take the other into account.

lucaromagnoli:
## Summary

- converge runtime Faust effects and instruments on one shared `FaustUIHarvester`
- move both device types onto the shared `FaustUI` + `ParamHostComponent` renderer
- use outermost Faust author groups as named tabs for both effects and instruments
- keep `[idx:N]` as the stable automation, modulation, MIDI Learn, and gate identity
- remove the bespoke `FaustInstrumentTabbedUI`
- cache group-page mappings and hide pagination when only one page exists

## Related issues

Closes #1930.

This also implements #1924 by decoupling tab membership from pool-slot page boundaries: author groups define tabs, while `[idx:N]` remains parameter identity only.

#1913 is intentionally unchanged because it belongs to the audio-path `applyToBuffer` behavior rather than harvesting or rendering.

## Verification

- `cmake-build-debug/tests/magda_tests "[faust]"` — 572 assertions across 85 test cases
- `cmake --build cmake-build-debug --target magda_daw_app -j 8`


magda/daw/ui/components/chain/layout/FaustDeviceLayout.cpp
```diff
@@ -38,17 +39,135 @@ int maxPoolIndex(const magda::DeviceInfo& device) {
...
+
+    for (const auto& param : device.parameters) {
+        if (param.paramIndex < firstSlot || param.paramIndex >= lastSlot)
+            continue;
+        foundParam = true;
+        if (param.group.isEmpty())
+            return juce::String(pageIndex + 1);
+        if (candidate.isEmpty())
+            candidate = param.group;
+        else if (candidate != param.group)
```
lucaromagnoli: This is the page-alignment authoring rule that my correction on #1924 explicitly asked to remove:

> Tabs should therefore be decoupled from pages. Group membership defines the tab, `[idx:N]` continues to mean slot identity only, and `cellFor` consumes a per-tab ordered list of pool indices rather than deriving the index from the page number. This removes the page-alignment authoring rule entirely.

`PageMode::Groups` below does exactly that. So the two modes encode opposite answers to the same question, and the effect gets the one I had already ruled out.

Concretely: `kSize` is 64 and a page is 32 cells, so `PoolSlots` caps at two pages. Since no bundled effect exceeds 14 controls, every one of them is a single page whose controls carry 2 or more distinct groups, so this branch trips and the tab is named `1`. The author's group labels reach `ParameterInfo::group` and die here.

Suggest deleting `PoolSlots` and giving effects `Groups` too. `[idx:N]` keeps meaning slot identity, group membership defines the tab, and #1924 closes with it.

magda/daw/ui/components/chain/params/ParamHostComponent.cpp
```diff
@@ -166,9 +204,11 @@ void ParamHostComponent::setGridVisible(bool visible) {
 
 void ParamHostComponent::setPaginationVisible(bool visible) {
     const bool effective = visible && layout_->wantsPagination();
-    prevPageButton_->setVisible(effective);
-    nextPageButton_->setVisible(effective);
-    pageLabel_->setVisible(effective);
+    const bool tabs = effective && layout_->wantsPageTabs();
```
lucaromagnoli, line 208: `wantsPageTabs()` is hardcoded true for `FaustDeviceLayout`, and `layoutContent` ends with an unconditional `setPaginationVisible(true)` (line 292), so page count never gates the strip.

A one-page effect therefore renders a `TabbedButtonBar` holding a single tab labelled `1`, where it previously showed prev/next arrows plus `1/1`. Every bundled effect DSP is one page, so that is the state they all land in.

Worth hiding the strip when `totalPages <= 1` and falling back to no pagination affordance at all, the way a single-page device reads today.

magda/daw/ui/components/chain/layout/DeviceParamLayout.hpp
```diff
@@ -56,9 +56,35 @@ class DeviceParamLayout {
...
+    }
+
     /// What to render in cell `cellIndex` on the current page.
     virtual ParamCell cellFor(const magda::DeviceInfo& device, int cellIndex,
                               int currentPage) const = 0;
+
+    /// Resolve a stable parameter identity to the page that displays it.
+    /// Generic implementation deliberately asks cellFor(), so custom layouts
+    /// do not need a second mapping implementation.
+    virtual int pageForParameter(const magda::DeviceInfo& device, int paramIndex) const {
```
lucaromagnoli, line 77: This default walks every page x every cell calling `cellFor`, which is fine for a cheap `cellFor` but expensive for the new `Groups` path: each call rebuilds the entire group list (a `std::vector<Group>`, `juce::String` copies, O(params) string compares).

`ParamHostComponent` already calls `cellFor` in three separate 32-cell loops per refresh (lines 123, 147, 161), so a refresh is roughly 96 rebuilds. `DeviceParameterChangeHandler.cpp:109` then calls `pageForParameter` on the MIDI-learn / parameter-move path, adding pages x 32 more per event.

The old `PoolSlots` path was a deliberately allocation-free linear scan, with a comment saying so. Two options: memoise `groupedPages` against the `DeviceInfo` and invalidate on parameter-list change, or override `pageForParameter` in `Groups` mode to look the group up directly instead of probing `cellFor`.

magda/daw/audio/plugins/FaustUIHarvester.cpp
```diff
@@ -0,0 +1,139 @@
...
+        return label;
+    }
+    return {};
+}
+
+void FaustUIHarvester::emitControl(FaustParamSlot::Kind kind, const char* rawLabel,
+                                   FAUSTFLOAT* zone, FAUSTFLOAT init, FAUSTFLOAT min,
+                                   FAUSTFLOAT max, FAUSTFLOAT step) {
+    auto parsed = parseFaustLabel(fromUtf8OrEmpty(rawLabel));
+    auto merged = mergedMetadataFor(zone);
```
lucaromagnoli, line 70: Ordering subtlety worth a comment: `mergedMetadataFor(zone)` erases the zone's pending declare as a side effect, so it has to run before the `PolyphonicVoices` early return below, otherwise a skipped proxy control leaves its declare in `pendingByZone_` and it silently attaches to the next control harvested.

As written that is correct, but it reads like wasted work on the skip path and is an easy thing for someone to "optimise" by moving the early return up.

tests/test_faust_device_layout.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    DeviceInfo interleaved;
+    interleaved.parameters = {param(0, "Filter"), param(1, "Envelope")};
+    REQUIRE(layout.pageName(interleaved, 0) == "1");
+
+    DeviceInfo straddled;
+    straddled.parameters = {param(31, "Filter"), param(32, "Filter")};
+    REQUIRE(layout.pageName(straddled, 0) == "1");
+    REQUIRE(layout.pageName(straddled, 1) == "2");
+}
```
lucaromagnoli, line 50: The two modes are tested separately, so nothing here fails when they disagree, which is the invariant #1930 is about.

Worth a test that runs one source through both configurations and asserts identical page names and identical cell-to-`targetParamIndex` mapping. That test would currently fail on the naming difference described in my `poolPageName` comment, which is the point.

lucaromagnoli (commented):
Reviewed against #1930.

The core ask lands cleanly. One `FaustUIHarvester` serves both paths, `FaustParamSlot::group` is no longer "unused by the Faust effect", `FaustInstrumentTabbedUI` is deleted outright (360 lines), and `[hidden:1]` is now one rule applied symmetrically in `NativeDeviceProcessors.cpp` with a comment tying the two sites together. Net -28 lines, with new coverage for the harvester and the layout. It also takes the renderer step #1930 deferred, choosing option 2 (instrument onto the shared grid), which #1930 called the smaller long-term surface.

Blocking on one point (GitHub will not let me formally request changes on my own PR, so this is a comment review): the effect's page naming re-implements the rule my own correction on #1924 rejected, and as written the feature is dead on every DSP I ship.

`poolPageName` names a page only when a single group occupies one complete, non-interleaved 32-slot block. Every bundled runtime effect has 2 to 5 author groups and 14 or fewer controls, so all of them land in page 0 and the interleave check trips:

| DSP | author groups | resulting effect tab |
| --- | --- | --- |
| beat_repeater | Mix, Roll | `1` |
| transient_shaper | Attack, Sustain, Output | `1` |
| vinyl_emulator | Texture, Transport | `1` |
| reverser | Engine, Mix | `1` |
| realistic_rain_generator | Rain, Surfaces | `1` |
| vintage_lofi_emulator | Output, Sampler | `1` |
| ultimate_glitcher | Crush, Global, Mod & Stop, Reverser | `1` |
| 8_stage_morphable_wavefolder | Filter, Fold, Morph, Output | `1` |

Those labels are harvested, carried through `FaustParamSlot::group` and `ParameterInfo::group`, then discarded at the final step. Load the same patch as an instrument and it gets real named tabs. That is the same class of divergence #1930 exists to remove, relocated from the harvester into the layout: the PR now ships two different answers to "what names a tab", one per device.

Using `PageMode::Groups` for both devices, or dropping `PoolSlots` entirely, settles this and #1924 in one go.

The rest is smaller: a one-tab strip regression on effects, `groupedPages()` being rebuilt per cell, and a missing test for the invariant this PR is actually about. Details inline.

#1913 is untouched, which is right; it lives in `applyToBuffer`, not the harvester. Worth adding a PR description saying this closes #1930 and how it relates to #1924.

magda/daw/ui/components/chain/layout/FaustDeviceLayout.cpp
lucaromagnoli: Resolved in d3e1f3ec, and resolved the way I wanted rather than patched: `PageMode`, `PoolSlots` and `poolPageName` are all gone, so there is no second answer left to diverge. `createDeviceSlotParamLayout` no longer branches on `isFaustInstrument`, and the class comment now states that `[idx:N]` is stable parameter identity while author groups define pages, which is the contract my #1924 correction asked for.

Worth having #1924 point at this PR and close with it.

magda/daw/ui/components/chain/params/ParamHostComponent.cpp
lucaromagnoli: Fixed in d3e1f3ec. `setPaginationVisible` now gates on `totalPages_ > 1`, and `updatePageControls` calls it at the end so the decision is re-applied once the real count is known.

I checked the ordering, since this is the kind of fix that trades one glitch for another: `totalPages_` defaults to 1, so the `layoutContent` pass that runs before any `updatePageControls` hides the strip rather than briefly showing a bogus one.

magda/daw/ui/components/chain/layout/DeviceParamLayout.hpp
lucaromagnoli: Fixed in d3e1f3ec, both halves: `pagesFor` caches behind an FNV-style signature over `parameters.size()` plus each `paramIndex` and `group.hashCode64()`, and `FaustDeviceLayout` now overrides `pageForParameter` to walk the cached pages directly instead of probing `cellFor`.

Two things I checked, because caching was my suggestion and this is where it would bite:

- `gateEnabled` only calls `findParamArrayIndex`, so it cannot re-enter `pagesFor` and reallocate the vector that `cellFor` is holding a `const&` into.
- The signature covers every field `pagesFor` reads, in order, so a reorder or a rename changes it. `cell.enabled` stays outside the cache and is recomputed per call, which is right, since gate state changes with parameter values and not with page structure.

magda/daw/audio/plugins/FaustUIHarvester.cpp
lucaromagnoli: Fixed in d3e1f3ec. The comment states the erase side effect and the ordering requirement explicitly, which is what this needed.

tests/test_faust_device_layout.cpp
lucaromagnoli: Fixed in d3e1f3ec by "Faust effects and instruments use identical page and parameter mappings", which goes through the real `createDeviceSlotParamLayout` for both traits and asserts equal `totalPages`, `pageName`, and per-cell `mode` / `targetParamIndex`. That is the invariant #1930 is about, and it would have failed before this commit.

One gap left in the sibling cache test: "invalidates cached pages when group metadata changes" only mutates a group name. `parameters.size()` and `paramIndex` are both in the signature but untested, and those are the ones that actually churn on recompile, which is when a stale page list would be visible. A couple of extra asserts on the same `DeviceInfo` would cover it.

PR #1952, opened 2026-07-31
Stop reserving a pagination row for single-page devices

lucaromagnoli:
The layouts hardcode wantsPagination() to true, so every device reserved 2 + 18 + 4 px above its first row of parameter cells and the content painter ruled a line under that band. The page controls themselves are gated on totalPages > 1, so a device whose parameters fit one page showed an empty band with a stray line under it.

ParamHostComponent::paginates() now answers what the layout wants AND what the device actually needs, and drives getChromeHeight, layoutContent and the control visibility from that one place. The painter takes 0 for the row height in that case and skips the rule. Single-page devices get those 24px back as grid.

Re-layouts on the transition, since gaining or losing the row changes the chrome height the parent divides the body by.


lucaromagnoli:
Superseded by #1953: this commit is merged into feat/1945-faust-device-header, which ships it alongside the Faust header rebuild that made the empty band visible.

Issue #1980, opened 2026-08-01
Replace the Tracktion IR Reverb with a native convolution device


lucaromagnoli:
Part of #1882 (Phase 0). Follow-up to #1888.

#1888 retired eight stock Tracktion effects onto their compiled-Faust successors. IR Reverb is one of the few stock TE devices left (with Test Tone, 4OSC and the hidden Volume/Pan) and the only browser-visible one with **no successor to alias onto**, so it needs a replacement built rather than a mapping written.

## Faust is not an option here

`fi.conv` is a plain FIR whose kernel is a **compile-time constant** signal bank — a 2 s IR at 48 kHz would unroll ~96k taps into the generated C++. faustlibraries has no partitioned/FFT convolution, and while `soundfile` reads audio at runtime it cannot feed `conv`'s compile-time kernel. A user-loadable IR loader therefore has to be a **native MAGDA device**, in the shape of the Mutable ports (Materia / Halo / Nimbus), not a compiled Faust device.

JUCE's `dsp::Convolution` is fair game: the epic's boundary rule for the engine target is "JUCE plus the model layer only".

## What the current device is

`te::ImpulseResponsePlugin`, a `juce::dsp::Convolution` behind five automatable parameters.

Frozen parameter indices (the compatibility surface a replacement has to map from):

| idx | name | note |
|-----|------|------|
| 0 | Gain | dB |
| 1 | Low Cut | **stored as a MIDI note number**, displayed as Hz |
| 2 | High Cut | same |
| 3 | Mix | 0–1 |
| 4 | Filter Q | 0.1–14 |

Non-automatable state: `name`, `normalise`, `trimSilence`.

The IR audio itself lives **inside the plugin state** as `irFileData` — the whole encoded source file as a binary blob, re-encoded to WAV on load. Device-state v2 base64-encodes binary properties, so it already round-trips through a `.mgd`; whatever replaces the device has to keep reading that blob or old projects lose their IR.

MAGDA side: `ImpulseResponseProcessor` (in `processors/internal/StockDeviceProcessors.hpp`), `ImpulseResponseUI` (file chooser + drag/drop), and the `impulseResponseLoadFile` device command that calls `loadImpulseResponse(File)`. No IRs ship with the app — users load their own.

## Convolution engine: what we can tap into

**`juce::dsp::Convolution` — already linked, no new dependency.** `juce::juce_dsp` is a
[...]

Issue #1882, opened 2026-07-27
Epic: native audio engine to replace Tracktion Engine


lucaromagnoli:
Replace Tracktion Engine with a MAGDA-owned engine covering exactly the feature set MAGDA uses today.

## Architecture
- The model compiles into an immutable, flat render plan. The audio thread executes the current plan; structural changes compile a new plan that is swapped in atomically.
- A plan differ keyed by MAGDA model IDs carries DSP state, plugin instances, and delay lines across swaps; only changed nodes crossfade. Rebuild clicks and identity-loss bugs collapse into this one component.
- Two executors over the same plan format: a single-threaded reference executor (bring-up, deterministic debugging, arbiter) and the parallel executor (compile-time dependency counts, fixed summing order, bit-identical across thread counts). The parallel executor must match or beat TE CPU on the benchmark corpus before cutover. Disk prefetch runs on its own thread with per-reader FIFOs.
- Parameters use two write lanes (host values, modulation) resolved once per block into a flat value buffer. Devices read resolved values only.
- Plan swap is the reclamation point for audio-thread object lifetime (epoch based).
- Every plan op carries a liveness-domain tag from day one, reserved for the post-cutover anticipative executor.

## Strategy
- Built in this repo as an isolated target, dark behind a CMake flag. Configure-time boundary check: no TE includes, no ui/ includes, dependencies are JUCE plus the model layer only.
- Primary verification is a null-diff harness: render the same .mgd through both engines and assert near-null difference. Plans are also dumpable data with golden tests, so the compiler is fully testable offline.
- Per-subsystem bring-up, then cutover, then the TE fork is removed.
- Scope guard until cutover: no new features that deepen TE coupling in clip playback, warp, launcher, or recording (host-side ARA and the audio editor source-domain restructure queue behind cutover). Model/UI/device-level milestones continue as normal.

## Phases
Phase 0 lands in normal releases and improves the current app regardless of the engine.

### Phase 0 (in normal releases)
- [x] #1883
- [x] #1884
- [x] #1885
- [x] #1886
- [x] #1887
- [x] #1888
- [x] #1901

### Engine build (dark target)
- [x] #1889
- [ ] #1896
- [x] #1890
- [x] #1891
- [x] #1892
- [ ] #1893
- [x] #1894
- [ ] #1895

### Cutover
- [ ] #1897

### Post-cutover
- [ ] #1898
- [ ] #1899

lucaromagnoli:
[...]

PR #1981, opened 2026-08-01
Retire the stock TE effects behind load-time aliases

### LINKED ISSUE #1888: Retire stock TE plugin wrappers behind load-time aliases
Part of #1882 (Phase 0).

Twelve stock-TE plugin wrappers live in processors/te/TeProcessorWrappers (965 lines). Seven hidden legacy registrations (EQ, Compressor, Delay, Chorus, Phaser, Reverb, PitchShift) exist only for old project loads and already have compiled-Faust successors.

- Convert the seven legacy device types to load-time aliases onto their compiled equivalents (with a parameter mapping per type), then delete TeProcessorWrappers.
- Keep the four user-facing stock devices working for now: Lowpass, IR Reverb, Test Tone, 4OSC. Plan native replacements; 4OSC is the hard one since user projects contain its state, so it needs a migration path, not just an alias.
- Acceptance: no processors/te directory; legacy projects load with equivalent-sounding devices.

lucaromagnoli:
Closes #1888. Part of #1882 (Phase 0).

Eight stock Tracktion effects had compiled-Faust successors but stayed registered so older projects would still open, which meant keeping a Tracktion plugin wrapper per device alive in `processors/te`. Each retired type is now a load-time alias instead: a project naming one loads the compiled device, with its saved parameter values converted into the successor's units.

| retired type | → | successor |
|---|---|---|
| `4bandEq` / `eq` / `equaliser` | → | `magda_eq` |
| `compressor` | → | `magda_compressor` |
| `delay` | → | `magda_delay` |
| `chorus` | → | `magda_chorus` |
| `phaser` | → | `magda_phaser` |
| `reverb` | → | `magda_reverb` |
| `pitchShifter` | → | `magda_pitch` |
| `lowpass` | → | `magda_filter` |

The issue named seven; Lowpass was added on top, so `Filter` is retired too and only Test Tone, IR Reverb, 4OSC and the hidden Volume/Pan remain on Tracktion.

## The mapping

`core/LegacyDeviceAliases.cpp` holds the table. Conversions work in **real units**, because that is the only space the two sides share — Tracktion's compressor threshold is a linear gain where the compiled one is dB, its ratio is reciprocal (`0` = INF → 50:1), its delay feedback is dB with −30 dB as silence, and its reverb splits wet and dry where the compiled one has a single Mix (folded through Freeverb's 3×wet / 2×dry weighting).

Controls the retired device had **baked in** become fixed slots, so the successor starts where its predecessor left off: the EQ's four fixed bands are explicitly enabled as LowShelf / Bell / Bell / HighShelf, the phaser gets four all-pass stages at 50 % mix, the compressor a hard knee. Parameters with no counterpart — reverb freeze, EQ phase invert, the compressor's sidechain gain — are dropped rather than approximated.

The rewrite runs on the staged model at load, so the retired id leaves the project on the next save and every downstream comparison sees the canonical id. It reaches every entry point: tracks, racks, post-FX, mixer-analysis, the master track, and device / chain / rack presets.

**Control graph:** links into a migrated device are dropped rather than re-pointed. Their `paramIndex` addresses a parameter list that no longer exists, and these devices were hidden long before macros and automation shipped, so in practice there is nothing to lose — but a stale link that writes the wrong knob would be worse than an absent one.

**Nested state:** a device embedded in another one — an FX chain on a Drum Grid pad — has no `DeviceInfo`; its owner restores it straight from the saved plugin tree. Tracktion registers all eight retired types as **built-in** plugin types and `PluginManager::createPlugin` checks that list *before* `EngineBehaviour::createCustomPlugin`, so load aliases alone could never reach it. The conversion runs at the seam MAGDA owns, in `DrumGridPlugin`'s restore, before the engine builds anything: the tree adopts the successor's id and the values are seated on the plugin once it exists. Both entry points run the same `SlotMapping` table through one `convertSlots()`, so a conversion fix cannot apply to only one shape.

Compiled specs also gained `loadAliases`, which feed the agent catalog — an instruction still worded `pitch shift` resolves to `magda_pitch`.

## Cleanup that came with it

- `processors/te` is gone. The surviving stock devices move to `processors/internal/StockDeviceProcessors`; 4OSC already had its own processor. `VolumeProcessor` was dead and went with them.
- The seven now-unreachable chain UIs (Equaliser, Compressor, Reverb, Delay, Chorus, Phaser, PitchShift) plus FilterUI are deleted, along with the `LegacyUiKind` suppression machinery that existed only to stop them claiming compiled slots via `containsIgnoreCase`.
- `InternalPluginAliases.cpp` no longer hand-authors parameter names: every device that curates aliases now declares them next to itself, so the file just collects what the registry holds.

Net **−1943 lines**.

## Verification

- Full suite green: **1701 cases / 163,508 assertions**, including 16 new ones in `test_legacy_device_aliases.cpp` covering each conversion, the identity rewrite, idempotency, and the link/lane pruning.
- End to end through `magda-cli`: a project holding all eight devices with genuine Tracktion parameter values and v1 XML plugin state loads through a real engine boot — no unknown-type warnings, every value converted (`Threshold=-20`, `Ratio=4`, `Feedback=0.501187`, `Decay=80`, `Mode=2` for a highpass, …), and the state re-saved as a v2 document under the canonical id.
- End to end for the nested case: a hand-built project whose Drum Grid pad holds a Tracktion delay restores as `magda_delay` with Time 375 ms, Feedback 0.5012 and Mix 0.4 carried across, guarded by a new engine-level regression test.
- Two tests asserted the pre-migration behaviour and were updated: `test_master_track_serialization` (now expects `magda_eq` / `magda_compressor` after a roundtrip) and `test_curated_alias_loader` (now checks the devices' own alias tables).

## Follow-up

IR Reverb is the one browser-visible stock device with no successor to alias onto — it is `te::ImpulseResponsePlugin` over JUCE convolution, and Faust cannot express a runtime-loadable IR (`fi.conv`'s kernel is compile-time constant). #1980 covers replacing it natively.


magda/daw/audio/plugins/compiled/CompiledPluginRegistry.cpp
```diff
@@ -72,6 +72,10 @@ const CompiledPluginSpec* findCompiledPluginSpec(const juce::String& pluginId) {
 
         if (spec->aliasKey != nullptr && pluginId.equalsIgnoreCase(spec->aliasKey))
             return spec;
+
+        for (int i = 0; i < spec->loadAliasCount; ++i)
```
lucaromagnoli, line 76: **[P1] Preserve legacy values for nested DrumGrid effects.** DrumGrid restores embedded plugins directly from their saved `ValueTree`, so these aliases instantiate the compiled successor without running `migrateRetiredDevice()`. The old state keys do not match the compiled keys—for example, Tracktion Delay stores `feedback`, `mix`, and `length`, while `magda_delay` reads `magda_delay_*` properties—so an embedded retired effect silently loads defaults and changes the sound of old projects/presets. Could the nested tree be converted before construction, or retain a compatibility restore path for this case?

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -551,6 +559,11 @@ bool ProjectSerializer::deserializeProject(const juce::var& json, ProjectInfo& o
         return false;  // Failed - no state modified
     }
 
+    // Retired stock Tracktion effects load as their compiled successors. The
+    // master track is deserialized separately below and migrated there.
+    legacy_devices::migrateRetiredDevicesInProject(stagedTracks, nullptr, stagedAutomation,
```
lucaromagnoli: **[P2] Include the master track when pruning migrated automation.** This migrates and then commits `stagedAutomation` with `masterTrack == nullptr`; the master is migrated separately below, but its returned paths are discarded. A lane targeting a retired master effect therefore survives and writes its old `paramIndex` into an unrelated compiled slot—the stale-link behavior the project-wide migration avoids for normal tracks. Please stage/deserialise the master before this pass, or prune lanes for the paths returned by the later master migration before committing.

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -559,6 +572,7 @@ bool ProjectSerializer::deserializeProject(const juce::var& json, ProjectInfo& o
     if (masterTrackVar.isObject()) {
         TrackInfo masterTrackData;
         if (deserializeTrackInfo(masterTrackVar, masterTrackData)) {
+            legacy_devices::migrateRetiredDevicesInTrack(masterTrackData);
```
lucaromagnoli: This migrates the master track's devices but discards the returned paths, so automation lanes targeting a retired device on the master are not pruned on this path. `loadAndStage` passes `outData.masterTrack.get()` into `migrateRetiredDevicesInProject` and does prune them, so the two entry points disagree.

Not hypothetical: `test_master_track_serialization` expects `magda_eq` and `magda_compressor` on the master chain, so two of the eight genuinely live there. A surviving lane keeps a `paramIndex` into a parameter list that no longer exists, which is the case this PR argues against elsewhere ("a stale link that writes the wrong knob would be worse than an absent one").

Limited in practice, since `ProjectManager` only ever calls `loadAndStage`, and awkward to fix in place because `commitStagedData` runs before the master is deserialized. Either move the master deserialization ahead of the commit so one project pass covers it, or say in the comment that lane pruning is deliberately skipped here. This is a public method and the comment currently reads as though the master is fully handled.

magda/daw/core/LegacyDeviceAliases.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * wet and dry where the compiled one has a single mix. Parameters with no
+ * counterpart (reverb freeze, EQ phase invert, the compressor's sidechain gain)
+ * are dropped rather than approximated, and controls the retired device had
+ * baked in (its filter topology, its stage count) become fixed slot values so
+ * the successor starts where its predecessor left off.
+ */
+namespace legacy_devices {
+
+/// True when `pluginId` names one of the retired stock Tracktion effects.
+bool isRetiredDeviceId(const juce::String& pluginId);
```
lucaromagnoli: No production caller: the only uses are the two assertions in `test_legacy_device_aliases.cpp`. Either a call site was meant to use it, or it is speculative API on a public header.

magda/daw/core/LegacyDeviceAliases.cpp
```diff
@@ -0,0 +1,584 @@
...
+    return juce::jlimit(0.0f, 100.0f, value * 100.0f);
+}
+
+float unitToWidthPercent(float value, float) {
+    // The compiled reverb's width is 0..200 % with 100 % as plain stereo, which
+    // is where Tracktion's 0..1 width sits at 1.
+    return juce::jlimit(0.0f, 200.0f, value * 100.0f);
+}
+
+float reverbMix(float wet, float dry) {
```
lucaromagnoli, line 109: Worth a comment for the degenerate case: when a saved reverb has wet and dry both at 0 (silent), `total <= 0.0f` returns 0, which is fully dry, so the device becomes a passthrough where it previously output nothing. Probably never occurs in a real project, but it converts silence into audio rather than preserving it, which is a behaviour change rather than a conversion.

lucaromagnoli (commented):
Reviewed the mapping table against each compiled device's own `AliasSpec` array rather than taking it on trust, since a wrong slot index or enum ordinal writes the wrong knob silently. It all checks out.

**Slot indices**: compressor (Engine 0 ... Output 8), delay (Time 0, Sync 2, Feedback 3, Mix 4), chorus (Voices 0 ... Width 7), reverb (Engine 0, Mix 1, Predelay 2, Decay 3, Damping 4, Width 7), phaser (Rate 0 ... Mix 6), pitch (Engine 0, Pitch 1, Mix 4), filter (Cutoff 0 ... Mode 4), EQ (5 slots per band, output at 40). Leaving delay slot 1 and chorus slot 3 (both Division) untouched is right given both pin Sync to 0.

**Enum ordinals used as fixed values**: Phaser `Stages` is `{"2","4","6","8"}` so 1 is four stages; Reverb `Engine` is `{"Plate","Hall","Room"}` so 2 is Room; EQ band type is `{"HP","LowShelf","Bell","HighShelf","LP","Notch"}` so 1/2/2/3 is LowShelf/Bell/Bell/HighShelf. All match their comments.

The filter had a trap that got avoided: Mode 2 is HP, but `modeChoicesForEngine` shows most engines only offer `{"LP"}`. Pinning Engine to 0 (SVF) is what makes Mode 2 legal, so that coupling is load-bearing if the SVF choice is ever revisited.

`uniqueId = targetPluginId` matches the internal-device convention (`TrackManager.cpp:400,411` do the same for midichordengine and 4osc), and `identifierForDevice` falls back to `pluginId` anyway.

Three comments inline, one of which is a real gap on the master track. Two smaller notes:

- `retiredValue` is a linear scan per slot (EQ: 20 slots by N parameters). Irrelevant once at load, noting it only so it does not get "optimised" into a map later and counted as a win.
- `addAlias` dedupes within an entry but not across entries. Safe now that the TE registrations are gone and "delay"/"reverb"/"eq" have exactly one claimant each, but if a future device ever claims one of those words the resolution becomes iteration-order dependent.

Test coverage is good: the 16 cases hit each conversion, the identity rewrite, idempotency, non-retired passthrough, user-renamed devices, and both link and lane pruning. The one gap maps to the master-track comment below.

lucaromagnoli:
All five addressed in bd98a37c.

**[P1] Nested DrumGrid effects** — right, and worse than the comment says. Tracktion registers all eight retired types as **built-in** plugin types, and `PluginManager::createPlugin` scans `builtInTypes` *before* falling through to `EngineBehaviour::createCustomPlugin`. So the load aliases were never reachable for a nested tree at all: a pad's delay came back as a real `te::DelayPlugin` — not defaults, but the retired plugin itself, now with no processor, no UI and no catalog entry. I confirmed this against a hand-built pad chain before fixing it.

The conversion now runs at the seam MAGDA actually owns, in `DrumGridPlugin::restorePluginStateFromValueTree`, before the engine is asked to build anything. The tree adopts the successor's id and drops the consumed properties; the converted values are seated once the plugin exists. That also required a property table per retired device, since a nested tree names its values where the model indexes them — and Tracktion stores the Lowpass mode as text, so a property can declare the string that reads as 1. Both entry points now run the same `SlotMapping` table through one `convertSlots()`, so a conversion fix can't apply to only one shape. The dead `applyRetiredDeviceState` hook in `createCustomPlugin` is gone.

**[P2] Master track lanes** — fixed as you suggested: the master is staged before the migration pass, so one pass covers it and the two entry points agree. The awkwardness you flagged was real, and hoisting the deserialization was cleaner than pruning twice.

**`isRetiredDeviceId`** — you were right that it was speculative. Removed; the nested path genuinely needs *which* device replaced it, so it became `retiredDeviceSuccessor`, and the test asserts the mapping rather than a bool.

**`reverbMix` degenerate case** — commented. It resolves to dry-through deliberately: the compiled reverb's Mix has no position for silence, so the choice is between dry-through and burying it at minimum output. A reverb saved with wet and dry both at zero is far more likely a mis-set device than an intended one, and dry-through is the recoverable outcome.

**`retiredValue` linear scan / `addAlias` cross-entry collisions** — noted, no change. Agreed on both, including that the scan shouldn't be "optimised" later and counted as a win.

Thanks for checking every slot index and enum ordinal against the `AliasSpec` arrays — and for spotting the Engine 0 / Mode 2 coupling on the filter. I've left that comment in place; it's exactly the kind of thing that would bite silently if the SVF choice is revisited.

Suites: 1704 Catch2 cases / 163,522 assertions, plus the JUCE suite, all green on JUCE 8.0.12.

PR #1985, opened 2026-08-02
Dev/0.18.0

lucaromagnoli:


resources/controllers/scripts/launchkey_mini_mk4.lua
```diff
@@ -431,6 +557,34 @@ local function handle_pad_press(e)
...
+
+  -- Log the raw pad note next to the injected one. If you hear two notes per
+  -- pad, the device's raw note is reaching the track as well as this injected
+  -- one, which means the DAW port is not set as this script's Port In.
+  magda.log.info(string.format(
+    "[launchkey] note mode  raw=%d -> inject=%d %s",
+    e.number, note, isOn and "on" or "off"))
+
+  if isOn then
+    held_pads[e.number] = true
```
lucaromagnoli: [P2] Keep note-off routed to the originating track

`held_pads` records only `true`, while both this handler and `release_held_notes()` resolve the selected track again. If a pad is pressed on track A, the user changes selection (including through this script’s track buttons), and then releases it, the note-off is injected into track B—or nowhere—leaving the note sounding on A. Please store the originating track, channel, and translated note for each held pad and use those saved values for both release paths.

magda/daw/audio/DeviceMeteringManager.cpp
```diff
@@ -1,9 +1,88 @@
...
+    auto* manager = DeviceMeteringManager::getInstanceForEdit(plugin.edit);
+    if (!manager)
+        return node;
+
+    const auto devicePath = manager->getDevicePathForPlugin(&plugin);
+    if (!devicePath.isValid())
+        return node;
+
+    auto& measurer = manager->getOrCreateMeasurer(devicePath);
+    if (auto* gain = manager->getGainAtomic(devicePath))
```
lucaromagnoli: **Use-after-free reachable from the audio thread, on every normal shutdown.**

`DeviceGainNode` stores `std::atomic<float>& gain_`, and `te::LevelMeasuringNode` below it stores a `te::LevelMeasurer&`. Both reference fields of an `Entry`, which `entries_` owns by `unique_ptr`. Almost every device in the graph gets one of these taps.

The teardown order frees those entries while the graph is still live:

1. `TracktionEngineWrapperInit.cpp:611` does `audioBridge_.reset()`.
2. `~AudioBridge()` at `AudioBridge.cpp:213` calls `deviceMetering_.clear()`, which runs `entries_.clear()` and destroys every `Entry` synchronously on the message thread.
3. Only afterwards does `TracktionEngineWrapperInit.cpp:626` call `transport.freePlaybackContext()` (which is what actually tears down the node graph), and `:653` `closeDevices()`.

Between steps 2 and 3 the audio device is still open and the old graph is still installed, so the audio callback can run `DeviceGainNode::process()` and read `gain_.load()` on freed memory. TE keeps processing the graph for live monitoring even when the transport is stopped, so this does not need playback to be running. Under ASan, or on any machine where device close is slow, this is a reliable crash rather than a silent read.

`unregisterForEdit()` on the line above does not help: it only stops future graph rebuilds from adding taps, it cannot reach into the graph that is already wired.

The fix already exists in this file. `getRealtimeTap()` returns `shared_ptr`-backed storage, and the doc comment on `DeviceMeteringManager.hpp:132-138` says it exists precisely so `removeMeasurer()` and `clear()` cannot invalidate an already-wired audio-thread tap. `getGainAtomic()` and the raw `measurer` reference did not get the same treatment. Having the tap node hold the `RealtimeTapStorage` `shared_ptr` would close both windows.

Worth noting `getGainAtomic()` existed on main but had zero callers. This PR is what makes it reachable, so this is a new defect rather than an inherited one. `AudioBridge::resetTestState()` at `AudioBridge.cpp:264` reaches the same path, which is likely to show up as flakiness in the shared-engine JUCE tests.

magda/daw/ui/components/clips/ClipComponent.cpp
```diff
@@ -1631,6 +1675,9 @@ void ClipComponent::mouseDown(const juce::MouseEvent& e) {
             dragMode_ = DragMode::ResizeLeft;
             dragStartClipSnapshot_ = *clip;
             resizePreviewClip_ = *clip;
+            // Capture original state of other selected clips so the drag can
+            // preview them resizing together (the commit path already does)
+            captureMultiResizeSnapshots();
```
lucaromagnoli, line 1684: **Multi-clip left resize has no neighbour clamp, and can silently trim or delete unselected clips.**

The `ResizeRight` branch just below (lines 1693-1717) computes `multiResizeMaxDelta_` by scanning each selected clip's own track for the nearest non-selected neighbour, then clamps the whole gesture to the tightest gap. `ResizeLeft` calls `captureMultiResizeSnapshots()` and stops there. The preview loop applies the same `lengthDelta` to every selected clip with only a per-clip `jmax(0.1, ...)` floor.

So during a left drag the preview happily runs through a preceding unselected clip with no visual stop, and on `mouseUp` each `ResizeClipCommand` calls `ClipManager::resizeClipBeats` -> `resolveOverlaps` independently, which trims or fully removes whatever was in the way.

Repro: select clip A on track 1 and clip B on track 2, each with a short unselected clip just before it. Drag A's left edge well to the left. B's neighbour gets trimmed or deleted, off screen, with nothing in the drag preview to suggest it happened. Right-edge resize on the same selection stops cleanly at the boundary instead.

Single-clip left resize had the same gap before this PR, but it was self-evident because the user was watching the one edge they were dragging. Multi-select turns it into a hidden side effect spread across the selection.

magda/daw/ui/components/clips/ClipComponent.cpp
```diff
@@ -66,6 +66,29 @@ double timelineEndSeconds(const ClipInfo& clip, double bpm) {
...
 
+// Left-resize moves the clip start, so the preview has to carry the derived
+// source-domain phase (offset/loopStart/midiOffset) across as well as the
+// placement. Used for the dragged clip and every other clip in the selection.
+void applyLeftResizePreview(ClipInfo& target, const ClipInfo& preview, double bpm) {
+    ClipOperations::setTimelinePlacement(target, timelineStartSeconds(preview, bpm),
+                                         timelineLengthSeconds(preview, bpm), bpm);
+    target.offset = preview.offset;
+    target.loopStart = preview.loopStart;
+    target.midiOffset = preview.midiOffset;
```
lucaromagnoli, line 77: **`applyLeftResizePreview` drops `midiTrimOffset`, so piano roll notes drift against the clip body mid-drag.**

This helper carries `offset`, `loopStart` and `midiOffset` across, but not `midiTrimOffset`. For a non-looped MIDI clip, `ClipOperations::resizeContainerFromLeft` (`ClipOperations.hpp:259`) accumulates the start move into exactly that field, and `PianoRollGridComponent` depends on it via `getMidiVisibleRange` for absolute-mode note positioning.

The sibling `restoreLeftResizePreview` on line 85 does restore it, and its own comment calls out that `midiTrimOffset` is what `resizeContainerFromLeft` accumulates. So the restore side knows about the field and the apply side does not.

Repro: open the piano roll on a non-looped MIDI clip in absolute mode and drag the clip's left edge. The notes slide out of alignment with the clip body for the duration of the drag, then snap back correct on mouse up when the real `ResizeClipCommand` recomputes the field from scratch.

The omission came across verbatim from the old single-clip code, but this helper now runs for every clip in the selection, so it shows up on more clips at once.

magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
```diff
@@ -452,6 +412,127 @@ std::shared_ptr<FaustInstrumentPlugin::FaustState> FaustInstrumentPlugin::compil
...
+            (voice->fCurNote == kLegatoVoice && voice->fNextNote == pitch))
+            voice->keyOff(/*hard*/ false);
+    }
+}
+
+bool FaustInstrumentPlugin::handleMonoNoteOn(const std::shared_ptr<FaustState>& state, int note,
+                                             int velocity, int mode) {
+    const float g = static_cast<float>(velocity) / 127.0f;
+    const bool wasEmpty = heldNotes_.empty();
+    heldNotes_.push_back({note, g});
```
lucaromagnoli, line 495: **Heap allocation on the audio thread.**

`heldNotes_` is a plain `std::vector<HeldNote>` (`FaustInstrumentPlugin.hpp:243`) and is never reserved. `handleMonoNoteOn` is called only from `applyToBuffer` (line 906) while walking the block's MIDI, so any note-on that pushes the held-note stack past its previous high water mark allocates on the real-time thread. That includes the very first note after load, and later any bigger chord held under sustain.

A `reserve(128)` in the constructor bounds it, since MIDI cannot hold more than 128 distinct notes at once.

Separately, and pre-existing rather than from this PR: `inPtrs_`/`outPtrs_` get `.resize()`d inside `applyToBuffer` (line 815 here, and `FaustPlugin.cpp:546-547`). Steady state that is a no-op, but the first render after swapping to a DSP with more channels than any previously seen will grow capacity on the audio thread. Same one-line fix if you want to close it while you are in here.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -845,8 +849,8 @@ void PluginBrowserContent::showPluginContextMenu(const PluginBrowserInfo& plugin
                     device.format = magda::PluginFormat::VST3;
                 } else if (plugin.format == "AU" || plugin.format == "AudioUnit") {
                     device.format = magda::PluginFormat::AU;
-                } else if (plugin.format == "VST") {
-                    device.format = magda::PluginFormat::VST;
+                } else if (plugin.format == "LV2") {
+                    device.format = magda::PluginFormat::LV2;
```
lucaromagnoli, line 853: **LV2 in the browser makes an existing `juce::File` assertion reachable.**

This line is what puts LV2 plugins into the browser list. The "Show in Finder" handler at line 897 then does `juce::File pluginFile(plugin.fileOrIdentifier);` on the raw identifier. LV2 identifies a plugin by URI, not a path, for example `http://gareus.org/oss/lv2/midifilter#cctonote`, and `juce::File`'s constructor hits `jassertfalse` ("Illegal absolute path") for any non-absolute-path string.

This PR already diagnosed the exact hazard: `PluginScanCoordinator.hpp:27` adds `pluginDisplayName()` with a comment saying an LV2 URI "is not a path, so passing it to juce::File trips the 'Illegal absolute path' assertion", and `PluginSettingsDialog.cpp` was updated to route through it. Line 897 is the call site that got missed.

Repro: right click any LV2 plugin in the browser and pick Show in Finder. Debug and ASan builds abort. Release builds are a harmless no-op, since the bogus `File` just fails `.exists()` and falls into the `DBG` branch.

Guarding with `juce::File::isAbsolutePath(plugin.fileOrIdentifier)` before constructing the `File` handles it, and would also let the menu item be greyed out for LV2 rather than silently doing nothing.

lucaromagnoli (commented):
Review of the 0.18.0 aggregate diff (541 files, 45 commits). Most of this landed through its own PR already, so I focused on defects that only show up in the integrated result. Five findings inline, ordered by severity. The metering one is the only release blocker in my view.

Checked and cleared along the way:
- The `PluginFormat::VST=2` to `LV2=2` enum slot reuse is safe. `JUCE_PLUGINHOST_VST=1` never appears anywhere in this repo's history and the scanner only ever handled VST3/AU, so no saved project can hold format value 2.
- `ReallocationInhibitor` is not lost in the export rewrite, it moved into `OfflineRenderHelper.cpp:186` and covers the session lifetime.
- All four locales are complete: 431 keys each, no empty values.
- No non-ASCII in any changed workflow or shell script.
- The new convolution device is clean. `applyToBuffer` allocates nothing, `loadImpulseResponse` is wait-free by design, the IR rides in project state as a blob rather than a path so it cannot go missing, and `kConvolutionSlots`/`kConvolutionProperties` remaps all five parameters off the retired device including the MIDI-note to frequency conversion on both cutoffs. Its sub-blocked smoothing is actually more correct than the device it replaces.

Still outstanding: I have not finished reviewing the device state schema v2 migration, the Faust custom UI layer, the rest of the legacy device alias tables, or the packaging rules for the new wasmtime dependency. Of those, the alias tables and the pluginState v2 read path are the two most likely to hide a silent data-loss bug, so I would not merge before those get a look.

Issue #1859, opened 2026-07-25
MagdaApi: complete the external device, rack, and automation control surface


lucaromagnoli:
Part of #701. Blocks full WebSocket and MCP feature parity.

## Summary

Fill the safe, serializable control gaps through `MagdaApi`. Remote adapters must not bypass the facade and reach into global managers.

## Current gaps

- `TrackApi` has no device removal
- No stable device-path lookup for remote callers
- No device parameter read/write methods
- No serializable internal/external device catalogue
- `addDeviceToTrack` accepts a complete `DeviceInfo` instead of a catalogue ID plus placement
- Incomplete operations for devices inside racks and chains
- Automation enumeration is missing
- Some facade mutations do not consistently use available undoable commands

## Scope

- [x] Define stable remote device and catalogue identifiers
- [x] List available devices without plugin file paths
- [x] Add, remove, move, and bypass devices through `MagdaApi`
- [x] Enumerate and read parameters
- [x] Write parameters by stable device path and parameter ID
- [x] Support top-level and rack-chain placement
- [x] Enumerate automation lanes and targets
- [x] Add safe bulk automation point and clip operations
- [x] Route mutations through undoable commands
- [x] Extend `MockMagdaApi` and facade tests

## Acceptance

- Remote callers never construct or submit raw `DeviceInfo`.
- External plugins use stable catalogue IDs while filesystem locations remain private.
- Device removal and parameter changes work for top-level and rack-contained devices.
- Automation is discoverable and editable without raw registry access.
- Every mutation is undoable and stale targets return explicit not-found or conflict errors.

lucaromagnoli:
## Addressing decision: the remote device identifier is a serialized `ControlTarget`

This issue asks to "define stable remote device and catalogue identifiers". Settling that before the parameter read/write work starts, because the same answer is needed by #1757 and #1857. Implemented in #1991.

**Don't invent a new device-path scheme.** `ControlTarget` (`core/ControlTarget.hpp`) is already the canonical address for everything writable — plugin params, device macros, mod params, track volume/pan, sends, tempo — and its own docs state that every consumer speaks it: macro/mod links, automation lanes, MIDI bindings, alias resolvers. `TargetResolver::resolveToParam` is the single resolution point. A parallel remote-only scheme would be a fourth encoding of an address that already has three.

#1991 adds what was missing — a canonical serializer:

- `toVar`/`fromVar` on `ChainNodePath` and `ControlTarget` for persistence
[...]

Issue #1858, opened 2026-07-25
MCP: expose the MAGDA Remote API as tools and resources


lucaromagnoli:
Part of #701. Related to closed exploratory issue #860. Depends on #1854 and #1855.

## Summary

Add an MCP server adapter so external AI clients can inspect and control the currently running MAGDA session through the same operations used by WebSocket.

## Architecture

Expose a localhost Streamable HTTP MCP endpoint:

- MCP tools map to shared Remote API operations
- MCP resources and templates map to shared read operations
- Shared JSON Schemas become MCP tool schemas
- Shared state events become resource update notifications where supported

An optional stdio-to-HTTP proxy can follow for hosts without Streamable HTTP. Do not treat the existing outbound `MCPClient` as a server implementation.

## Initial resources

- `magda://project/current`
- `magda://tracks`
- `magda://tracks/{track_id}`
- `magda://clips/{clip_id}`
- `magda://selection`
- `magda://transport`
- `magda://session`
- `magda://devices/catalog`

## Scope

- [x] Implement initialization and protocol-version negotiation
- [x] Implement `tools/list` and `tools/call`
- [x] Implement `resources/list`, templates, and `resources/read`
- [x] Return structured results using shared output schemas
- [x] Share authentication, permissions, errors, and audit context
- [x] Send resource updates where client capabilities permit
- [ ] Test against at least two MCP hosts  
- [x] Document setup

## Acceptance

- A client can inspect the current project and transport.
- With edit permission it can create a track and MIDI clip, add notes, control playback, and undo each mutation as one action.
- WebSocket and MCP return equivalent structured data for the same operation.
- Protocol negotiation is not hardcoded to the existing `2024-11-05` client version.
- No MCP handler calls MAGDA managers directly.

lucaromagnoli:
Shipped in #2049, with the permissions, error and audit context of its fifth scope item completed by #1860 (#2056).

One item is deliberately left unticked: **testing against at least two MCP hosts** cannot be established from the code or from CI, and I have no evidence it has been done. It has moved to #2059 rather than being ticked on trust — closing this issue reflects that the engineering is complete, not that the integration has been observed.

Issue #1855, opened 2026-07-25
Remote API: add safe message-thread dispatch, revisions, and undo transactions


lucaromagnoli:
Part of #701. Depends on #1854.

## Summary

Implement the execution service between remote transports and `MagdaApi`. Network threads must never touch DAW model objects directly, and remote work must never block the audio thread.

## Scope

- [x] Add an asynchronous `RemoteApiService` dispatcher
- [x] Validate before hopping DAW reads and writes to the JUCE message thread
- [x] Copy results into remote DTOs on the message thread and serialize off-thread
- [x] Cancel queued work safely during shutdown or project replacement
- [x] Add monotonically increasing project revisions
- [x] Support `expected_revision` and reject stale writes
- [x] Make each mutating request one named undo step
- [x] Add request-ID idempotency for safe retries
- [x] Map validation, not-found, conflict, timeout, and internal failures to structured errors
- [x] Add unit and stress tests

Request context carries client identity, permission scopes, request ID, deadline, and expected project revision.

## Acceptance

- Worker-thread requests execute DAW access on the message thread.
- Shutdown and project replacement cannot leave callbacks accessing destroyed state.
- Retrying a completed mutation does not apply it twice.
- One remote mutation is reversed by one Undo action.
- No `MessageManagerLock`, blocking audio-thread work, or synchronous network I/O is introduced.

lucaromagnoli:
## Add the coalesced change source here, alongside revisions

#1857 (WebSocket subscriptions) and #1757 §4 (OSC feedback) both need model listeners collapsed into a coalesced stream of "what changed" — so a 100 Hz fader move produces one notification, not 100 JUCE callbacks per subscriber. Building it in either adapter means building it twice, or retrofitting the second one.

It belongs in this issue because **revisions are the coalescing key** — you collapse all changes since revision N — and "add monotonically increasing project revisions" is already in scope here. Splitting them means this issue ships revisions and #1857 immediately reopens that code.

**Share the change source; keep delivery private.** The two consumers are not the same downstream:

| | OSC feedback | WebSocket subscriptions |
|---|---|---|
| Transport | fire-and-forget UDP | ordered, reliable TCP |
| Client state | none | per-client queue + revision cursor |
| Failure mode | drop; next value corrects it | resync from known revision |
| Dedup key | OSC address | topic + revision |

Forcing OSC's stateless per-address push through a revisioned per-client queue would make it worse. So: one shared source, two private delivery layers.

[...]

Issue #1757, opened 2026-07-14
OSC support: bidirectional remote control (TouchOSC, hardware surfaces)


lucaromagnoli:
## Summary

Add OSC (Open Sound Control) support so MAGDA can be remote-controlled over UDP by TouchOSC/Open Stage Control-style touch surfaces, hardware consoles, and show-control software — with bidirectional feedback (moving a fader in MAGDA updates the surface, and vice versa).

OSC complements the planned WebSocket API (#701): WebSocket is the query/scripting/structured-data surface, OSC is the low-latency continuous-control surface. They should not share a protocol layer, but they address the same underlying model.

## Use cases

- **TouchOSC / Open Stage Control / Lemur** on a tablet: mixer page (faders, pans, mutes, solos), transport page, macro knobs for the focused device.
- **Hardware OSC consoles** (Behringer X-Touch in OSC mode, monome/grid bridges).
- **Show control / interop**: QLab, Max/MSP, SuperCollider, Ableton Link-adjacent rigs sending transport and parameter cues.

## What this builds on (analysis, July 2026)

MAGDA already has a complete MIDI control-surface stack, and OSC should be a **peer input into that same pipeline**, not a new one:

- `Binding` model (`core/controllers/Binding.hpp`): `BindingSource` → `Target` + mode + range/curve, persisted globally (`Config`) or per-project (`BindingRegistry::saveProject`).
- `ControlTarget` (`core/ControlTarget.hpp`) is the canonical address for everything writable — plugin params, device macros, mod params, track volume/pan, send levels, tempo — resolved by `TargetResolver::resolveToParam`.
- `ControllerRouter` (`audio/controllers/ControllerRouter.cpp`): receive on input thread → binding lookup + transform → atomic coalescing → `MessageManager::callAsync` → `ControllerParamWriter::write` on the message thread. `ControllerParamWriter` already writes via `setParameterFromHost` (ControllerParamWriter.cpp:69), so OSC parameter writes won't be dropped when LFOs/macros are active.
- MIDI-learn session machinery (`beginLearnSession`/`LearnCapture`) and the `ControllersDialog` preferences UI.
- `ControllerFeedback` sink — the hook for echoing value changes back out (currently for motorized faders/LEDs; OSC feedback is the same shape).
- Transport: `TransportApi` with UI-consistent play/stop/loop dispatchers (wired through `TimelineController` in `MainWindow.cpp`); position is in **beats** per the repo convention.
[...]

PR #1991, opened 2026-08-02
Canonical ControlTarget serialization and unambiguous remote device paths

lucaromagnoli:
## Why

`ControlTarget` is the declared canonical address for every writable parameter — its own docs say "every consumer — macro/mod links, automation lanes, MIDI bindings, alias resolvers — speaks `ControlTarget`". But it had no serializer, so the repo had grown three encodings of the same address and no canonical one:

1. `ControlTarget` itself — the intended form, unserializable.
2. `AliasRegistry.cpp:305` — a hand-rolled inline copy, comment and all (`// Inline ChainNodePath serialization`).
3. `remote_api_projection.cpp` — a third form that dropped `ChainNodePath::steps`.

Dropping the steps loses the section discriminator, and that turns out to be load-bearing.

## The bug

`DeviceId` is allocated from three separate counters — `nextFxDeviceId_`, `nextPostFxDeviceId_`, `nextMixerAnalysisDeviceId_` (TrackManager.hpp:431: "Post-fx and mixer-analysis keep their own ID spaces"). So id 3 legitimately exists in all three sections of one track. The discriminator lives in `ChainNodePath::steps.front()`.

Two places dropped it:

- **`AutomationTargetDto`** projected `kind + trackId + deviceId()` — so a lane on fx device #3 and one on post-fx device #3 of the same track serialized to byte-identical JSON. `automation.createLane` already publishes that shape under `API_VERSION 1.0`.
- **`DeviceDto`** projected `rackId`/`chainId` only, both null for top-level *and* post-fx devices — so `devices.list` returned the two as rows differing only in `name`. It also can't express a device nested more than one rack deep.

Nothing decodes either shape yet (the handlers are #1855/#1859 work), so the wire contract is still free to change. Once clients exist it isn't.

## What changed

- `toVar`/`fromVar` for `ChainNodePath` and `ControlTarget`, living next to the types. The path form is byte-identical to what alias files already contain, so existing saves keep loading; `kind` serializes as a stable string, not an enum ordinal.
- `AliasRegistry` routed through the canonical pair (45 lines → 8).
- `DevicePathDto` — the public projection, carrying `section` plus the full rack/chain route. Step types are strings (`"rack"`, `"post_fx"`) rather than enum ordinals: this is a wire contract consumed by scripts and MCP clients (#1858) and must not couple to internal `ChainStepType` ordering.
- `AutomationTargetDto.deviceId` → `devicePath`; `DeviceDto` gains `devicePath`, with a real `ChainNodePath` threaded through the graph walk. `rackId`/`chainId` stay as immediate-parent pointers for tree rendering, now documented as not-an-address.
- `makeDevicePathDto`/`toChainNodePath` for the write direction #1859 needs.
- One shared `devicePathFromJson` between device listings and automation targets, so the two can't drift.

The schema validator has no `$ref` support, so `devicePath` composes into both schemas programmatically via the existing pattern at `remote_api.cpp:412`.

## Also settled

Project IDs are **durable, not session-scoped**: `restoreTrack` keeps the saved id and only bumps the counter past it (TrackManager.cpp:891), and `recalculateIdCounters` rescans on load (:3075-3082). So `DevicePathDto` needs no secondary key. This contradicts the premise in #1757 — its 1-based-visible-order choice for OSC is still right, but because IDs are sparse, not because they're unstable.

## Tests

12 new cases: the three-space ambiguity at both the path and `devices.list` level, a two-rack-deep device, nested-rack round-trips, edit-scoped tempo carrying no path, and malformed-input rejection.

Full suite: **1724 cases, 1711 passed, 13 skipped, 0 failures** (163,690 assertions).

## Notes for review

- `magda_daw` and all tests compile clean. The IDE's clang index can't resolve JUCE headers in this checkout, so inline diagnostics are noise — the `make test-build` output is the real signal.
- Unrelated files that were already unformatted on `dev/0.19.0` are deliberately untouched; `make format` would sweep them into this diff.

Part of #701. Unblocks #1859 and #1757, which both needed one agreed parameter address.

### DIFF
--- a/magda/daw/api/remote_api.cpp
+++ b/magda/daw/api/remote_api.cpp
@@ -357,24 +357,57 @@ const juce::var& sessionSchema() {
     return value;
 }
 
-const juce::var& automationTargetSchema() {
+const juce::var& devicePathSchema() {
     static const auto value = parseSchema(R"json({
         "type":"object",
         "properties":{
-            "kind":{"type":"string","enum":["plugin_param","device_macro","mod_param",
-                "track_volume","track_pan","send_level","tempo"]},
-            "trackId":{"anyOf":[{"type":"null"},{"type":"integer","const":-2},
+            "trackId":{"anyOf":[{"type":"integer","const":-2},
                                 {"type":"integer","minimum":0}]},
-            "deviceId":{"type":["integer","null"],"minimum":0},
-            "parameterIndex":{"type":"integer","minimum":-1},
-            "modId":{"type":"integer","minimum":-1},
-            "modParameterIndex":{"type":"integer","minimum":-1},
-            "sendBusIndex":{"type":"integer","minimum":-1}
+            "section":{"type":"string","enum":["fx","post_fx","mixer_analysis"]},
+            "trackLevel":{"type":"boolean"},
+            "topLevelDeviceId":{"type":["integer","null"],"minimum":0},
+            "steps":{"type":"array","items":{
+                "type":"object",
+                "properties":{
+                    "type":{"type":"string","enum":["rack","chain","device"]},
+                    "id":{"type":"integer","minimum":0}
+                },
+                "required":["type","id"],
+                "additionalProperties":false
+            }}
         },
-        "required":["kind","trackId","deviceId","parameterIndex","modId","modParameterIndex",
-                    "sendBusIndex"],
+        "required":["trackId","section","trackLevel","topLevelDeviceId","steps"],
         "additionalProperties":false
     })json");
+    return value;
+}
+
+const juce::var& automationTargetSchema() {
+    static auto value = [] {
+        auto schema = parseSchema(R"json({
+            "type":"object",
+            "properties":{
+                "kind":{"type":"string","enum":["plugin_param","device_macro","mod_param",
+                    "track_volume","track_pan","send_level","tempo"]},
+                "devicePath":{},
+                "parameterIndex":{"type":"integer","minimum":-1},
+                "modId":{"type":"integer","minimum":-1},
+                "modParameterIndex":{"type":"integer","minimum":-1},
+                "sendBusIndex":{"type":"integer","minimum":-1}
+            },
+            "required":["kind","devicePath","parameterIndex","modId","modParameterIndex",
+                        "sendBusIndex"],
+            "additionalProperties":false
+        })json");
+        // Edit-scoped targets (tempo) carry no path, so null is permitted.
+        auto* nullable = new juce::DynamicObject();
+        juce::Array<juce::var> alternatives;
+        alternatives.add(parseSchema(R"json({"type":"null"})json"));
+        alternatives.add(devicePathSchema());
+        nullable->setProperty("anyOf", juce::var(alternatives));
+        schema["properties"].getDynamicObject()->setProperty("devicePath", juce::var(nullable));
+        return schema;
+    }();
     return value;
 }
 
@@ -866,11 +899,29 @@ juce::var toJson(const AutomationPointDto& dto) {
     return object;
 }
 
+juce::var toJson(const DevicePathDto& dto) {
+    auto object = new juce::DynamicObject();
+    object->setProperty("trackId", dto.trackId);
+    object->setProperty("section", dto.section);
+    object->setProperty("trackLevel", dto.trackLevel);
+    object->setProperty("topLevelDeviceId", nullableId(dto.topLevelDeviceId));
+
+    juce::Array<juce::var> steps;
+    for (const auto& step : dto.steps) {
+        auto* stepObject = new juce::DynamicObject();
+        stepObject->setProperty("type", step.type);
+        stepObject->setProperty("id", step.id);
+        steps.add(juce::var(stepObject));
+    }
+    object->setProperty("steps", juce::var(steps));
+    return object;
+}
+
 juce::var toJson(const AutomationTargetDto& dto) {
     auto object = new juce::DynamicObject();
     object->setProperty("kind", dto.kind);
-    object->setProperty("trackId", nullableId(dto.trackId));
-    object->setProperty("deviceId", nullableId(dto.deviceId));
+    object->setProperty("devicePath",
+                        dto.devicePath.has_value() ? toJson(*dto.devicePath) : juce::var());
     object->setProperty("parameterIndex", dto.parameterIndex);
     object->setProperty("modId", dto.modId);
     object->setProperty("modParameterIndex", dto.modParameterIndex);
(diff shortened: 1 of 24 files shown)


lucaromagnoli:
Both findings confirmed and fixed in b00b6f1d.

## P1 — track-level automation loses its address after reload

Correct, and worse than the review states in one respect: `ProjectSerializer::serializeChainNodePath` (AutomationModSerializer.cpp:349) was a **fourth** encoding of `ChainNodePath`, which this branch hadn't found. The PR description says three.

It wrote `trackId`, `topLevelDeviceId`, and `steps` but not `isTrackLevel`, so a saved `TrackVolume`/`TrackPan`/`SendLevel` target reloaded with `getType() == None`. The old projection emitted a bare `trackId` for those; the new `isValid()` guard emits `devicePath: null`. So this branch turned a degraded target into a lost one — a regression I introduced, not a pre-existing bug I merely exposed.

Fixed by taking the route the review suggested — project persistence now shares the canonical serializer, and `fromVar` decodes legacy files. The legacy inference is scoped deliberately: it fires only when `isTrackLevel` is **absent** from the JSON and the path names a track but has no steps and no device. That shape addresses nothing, and no producer other than the old writer emits it. An explicit `false` is honoured, so new data never takes the branch.

## P2 — malformed paths can silently retarget another device

Correct. Confirmed the mechanism at `TrackManager::getDeviceInChainByPath` (TrackManagerDevices.cpp:1035): `isPostFx()` tests `steps.front()`, so dropping a corrupt leading segment silently re-resolves a post-fx path against the main FX chain.

`fromVar` now fails closed on all four cases, including the two the review asked for beyond plain rejection:

- non-object step
- step type outside `ChainStepType`
- **segment id outside `ChainSegment`** — previously fell through `makeDevicePathDto`'s switch and read as the default `"fx"` section
- **segment anywhere but the front** — previously accepted and then ignored by every `isPostFx()`-style accessor

`AliasRegistry` also rejects an alias whose `path` is present but unreadable, instead of dropping it to `nullopt` — which is `StoredAlias`'s encoding for a user-global alias, so a corrupt path would have silently widened the alias's scope rather than failing.

## Tests

- `ChainNodePath - malformed input fails closed` — the four rejection cases plus a positive control
- `ChainNodePath - legacy paths without isTrackLevel are recovered` — including that an explicit `false` is not overridden, and that the inference doesn't fire on a path that already addresses something
- `Saved automation targets keep the flag that makes a track path valid` — goes through the public `serializeAutomation()` seam and asserts the written JSON parses back to the saved path

Full suite: 1726 cases, 1713 passed, 13 skipped, 0 failures.

The one test that asserted the old skip-and-continue behaviour was inverted rather than deleted — it had encoded the bug.

lucaromagnoli:
Both correct. Fixed in a10385f0.

## P2.1 — an empty object parsed as a valid address

`fromVar({})` read the missing `trackId` as a void var, which converts to `0`, then the legacy `isTrackLevel` inference promoted the result to `trackLevel(0)`. So my own legacy-recovery branch was the thing that turned a malformed object into an address — the previous round's fix supplied the second half of this bug.

Same mechanism inside a step: `{}` decoded as `Rack(0)`. And a present-but-non-array `steps` was skipped rather than rejected.

Now enforced via a `readRequiredInt` helper:

- `trackId` required and numeric
- `topLevelDeviceId` numeric when present
- each step's `type` and `id` required and numeric
- `isTrackLevel` must be a bool when present
- a present `steps` must be an array

The inference is unchanged but is now unreachable from a malformed object, since `trackId` has to be there first.

## P2.2 — a present null alias path still widened to global scope

Correct, and it makes the comment I added last round untrue. JSON `null` decodes to a void var, so the `isVoid()` guard I wrote read `"path": null` as absent — and `StoredAlias` uses an absent path to mean *user-global*, with `isValid()` not requiring one. A corrupt project-scoped alias therefore became a context-resolved global alias, which is exactly the outcome the fix claimed to prevent.

The `isVoid()` escape is gone: if `path` is present it must parse, `null` included. Nothing writes `"path": null` — `serializeStoredAlias` omits the property entirely for a global alias — so no legitimate data takes the new failure branch.

## Tests

`ChainNodePath - malformed input fails closed` gains: empty object, missing/non-numeric `trackId`, non-array `steps`, step missing `type` or `id`, non-boolean `isTrackLevel`.

New in `test_alias_registry.cpp`: path round-trip, absent path stays global, and a present-but-unreadable path rejected across `null`, non-object, and empty-object.

Full suite: 1729 cases, 1716 passed, 13 skipped, 0 failures.

Worth noting both rounds of findings were the same failure mode — treating absent or unreadable data as a default value instead of an error — and I reintroduced it in the fix for the first one. The parser is now uniformly required-and-typed rather than tolerant, which should close the class rather than the instances.

lucaromagnoli:
Both correct. Fixed in 5813df6f, plus one thing your finding led me to.

## P2.1 — out-of-range int64 narrowed instead of failing

Right, and it's the same silent-retargeting problem one layer down from where I'd been fixing it: `trackId: 4294967297` narrowed to track 1. Now rejected rather than narrowed.

Note I did **not** copy the remote API decoder exactly. `decodeBoundedInt` asserts and clamps, which is correct there because `validateJson` has already bounded the value against the schema and the clamp is a second line of defence. The path decoder has no schema in front of it, so clamping would invent an in-range id from out-of-range input. It rejects instead.

## P2.2 — mixed representations passed parsing

Confirmed. `getType()` resolves the three by precedence, so such a path is "valid" while `getDeviceId()` (prefers `topLevelDeviceId`) and `isPostFx()` (reads `steps.front()`) disagree about where it points. Track-level, top-level-device, and stepped routes are now mutually exclusive.

Checked this doesn't reject legitimate data before implementing: no factory produces a combination, and the direct `topLevelDeviceId` writers (`TrackChainContent`, `ChainPanel`, `DeviceMeteringManager`, `ParameterQueue`) all round-trip whatever was stored rather than synthesizing hybrids. A default path — all three empty — still parses, since edit-scoped targets serialize one and callers check `isValid()`.

## A fifth copy, found while checking the above

`aliases/Target.cpp` had its own `encodePath`/`decodePath` with a **byte-identical wire shape** to the canonical serializer and the old tolerant decoder — no bound checks, `continue` on unreadable steps. Routed through `toVar`/`fromVar`. That was the last remaining path-object parser; the PR description's "three encodings" is now five found and four retired.

The two `ChainNodePath` writers left are `TrackChainContent` and `ChainPanel`, which persist UI state as flat suffixed keys (`trackId0`, `stepTypes0`) rather than path objects. Different format, no parsing of untrusted structure, so out of scope here — but they are the last places that reconstruct a path without validation, and worth folding in if this shape gets touched again.

## Tests

Added: out-of-range `trackId` and step `id`; all three mixed-representation pairs; each representation alone still parsing; default path still parsing but invalid.

Full suite: 1730 cases, 1717 passed, 13 skipped, 0 failures.

lucaromagnoli:
The UI drag readers are done too — fde722f9. They were worse than I described when I scoped them out.

## Not UI state — drag payloads, and there were five copies

I called these "UI state persistence, flat suffixed keys, out of scope". They're actually chain-element **drag-and-drop payload decoders**, and the duplication was worse than the two files I named: `ChainPanel`, `TrackChainContent`, and `TrackHeadersPanel` held **five** near-identical copies between them, two of them inside a single file.

They decoded a path from flat `trackId`/`topLevelDeviceId`/`isTrackLevel` keys plus comma-joined `stepTypes`/`stepIds` strings — so they carried the same defects the last three rounds have been closing:

- missing property read as `0`
- `getIntValue()` on a non-numeric token yields `0`, turning an unreadable step into `Rack(0)`
- no range checks, no mutual-exclusivity check

## An existing bug they were hiding

All five rejected any `Segment` step (their range check is `Rack..Device`), but the writer encoded every step including `Segment`. So dragging a post-fx device serialized fine and then failed to decode — the drag silently did nothing.

That restriction is *correct* — post-fx and mixer-analysis are flat rail-managed sections with no reorder or reparent semantics — so I preserved it rather than "fixing" drag behaviour inside a serialization PR. It's now stated directly (`isPostFx() || isMixerAnalysis()`) instead of falling out of a step-type range check.

## What changed

New `ChainNodePathDrag.{hpp,cpp}` holds the payload layout; the path goes under a single key as the canonical object. One writer and one reader, used by all four call sites. The flat encoding is gone from the tree entirely.

Safe to change the format: a drag description is in-memory and lives for one gesture, so there's nothing persisted to stay compatible with. I checked the payload's flat keys weren't read anywhere else first — the one other `getProperty("trackId")` in the UI belongs to the unrelated `sessionClip` drag.

Malformed entries still drop only themselves, so a corrupt entry in a multi-node selection doesn't kill the whole drag.

## Tests

`test_chain_node_path_drag.cpp`, 7 cases: single and multi-node round-trips, unsuffixed fallback, corrupt entry dropping only itself, absent entry not read as a default path, out-of-range id rejected not narrowed, rail-managed sections refused, invalid path never reaching a drop handler.

Full suite: 1736 cases, 1723 passed, 13 skipped, 0 failures. App builds clean.

That accounts for every `ChainNodePath` decoder in the tree: seven found, six retired into the canonical serializer, and the seventh — `ParameterQueue`'s `PackedChainNodePath` — is a fixed-size in-memory struct that never parses untrusted input.

PR #1994, opened 2026-08-02
Clip model: separate container from content (#1901)

### LINKED ISSUE #1901: Clip model: separate container from content
Part of #1882 (Phase 0: lands in normal releases, against the current engine, so the plan compiler is designed against the final model).

Today a clip and its audio file are the same object: one ClipInfo, one file, boundaries coupled. Restructure to the event model (closest references: Cubase events inside parts, Bitwig audio events inside clips): the clip is a container hosting one or more audio events, and the audio inside a clip does not need to match the clip in length or start point.

**Three levels**

1. **Clip (container)**
- Timeline/session placement in beats: start, length, loop region.
- Launch properties (quantize, follow actions), enabled, clip-level gain.
- Owns an ordered list of audio events. Knows nothing about files directly.

2. **AudioEvent (placement of a source inside a clip)**
- Clip-relative start and length in beats.
- Anchor into the source (source-domain, samples, mapped through the warp map).
- Per-event interpretation: warp markers, stretch mode, transpose, reverse.
- Per-event gain and fades; crossfades where adjacent events overlap.

3. **Source (immutable file facts, pooled per file)**
- File reference and on-disk duration (the one allowed seconds value).
- Analysis: detected transients, detected BPM, waveform peaks reference.
- Not user-editable; media DB join point; seeds default interpretation for new events.

**Semantics**
- Clip bounds are a window: resizing a clip crops or reveals events, it never moves or stretches them. Events may extend past clip bounds (cropped at render) or leave gaps (silence).
- Multiple events sequence or overlap inside one clip: slicing, comping, and consolidation-without-render are event operations that never split the timeline clip.
- Takes: a take is a lane of events; the comp is the visible event sequence. Replaces the TE TAKES tree.
- Recording appends or replaces events inside the target clip.

**Mechanics**
- Serialization v2 with migration: a v1 clip loads as a clip containing exactly one event spanning it; sources pooled and deduplicated per file.
- Current engine bridge: each event maps to its own TE WaveAudioClip on the arranger (TE never sees the MAGDA clip grouping). Session slots on TE hold one clip, so session clips stay single-event until the native engine; the model allows more, the bridge restricts.
- Native engine: the clip player op consumes an event-list snapshot per clip: windowed, warped source reads per event with boundary crossfades. Consistent with the topology-only plan rule (events are data, not plan structure).
- Stretch and proxy caches key on source id plus interpretation hash; identical interpretations share cache entries as an implementation detail.
- Event geometry is beats; only the source anchor is source-domain, consistent with the beats-canonical constraint on #1889.
- Source ids are type-agnostic so a MIDI equivalent can reuse the scheme later.

lucaromagnoli:
Closes #1901. Phase 0 of #1882.

Restructures the clip model into the three levels the plan compiler will be designed against, without changing any behaviour.

## The model

- **`Source`** — immutable file facts (path, duration, sample rate, detected BPM and key), pooled per file in `SourcePool` and deduplicated on canonical path. Additive within a session, so an undone clip deletion still finds its source; garbage collected at save.
- **`AudioEvent`** — the placement of a source inside a clip. Clip-relative geometry in beats; the read anchor and the loop region are the only source-domain values, held as samples at the source's own rate. Every field that said how the audio is interpreted (BPM, warp, stretch mode, pitch, reverse, fades, channels) moved here off `ClipInfo`.
- **`ClipInfo`** — the container: placement, launch properties, enabled, clip gain, and an ordered list of events.

Storing the source domain in samples removes the parallel beats/seconds representations the old model kept in sync by hand. A source-BPM reinterpretation now leaves the audible region untouched and simply re-reads its musical length, which is exactly what the old reconciliation code in `applyAudioClipBeats` was written to achieve.

## Container and content are the same extent

`setPlacementBeats` keeps a single-event clip's event spanning it, so every existing resize, trim and split path behaves as before. Multi-event clips are representable, serializable and testable, but nothing in the app builds one yet: slicing, comping without render and multi-event session clips switch on later.

## Serialization

v2, with permanent read support for v1 (all three of its historical layouts). Sources are written once per project and referenced by id; a v1 clip loads as a clip holding exactly one event spanning it. Anchors convert from the old seconds, or autoTempo beats, into samples. A file that cannot be opened migrates at a nominal rate and rescales on the first successful open. v2 files do not open in older builds; there is no downgrade path.

## Deliberately unchanged

To keep the split behaviour-neutral, three things I had planned differently in the issue comment are left alone:

- fades stay in **seconds** rather than beats
- warp markers keep their `{sourceTime, warpTime}` seconds pair
- the engine bridge still keys its id map on `ClipId`: one clip is one event is one TE clip, so a `(ClipId, EventId)` key would be churn with no behaviour behind it. It grows the `EventId` when multi-event clips do.

MIDI clips keep their content as it was, and takes and comping keep their render-based shape. The MIDI loop moved into `MidiClipModel` but stays per-instance across ghost-clip propagation, as it was when it sat on the clip.

## Bugs found on the way

- `SourcePool` built a `juce::File` from a possibly-relative path in two places, which asserts in debug.
- The pool clear sat in `ClipManager::clearAllClips`, which the project load path calls during its commit phase, **after** the incoming sources are staged. Every restored event's source id was left dangling. Clearing now happens where a project is actually replaced.
- Moving the MIDI loop into `MidiClipModel` put it inside the ghost-clip shared `content`, so a ghost's loop toggle started propagating to its siblings. `copySharedContentFrom` now keeps it per-instance.

## Testing

Full suite green: 1698 passed, 0 failed, 13 skipped, 163587 assertions.

Not covered by the suite, worth exercising on a real project before merge:

- the sample-rate probe path, which needs real audio on disk
- migrating a v1 project whose source file is missing (nominal-rate anchor, rescale on first successful open)

### DIFF
--- a/magda/daw/audio/session/ClipSynchronizer.cpp
+++ b/magda/daw/audio/session/ClipSynchronizer.cpp
@@ -71,8 +71,8 @@ double followActionBaseLengthBeats(const ClipInfo& clip, double bpm) {
                 return loopLengthBeats;
         }
 
-        if (audioEventRef(clip).loopLengthBeats() > 0.0)
-            return audioEventRef(clip).loopLengthBeats();
+        if (clip.loopLengthInBeats() > 0.0)
+            return clip.loopLengthInBeats();
 
         const double sourceLength =
             audioEventRef(clip).sourceLengthSeconds(clip.getTimelineLength(bpm));
@@ -1639,9 +1639,10 @@ bool ClipSynchronizer::syncMidiClipToEngine(ClipId clipId, const ClipInfo* clip)
     }
 
     // Set up internal looping on the TE clip
-    if (clip->loopEnabled && audioEventRef(*clip).loopLengthBeats() > 0.0) {
-        // Use the stored loop region length, not the clip container length
-        double loopBeats = audioEventRef(*clip).loopLengthBeats();
+    if (clip->loopEnabled && clip->loopLengthBeats > 0.0) {
+        // Use the stored loop region length, not the clip container length.
+        // This is a MIDI clip: its loop is clip beats on the clip itself.
+        double loopBeats = clip->loopLengthBeats;
         auto& tempoSeq = edit_.tempoSequence;
         auto loopStartTime = tempoSeq.beatsToTime(te::BeatPosition::fromBeats(0.0));
         auto loopEndTime = tempoSeq.beatsToTime(te::BeatPosition::fromBeats(loopBeats));
--- a/magda/daw/audio/session/ClipSynchronizer.cpp
+++ b/magda/daw/audio/session/ClipSynchronizer.cpp
@@ -271,9 +271,13 @@ static void seedInterpretationFromLoopInfo(ClipInfo& clip, double numBeats, doub
 
     event->seedInterpretation(numBeats, bpm);
 
+    // A Source is shared by every clip on the file, so this estimate is only
+    // ever a stand-in for facts we could not read: an unresolved source with no
+    // duration at all. Once the file opens, probe() owns both numbers, and one
+    // event's odd interpBpm must not restate them for unrelated clips.
     if (auto* source = SourcePool::getInstance().getMutable(event->sourceId);
-        source != nullptr && source->durationSeconds <= 0.0 && event->interpTotalBeats > 0.0 &&
-        event->interpBpm > 0.0) {
+        source != nullptr && !source->isResolved() && source->durationSeconds <= 0.0 &&
+        event->interpTotalBeats > 0.0 && event->interpBpm > 0.0) {
         source->durationSeconds = event->interpTotalBeats * 60.0 / event->interpBpm;
     }
 }
--- a/magda/daw/audio/session/ClipSynchronizer.cpp
+++ b/magda/daw/audio/session/ClipSynchronizer.cpp
@@ -71,8 +71,8 @@ double followActionBaseLengthBeats(const ClipInfo& clip, double bpm) {
                 return loopLengthBeats;
         }
 
-        if (clip.loopLengthInBeats() > 0.0)
-            return clip.loopLengthInBeats();
+        if (clip.loopLengthInBeats(bpm) > 0.0)
+            return clip.loopLengthInBeats(bpm);
 
         const double sourceLength =
             audioEventRef(clip).sourceLengthSeconds(clip.getTimelineLength(bpm));
@@ -1087,10 +1087,17 @@ bool ClipSynchronizer::syncSessionClipToSlot(ClipId clipId) {
         }
         syncFollowActionToTracktionClip(*midiClipPtr, *clip, bpm);
 
-        // Set LaunchHandle looping state at creation time
+        // Set LaunchHandle looping state at creation time. Same whole-clip
+        // fallback as the loop range above: a legacy clip carrying the 0
+        // sentinel would otherwise get a loop range but a one-shot handle,
+        // and stop after a single pass.
         if (auto lh = midiClipPtr->getLaunchHandle()) {
-            if (clip->loopEnabled && loopLengthBeats > 0.0)
-                lh->setLooping(te::BeatDuration::fromBeats(loopLengthBeats));
+            if (clip->loopEnabled) {
+                const double handleBeats =
+                    loopLengthBeats > 0.0 ? loopLengthBeats : clip->getLengthInBeats(bpm);
+                if (handleBeats > 0.0)
+                    lh->setLooping(te::BeatDuration::fromBeats(handleBeats));
+            }
         }
 
         return true;
@@ -1141,9 +1148,13 @@ void ClipSynchronizer::launchSessionClip(ClipId clipId, bool forceImmediate) {
                     (srcLength / audioEventRef(*clip).speedRatio) * (bpm / 60.0);
                 launchHandle->setLooping(te::BeatDuration::fromBeats(loopDurationBeats));
             } else if (clip->isMidi()) {
-                // Already clip beats.
-                if (clip->loopLengthBeats > 0.0)
-                    launchHandle->setLooping(te::BeatDuration::fromBeats(clip->loopLengthBeats));
+                // Already clip beats, with the whole-clip fallback for the 0
+                // sentinel a legacy project can still carry.
+                const double handleBeats = clip->loopLengthBeats > 0.0
+                                               ? clip->loopLengthBeats
+                                               : clip->getLengthInBeats(bpm);
+                if (handleBeats > 0.0)
+                    launchHandle->setLooping(te::BeatDuration::fromBeats(handleBeats));
             }
         } else {
             launchHandle->setLooping(std::nullopt);
--- a/magda/daw/audio/session/ClipSynchronizer.cpp
+++ b/magda/daw/audio/session/ClipSynchronizer.cpp
@@ -462,7 +462,17 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
 
                         // Time-based loop state (existing behavior)
                         double projectBpm = projectBpmAtClip(edit_, *clip);
-                        if (clip->loopEnabled) {
+                        if (clip->loopEnabled && clip->isMidi()) {
+                            // MIDI loops in clip beats. Routing it through the
+                            // audio event's source region reads an empty one
+                            // and loops the whole container instead.
+                            const double loopBeats = clip->loopLengthBeats > 0.0
+                                                         ? clip->loopLengthBeats
+                                                         : clip->getLengthInBeats(projectBpm);
+                            teClip->setLoopRangeBeats(
+                                {te::BeatPosition::fromBeats(clip->loopStartBeats),
+                                 te::BeatPosition::fromBeats(clip->loopStartBeats + loopBeats)});
+                        } else if (clip->loopEnabled) {
                             const double timelineLength = clip->getTimelineLength(projectBpm);
                             if (audioEventRef(*clip).sourceLengthSeconds(timelineLength) > 0.0) {
                                 teClip->setLoopRange(te::TimeRange(
@@ -500,6 +510,13 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                                 if (loopLengthBeats > 0.0)
                                     launchHandle->setLooping(
                                         te::BeatDuration::fromBeats(loopLengthBeats));
+                            } else if (clip->isMidi()) {
+                                // Already clip beats, same as launchSessionClip.
+                                const double loopBeats =
+                                    clip->loopLengthBeats > 0.0
+                                        ? clip->loopLengthBeats
+                                        : clip->getLengthInBeats(projectBpmAtClip(edit_, *clip));
+                                launchHandle->setLooping(te::BeatDuration::fromBeats(loopBeats));
                             } else {
                                 double bpm = projectBpmAtClip(edit_, *clip);
                                 double loopLengthSeconds = audioEventRef(*clip).sourceLengthSeconds(
@@ -1049,10 +1066,14 @@ bool ClipSynchronizer::syncSessionClipToSlot(ClipId clipId) {
                                  te::BeatDuration::fromBeats(length), note.velocity, 0, nullptr);
         }
 
-        // Set looping if enabled
+        // Set looping if enabled. A v1 project can still carry the 0
+        // sentinel, which would otherwise ask TE to loop nothing.
         if (clip->loopEnabled) {
-            midiClipPtr->setLoopRangeBeats({te::BeatPosition::fromBeats(loopStartBeat),
-                                            te::BeatPosition::fromBeats(loopEndBeat)});
+            const double rangeLengthBeats =
+                loopLengthBeats > 0.0 ? loopLengthBeats : clip->getLengthInBeats(bpm);
+            midiClipPtr->setLoopRangeBeats(
+                {te::BeatPosition::fromBeats(loopStartBeat),
+                 te::BeatPosition::fromBeats(loopStartBeat + rangeLengthBeats)});
         }
 
         // Set per-clip launch quantization
@@ -1951,7 +1972,7 @@ bool ClipSynchronizer::syncAudioClipToEngine(ClipId clipId, const ClipInfo* clip
         }
 
         audioClipPtr->setOffset(te::TimeDuration::fromSeconds(
-            audioEventRef(*clip).engineOffsetSeconds(clip->loopEnabled)));
+            audioEventRef(*clip).engineOffsetSeconds(clip->loopEnabled, bpm)));
     }
 
     // 3b. REVERSE — must be handled before position/loop/offset sync.
--- a/magda/daw/audio/session/ClipSynchronizer.cpp
+++ b/magda/daw/audio/session/ClipSynchronizer.cpp
@@ -54,23 +54,30 @@ te::FollowAction toTracktionFollowAction(FollowAction action) {
     return te::FollowAction::none;
 }
 
+// The bridge keys its engine-id map on ClipId because Phase A of #1901 gives
+// every audio clip exactly one event: one clip is one TE clip. When a clip can
+// hold several events, the audioEventRef() calls below become a loop over
+// ClipInfo::events() and the map grows an EventId.
+
 bool isSessionLooping(const ClipInfo& clip) {
     return clip.view == ClipView::Session || clip.loopEnabled;
 }
 
 double followActionBaseLengthBeats(const ClipInfo& clip, double bpm) {
     if (isSessionLooping(clip)) {
-        if (clip.isAudio() && clip.autoTempo) {
-            auto [_, loopLengthBeats] = ClipOperations::getAutoTempoBeatRange(clip, bpm);
+        if (clip.isAudio() && audioEventRef(clip).autoTempo) {
+            auto [_, loopLengthBeats] = ClipOperations::getAutoTempoBeatRange(audioEventRef(clip));
             if (loopLengthBeats > 0.0)
                 return loopLengthBeats;
         }
 
-        if (clip.loopLengthBeats > 0.0)
-            return clip.loopLengthBeats;
+        if (audioEventRef(clip).loopLengthBeats() > 0.0)
+            return audioEventRef(clip).loopLengthBeats();
 
-        const double sourceLength = clip.getSourceLength(bpm);
-        const double speed = clip.speedRatio > 0.0 ? clip.speedRatio : 1.0;
+        const double sourceLength =
+            audioEventRef(clip).sourceLengthSeconds(clip.getTimelineLength(bpm));
+        const double speed =
+            audioEventRef(clip).speedRatio > 0.0 ? audioEventRef(clip).speedRatio : 1.0;
         if (sourceLength > 0.0)
             return (sourceLength / speed) * bpm / 60.0;
     }
@@ -146,7 +153,7 @@ bool syncFollowActionToTracktionClip(te::Clip& teClip, const ClipInfo& clip, dou
 constexpr double warpMarkerSyncEpsilonSeconds = 1.0e-6;
 
 bool warpMarkerMapsMatch(const juce::Array<te::WarpMarker*>& engineMarkers,
-                         const std::vector<ClipInfo::WarpMarker>& savedMarkers) {
+                         const std::vector<WarpMarker>& savedMarkers) {
     if (engineMarkers.size() != static_cast<int>(savedMarkers.size()))
         return false;
 
@@ -166,7 +173,7 @@ bool warpMarkerMapsMatch(const juce::Array<te::WarpMarker*>& engineMarkers,
 }
 
 bool restoreWarpMarkersIfNeeded(te::WarpTimeManager& warpManager,
-                                const std::vector<ClipInfo::WarpMarker>& markers) {
+                                const std::vector<WarpMarker>& markers) {
     // A valid map needs both boundary markers. In particular, do not insert a
     // lone saved marker on top of TE's recreated boundaries: that can create a
     // zero-length segment in WarpTimeManager's stretch-ratio calculation.
@@ -201,7 +208,7 @@ bool restoreWarpMarkersIfNeeded(te::WarpTimeManager& warpManager,
 bool syncWarpStateToTracktionClip(te::WaveAudioClip& audioClip, const ClipInfo& clip) {
     bool changed = false;
 
-    if (!clip.warpEnabled) {
+    if (!audioEventRef(clip).warpEnabled) {
         if (audioClip.getWarpTime()) {
             audioClip.setWarpTime(false);
             changed = true;
@@ -214,13 +221,13 @@ bool syncWarpStateToTracktionClip(te::WaveAudioClip& audioClip, const ClipInfo& 
         changed = true;
     }
 
-    if (!clip.warpMarkers.empty()) {
+    if (!audioEventRef(clip).warpMarkers.empty()) {
         auto& warpManager = audioClip.getWarpTimeManager();
 
         // A newly-created WarpTimeManager contains only its two default
         // boundary markers. Replace those with the marker map copied from the
         // source clip, but never overwrite an already-live edited map.
-        if (restoreWarpMarkersIfNeeded(warpManager, clip.warpMarkers))
+        if (restoreWarpMarkersIfNeeded(warpManager, audioEventRef(clip).warpMarkers))
             changed = true;
     }
 
@@ -244,36 +251,46 @@ static void syncAudioSourceInterpretationToLoopInfo(te::WaveAudioClip& audioClip
     auto waveInfo = audioClip.getWaveInfo();
     auto& li = audioClip.getLoopInfo();
 
-    if (clip.audio().interpretation.bpm > 0.0 &&
-        std::abs(li.getBpm(waveInfo) - clip.audio().interpretation.bpm) > 0.1) {
-        li.setBpm(clip.audio().interpretation.bpm, waveInfo);
+    if (audioEventRef(clip).interpBpm > 0.0 &&
+        std::abs(li.getBpm(waveInfo) - audioEventRef(clip).interpBpm) > 0.1) {
+        li.setBpm(audioEventRef(clip).interpBpm, waveInfo);
     }
 
-    if (clip.audio().interpretation.totalBeats > 0.0 &&
-        std::abs(li.getNumBeats() - clip.audio().interpretation.totalBeats) > 0.001) {
-        li.setNumBeats(clip.audio().interpretation.totalBeats);
+    if (audioEventRef(clip).interpTotalBeats > 0.0 &&
+        std::abs(li.getNumBeats() - audioEventRef(clip).interpTotalBeats) > 0.001) {
+        li.setNumBeats(audioEventRef(clip).interpTotalBeats);
     }
 }
 
-static void initialiseSourceLoopBeatsFromMetadata(ClipInfo& clip) {
-    if (!clip.isAudio() || !clip.autoTempo || clip.loopLengthBeats > 0.0)
+/// Seed the interpretation from Tracktion's loopInfo, filling gaps only. The
+/// file duration is a Source fact, so it lands on the pooled source.
+static void seedInterpretationFromLoopInfo(ClipInfo& clip, double numBeats, double bpm) {
+    auto* event = clip.primaryEvent();
+    if (event == nullptr)
         return;
 
-    const double sourceBpm = clip.audio().interpretation.bpm;
-    if (sourceBpm <= 0.0)
-        return;
+    event->seedInterpretation(numBeats, bpm);
 
-    const double sourceTotalBeats = clip.audio().interpretation.totalBeats;
-    clip.loopStartBeats = juce::jmax(0.0, clip.loopStart * sourceBpm / 60.0);
-
-    double loopBeats =
-        clip.loopLength > 0.0 ? clip.loopLength * sourceBpm / 60.0 : sourceTotalBeats;
-    if (sourceTotalBeats > 0.0) {
-        clip.loopStartBeats = juce::jmin(clip.loopStartBeats, sourceTotalBeats);
-        loopBeats = juce::jmin(loopBeats, sourceTotalBeats - clip.loopStartBeats);
+    if (auto* source = SourcePool::getInstance().getMutable(event->sourceId);
+        source != nullptr && source->durationSeconds <= 0.0 && event->interpTotalBeats > 0.0 &&
+        event->interpBpm > 0.0) {
+        source->durationSeconds = event->interpTotalBeats * 60.0 / event->interpBpm;
     }
+}
 
-    clip.loopLengthBeats = juce::jmax(0.0, loopBeats);
+/// A session clip imported before its source beat domain was known carries a
+/// zero-length loop region, meaning "the whole source". Give it a real one now
+/// that the interpretation has arrived.
+static void initialiseSourceLoopRegionFromMetadata(ClipInfo& clip) {
+    auto* event = clip.primaryEvent();
+    if (event == nullptr || !event->autoTempo || event->loopLengthSamples > 0)
+        return;
+    if (event->interpBpm <= 0.0 || event->interpTotalBeats <= 0.0)
+        return;
+
+    const double startBeats = juce::jmin(event->loopStartBeats(), event->interpTotalBeats);
+    event->setLoopStartBeats(startBeats);
+    event->setLoopLengthBeats(juce::jmax(0.0, event->interpTotalBeats - startBeats));
 }
 
 ClipSynchronizer::ClipSynchronizer(te::Edit& edit, TrackController& trackController,
@@ -424,7 +441,7 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                         needsGraphReallocation;
 
                     // AutoTempo handling for audio clips
-                    bool isAutoTempoAudio = clip->isAudio() && clip->autoTempo;
+                    bool isAutoTempoAudio = clip->isAudio() && audioEventRef(*clip).autoTempo;
 
                     // configureSessionAutoTempo applies the stretch mode
                     // itself; remember what it was so the stretch-mode block
@@ -446,10 +463,14 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                         // Time-based loop state (existing behavior)
                         double projectBpm = projectBpmAtClip(edit_, *clip);
                         if (clip->loopEnabled) {
-                            if (clip->getSourceLength(projectBpm) > 0.0) {
+                            const double timelineLength = clip->getTimelineLength(projectBpm);
+                            if (audioEventRef(*clip).sourceLengthSeconds(timelineLength) > 0.0) {
                                 teClip->setLoopRange(te::TimeRange(
-                                    te::TimePosition::fromSeconds(clip->getTeLoopStart()),
-                                    te::TimePosition::fromSeconds(clip->getTeLoopEnd(projectBpm))));
+                                    te::TimePosition::fromSeconds(
+                                        audioEventRef(*clip).engineLoopStartSeconds()),
+                                    te::TimePosition::fromSeconds(
+                                        audioEventRef(*clip).engineLoopEndSeconds(
+                                            timelineLength))));
                             }
                         } else if (teClip->isLooping()) {
                             teClip->disableLooping();
@@ -475,14 +496,15 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                                 // AutoTempo: loop beats come from beat fields
                                 double bpm = projectBpmAtClip(edit_, *clip);
                                 auto [loopStartBeats, loopLengthBeats] =
-                                    ClipOperations::getAutoTempoBeatRange(*clip, bpm);
+                                    ClipOperations::getAutoTempoBeatRange(audioEventRef(*clip));
                                 if (loopLengthBeats > 0.0)
                                     launchHandle->setLooping(
                                         te::BeatDuration::fromBeats(loopLengthBeats));
                             } else {
                                 double bpm = projectBpmAtClip(edit_, *clip);
-                                double loopLengthSeconds =
-                                    clip->getSourceLength(bpm) / clip->speedRatio;
+                                double loopLengthSeconds = audioEventRef(*clip).sourceLengthSeconds(
+                                                               clip->getTimelineLength(bpm)) /
+                                                           audioEventRef(*clip).speedRatio;
                                 double bps = bpm / 60.0;
                                 double loopLengthBeats = loopLengthSeconds * bps;
                                 launchHandle->setLooping(
@@ -498,22 +520,24 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                         auto* audioClip = dynamic_cast<te::WaveAudioClip*>(teClip);
                         if (audioClip) {
                             // Pitch
-                            bool isAnalog = clip->isAnalogPitchActive();
-                            if (clip->autoPitch != audioClip->getAutoPitch())
-                                audioClip->setAutoPitch(isAnalog ? false : clip->autoPitch);
+                            bool isAnalog = audioEventRef(*clip).isAnalogPitchActive();
+                            if (audioEventRef(*clip).autoPitch != audioClip->getAutoPitch())
+                                audioClip->setAutoPitch(isAnalog ? false
+                                                                 : audioEventRef(*clip).autoPitch);
                             if (isAnalog) {
                                 if (std::abs(audioClip->getPitchChange()) > 0.001f)
                                     audioClip->setPitchChange(0.0f);
                             } else {
-                                if (std::abs(audioClip->getPitchChange() - clip->pitchChange) >
-                                    0.001f)
-                                    audioClip->setPitchChange(clip->pitchChange);
+                                if (std::abs(audioClip->getPitchChange() -
+                                             audioEventRef(*clip).pitchChange) > 0.001f)
+                                    audioClip->setPitchChange(audioEventRef(*clip).pitchChange);
                             }
-                            if (audioClip->getTransposeSemiTones(false) != clip->transpose)
-                                audioClip->setTranspose(clip->transpose);
+                            if (audioClip->getTransposeSemiTones(false) !=
+                                audioEventRef(*clip).transpose)
+                                audioClip->setTranspose(audioEventRef(*clip).transpose);
                             // Playback
-                            if (clip->isReversed != audioClip->getIsReversed())
-                                audioClip->setIsReversed(clip->isReversed);
+                            if (audioEventRef(*clip).reversed != audioClip->getIsReversed())
+                                audioClip->setIsReversed(audioEventRef(*clip).reversed);
                             // Per-Clip Mix
                             {
                                 float combinedGain = clip->volumeDB + clip->gainDB;
@@ -526,11 +550,12 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                             if (audioClip->getLaunchFadeSamples() != clip->launchFadeSamples)
                                 audioClip->setLaunchFadeSamples(clip->launchFadeSamples);
 
-                            auto desiredMode =
-                                static_cast<te::TimeStretcher::Mode>(clip->timeStretchMode);
+                            auto desiredMode = static_cast<te::TimeStretcher::Mode>(
+                                audioEventRef(*clip).timeStretchMode);
                             if (!isAnalog && desiredMode == te::TimeStretcher::disabled &&
-                                (clip->autoTempo || clip->warpEnabled ||
-                                 std::abs(clip->speedRatio - 1.0) > 0.001))
+                                (audioEventRef(*clip).autoTempo ||
+                                 audioEventRef(*clip).warpEnabled ||
+                                 std::abs(audioEventRef(*clip).speedRatio - 1.0) > 0.001))
                                 desiredMode = te::TimeStretcher::defaultMode;
                             if (isAnalog)
                                 desiredMode = te::TimeStretcher::disabled;
@@ -749,7 +774,7 @@ bool ClipSynchronizer::syncSessionClipToSlot(ClipId clipId) {
     // migrated temp project media), recreate it so TE follows ClipManager.
     if (auto* existingSlotClip = slot->getClip()) {
         if (clip->isAudio()) {
-            juce::File desiredAudioFile(clip->audio().source.filePath);
+            juce::File desiredAudioFile(audioEventRef(*clip).sourceFilePath());
             if (desiredAudioFile.existsAsFile()) {
                 if (auto* existingAudioClip = dynamic_cast<te::WaveAudioClip*>(existingSlotClip)) {
                     if (existingAudioClip->getOriginalFile() != desiredAudioFile) {
@@ -778,13 +803,13 @@ bool ClipSynchronizer::syncSessionClipToSlot(ClipId clipId) {
     // TE's free functions insertWaveClip(ClipOwner&, ...) and insertMIDIClip(ClipOwner&, ...)
     // accept ClipSlot as a ClipOwner, creating the clip's ValueTree directly in the slot.
     if (clip->isAudio()) {
-        if (clip->audio().source.filePath.isEmpty())
+        if (audioEventRef(*clip).sourceFilePath().isEmpty())
             return false;
 
-        juce::File audioFile(clip->audio().source.filePath);
+        juce::File audioFile(audioEventRef(*clip).sourceFilePath());
         if (!audioFile.existsAsFile()) {
             DBG("ClipSynchronizer::syncSessionClipToSlot: Audio file not found: "
-                << clip->audio().source.filePath);
+                << audioEventRef(*clip).sourceFilePath());
             return false;
         }
 
@@ -821,12 +846,11 @@ bool ClipSynchronizer::syncSessionClipToSlot(ClipId clipId) {
             auto& loopInfoRef = audioClipPtr->getLoopInfo();
             auto waveInfo = audioClipPtr->getWaveInfo();
             if (auto* mutableClip = cm.getClip(clipId)) {
-                bool sourceInterpretationBpmWasUnset =
-                    mutableClip->audio().interpretation.bpm <= 0.0;
-                mutableClip->setSourceMetadata(loopInfoRef.getNumBeats(),
+                bool sourceInterpretationBpmWasUnset = audioEventRef(*mutableClip).interpBpm <= 0.0;
+                seedInterpretationFromLoopInfo(*mutableClip, loopInfoRef.getNumBeats(),
                                                loopInfoRef.getBpm(waveInfo));
-                initialiseSourceLoopBeatsFromMetadata(*mutableClip);
-                if (sourceInterpretationBpmWasUnset && mutableClip->autoTempo) {
+                initialiseSourceLoopRegionFromMetadata(*mutableClip);
+                if (sourceInterpretationBpmWasUnset && audioEventRef(*mutableClip).autoTempo) {
                     double projectBpm = projectBpmAtClip(edit_, *clip);
                     cm.refreshDerivedSeconds(clipId, projectBpm);
                     cm.forceNotifyClipPropertyChanged(clipId);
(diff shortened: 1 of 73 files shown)


magda/daw/core/SourcePool.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+bool SourcePool::resolveFacts(SourceId id) {
+    auto* source = getMutable(id);
+    if (source == nullptr)
+        return false;
+
+    const bool wasResolved = source->isResolved();
+    probe(*source);
+    return !wasResolved && source->isResolved();
```
lucaromagnoli, line 160: The transition is reported, but no rescale actually happens. `resolveFacts` and `relink` change the effective sample rate while every referring event keeps sample counts computed at the nominal 48 kHz; the relink callers also ignore this return value. Resolving a 44.1 kHz file therefore shifts anchors and loop regions. Please rescale `sourceAnchorSamples`, `loopStartSamples`, and `loopLengthSamples` for every event referencing this source (and test the resulting values, not only the boolean).

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -446,10 +463,14 @@ bool ClipSynchronizer::syncClipPropertyToEngine(ClipId clipId) {
                         // Time-based loop state (existing behavior)
                         double projectBpm = projectBpmAtClip(edit_, *clip);
                         if (clip->loopEnabled) {
-                            if (clip->getSourceLength(projectBpm) > 0.0) {
+                            const double timelineLength = clip->getTimelineLength(projectBpm);
+                            if (audioEventRef(*clip).sourceLengthSeconds(timelineLength) > 0.0) {
```
lucaromagnoli, line 481: This branch also runs for session MIDI clips. `audioEventRef` then returns the default event, so this sets the TE loop to `[0, containerLength]`; the launch-handle block below similarly derives the full container length and ignores `clip.loopStartBeats` / `clip.loopLengthBeats`. A 4-beat MIDI loop in a 16-beat clip becomes a 16-beat loop after any property sync. Please split the MIDI path and apply its beat-domain loop range directly.

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -707,6 +722,64 @@ juce::var ProjectSerializer::serializeTracks() {
...
+    // within a session so an undone delete can find its source again, and save
+    // is where that garbage is collected.
+    std::unordered_set<SourceId> live;
+    for (const auto& clip : ClipManager::getInstance().getClips()) {
+        if (!clip.isAudio())
+            continue;
+        for (const auto& event : clip.audio().events)
+            live.insert(event.sourceId);
+    }
+    SourcePool::getInstance().retainOnly(live);
```
lucaromagnoli: This mutates live session state during serialization and breaks delete undo across a save. The undo stack survives saving, but `DeleteClipCommand` only snapshots the `SourceId`; after this GC, undo restores a clip whose source no longer exists. It also happens before the file write succeeds. Could serialization filter the emitted snapshot without deleting pool entries, or retain source data in undo snapshots?

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -251,6 +253,13 @@ bool ProjectSerializer::loadAndStage(const juce::File& file, StagedProjectData&
             return false;
         }
 
+        // Sources must be pooled before the clips that reference them by id,
+        // and after the outgoing project's pool is dropped. This cannot live in
+        // ClipManager::clearAllClips: the commit phase calls that AFTER the
+        // sources are staged, which would strand every event's source id.
+        SourcePool::getInstance().clear();
```
lucaromagnoli: This breaks the atomic-staging guarantee: the global pool is cleared before clips and automation have validated. If either later staging step returns `false`, the currently open clips remain but all of their source IDs are now dangling. Could sources be staged independently and installed only in `commitStaged` (or the previous pool restored on failure)?

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -707,6 +722,64 @@ juce::var ProjectSerializer::serializeTracks() {
...
+        source.detectedKeyRoot = sourceObj->getProperty("detectedKeyRoot").toString().toStdString();
+        source.detectedKeyScale =
+            sourceObj->getProperty("detectedKeyScale").toString().toStdString();
+        const auto id = pool.insert(source);
+
+        // A project saved while the file was missing carries sampleRate 0, so
+        // its events' anchors were computed at the nominal rate. Re-probing now
+        // resolves the source and rescales them.
+        if (id != INVALID_SOURCE_ID && !source.isResolved())
+            pool.resolveFacts(id);
```
lucaromagnoli: **High.** The rescale this comment promises does not exist. resolveFacts returns the 0 to real transition, but this call discards it, as do the relink calls in MediaCollector and ProjectManager; no code anywhere scales sourceAnchorSamples / loopStartSamples / loopLengthSamples by realRate over nominal. Worse, resolving here is actively harmful while the rescale is missing: a project saved while a 44.1 kHz file was missing stores anchors computed at the 48 kHz nominal rate, and this re-probe makes them all reinterpret at 44.1 kHz on load, shifting every position by about 9%. Leaving the source unresolved would at least keep the project self-consistent. The migration and source-pool tests assert only the returned bool, which is how this survived the suite.

magda/daw/project/serialization/ProjectSerializer.cpp
lucaromagnoli: **High.** This GC runs on every save including autosave, and `live` is built from current clips only. Delete a clip, let autosave fire, undo: DeleteClipCommand restores the ClipInfo verbatim through restoreClip, which never re-acquires, so the restored event's sourceId dangles and the next save persists it. That contradicts the lifetime contract in SourcePool.hpp ("an undone clip deletion finds its source still there"). The clipboard has the same exposure. Either include undo-stack and clipboard sources in `live`, or make restoreClip re-acquire by path.

magda/daw/project/serialization/ProjectSerializer.cpp
lucaromagnoli: **High.** Two problems with clearing the live pool here. First, it runs mid-staging: if clip or automation staging fails after this line, the still-open project keeps its clips but their sourceIds now point into a cleared pool repopulated from the failed file, where ids can collide numerically and silently repoint live clips at wrong files; the failure comment below still claims no state was modified. Second, loadProjectAsync runs loadAndStage on a background std::thread, so this mutates a singleton documented as message-thread-only while the open project's UI reads it through sourceRateOf on every paint. Staging sources into StagedProjectData and swapping the pool during the message-thread commit fixes both.

magda/daw/project/serialization/ProjectSerializer.hpp
```diff
@@ -24,6 +24,17 @@ struct StagedProjectData {
...
+/**
+ * @brief Project file schema version.
+ *
+ * 1 (written as no key at all): a clip carried one audio source and one set of
+ *   playback fields.
+ * 2: the clip hosts a list of events referencing pooled sources (#1901). Read
+ *   support for 1 is permanent; nothing writes it any more, so a v2 file does
+ *   not open in an older build.
+ */
+constexpr int kProjectSchemaVersion = 2;
```
lucaromagnoli, line 51: **Medium.** This claim does not hold: old builds validate only that magdaVersion is a non-empty string and ignore unknown keys, so a v2 file opens in an old build with every audio clip silently stripped of its source, and a save from there destroys the project. Nothing the v2 writer emits can make already-shipped readers reject, so the comment should state the real failure mode rather than promise a gate.

magda/daw/ui/components/clips/ClipComponent.cpp
```diff
@@ -3559,7 +3591,7 @@ void ClipComponent::showContextMenu() {
             if (c && c->isAudio() && takeIndex >= 0 &&
                 takeIndex < static_cast<int>(c->audio().takes.size())) {
                 c->audio().currentTakeIndex = takeIndex;
-                c->audio().source.filePath = c->audio().takes[takeIndex].filePath;
+                magda::audioEventRef(*c).sourceFilePath() = c->audio().takes[takeIndex].filePath;
```
lucaromagnoli: **High.** This assignment is a no-op: audioEventRef returns a const ref and sourceFilePath() returns a juce::String by value, so this writes into a temporary and the event keeps its old sourceId. Only currentTakeIndex changes; the old take keeps playing and painting. ClipManager::setAudioClipCurrentTake already does this correctly (re-acquires the source, pushes an undo snapshot), so this branch can collapse to that call.

magda/daw/ui/panels/content/inspector/clip/ClipInspectorSections.cpp
```diff
@@ -1095,7 +1108,8 @@ void ClipInspector::initClipPropertiesSection() {
         double newLoopStartBeats = clipLoopStartValue_->getValue();
         if (clip->isMidi()) {
             const double newOffsetBeats =
-                juce::jmax(0.0, newLoopStartBeats + clip->midiOffset - clip->loopStartBeats);
+                juce::jmax(0.0, newLoopStartBeats + clip->midiOffset -
```
lucaromagnoli: **High.** This is the isMidi() branch, but audioEventRef on a MIDI clip returns the static default event, whose loopStartBeats() is always 0. The MIDI loop start lives in clip->loopStartBeats on the container. Any MIDI clip with a nonzero loop start now gets a wrong midiOffset written, shifting notes.

magda/daw/ui/panels/content/inspector/clip/ClipInspectorSections.cpp
```diff
@@ -1151,9 +1166,10 @@ void ClipInspector::initClipPropertiesSection() {
         if (clip->isAudio()) {
             const double newLoopEndSeconds =
                 displayBeatsToAudioSourceSeconds(*clip, newLoopEndBeats, bpm);
-            newLoopLengthSeconds = juce::jmax(0.0, newLoopEndSeconds - clip->getSourceLoopStart());
+            newLoopLengthSeconds =
+                juce::jmax(0.0, newLoopEndSeconds - magda::audioEventRef(*clip).loopStartSeconds());
         } else {
-            double loopStartBeats = clip->loopStartBeats;
+            double loopStartBeats = magda::audioEventRef(*clip).loopStartBeats();
```
lucaromagnoli: **High.** Same swap as the loop-start handler above: this is the non-audio branch, so audioEventRef(*clip).loopStartBeats() is always 0 and the loop length comes out inflated by the loop start. Should read clip->loopStartBeats.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -1892,26 +1930,28 @@ bool ClipSynchronizer::syncAudioClipToEngine(ClipId clipId, const ClipInfo* clip
...
+                    te::TimePosition::fromSeconds(audioEventRef(*clip).engineLoopStartSeconds()),
+                    te::TimePosition::fromSeconds(
+                        audioEventRef(*clip).engineLoopEndSeconds(clip->getTimelineLength(bpm)))));
             }
         }
 
-        audioClipPtr->setOffset(
-            te::TimeDuration::fromSeconds(clip->getTeOffset(clip->loopEnabled, bpm)));
+        audioClipPtr->setOffset(te::TimeDuration::fromSeconds(
+            audioEventRef(*clip).engineOffsetSeconds(clip->loopEnabled)));
```
lucaromagnoli: **Medium.** This call dropped the bpm argument the old getTeOffset received; bpm is in scope and the parallel call sites pass it. With projectBpm defaulting to 0, engineOffsetSeconds skips its autoTempo branch and returns raw source seconds, so reversed beat-mode clips get the wrong trim whenever interpBpm differs from the project tempo, which is exactly the bug this block's comment says it prevents.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -983,9 +1013,10 @@ bool ClipSynchronizer::syncSessionClipToSlot(ClipId clipId) {
         // Apply midiOffset: exclude notes before offset, shift remaining notes
         auto& sequence = midiClipPtr->getSequence();
         double bpm = projectBpmAtClip(edit_, *clip);
-        double srcLength = clip->getSourceLength(bpm);
-        double loopStartBeat = clip->loopStart * (bpm / 60.0);
-        double loopLengthBeats = srcLength * (bpm / 60.0);
+        // A MIDI clip's loop is already clip beats. Deriving it from an audio
+        // event would read an empty region and loop the whole clip instead.
+        double loopStartBeat = clip->loopStartBeats;
+        double loopLengthBeats = clip->loopLengthBeats;
```
lucaromagnoli, line 1040: **Medium.** A MIDI session clip with loopEnabled and loopLengthBeats == 0 used to fall back to the whole clip length; now loopEndBeat == loopStartBeat and the setLoopRangeBeats below runs unguarded, producing a zero-length loop range, and launchSessionClip skips setLooping. Current creation paths always populate the field, but a v1 project or a duplicate of a legacy clip can still carry the 0 sentinel. Guard the setLoopRangeBeats like the note truncation above and restore a whole-clip fallback.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -244,36 +251,46 @@ static void syncAudioSourceInterpretationToLoopInfo(te::WaveAudioClip& audioClip
...
-    const double sourceTotalBeats = clip.audio().interpretation.totalBeats;
-    clip.loopStartBeats = juce::jmax(0.0, clip.loopStart * sourceBpm / 60.0);
+    event->seedInterpretation(numBeats, bpm);
 
-    double loopBeats =
-        clip.loopLength > 0.0 ? clip.loopLength * sourceBpm / 60.0 : sourceTotalBeats;
-    if (sourceTotalBeats > 0.0) {
-        clip.loopStartBeats = juce::jmin(clip.loopStartBeats, sourceTotalBeats);
-        loopBeats = juce::jmin(loopBeats, sourceTotalBeats - clip.loopStartBeats);
+    if (auto* source = SourcePool::getInstance().getMutable(event->sourceId);
```
lucaromagnoli, line 278: **Low.** This writes an interpretation-derived estimate into the pooled Source, which is shared by every clip on the file, so one event's odd interpBpm can stamp a wrong duration onto unrelated clips; the old setSourceMetadata wrote to the clip's private copy. Probably fine given probe overwrites it, but worth making the shared write deliberate (e.g. only while unresolved).

magda/daw/audio/session/SessionRecorder.cpp
```diff
@@ -236,31 +236,28 @@ void SessionRecorder::finalizeRecording(const ActiveRecording& rec, double stopT
...
-        newClip->loopStartBeats = sessionClip->loopStartBeats;
-        newClip->loopLengthBeats = sessionClip->loopLengthBeats;
+    auto* newEvent = newClip->primaryEvent();
+    const auto* sessionEvent = sessionClip->primaryEvent();
+    if (newEvent != nullptr && sessionEvent != nullptr) {
+        // Read position, source region, stretch and interpretation come across
+        // verbatim; the pooled source and event id belong to the new clip.
+        const EventId keepId = newEvent->id;
+        const SourceId keepSourceId = newEvent->sourceId;
+        *newEvent = *sessionEvent;
```
lucaromagnoli, line 246: **Medium.** *newEvent = *sessionEvent copies the session event's startBeat and lengthBeats onto the arrangement clip's event, and the mirror-field writes below never go through setPlacementBeats, so syncSingleEventToClipBounds never re-runs. A 16-beat recording from a 4-beat slot leaves the arrangement event 4 beats long, and v2 serialization persists that geometry. Restore the container-spanning geometry after the copy, or exclude geometry from it.

magda/daw/core/ClipManager.cpp
```diff
@@ -3560,26 +3474,27 @@ std::vector<ClipId> ClipManager::pasteFromClipboardBeats(double pasteBeat, Track
...
-                            const double newLength = clipData.loopLengthBeats * 60.0 / bpm;
-                            ClipOperations::setTimelinePlacement(
-                                *newClip, newClip->getTimelineStart(bpm), newLength, bpm);
-                            newClip->loopLength = newClip->getTimelineLength(bpm);
-                        } else if (clipData.loopEnabled && clipData.loopLength > 0.0 &&
-                                   clipData.getTimelineLength(currentProjectTempoOrDefault()) >
-                                       clipData.sourceToTimeline(clipData.loopLength)) {
-                            const double bpm = currentProjectTempoOrDefault();
+                        const double bpm = currentProjectTempoOrDefault();
+                        const double loopBeats = clipData.loopLengthBeats;
```
lucaromagnoli: **Medium.** clipData.loopLengthBeats is the MIDI loop field and is always 0 for audio clips now, so a beat-mode audio session clip always takes the seconds fallback below, which scales by speedRatio (pinned to 1 under autoTempo) instead of the projectBpm over interpBpm stretch, landing on the wrong length whenever the two tempos differ. clipData.loopLengthInBeats() is the content-aware read; beat-mode events should go through the beat branch.

magda/daw/core/SourcePool.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+bool SourcePool::relink(SourceId id, const juce::String& newFilePath) {
```
lucaromagnoli: **Medium.** Two things. Relinking to a path already pooled under another id silently steals that key: idByPathKey_ ends up pointing at this id while the other Source keeps the same filePath, breaking the per-file dedup invariant (insert() guards exactly this case). And a resolved to resolved relink to a file at a different sample rate keeps the anchor samples unchanged, which moves every event position, while the 0-to-real return contract guarantees no caller will rescale for that case. Worth handling both while wiring up the rescale.

magda/daw/project/serialization/DawProjectArchive.cpp
```diff
@@ -157,18 +157,21 @@ bool DawProjectArchive::readFromFile(const juce::File& file, ProjectDocument& ou
...
             continue;
         const int entryIndex = zip.getIndexOfFileName(entryName, false);
         if (entryIndex < 0)
             continue;  // external/absolute reference, nothing to extract
 
         mediaDir.createDirectory();
-        if (zip.uncompressEntry(entryIndex, mediaDir, true).wasOk())
-            clip.audio().source.filePath = mediaDir.getChildFile(entryName).getFullPathName();
+        if (zip.uncompressEntry(entryIndex, mediaDir, true).wasOk()) {
+            event->sourceId = SourcePool::getInstance().acquire(
```
lucaromagnoli, line 177: **Medium.** The anchors were computed in clipFromXml against the archive-relative path, i.e. at the 48 kHz nominal rate, and swapping the sourceId here re-probes the extracted file at its real rate, so imported play-start and loop regions scale by 48000 over realRate for any non-48k embedded audio. Extracting and acquiring the real file before clipFromXml computes anchors avoids it, and stops the archive-entry source lingering until save GC.

magda/daw/project/MediaCollector.cpp
```diff
@@ -165,13 +165,17 @@ MediaCollector::Summary MediaCollector::apply(const Plan& plan) {
...
-                clipManager.forceNotifyClipPropertyChanged(clipId);
-            }
+            auto* clip = clipManager.getClip(clipId);
+            auto* event = primaryEventOf(clip);
+            if (event == nullptr)
+                continue;
+            const auto oldPath = event->sourceFilePath();
+            // Repoint the pooled source, not the clip: every clip sharing this
+            // file follows automatically.
+            SourcePool::getInstance().relink(event->sourceId, newPath);
```
lucaromagnoli: **Low.** scan() pushes clip.id once per matching event, but apply() always relinks the primary event's source, so for a future multi-event clip a match on event 2 would relink event 1's unrelated source. Latent while everything is single-event; carrying (ClipId, EventId) or the SourceId itself in the work list makes it correct now.

lucaromagnoli (commented):
Reviewed the full diff. The model design is right: a samples-authoritative source domain with beats and seconds as derived views kills the applyAudioClipBeats reconciliation exactly as intended, applyAutoTempoStretch collapsing to interpretation-only scaling is the proof that it works, and the three v1 layouts funnelling into one migrateLegacyAudioClip is clean. The mechanical conversion is faithful almost everywhere, and the test port is honest: the few changed expectations are individually justified.

The findings cluster into four themes, in descending order of concern:

1. **The missing-file story is half-built.** resolveFacts and relink return the rescale signal, but nothing consumes it, and the load-time re-probe actively corrupts anchors that were saved at the nominal rate. The PR text lists this path as untested on real projects; the handler was never written at all. Inline comments on ProjectSerializer, DawProjectArchive and SourcePool.
2. **SourcePool lifecycle.** GC-on-save runs on autosave and breaks the pool's own undo contract; a failed load clears the live pool mid-staging; loadProjectAsync now mutates a message-thread-only singleton from a background thread. Three inline comments on ProjectSerializer.
3. **Reads of the now-MIDI-only clip loop fields on audio paths, and audioEventRef inside MIDI branches.** The two inspector loop handlers, the session paste loop reset, the drum grid and the piano roll all hit the exact failure mode the loopLengthInBeats() doc comment warns about.
4. **Individual regressions:** the clip context menu take switch is a silent no-op, the reversed-clip offset dropped its bpm argument, the session recorder copies session geometry onto the arrangement event, and MIDI clips with a zero loop-length sentinel get a degenerate loop range.

Two PR-description corrections: the MIDI loop did not move into MidiClipModel, it stayed on the container, which is why it is per-instance (the fix is real, the description is stale). And "v2 files do not open in older builds" is not true, see the inline comment on ProjectSerializer.hpp.

Small things not worth their own thread: getAutoTempoBeatRange returns {0,0} when interpBpm is unknown where the old code fell back to project beats (transient until detection lands); splitClipAtBeat's loop truncation silently no-ops through setLoopLengthBeats when interpBpm is unknown; paste now copies event fades, gain and beat-detection settings wholesale where the old paste did not (probably an improvement, but a behaviour change to bless knowingly); deserializeAudioEvent's missing-id guard compares against INVALID_EVENT_ID but a missing property reads as 0, so it never fires; resetLoopedClipLength is uncalled on both branches and its audio path now reads the MIDI field, so fix it or delete it.

One test gap worth closing together with the rescale fix: kUnresolvedSourceSampleRate, the test helper rate and every migration case are all 48 kHz, so no test can tell the probed rate from the nominal fallback. A single migration case at 44.1 kHz asserting anchorSeconds() after resolve covers the whole first cluster.

magda/daw/core/ClipInfo.hpp
```diff
@@ -357,146 +639,77 @@ struct ClipInfo {
...
-    std::vector<WarpMarker> warpMarkers;
+    AudioEvent* primaryEvent() {
+        return isAudio() ? audio().primaryEvent() : nullptr;
+    }
+    const AudioEvent* primaryEvent() const {
+        return isAudio() ? audio().primaryEvent() : nullptr;
+    }
+
+    /// Every audio event, empty for MIDI clips.
+    std::vector<AudioEvent>& events() {
```
lucaromagnoli, line 735: **Low.** The comment says "empty for MIDI clips" but audio() is std::get on the variant, so this throws bad_variant_access for MIDI. Either return an empty list (mirroring primaryEvent's isAudio() guard) or fix the comment so nobody trusts it.

magda/daw/ui/panels/content/DrumGridClipContent.cpp
```diff
@@ -2840,10 +2840,12 @@ void DrumGridClipContent::updateGridSize() {
...
-        double loopOffsetBeats = clip->loopStart * beatsPerSecond;
-        // MIDI clips use loopLengthBeats directly; audio clips derive from loopLength (seconds)
-        double sourceLengthBeats =
-            clip->loopLengthBeats > 0.0 ? clip->loopLengthBeats : clip->loopLength * beatsPerSecond;
+        // The loop is read through the content-aware accessors: a MIDI clip's
+        // loop lives in clip beats, an audio event's in its source domain.
+        juce::ignoreUnused(beatsPerSecond);
+        const double loopOffsetBeats =
+            clip->isMidi() ? clip->loopStartBeats : magda::audioEventRef(*clip).loopStartBeats();
+        const double sourceLengthBeats = clip->loopLengthBeats;
```
lucaromagnoli: **Low.** For a non-MIDI clip this reads the MIDI-only container field (always 0 now) while the loopOffsetBeats ternary right below anticipates the audio case; clip->loopLengthInBeats() is the content-aware accessor.

magda/daw/ui/panels/content/PianoRollContent.cpp
```diff
@@ -1344,10 +1344,12 @@ void PianoRollContent::updateGridSize() {
...
-        double loopOffsetBeats = clip->loopStart * beatsPerSecond;
-        // MIDI clips use loopLengthBeats directly; audio clips derive from loopLength (seconds)
-        double sourceLengthBeats =
-            clip->loopLengthBeats > 0.0 ? clip->loopLengthBeats : clip->loopLength * beatsPerSecond;
+        // The loop is read through the content-aware accessors: a MIDI clip's
+        // loop lives in clip beats, an audio event's in its source domain.
+        juce::ignoreUnused(beatsPerSecond);
+        const double loopOffsetBeats =
+            clip->isMidi() ? clip->loopStartBeats : magda::audioEventRef(*clip).loopStartBeats();
+        const double sourceLengthBeats = clip->loopLengthBeats;
```
lucaromagnoli: **Low.** Same as the drum grid: clip->loopLengthBeats is 0 for audio clips in the new model, and the comment above claims the loop is read through content-aware accessors, but this line is not.

magda/daw/project/serialization/DawProjectXmlAdapter.cpp
```diff
@@ -744,26 +744,35 @@ juce::XmlElement& addClipElement(juce::XmlElement& parent, const ClipInfo& clip,
...
+    // one-source-per-clip shape, and Phase A of #1901 never builds more.
+    const auto* event = clip.primaryEvent();
+    const bool beatContent = clip.isMidi() || (event != nullptr && event->autoTempo);
     clipElement->setAttribute("contentTimeUnit", beatContent ? "beats" : "seconds");
 
     // Playback offset + loop region, in the clip's content time unit.
-    if (beatContent) {
-        clipElement->setAttribute("playStart", clip.offsetBeats);
+    if (clip.isMidi()) {
+        const auto& midi = clip.midi();
```
lucaromagnoli: **Low.** `midi` is unused in this branch now; it will trip -Wunused-variable.

magda/daw/project/serialization/ClipSerializer.cpp
```diff
@@ -32,6 +32,293 @@ double getValidProjectTempo(double projectTempo) {
...
+/// Turn a v1 audio clip into a clip holding exactly one event that spans it.
+///
+/// The source-domain values were seconds (or beats, for autoTempo clips) and
+/// become samples at the source's own rate. When the file cannot be opened the
+/// pool reports a nominal rate and rescales the anchors on the first successful
+/// open, so an offline project still round-trips.
+void migrateLegacyAudioClip(const LegacyAudioSource& v1, const LegacyAudioFields& legacy,
+                            ClipInfo& outClip) {
+    AudioEvent event;
+    event.sourceId = SourcePool::getInstance().acquire(v1.filePath);
```
lucaromagnoli: Staging is still not isolated for v1 files. `loadAndStage` calls this migration through `deserializeClipsToStaging` on its background thread, and this `acquire` plus the `getMutable` below mutate the global message-thread-only pool. A failed v1 load therefore still changes the open project pool and races its UI reads. `installStagedSources` carrying these live entries forward confirms the migration bypasses the new staging collection. Please allocate legacy sources and bind their IDs inside the staging context, without touching the singleton until commit.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -1072,20 +1128,22 @@ void ClipSynchronizer::launchSessionClip(ClipId clipId, bool forceImmediate) {
...
+                double loopDurationBeats =
+                    (srcLength / audioEventRef(*clip).speedRatio) * (bpm / 60.0);
                 launchHandle->setLooping(te::BeatDuration::fromBeats(loopDurationBeats));
-            } else {
-                // MIDI
-                double loopLengthBeats = srcLength * (bpm / 60.0);
-                launchHandle->setLooping(te::BeatDuration::fromBeats(loopLengthBeats));
+            } else if (clip->isMidi()) {
+                // Already clip beats.
+                if (clip->loopLengthBeats > 0.0)
```
lucaromagnoli: The zero-length sentinel fallback is still missing at the launch boundary. Creation/property sync now use the container length when `loopLengthBeats == 0`, but both the creation-time LaunchHandle block above and this branch skip `setLooping` in that case. A legacy session MIDI clip therefore gets a valid TE loop range but its launch handle remains one-shot and stops after one pass. Please use the same whole-clip fallback in both LaunchHandle paths.

magda/daw/core/ClipInfo.hpp
```diff
@@ -357,146 +658,80 @@ struct ClipInfo {
...
-    int autoPitchMode = 0;     // 0=pitchTrack, 1=chordTrackMono, 2=chordTrackPoly
-    float pitchChange = 0.0f;  // -48 to +48 semitones
-    int transpose = 0;         // -24 to +24 semitones (only when !autoPitch)
+    /// Loop length in beats for whichever content the clip holds.
+    ///
+    /// A MIDI clip keeps it in the field above. An audio clip's loop is a
+    /// source region in samples, so its beat length is a view derived through
+    /// the event's interpretation. Reading only one of the two silently
+    /// returns zero for the other, which reads as "not looping".
+    double loopLengthInBeats() const {
```
lucaromagnoli: **Medium-high.** This conflates two beat domains: for MIDI it returns timeline beats, for audio it returns the SOURCE-beat view (loopLengthSeconds * interpBpm / 60). The two coincide only under beat mode. For a looped, beat-mode-OFF audio clip the timeline loop span is loopLengthSeconds / speedRatio, and interpBpm is now seeded on every event at creation via seedInterpretationFromSource, so any analyzed library loop has it set even when the user never enables BEAT. On the base branch non-autoTempo audio left clip-level loopLengthBeats at 0, so every consumer fell through to the correct seconds-derived path. Consumers that regress: followActionBaseLengthBeats (inline at ClipSynchronizer), getTimelineLoopLength (slot progress overlay), and the fixed call sites in the drum grid, piano roll and session paste reset inherited the same skew. Suggest gating the event-derived beat view on event->autoTempo (or warp) and deriving from sourceToTimeline(loopLengthSeconds()) otherwise, so the round-1 fixes land on a correct accessor.

magda/daw/audio/session/ClipSynchronizer.cpp
```diff
@@ -54,23 +54,30 @@ te::FollowAction toTracktionFollowAction(FollowAction action) {
...
-            auto [_, loopLengthBeats] = ClipOperations::getAutoTempoBeatRange(clip, bpm);
+        if (clip.isAudio() && audioEventRef(clip).autoTempo) {
+            auto [_, loopLengthBeats] = ClipOperations::getAutoTempoBeatRange(audioEventRef(clip));
             if (loopLengthBeats > 0.0)
                 return loopLengthBeats;
         }
 
-        if (clip.loopLengthBeats > 0.0)
-            return clip.loopLengthBeats;
+        if (clip.loopLengthInBeats() > 0.0)
```
lucaromagnoli: **Medium-high, the audible consumer of the loopLengthInBeats issue.** For a non-beat-mode looped audio clip this branch now returns source beats and treats them as timeline beats: a 174 BPM loop in a 120 BPM project schedules its follow action about 45% late. On base, clip-level loopLengthBeats was 0 for non-beat-mode audio, so this fell through to the (sourceLength / speed) * bpm / 60 path below, which is the correct value. Fixing the accessor fixes this site.

magda/daw/core/ClipManager.cpp
```diff
@@ -3422,13 +3398,13 @@ std::vector<ClipId> ClipManager::pasteFromClipboardBeats(double pasteBeat, Track
 
         // Create new clip based on type, using targetView instead of clipData.view
         ClipId newClipId = INVALID_CLIP_ID;
-        if (clipData.isAudio()) {
-            if (clipData.audio().source.filePath.isNotEmpty()) {
+        if (const auto* pastedEvent = clipData.primaryEvent()) {
+            if (pastedEvent->sourceFilePath().isNotEmpty()) {
```
lucaromagnoli: **High.** The clipboard deliberately survives project switches (clearAllClips scrubs only linkGroupId), but a clipboard event now carries only a sourceId, and loading another project clears and renumbers the pool with small sequential ids. So a paste after switching projects resolves sourceFilePath() against the new project's pool: an id collision quietly creates a clip from the wrong project's file, and a dangling id makes the paste silently vanish. On base the clipboard ClipInfo carried filePath inline, so cross-project paste was correct. The new SourcePool.hpp lifetime comment also cites the clipboard as protected, which only holds within one project. Simplest fix is to resolve and store the path (or a path-carrying copy of the Source) at clipboard-store time.

magda/daw/project/serialization/ClipSerializer.cpp
lucaromagnoli: **Medium.** This acquire still mutates the live pool, and for a v1 project it runs inside deserializeClipsToStaging on loadProjectAsync's background thread, so the new staging invariant ("staging must not touch the live pool") holds only for pure-v2 files. The v1 path keeps both the data race with paint-time sourceRateOf reads and a share of the old failed-load pollution (acquired sources linger in the live pool if staging fails later). installStagedSources already carries these entries across, so staging-local acquisition (a staged pool or a path-keyed side table resolved at install time) would close the gap without changing the shape.

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -707,6 +728,105 @@ juce::var ProjectSerializer::serializeTracks() {
...
+            if (event.sourceId == INVALID_SOURCE_ID || !known.insert(event.sourceId).second)
+                continue;
+            if (const auto* live = pool.get(event.sourceId))
+                carried.push_back(*live);
+        }
+    }
+
+    pool.clear();
+    for (const auto& source : sources)
+        pool.insert(source);
```
lucaromagnoli: **Medium.** Both insert loops discard the return, but SourcePool::insert dedups by canonical path and returns the WINNING id exactly so the caller can repoint; MediaCollector and migrateMediaFiles both honor that contract via repointEventsToSource, this site does not. The collision is reachable by this function's own supported scenario: a v2 source-table entry for path P plus a v1-migrated clip whose acquire(P) pooled P under a different live id; the carried entry loses the insert, and the migrated clip's event dangles, silently losing its audio. Also reachable via two saved sources whose paths differ only by case on the case-folding platforms. Same fix as the siblings: capture the returned owner and repoint staged clip events when it differs.

magda/daw/core/ClipOperations.hpp
```diff
@@ -206,44 +205,31 @@ class ClipOperations {
...
-                        clip.loopLength > 0.0 ? clip.loopLength : clipLength * toSource;
-                    if (sourceLength > 0.0) {
-                        double phaseDelta = actualDelta * toSource;
-                        double relOffset = clip.offset - clip.loopStart;
-                        clip.offset =
-                            clip.loopStart + wrapPhase(relOffset + phaseDelta, sourceLength);
-                    }
+                const double sourceLength = event->loopLengthSamples > 0
+                                                ? event->loopLengthSeconds()
+                                                : clipLength * event->speedRatio;
```
lucaromagnoli: **Medium-low.** For a beat-mode clip with the loop-region-0 sentinel (the session default when the file's BPM was already pooled at creation, and carried onto arrangement clips by session-to-arrangement paste), this wraps the anchor using clipLength, which is TIMELINE seconds at project tempo, as a SOURCE-seconds period, while sourceDelta above was correctly converted through interpBpm. Base had a sentinel guard in the autoTempo branch (else if loopLengthBeats > 0) and left the phase untouched. Suggest restoring the sentinel no-op for beat mode, or deriving the period through the beat domain.

magda/daw/project/serialization/ClipSerializer.cpp
```diff
@@ -32,6 +32,293 @@ double getValidProjectTempo(double projectTempo) {
...
+    event.autoTempo = legacy.autoTempo;
+    event.speedRatio = v1.speedRatio > 0.0 ? v1.speedRatio : 1.0;
+    event.timeStretchMode = v1.timeStretchMode;
+    event.warpEnabled = v1.warpEnabled;
+    event.warpMarkers = v1.warpMarkers;
+    event.analogPitch = v1.analogPitch;
+
+    // v1 kept beats authoritative under autoTempo and seconds otherwise. Both
+    // resolve to the same source seconds, which is what becomes the anchor.
+    const bool beatsAuthoritative = legacy.autoTempo && v1.interpBpm > 0.0;
```
lucaromagnoli: **Low-medium.** A v1 clip saved with autoTempo on but interpBpm still 0 (detection pending) kept its beat fields authoritative and its seconds mirrors deliberately stale: base refreshDerivedSeconds skipped the source-domain sync in that state while playback consumed offsetBeats/loopLengthBeats directly via project tempo. beatsAuthoritative false here anchors from those stale seconds and discards the beat values entirely, so such a project migrates to a different region than it played. Worth folding the beat fields in when autoTempo is set even without interpBpm, converting through the project tempo the way base playback did.

magda/daw/project/serialization/ClipSerializer.cpp
```diff
@@ -32,6 +32,293 @@ double getValidProjectTempo(double projectTempo) {
...
+    AudioEvent event;
+    event.sourceId = SourcePool::getInstance().acquire(v1.filePath);
+
+    // Duration recorded in the project is a fallback for a file we cannot read.
+    if (auto* source = SourcePool::getInstance().getMutable(event.sourceId);
+        source != nullptr && source->durationSeconds <= 0.0) {
+        double duration = v1.durationSeconds;
+        if (duration <= 0.0 && v1.interpTotalBeats > 0.0 && v1.interpBpm > 0.0)
+            duration = v1.interpTotalBeats * 60.0 / v1.interpBpm;
+        source->durationSeconds = juce::jmax(0.0, duration);
```
lucaromagnoli: **Low.** When the file is missing, the first migrated clip's recorded duration becomes the pooled duration for every clip of that file (first non-zero wins), where v1 kept durationSeconds per clip. A stale short value from clip A then clamps clip B's anchors and loop region via sanitizeAudioClip and caps its right-extension. Minor, but worth preferring the longest recorded duration across the migrating clips, or leaving it 0 and letting resolve fill it.

magda/daw/core/ClipManager.cpp
```diff
@@ -637,7 +710,10 @@ void ClipManager::setAudioClipCurrentTake(ClipId clipId, int takeIndex) {
     a.compActive = false;
     a.comp.clear();
     a.currentTakeIndex = takeIndex;
-    a.source.filePath = a.takes[static_cast<size_t>(takeIndex)].filePath;
+    if (auto* event = a.primaryEvent()) {
+        event->sourceId =
```
lucaromagnoli, line 745: **Low.** This swaps the event's sourceId without rescaling the sample-domain positions, unlike repointEventsToSource which was added for exactly this hazard. Takes normally share the engine rate so it is latent, but an imported or comp-rendered take at another rate would silently re-denominate the anchors. Cheap to route through the same rescaleSourcePositions(oldRate, newRate) discipline here and in deleteClipTake's re-front.

magda/daw/core/ClipManager.cpp
```diff
@@ -293,63 +364,65 @@ ClipId ClipManager::createAudioClipBeats(TrackId trackId, double startBeats, dou
...
+                    // A beat-mode clip's natural length is its musical length
+                    // (the saved beat count), not the raw audio-file duration
+                    // the caller derived the placement from. At a project tempo
+                    // different from the sample's, file-duration-in-beats is not
+                    // the real beat count, so snap both the timeline placement
+                    // and the loop region to the beat count and let autoTempo
+                    // stretch the audio to fit.
+                    const double beats = savedEvent->interpTotalBeats;
+                    savedClip.setPlacementBeats(startBeats, beats);
+                    if (savedEvent->loopLengthSamples <= 0 && savedEvent->interpBpm > 0.0)
```
lucaromagnoli: **Low.** For an Arrangement drop this guard is now always false: line ~355 already seeded the loop region to the file-duration extent for every clip, so a beat-mode library file no longer snaps its region to the saved beat count (base guarded on loopLengthBeats <= 0, which was still 0 at this point, so base did snap). Audible only when the saved totalBeats was user-locked to something other than duration * bpm / 60. Session drops are unaffected since the region resets to 0 before this branch.

magda/daw/core/ClipInfo.hpp
```diff
@@ -555,39 +790,98 @@ struct ClipInfo {
...
+                copySharedEventFieldsFrom(events()[i], src.events()[i]);
+            auto& dstAudio = audio();
+            const auto& srcAudio = src.audio();
+            dstAudio.nextEventId = srcAudio.nextEventId;
+            dstAudio.takes = srcAudio.takes;
+            dstAudio.currentTakeIndex = srcAudio.currentTakeIndex;
+            dstAudio.comp = srcAudio.comp;
+            dstAudio.compActive = srcAudio.compActive;
+        } else {
+            content = src.content;
```
lucaromagnoli, line 962: **Low, latent until multi-event.** Two notes for when event counts can diverge: this wholesale branch clobbers every per-instance event field (anchor, region, speedRatio, fades, channels, gainDB, geometry) and replaces dst event ids with src's; and in the match branch above, mirroring nextEventId while ids stay per-instance can set the allocator at or below an existing local id if siblings' event histories ever diverge, minting a duplicate id on the next addEvent. Both are unreachable in Phase A (verified: every live flow keeps link-group members at exactly one event). An assert or prefix-match here would be cheap insurance for the slicing milestone.

magda/daw/project/serialization/ClipSerializer.cpp
```diff
@@ -32,6 +32,293 @@ double getValidProjectTempo(double projectTempo) {
...
+    if (!event.keyRoot.empty())
+        obj->setProperty("keyRoot", juce::String(event.keyRoot));
+    if (!event.keyScale.empty())
+        obj->setProperty("keyScale", juce::String(event.keyScale));
+
+    obj->setProperty("autoTempo", event.autoTempo);
+    obj->setProperty("speedRatio", event.speedRatio);
+    if (event.timeStretchMode != 0)
+        obj->setProperty("timeStretchMode", event.timeStretchMode);
+    if (event.warpEnabled) {
```
lucaromagnoli: **Low, pre-existing but now in tension with itself.** Markers are only written when warpEnabled, but warp-off-with-markers is a state the code cultivates: setClipWarpEnabled(false) keeps them and paste explicitly preserves them, and the ghost shared-field set treats markers as independent of the toggle. So re-enabling warp in-session restores the map, re-enabling after a save/reload loses it. v1 had the same write condition, so not a regression, just worth deciding: either persist markers unconditionally or stop preserving them when warp goes off.

magda/daw/project/serialization/DawProjectXmlAdapter.cpp
```diff
@@ -823,46 +843,53 @@ ClipInfo clipFromXml(const juce::XmlElement& clipElement, TrackId trackId, ClipI
...
             maxSeconds = juce::jmax(maxSeconds, w->getDoubleAttribute("contentTime", 0.0));
         }
-        clip.audio().interpretation.totalBeats = maxBeats;
+        event.interpTotalBeats = maxBeats;
         if (maxSeconds > 0.0)
-            clip.audio().interpretation.bpm = maxBeats * 60.0 / maxSeconds;
+            event.interpBpm = maxBeats * 60.0 / maxSeconds;
 
-        clip.offsetBeats = clipElement.getDoubleAttribute("playStart", 0.0);
+        event.setAnchorBeats(clipElement.getDoubleAttribute("playStart", 0.0));
```
lucaromagnoli, line 876: **Low.** When the Warps element yields no usable point, interpBpm stays 0 and setAnchorBeats / setLoopStartBeats / setLoopLengthBeats are all silent no-ops while loopEnabled is still set, so the clip loops the whole source from anchor 0 instead of the file's stated region. The old importer stored the beat fields verbatim with no BPM dependency and they took effect once the interpretation arrived. Worth stashing the beat values through a project-tempo conversion when interpBpm is unknown, or skipping loopEnabled when the region could not be applied.

magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp
```diff
@@ -39,23 +39,17 @@ double timelineEndBeats(const ClipInfo& clip, double bpm) {
...
-        return clip.loopStartBeats;
-
-    if (clip.loopStart > 0.0 && bpm > 0.0)
-        return clip.loopStart * bpm / 60.0;
-
-    return 0.0;
+// The piano roll only ever edits MIDI, so both of these read the clip's own
+// loop in clip beats. The audio event's loop is a region of a source file and
+// has nothing to say about a note grid.
+double effectiveLoopStartBeats(const ClipInfo& clip, double) {
```
lucaromagnoli: **Low.** These helpers read the clip-level loop fields, and the comment claims the piano roll only ever edits MIDI, but the sibling editors on this branch explicitly route audio clips through the event accessors (PianoRollContent 1351, DrumGridClipContent 2847, MidiEditorContent 97), so audio clips do reach this component; base had a seconds fallback here that served them. For an audio clip in relative mode the loop resolves as 0 / clip-length instead of the event region. Either the comment is right and the sibling audio branches are dead (then delete those), or these helpers need the same content-aware treatment as the round-1 sites.

lucaromagnoli (commented):
Second pass, against the head with the four follow-up commits.

**Fix verification: every round-1 finding is properly fixed.** I checked each site rather than trusting the commit messages. The rescale now fires from inside SourcePool via a rate-change handler, so no resolveFacts/relink caller can forget it; relink returns the owning id and refuses to steal path keys, and MediaCollector/ProjectManager repoint events on collision; GC-at-save is gone in favour of filtering the emitted snapshot; load stages sources off-thread and installs them in the message-thread commit, with resolveStagedSources ordered after the clips exist so the rescale can reach them; the recorder re-spans through setPlacementBeats; the take switch collapses to setAudioClipCurrentTake and gains undo; both inspector handlers, the drum grid, the piano roll content, the paste reset, the MIDI zero-sentinel loop paths, the reversed-clip bpm, the shared-Source duration write, and the doc corrections are all fixed as suggested. Good, thorough follow-up.

**New findings.** The second sweep ran different lenses (whole-branch pattern sweep, serialization field symmetry, ghost/undo invariants, and an independent re-read of the core files). The biggest item is one the fix commits made more load-bearing: `loopLengthInBeats()` conflates two beat domains (timeline beats for MIDI, SOURCE beats for audio), which is only correct under beat mode, and interpBpm is now seeded on every analyzed file at creation, so non-beat-mode looped audio hits it. Details inline; the follow-action consumer is the audible one. Second is the cross-project clipboard: the clipboard deliberately survives project switches, but its events now carry only a sourceId into a pool that load clears and renumbers, so pasting after switching projects resolves the wrong project's file or silently drops the clip (old code carried the path inline). The rest are inline with severities; the latent multi-event notes (paste/range-copy/session-to-arrangement collapse clips to the primary event; the ghost wholesale-copy branch clobbers per-instance state if event counts ever diverge) are Phase-A-safe and flagged for when slicing lands.

**Verified clean in this pass:** ghost shared/per-instance split matches the old model field-for-field with copy and compare in exact lockstep; undo symmetry holds for every command (multi-field setters are wrapped in whole-ClipInfo snapshots); all 34 AudioEvent fields round-trip v2 including the conditional writes; the three v1 layouts map field-for-field; the autoTempo/warp triple is consistent with the base at every call site; no quantization drift on drag paths (previews recompute from drag-start snapshots); no further instances of the by-value-assignment or discarded-return families beyond what is inline here.

**Still open from round 1, minor:** the deserializeAudioEvent missing-id guard still compares against INVALID_EVENT_ID while a missing property reads as 0; resetLoopedClipLength is still dead code with the stale audio branch; getAutoTempoBeatRange still returns {0,0} while interpBpm is unknown; splitClipAtBeat's truncation still no-ops via the silent beat setters when interpBpm is unknown; paste still copies event fades/gain wholesale (fine if that is now the blessed behaviour).

lucaromagnoli (commented):
Third pass, on the round-2 fix commit. Every round-2 finding is properly fixed, and this time the sweep found nothing new. Site-by-site:

- **loopLengthInBeats / loopStartInBeats**: now take projectBpm and gate on autoTempo, with the source-region-to-timeline conversion for the off case. Every consumer was moved over: follow actions, both grid panels, the piano-roll grid helpers (fixed content-aware, which also resolves the comment-vs-siblings contradiction), both MidiEditorContent helpers (now consistent with each other), the inspector display, and the paste reset. The waveform hash switching to the raw fields instead of the bpm-dependent accessor is the right call for a hash. I also chased the inspector display domains: the audio path returns early through ClipDisplayInfo, so the fall-through mixing I went looking for cannot happen; the fall-through is MIDI-only.
- **Cross-project clipboard**: paths are stashed at all three clipboard population sites, the seconds shim delegates to the beats path so it is covered too, and paste resolves by stashed path with the pool as fallback. Verified there is no fourth population path.
- **v1 migration on the staging thread**: staging now allocates provisional negative ids into StagedProjectData::legacySources and never touches the live pool; install acquires for real on the message thread. This also closes the failed-load pollution the v1 path still had.
- **installStagedSources**: the remap table honors insert()'s winner and rescales across the loser/owner rate difference, which covers both the v1-plus-v2-mixed collision and the case-fold collision.
- The sentinel wrap is a no-op again for beat mode, the migration folds beat-authoritative v1 values through the project tempo, stageLegacySource keeps the longest duration, both take-swap sites rescale, the arrangement drop snaps unconditionally, the wholesale ghost branch got its assert plus the nextEventId jmax, warp markers persist independently of the toggle, and the dawproject import falls back to DEFAULT_BPM so the region survives.

Two observations, neither blocking: installStagedSources now probes files on the message thread during commit (fine, the commit phase already does IO, just worth knowing it is there); and persisting warp markers unconditionally changes the v2 write format, which is fine while v2 is unreleased.

Known-opens carried from earlier rounds, unchanged and all minor: the deserializeAudioEvent missing-id guard, dead resetLoopedClipLength, the silent split truncation and engine {0,0} loop range while interpBpm is unknown (the migration fix preserves the DATA for that state now; playback still waits for detection), and paste copying event fades wholesale. None of these block merging.

From my side this is done: three passes, all findings fixed and verified at the site level. Good PR.

magda/daw/core/ClipInfo.hpp
```diff
@@ -357,146 +658,107 @@ struct ClipInfo {
...
-    // not the raw field, so the inspector and the audio editor agree — e.g.
-    // after a session drop auto-enables beat mode.
-    int getEffectiveTimeStretchMode() const {
-        if (timeStretchMode == 0 && !isAnalogPitchActive() &&
-            (autoTempo || warpEnabled || std::abs(speedRatio - 1.0) > 0.001 ||
-             std::abs(pitchChange) > 0.001f)) {
-            return time_stretch_mode::kSignalsmith;
-        }
-        return timeStretchMode;
+        if (event->autoTempo)
```
lucaromagnoli: **Medium-high.** This still misses warped-but-not-auto-tempo events. warpEnabled is independent (setClipWarpEnabled does not enable autoTempo), and ClipSynchronizer deliberately treats autoTempo || warpEnabled as source-beat processing. Falling through here applies the linear speedRatio conversion and ignores warpMarkers, so a nonlinear warped loop reports the wrong timeline start/length; follow-action timing and the grid/paste consumers of these accessors can drift. Please map loop boundaries through the warp-time mapping for warped events, and add a warpEnabled=true/autoTempo=false test.

magda/daw/project/serialization/DawProjectXmlAdapter.cpp
```diff
@@ -823,46 +843,65 @@ ClipInfo clipFromXml(const juce::XmlElement& clipElement, TrackId trackId, ClipI
...
+            event.interpBpm = maxBeats * 60.0 / maxSeconds;
+
+        // The beat setters below all convert through interpBpm, so without one
+        // they are silent no-ops: the anchor and region would stay at zero
+        // while loopEnabled was set, and the clip would loop the whole source
+        // instead of the region the file states. The importer knows no better
+        // tempo than the project's, which is what the old beat-field import
+        // effectively played at.
+        if (!isValidBpm(event.interpBpm) && maxBeats > 0.0)
+            event.interpBpm = DEFAULT_BPM;
```
lucaromagnoli: **Low-medium.** The comment says this falls back to the project tempo, but the implementation hard-codes DEFAULT_BPM. fromProjectXml has already parsed document.info.tempo before calling clipFromXml, so on a non-120 BPM project with no derivable marker ratio the beat-domain playStart and loop values are converted into source positions at 120 BPM, moving the audible region. Please pass the imported project tempo into clipFromXml and use it here, falling back to DEFAULT_BPM only when the project tempo itself is invalid.

PR #1998, opened 2026-08-03
Fix/linux file drag and drumgrid gain

### LINKED ISSUE #1995: [Bug]: Drum Grid & Sampler : Samples cannot be drag and dropped from internal browser
### What happened?

Dragging a sample from MAGDA's internal browser doesn't work.

### Steps to reproduce

[1] Load a Drum Grid instrument onto a track
[2] Drag a sample from the browser and try to drop it onto one of the drum pads of the Drum Grid
[3] here we see that the pad doesn't get focus for the drop action and the overall operation fails

### Expected behavior

Pad should highlight once the audio file is hovered with the mouse over it and the drop should work.

* Note that i can actually drag audio files from my Desktop's browser onto the pads and also instruments from the browser too and all works great

### MAGDA version

0.17.1

### Operating system

Linux

lucaromagnoli:


magda/daw/audio/plugins/DrumGridPlugin.cpp
```diff
@@ -358,10 +358,15 @@ void DrumGridPlugin::processChain(const AudioChainEntry& entry,
...
-        levelLinear * std::sin((panValue + 1.0f) * juce::MathConstants<float>::halfPi * 0.5f);
+    // Linear pan law, matching Tracktion's default (PanLawLinear) so a pad at
+    // centre is unity — the same as a device on a track. A constant-power law
+    // was used here previously, which cost every pad 3 dB (cos(pi/4) = 0.707)
+    // at the default centre pan and made Drum Grid devices quieter than the
+    // identical device loaded standalone.
+    panValue = juce::jlimit(-1.0f, 1.0f, panValue);
+    const float panGain = panValue * levelLinear;
+    float leftGain = levelLinear - panGain;
+    float rightGain = levelLinear + panGain;
```
lucaromagnoli: The existing Drum Grid pan-law test still independently implements and asserts the old equal-power formula (0.707 at centre, 1.0 at hard pan), so it will keep passing while contradicting this new Tracktion-linear behavior (1.0 at centre, 2.0 on the favored channel at hard pan). Please update the test—ideally by extracting/calling the production gain calculation or rendering through the plugin—so this change has real regression coverage.

Issue #1889, opened 2026-07-27
Engine core: render plan IR, compiler, executor, PDC


lucaromagnoli:
Part of #1882.

The core of the native engine: plan IR, compiler, executor, and the differ.

- Plan IR: flat, dependency-ordered op list; ops carry model-ID identity and a liveness-domain tag (reserved for the anticipative executor).
- Compiler from the MAGDA model (tracks, chains, sends, master) with latency compensation computed at compile time.
- Single-threaded RT executor; disk prefetch on a dedicated thread with per-reader FIFOs.
- Plan differ with state carry-over keyed by model IDs; crossfade only changed nodes. This is the component that owns the rebuild-click problem.
- Transport and tempo (a native TempoMap implementation, loop, count-in, click), offline plan rendering, native meter/tap ops (master, track, device).
- Plans are dumpable; golden tests assert compiled plans for model fixtures (for example: adding a send inserts these ops, PDC is N samples here).
- Milestone: a static arrangement (audio clips plus internal FX chains) renders null against TE.

lucaromagnoli:
Design constraints locked in from the architecture discussion, so they survive into implementation:

**Immutability**
- A published plan is never mutated. No splicing ops into a structure the audio thread can see. Every structural change builds a new plan and swaps it atomically. This is the property the whole architecture rests on.

**Incremental compilation is caching, not splicing**
- Front-end: per-subtree op emission may be memoized, keyed by a subtree fingerprint (ChainFingerprint is the natural key). Cache reuse only, never partial mutation.
- Back-end: the global passes (PDC, buffer assignment, dead-branch elimination) always rerun over the full assembled op list. They are linear and cheap; skipping them for locality is how incremental graph engines get wrong-latency bugs.
- The subtree cache is an optimization to add when profiling asks (continuous device drag is the one hot case, and it also gets a debounced fast path). Not a day-one requirement.

**The plan encodes signal topology only**
- Tracks, devices, routing, sends. Nothing else.
- Clip positions, automation curves, and modulation are data that ops read through swapped snapshots. Editing them never triggers a compile.
- Sample-accurate events (notes, parameter ramps) flow as event streams into ops, never as recompiles.
- Consequence: structural recompiles happen only at human speed (add/remove/rewire devices and routing), so a full compile just has to stay in the microsecond-to-sub-millisecond range at realistic session scale. Worth a perf assertion in the test suite.

**Differ**
[...]

Issue #1890, opened 2026-07-27
Engine: arranger clip playback, streaming, stretch, warp


lucaromagnoli:
Part of #1882. Depends on the engine core.

Arranger audio clip playback at parity with what MAGDA uses from TE today.

- Disk streaming and cache with per-reader prefetch (arranger and launcher readers must not share a starvable cache).
- Time-stretch integration: Signalsmith (already vendored) and SoundTouch modes; the persisted mode values stay pinned.
- Auto-tempo and loop-info semantics (source BPM / beat count interpretation), warp markers with transient detection, reverse playback, fades, takes.
- Verified by null-diff renders across a corpus of stretch/warp/auto-tempo projects.

lucaromagnoli:
Requirement from the clip layering work (#2003, branch `feat/2003-clip-layering`).

Arrangement clips now overlap instead of erasing each other, and nothing on a lane is ever cut: a clip keeps its placement and its content whatever lands on it, and what it actually plays is derived from the lane on every read.

So the clip snapshot a `ClipAudio` / `ClipMidi` op reads has to carry **resolved playback**, not raw placements:

- the audible span per clip (start and length, beats),
- interior silenced ranges, for audio as well as MIDI (a clip dropped inside another leaves a hole in the middle of it),
- resolved fade in and fade out per clip.

All three come from the model layer at snapshot compile time: `computeAudibleSpans` in `magda/daw/core/ClipOcclusion.hpp` and `ClipManager::effectiveFadesIn`. Both are pure and beat domain, no engine, no UI.

The op then never sees an overlap: it plays a span with fades. An edit recompiles the snapshot the same way a structural change recompiles the plan, so there is no sync protocol between two representations and no state that can go stale.

The rules those two functions implement, for reference:

- `overlapPlaysBoth` on either clip of a pair: both sound across the overlap. Otherwise only the clip on top sounds there.
- `autoCrossfade` shapes an overlap that already plays both, and cannot keep a clip alive on its own. The clip that starts first fades out across the overlap, the one that starts later fades in, one curve each. With play-through off there is no fade at all.

Tracktion is deliberately not taught any of this, so overlaps just sum in dev builds until this issue lands. The rules are pinned by `tests/test_clip_occlusion.cpp` and `tests/test_clip_crossfades.cpp`.

Issue #1834, opened 2026-07-24
Devices decouple 4: carve magda_devices library + include-boundary guard


lucaromagnoli:
Part of the device-decouple epic.

## Problem

The device sources compile straight into the monolithic `magda_daw` STATIC library via `target_sources(magda_daw PRIVATE ...)` (e.g. `magda/daw/audio/CMakeLists.txt`). There is no separate device target, so the device layer cannot be linked or reused headless, and `magda_cli` needlessly drags in the whole UI to use engine/device code.

Depends on the value-type, registration, and injection issues landing first (so the carved library has a clean dependency set).

## Scope

- Create a new `magda_devices` STATIC library from: `magda/daw/audio/plugins/`, `magda/daw/audio/processors/`, `magda/daw/audio/faust_dsp/`, `plugins/compiled/`, `plugins/mutable/`.
- Move the Faust build machinery with it: the `magda_compile_faust_dsp()` codegen function and the `faust` / `eurorack` third-party build blocks, plus the compiled-DSP include dir and `MagdaAssets`/`faustlibraries` staging that the runtime-interpreted Faust devices need.
- Public API of the library = the registration hook + `DeviceProcessorFactory` + `InternalPluginRegistry`.
- `magda_devices` links only: `magda_types` + Tracktion Engine + JUCE (non-gui modules) + `faust::libfaust` + `magda::mutable`.
- `magda_daw` depends on `magda_devices` one-way. `PluginManagerSync`'s `dynamic_cast` special-casing of concrete plugin types stays in `magda_daw` and keeps consuming the device library's public plugin headers (one-directional, fine).
- Add an include-boundary guard (same pattern as the DAW/agents cycle fix) that fails configure if any `magda_devices` source includes a `core/` (non-types), `engine/`, `project/`, or `ui/` header.

## Acceptance

- `magda_devices` configures and links against only the dependencies listed above; the boundary guard passes.
- `magda_daw_app` and `magda_cli` build; `magda_cli` no longer pulls UI in order to use device code.
- No behavior change; tests green.

PR #2001, opened 2026-08-03
Native engine: render plan IR, compiler and canonical dump

lucaromagnoli:
Part of #1889 (epic #1882). First slice of the delivery split agreed on the issue: plan IR, deterministic compiler and canonical dump. No executor, no PDC, no buffer assignment.

## The IR

`magda/engine/plan/RenderPlan.hpp` — a flat, dependency-ordered op list.

- `OpKey` (track / rack / chain / device ids plus a structural role) is the differ's identity key. A device contributes a process op, a gain op and a meter op, so model id alone cannot identify an op; the role is the other half.
- `LivenessDomain` is on every op from day one, reserved for the anticipative executor. Live sources tag themselves and liveness propagates downstream.
- Input arity is fixed per op kind, so slot meaning stays positional: a device is always `[audio, MIDI, sidechain]` with unfilled slots holding an invalid `PortRef` rather than shifting the rest down.
- `bakeScheduling()` precomputes dependency counts, reversed edges in CSR form and the initial ready list. TE walks every node on the audio thread each block to reseed its queue; here that becomes a memcpy of a plan constant.
- `validatePlan()` checks dependency ordering, arity, port existence, producer/consumer signal-kind agreement and liveness monotonicity.

The plan encodes signal topology only. Clip positions, automation curves, modulation and parameter values stay out, so editing them never triggers a compile.

## The compiler

`magda/engine/plan/PlanCompiler.cpp` compiles the track model directly, taking `TrackInfo` values rather than reaching into `TrackManager`, so it stays pure and offline-testable.

Covered: track chains with nested racks, post-FX and mixer-analysis sections in the order the current engine sequences them, chain power as a gate over the insert chain (post-FX unaffected), device bypass, pre and post-fader sends to aux tracks, group summing, audio and MIDI sidechains, master, hardware output. MIDI stream policy comes from the existing `routing::ChainRoutingModel` rather than being re-derived.

Track emission order is a topological sort over routing, send and sidechain dependencies. A routing cycle is broken deterministically at the lowest-numbered track and reported.

Anything the compiler cannot express lands in `RenderPlan::diagnostics` instead of being dropped silently: rack aux outputs, multi-out pairs, sends with no resolvable destination. Nothing is ever summed into the wrong place to avoid an empty result.

## Isolation

The target is dark behind `MAGDA_BUILD_NATIVE_ENGINE` (default ON so it cannot rot) and linked by nothing but the tests. A configure-time boundary check rejects any Tracktion Engine include and any quoted include outside the model layer, so the isolation the epic asks for is enforced by the build rather than by review. Both branches of the check were verified by temporarily injecting a forbidden include of each kind.

## Tests

`tests/test_render_plan_compiler.cpp`, 17 model-level Catch2 cases including a full-dump golden, a determinism check, and a scheduling test that runs the executor's countdown loop at compile time to prove the baked constants drain every op exactly once. Every fixture also asserts `validatePlan()` is clean.

Full suite locally: 1805 cases, 1792 passed, 13 skipped, 0 failed.

## Next

The minimal single-threaded executor over this plan, then runtime-state ownership and epoch retirement, then PDC and buffer assignment.

### DIFF
--- a/magda/engine/CMakeLists.txt
+++ b/magda/engine/CMakeLists.txt
@@ -1,0 +1,84 @@
+# MAGDA native audio engine (epic #1882).
+#
+# Dark target: nothing in the app links it yet. It is built on every configure
+# so it cannot rot, and it is deliberately isolated — the C++ standard library,
+# JUCE headers and the MAGDA model layer, nothing else. No Tracktion Engine, no
+# UI, no app services. The boundary check below enforces that at configure time
+# rather than leaving it to review.
+
+add_library(magda_engine STATIC
+    plan/RenderPlan.cpp
+    plan/PlanCompiler.cpp
+    plan/PlanDump.cpp
+    plan/RenderPlan.hpp
+    plan/PlanCompiler.hpp
+    plan/PlanDump.hpp
+)
+add_library(magda::engine ALIAS magda_engine)
+
+target_compile_features(magda_engine PUBLIC cxx_std_20)
+
+# magda_types carries the model-layer headers and publishes magda/daw as its
+# include root, so engine sources resolve "core/<model>.hpp".
+target_link_libraries(magda_engine PUBLIC magda_types)
+
+target_include_directories(magda_engine PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+
+# Borrow JUCE's headers and defines without linking the modules: the model
+# headers name juce::String and juce::Colour, and the juce_core / juce_graphics
+# unity sources are already compiled once in magda_daw, so the undefined
+# symbols resolve at the final link instead of duplicating those objects here.
+# Same arrangement magda_types uses.
+target_include_directories(magda_engine PRIVATE
+    $<TARGET_PROPERTY:juce::juce_core,INTERFACE_INCLUDE_DIRECTORIES>
+    $<TARGET_PROPERTY:juce::juce_graphics,INTERFACE_INCLUDE_DIRECTORIES>)
+target_compile_definitions(magda_engine PRIVATE
+    $<TARGET_PROPERTY:juce::juce_core,INTERFACE_COMPILE_DEFINITIONS>
+    $<TARGET_PROPERTY:juce::juce_graphics,INTERFACE_COMPILE_DEFINITIONS>)
+
+# --- Dependency boundary check -----------------------------------------------
+#
+# Angle-bracket includes may name the standard library or JUCE, never Tracktion
+# Engine. Quoted includes may only reach the model layer ("core/...") or the
+# engine's own headers.
+
+set(MAGDA_ENGINE_ALLOWED_QUOTED_PREFIXES "core/" "plan/")
+
+file(GLOB_RECURSE MAGDA_ENGINE_SOURCES CONFIGURE_DEPENDS
+    "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/*.hpp")
+
+foreach(_engine_source ${MAGDA_ENGINE_SOURCES})
+    file(RELATIVE_PATH _engine_relative "${CMAKE_SOURCE_DIR}" "${_engine_source}")
+    file(STRINGS "${_engine_source}" _engine_includes REGEX "^[ \t]*#[ \t]*include[ \t]*[<\"]")
+
+    foreach(_include_line ${_engine_includes})
+        string(REGEX REPLACE "^[ \t]*#[ \t]*include[ \t]*[<\"]([^>\"]+)[>\"].*$" "\\1"
+            _included_header "${_include_line}")
+
+        if(_included_header MATCHES "^tracktion")
+            message(FATAL_ERROR
+                "\n"
+                "Native engine boundary violation in ${_engine_relative}:\n"
+                "    includes \"${_included_header}\"\n"
+                "The engine is built to replace Tracktion Engine and must never depend on it.\n")
+        endif()
+
+        if(_include_line MATCHES "#[ \t]*include[ \t]*\"")
+            set(_include_allowed FALSE)
+            foreach(_allowed_prefix ${MAGDA_ENGINE_ALLOWED_QUOTED_PREFIXES})
+                if(_included_header MATCHES "^${_allowed_prefix}")
+                    set(_include_allowed TRUE)
+                endif()
+            endforeach()
+
+            if(NOT _include_allowed)
+                message(FATAL_ERROR
+                    "\n"
+                    "Native engine boundary violation in ${_engine_relative}:\n"
+                    "    includes \"${_included_header}\"\n"
+                    "The engine may only include the model layer (core/) and its own headers.\n")
+            endif()
+        endif()
+    endforeach()
+endforeach()
--- a/magda/engine/CMakeLists.txt
+++ b/magda/engine/CMakeLists.txt
@@ -1,7 +1,7 @@
 # MAGDA native audio engine (epic #1882).
 #
 # Dark target: nothing in the app links it yet. It is built on every configure
-# so it cannot rot, and it is deliberately isolated — the C++ standard library,
+# so it cannot rot, and it is deliberately isolated: the C++ standard library,
 # JUCE headers and the MAGDA model layer, nothing else. No Tracktion Engine, no
 # UI, no app services. The boundary check below enforces that at configure time
 # rather than leaving it to review.
--- a/magda/engine/plan/PlanCompiler.cpp
+++ b/magda/engine/plan/PlanCompiler.cpp
@@ -1,0 +1,580 @@
+#include "plan/PlanCompiler.hpp"
+
+#include <algorithm>
+#include <map>
+#include <optional>
+#include <set>
+#include <string>
+#include <vector>
+
+#include "core/ChainRoutingModel.hpp"
+#include "core/RackInfo.hpp"
+#include "core/TrackChain.hpp"
+#include "core/TrackInfo.hpp"
+
+namespace magda::engine {
+namespace {
+
+/// The audio + MIDI pair flowing through a chain at one point.
+struct ChainSignal {
+    PortRef audio;
+    PortRef midi;
+};
+
+/// Where a chain element lives, for op keys. Only the innermost rack and chain
+/// are recorded: device ids are project-unique, so deeper nesting needs no
+/// further qualification.
+struct ChainSite {
+    TrackId trackId = INVALID_TRACK_ID;
+    RackId rackId = INVALID_RACK_ID;
+    ChainId chainId = INVALID_CHAIN_ID;
+};
+
+/// Analysis devices are transparent passthroughs with no gain trim and no
+/// meter of their own, so they compile to the process op alone.
+bool isTransparentTap(const DeviceInfo& device) {
+    return device.deviceType == DeviceType::Analysis;
+}
+
+bool consumesMidi(const DeviceInfo& device) {
+    return device.isInstrument || device.canReceiveMidi || device.deviceType == DeviceType::MIDI;
+}
+
+bool elementsConsumeMidi(const std::vector<ChainElement>& elements) {
+    return std::ranges::any_of(elements, [](const ChainElement& element) {
+        if (isDevice(element))
+            return consumesMidi(getDevice(element));
+        if (isRack(element))
+            return std::ranges::any_of(getRack(element).chains, [](const ChainInfo& chain) {
+                return elementsConsumeMidi(chain.elements);
+            });
+        return false;
+    });
+}
+
+bool chainConsumesMidi(const TrackInfo& track) {
+    return elementsConsumeMidi(track.chain.fxChainElements) ||
+           std::ranges::any_of(
+               track.chain.postFxChainElements,
+               [](const PostFxChainElement& element) { return consumesMidi(element.device); });
+}
+
+/// Tracks whose signal a device in `elements` reads as a sidechain input.
+/// `type` narrows the search to audio or MIDI sidechains; nullopt takes both.
+void collectSidechainSources(const std::vector<ChainElement>& elements,
+                             std::optional<SidechainConfig::Type> type, std::set<TrackId>& out) {
+    const auto collect = [&](const SidechainConfig& sidechain) {
+        if (sidechain.isActive() && (!type.has_value() || sidechain.type == *type))
+            out.insert(sidechain.sourceTrackId);
+    };
+
+    for (const auto& element : elements) {
+        if (isDevice(element)) {
+            collect(getDevice(element).sidechain);
+        } else if (isRack(element)) {
+            for (const auto& chain : getRack(element).chains)
+                collectSidechainSources(chain.elements, type, out);
+        }
+    }
+}
+
+void collectSidechainSources(const TrackInfo& track, std::optional<SidechainConfig::Type> type,
+                             std::set<TrackId>& out) {
+    collectSidechainSources(track.chain.fxChainElements, type, out);
+    for (const auto& element : track.chain.postFxChainElements)
+        if (element.device.sidechain.isActive() &&
+            (!type.has_value() || element.device.sidechain.type == *type))
+            out.insert(element.device.sidechain.sourceTrackId);
+}
+
+/// The track a track's audio output feeds. "track:<id>" routes into a group;
+/// everything else (including the default "master" and an empty string) goes
+/// straight to the master track.
+TrackId resolveAudioDestination(const TrackInfo& track) {
+    const auto& routing = track.audioOutputDevice;
+    if (routing.startsWith("track:")) {
+        const auto id = routing.substring(6);
+        if (id.isNotEmpty() && id.containsOnly("-0123456789"))
+            return id.getIntValue();
+    }
+    return MASTER_TRACK_ID;
+}
+
+class Compiler {
+  public:
+    Compiler(const std::vector<TrackInfo>& tracks, const TrackInfo& master,
+             const CompileOptions& options)
+        : tracks_(tracks), master_(master), options_(options) {}
+
+    RenderPlan run();
+
+  private:
+    OpId addOp(OpKind kind, const OpKey& key, std::vector<PortRef> inputs,
+               std::vector<SignalKind> outputs);
+    void diagnose(std::string message);
+
+    const TrackInfo* findTrack(TrackId id) const;
+    TrackId resolveSendDestination(const SendInfo& send) const;
+    std::vector<const TrackInfo*> computeTrackOrder();
+
+    void emitTrack(const TrackInfo& track);
+    ChainSignal emitElements(const std::vector<ChainElement>& elements, const ChainSite& site,
+                             ChainSignal signal);
+    ChainSignal emitDevice(const DeviceInfo& device, const ChainSite& site, ChainSignal signal);
+    ChainSignal emitRack(const RackInfo& rack, const ChainSite& site, ChainSignal signal);
+
+    /// Sums `sources` into one audio port, always through an op so the op's
+    /// identity survives sources appearing and disappearing.
+    PortRef emitMix(const OpKey& key, const std::vector<PortRef>& sources);
+
+    const std::vector<TrackInfo>& tracks_;
+    const TrackInfo& master_;
+    CompileOptions options_;
+
+    RenderPlan plan_;
+    /// Post-fader output of each emitted track: audio sidechain sources read it.
+    std::map<TrackId, PortRef> trackOutput_;
+    /// Chain-head MIDI of each emitted track: MIDI sidechain sources read it.
+    std::map<TrackId, PortRef> trackMidiInput_;
+    /// Signals waiting to be summed into a track that has not been emitted yet:
+    /// child track outputs and incoming send taps.
+    std::map<TrackId, std::vector<PortRef>> pendingInputs_;
+    /// Tracks another track's device reads MIDI from. Their MIDI clips have to
+    /// be compiled even when their own chain has nothing that consumes MIDI.
+    std::set<TrackId> midiSidechainSources_;
+};
+
+OpId Compiler::addOp(OpKind kind, const OpKey& key, std::vector<PortRef> inputs,
+                     std::vector<SignalKind> outputs) {
+    PlanOp op;
+    op.kind = kind;
+    op.key = key;
+    op.inputs = std::move(inputs);
+    op.outputs = std::move(outputs);
+
+    // Liveness flows downstream: an op is live if it reads anything live. Live
+    // sources set their own domain at the call site.
+    op.liveness = LivenessDomain::Deterministic;
+    for (const auto& input : op.inputs) {
+        if (input.valid() &&
+            plan_.ops[static_cast<std::size_t>(input.op)].liveness == LivenessDomain::Live) {
+            op.liveness = LivenessDomain::Live;
+            break;
+        }
+    }
+
+    plan_.ops.push_back(std::move(op));
+    return static_cast<OpId>(plan_.ops.size()) - 1;
+}
+
+void Compiler::diagnose(std::string message) {
+    plan_.diagnostics.push_back(std::move(message));
+}
+
+const TrackInfo* Compiler::findTrack(TrackId id) const {
+    if (id == master_.id)
+        return &master_;
+    const auto found =
+        std::ranges::find_if(tracks_, [id](const TrackInfo& track) { return track.id == id; });
+    return found == tracks_.end() ? nullptr : &*found;
+}
+
+TrackId Compiler::resolveSendDestination(const SendInfo& send) const {
+    if (send.destTrackId != INVALID_TRACK_ID)
+        return send.destTrackId;
+
+    // Older projects stored the aux bus index only.
+    const auto found = std::ranges::find_if(tracks_, [&send](const TrackInfo& track) {
+        return track.auxBusIndex >= 0 && track.auxBusIndex == send.busIndex;
+    });
+    return found == tracks_.end() ? INVALID_TRACK_ID : found->id;
+}
+
+std::vector<const TrackInfo*> Compiler::computeTrackOrder() {
+    const auto numTracks = tracks_.size();
+    std::map<TrackId, std::size_t> indexById;
+    for (std::size_t i = 0; i < numTracks; ++i)
+        indexById[tracks_[i].id] = i;
+
+    // The chord track is a monitor-only progression lane and never renders.
+    std::vector<bool> skipped(numTracks, false);
+    for (std::size_t i = 0; i < numTracks; ++i)
+        skipped[i] = tracks_[i].type == TrackType::Chord;
+
+    std::vector<std::vector<std::size_t>> successors(numTracks);
+    std::vector<int> indegree(numTracks, 0);
+    auto addEdge = [&](std::size_t from, std::size_t to) {
+        if (from == to || skipped[from] || skipped[to])
+            return;
+        successors[from].push_back(to);
+        ++indegree[to];
+    };
+    auto indexOf = [&](TrackId id) -> long {
+        const auto it = indexById.find(id);
+        return it == indexById.end() ? -1 : static_cast<long>(it->second);
+    };
+
+    for (std::size_t i = 0; i < numTracks; ++i) {
+        const auto& track = tracks_[i];
+        if (skipped[i])
+            continue;
+
+        // A track's audio has to exist before whatever sums it.
+        if (const auto destination = indexOf(resolveAudioDestination(track)); destination >= 0)
+            addEdge(i, static_cast<std::size_t>(destination));
+
+        for (const auto& send : track.sends)
+            if (const auto destination = indexOf(resolveSendDestination(send)); destination >= 0)
+                addEdge(i, static_cast<std::size_t>(destination));
+
+        // Sidechain and multi-out read another track, so that track comes first.
+        std::set<TrackId> sidechainSources;
+        collectSidechainSources(track, std::nullopt, sidechainSources);
+        for (const auto sourceId : sidechainSources)
+            if (const auto source = indexOf(sourceId); source >= 0)
+                addEdge(static_cast<std::size_t>(source), i);
+
+        if (track.multiOutLink)
+            if (const auto source = indexOf(track.multiOutLink->sourceTrackId); source >= 0)
+                addEdge(static_cast<std::size_t>(source), i);
+    }
+
+    std::set<std::size_t> ready;
+    std::vector<bool> emitted(numTracks, false);
+    std::size_t remaining = 0;
+    for (std::size_t i = 0; i < numTracks; ++i) {
+        if (skipped[i])
+            continue;
+        ++remaining;
+        if (indegree[i] == 0)
+            ready.insert(i);
+    }
+
+    std::vector<const TrackInfo*> order;
+    order.reserve(remaining);
+    while (order.size() < remaining) {
+        if (ready.empty()) {
+            // Every remaining track waits on another remaining track. Force the
+            // lowest-numbered one and drop the connections that close the cycle.
+            for (std::size_t i = 0; i < numTracks; ++i) {
+                if (skipped[i] || emitted[i])
+                    continue;
+                diagnose("routing cycle through track " + std::to_string(tracks_[i].id) +
+                         ": compiled ahead of its sources, connections closing the cycle dropped");
+                indegree[i] = 0;
+                ready.insert(i);
+                break;
+            }
+        }
+
+        const auto next = *ready.begin();
+        ready.erase(ready.begin());
+        emitted[next] = true;
+        order.push_back(&tracks_[next]);
+        for (const auto successor : successors[next])
+            if (--indegree[successor] == 0 && !emitted[successor])
+                ready.insert(successor);
+    }
+
+    return order;
+}
+
+PortRef Compiler::emitMix(const OpKey& key, const std::vector<PortRef>& sources) {
+    return PortRef{addOp(OpKind::MixAudio, key, sources, {SignalKind::Audio}), 0};
+}
+
+ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site,
+                                 ChainSignal signal) {
+    if (device.bypassed)
+        return signal;
+
+    const auto node = routing::makeRoutingNode(device);
+
+    PortRef midiIn;
+    if (node.usesExternalMidiSidechain()) {
+        const auto source = trackMidiInput_.find(node.midiSidechainSourceTrackId);
+        if (source != trackMidiInput_.end())
+            midiIn = source->second;
+        else
+            diagnose("device " + std::to_string(device.id) + " on track " +
+                     std::to_string(site.trackId) + ": MIDI sidechain source track " +
+                     std::to_string(node.midiSidechainSourceTrackId) + " produces no MIDI");
+    } else if (node.receivesChainMidi()) {
+        midiIn = signal.midi;
+    }
+
+    PortRef sidechainIn;
+    if (device.sidechain.type == SidechainConfig::Type::Audio && device.sidechain.isActive()) {
+        const auto source = trackOutput_.find(device.sidechain.sourceTrackId);
+        if (source != trackOutput_.end())
+            sidechainIn = source->second;
+        else
+            diagnose("device " + std::to_string(device.id) + " on track " +
+                     std::to_string(site.trackId) + ": audio sidechain source track " +
+                     std::to_string(device.sidechain.sourceTrackId) + " is not routed");
+    }
+
+    const auto producesMidi = node.outputsPluginMidi();
+    std::vector<SignalKind> outputs{SignalKind::Audio};
+    if (producesMidi)
+        outputs.push_back(SignalKind::Midi);
+
+    OpKey key{site.trackId, site.rackId, site.chainId, device.id, OpRole::DeviceProcess, 0};
+    const auto processOp =
+        addOp(OpKind::Device, key, {signal.audio, midiIn, sidechainIn}, std::move(outputs));
+
+    ChainSignal out;
+    out.audio = PortRef{processOp, 0};
+
+    if (!isTransparentTap(device)) {
+        key.role = OpRole::DeviceGain;
+        out.audio = PortRef{addOp(OpKind::Gain, key, {out.audio}, {SignalKind::Audio}), 0};
+
+        if (options_.deviceMeters) {
+            key.role = OpRole::DeviceMeter;
+            out.audio = PortRef{addOp(OpKind::Meter, key, {out.audio}, {SignalKind::Audio}), 0};
+        }
+    }
+
+    out.midi = signal.midi;
+    if (producesMidi) {
+        const PortRef deviceMidi{processOp, 1};
+        if (node.passesRawMidiInput() && signal.midi.valid()) {
+            key.role = OpRole::ChainMidiMerge;
+            out.midi = PortRef{
+                addOp(OpKind::MergeMidi, key, {signal.midi, deviceMidi}, {SignalKind::Midi}), 0};
+        } else {
+            out.midi = deviceMidi;
+        }
+    }
+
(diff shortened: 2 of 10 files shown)


magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+        audioSources.insert(audioSources.end(), pending->second.begin(), pending->second.end());
+        pendingInputs_.erase(pending);
+    }
+
+    const OpKey inputKey{track.id,          INVALID_RACK_ID,         INVALID_CHAIN_ID,
+                         INVALID_DEVICE_ID, OpRole::TrackAudioInput, 0};
+    ChainSignal signal;
+    signal.audio = emitMix(inputKey, audioSources);
+
+    std::vector<PortRef> midiSources;
```
lucaromagnoli, line 648: `midiInputDevice` can also be `track:<sourceId>` (see `routeMidiOutputToTrack`), but this path is never parsed here. A monitored/armed destination therefore turns the internal route into a live hardware `MidiInput`; an unmonitored destination ignores it entirely. Please add the source→destination ordering edge and feed the source track MIDI into this merge instead of treating every non-empty value as hardware input.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+void collectSidechainSources(const std::vector<ChainElement>& elements,
+                             std::optional<SidechainConfig::Type> type, std::set<TrackId>& out) {
+    const auto collect = [&](const SidechainConfig& sidechain) {
+        if (sidechain.isActive() && (!type.has_value() || sidechain.type == *type))
+            out.insert(sidechain.sourceTrackId);
+    };
+
+    for (const auto& element : elements) {
+        if (isDevice(element)) {
+            collect(getDevice(element).sidechain);
```
lucaromagnoli: Dependency discovery needs to apply the same activity rules as emission. This currently collects sidechains from bypassed devices and from a disabled insert chain, even though those devices are skipped later. Such a stale edge can create a false cycle with an active send/routing edge; cycle breaking can then emit the destination too early and drop the real connection as late-arriving. Please filter bypassed devices and disabled chain sections here.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+        // A chain that leaves the MIDI stream untouched is transparent to it;
+        // only chains that generate MIDI contribute to the rack's MIDI output.
+        if (chainSignal.midi.valid() && !(chainSignal.midi == signal.midi))
+            chainMidi.push_back(chainSignal.midi);
+    }
+
+    ChainSignal out;
+    const OpKey mixKey{site.trackId,      rack.id,         INVALID_CHAIN_ID,
+                       INVALID_DEVICE_ID, OpRole::RackMix, 0};
+    const auto mixed = emitMix(mixKey, chainAudio);
```
lucaromagnoli, line 544: Parity gap: when `rack.chains` is empty this compiles to a zero-input MixAudio, i.e. silence. The current engine passes audio and MIDI straight through when a rack has no chains (RackSyncManager.cpp:957). MIDI is already transparent here via `out.midi = signal.midi`, audio is not. Suggest special-casing `rack.chains.empty()` to pass `signal` through, plus an empty-rack test. Chains present but all aux-routed compiling to main-out silence does match the engine, so that path is fine.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+    return order;
+}
+
+PortRef Compiler::emitMix(const OpKey& key, const std::vector<PortRef>& sources) {
+    return PortRef{addOp(OpKind::MixAudio, key, sources, {SignalKind::Audio}), 0};
+}
+
+ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site,
+                                 ChainSignal signal) {
+    if (device.bypassed)
```
lucaromagnoli, line 413: Design note for #1890/#1891, nothing to change here: treating bypass (and chain power) as structure means a bypassed device contributes no ops, so under the differ rules (no match = dead, state released at epoch) toggling chain power would tear down and re-instantiate every plugin on the track. The current engine keeps instances alive across both toggles (setEnabled, and setChainEnabled deliberately leaves flags and instances untouched). I'd resolve this with a model-aware eviction policy in the state store (release on model deletion, retain across bypass and power toggles) rather than putting bypassed ops in the plan. Worth a note on the epic so it lands as a decision.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+    std::vector<const TrackInfo*> order;
+    order.reserve(remaining);
+    while (order.size() < remaining) {
+        if (ready.empty()) {
+            // Every remaining track waits on another remaining track. Force the
+            // lowest-numbered one and drop the connections that close the cycle.
+            for (std::size_t i = 0; i < numTracks; ++i) {
+                if (skipped[i] || emitted[i])
+                    continue;
+                diagnose("routing cycle through track " + std::to_string(tracks_[i].id) +
```
lucaromagnoli: When the ready set empties, the forced track's remaining producers can include tracks that are merely blocked by the cycle rather than on it; their sends then arrive after emission and get dropped too. Everything is diagnosed so nothing is silent, but this message over-claims: it can also drop connections that do not close the cycle. Fine for v1; a stricter fix would force a track from an actual strongly connected component, or at least soften the wording.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+            diagnose("device " + std::to_string(device.id) + " on track " +
+                     std::to_string(site.trackId) + ": MIDI sidechain source track " +
+                     std::to_string(node.midiSidechainSourceTrackId) + " produces no MIDI");
+    } else if (node.receivesChainMidi()) {
+        midiIn = signal.midi;
+    }
+
+    PortRef sidechainIn;
+    if (device.sidechain.type == SidechainConfig::Type::Audio && device.sidechain.isActive()) {
+        const auto source = trackOutput_.find(device.sidechain.sourceTrackId);
```
lucaromagnoli: Tap-point parity to verify: this wires the source track's post-fader, post-meter output into the sidechain slot. The current engine delegates to TE setSidechainSourceID and I could not confirm from this repo whether TE taps pre or post fader on the source track. If TE taps pre-fader this is a parity break. Worth pinning down before the null-diff harness, or at latest flagging it there.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+RenderPlan Compiler::run() {
+    // A track feeding someone else's MIDI sidechain has to produce MIDI even
+    // when nothing in its own chain consumes it, so this is collected up front.
+    for (const auto& track : tracks_)
+        collectSidechainSources(track, SidechainConfig::Type::MIDI, midiSidechainSources_);
+    collectSidechainSources(master_, SidechainConfig::Type::MIDI, midiSidechainSources_);
+
+    for (const auto* track : computeTrackOrder())
+        emitTrack(*track);
+    emitTrack(master_);
```
lucaromagnoli, line 801: Because the master is always emitted last, a device sidechained to the master's output can never resolve: the trackOutput_ lookup misses and it lands in diagnostics as not routed. Only matters if the sidechain source picker offers the master. If it does, this needs the master hoisted into the ordering rather than appended.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,580 @@
...
+            // silently, and compiling it nowhere would leave dead ops behind.
+            diagnose("rack " + std::to_string(rack.id) + " chain " + std::to_string(chain.id) +
+                     ": aux output " + std::to_string(chain.outputIndex) + " is not routed yet");
+            continue;
+        }
+
+        const ChainSite chainSite{site.trackId, rack.id, chain.id};
+
+        auto chainSignal = signal;
+        if (!chain.bypassed)
```
lucaromagnoli, line 526: Chain mute and solo are absent from the plan, which I read as intentional: they resolve at the RackChainFader in the value layer, so toggling mute never recompiles. Two things to carry forward: chain solo needs cross-chain resolution, so the value-snapshot spec has to cover it before the executor lands; and a muted chain's devices still execute under the plan where the current engine disconnects them entirely, which is a CPU difference only.

magda/engine/CMakeLists.txt
```diff
@@ -0,0 +1,84 @@
...
+    $<TARGET_PROPERTY:juce::juce_core,INTERFACE_COMPILE_DEFINITIONS>
+    $<TARGET_PROPERTY:juce::juce_graphics,INTERFACE_COMPILE_DEFINITIONS>)
+
+# --- Dependency boundary check -----------------------------------------------
+#
+# Angle-bracket includes may name the standard library or JUCE, never Tracktion
+# Engine. Quoted includes may only reach the model layer ("core/...") or the
+# engine's own headers.
+
+set(MAGDA_ENGINE_ALLOWED_QUOTED_PREFIXES "core/" "plan/")
```
lucaromagnoli, line 45: Nit: the quoted-include allowlist admits all of core/, which includes service headers like TrackManager.hpp, not just value types. Directory granularity is fine for now, just noting the boundary is weaker than the model-values-only intent; a later tightening could split the value-type headers into their own directory.

lucaromagnoli (commented):
Reviewed against the current engine's semantics, not just the PR text. Verified matches: chain power gating is exactly isDeviceEffectivelyEnabled (device bypass always, chain power gates inserts only); rack bypass as dry passthrough matches the wet/dry implementation; the fx -> post-FX -> mixer-analysis -> fader -> meter order matches ensureVolumePluginPosition; the MIDI sidechain tap at chain head matches the monitor being inserted at index 0 of the source track.

Verdict: solid foundation. One real parity gap (empty rack renders silence, inline comment), the rest are design notes and follow-ups. Nits not worth their own threads: comments across the new files use em-dashes; containsOnly("-0123456789") in resolveAudioDestination accepts strings like "1-2" and getIntValue parses them as 1, harmless.

Test suggestions beyond the inline ones: a nested-rack case (the emitRack recursion is untested), rack bypass passthrough, and a MIDI sidechain wiring case.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,668 @@
...
+
+/// The track id in a "track:<id>" routing string, or nullopt for a hardware
+/// device name, the default "master", or an empty field.
+std::optional<TrackId> parseTrackReference(const juce::String& routing) {
+    if (!routing.startsWith("track:"))
+        return std::nullopt;
+
+    const auto id = routing.substring(6).trim();
+    if (id.isEmpty() || !id.containsOnly("-0123456789"))
+        return std::nullopt;
```
lucaromagnoli: `nullopt` now conflates “not a track route” with “malformed track route.” Consequently `track:1-2` is rejected here but then compiled as a live hardware input at the input call sites; as an audio output it silently falls back to master. That still produces incorrect topology without a diagnostic. Please preserve a distinct malformed result (or check the `track:` prefix at each caller) so invalid internal routes are diagnosed and not reinterpreted as hardware/default routing.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,668 @@
...
+
+        for (const auto& send : track.sends)
+            if (const auto destination = indexOf(resolveSendDestination(send)); destination >= 0)
+                addEdge(i, static_cast<std::size_t>(destination));
+
+        // Sidechains, internal input routes and multi-out all read another
+        // track, so that track comes first.
+        std::set<TrackId> upstream;
+        collectSidechainSources(track, std::nullopt, upstream);
+        if (const auto source = parseTrackReference(track.audioInputDevice))
```
lucaromagnoli: These internal-input edges need the same monitoring/activity gate as emission below. `emitTrack` only reads an internal route when `carriesClips && monitorsInput`, but ordering adds it unconditionally. For example, if track B has an unmonitored input from A while B actively sends to A, this creates a false A→B→A cycle; forcing A first then causes B’s real send to arrive after A was compiled and be dropped. Please add these upstream edges only when the corresponding route will actually be emitted (and mirror that condition in `midiSourceTracks_`).

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,668 @@
...
+            out.insert(sidechain.sourceTrackId);
+    };
+
+    for (const auto& element : elements) {
+        if (isDevice(element)) {
+            const auto& device = getDevice(element);
+            if (!device.bypassed)
+                collect(device.sidechain);
+        } else if (isRack(element)) {
+            const auto& rack = getRack(element);
```
lucaromagnoli, line 100: This handles sidechains on devices inside the rack but silently ignores `rack.sidechain` itself. Rack-level audio/MIDI sources are model-supported and drive rack-wide follower/LFO modulation, so the source dependency and liveness never reach this plan. Since the current IR has no rack-sidechain input op, either represent that dependency here or emit an explicit unsupported diagnostic; otherwise a valid rack sidechain is compiled as if it did not exist, contrary to the no-silent-drop contract.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fixed in 6993a351. `midiInputDevice` is now parsed for `track:<id>`, gets the source to destination ordering edge, and reads the source track chain-head MIDI rather than compiling a hardware input.

The same bug was on `audioInputDevice`, which takes the identical `track:<id>` form, so that is fixed alongside it: an internal audio route reads the source track post-mute output. Both keep the monitoring/arm gate, since both go through the input path in the current engine. A track whose MIDI another track reads now compiles its MIDI clips even when nothing in its own chain consumes MIDI, otherwise the route had nothing to connect to.

Tests: "An internal MIDI route reads the source track MIDI" and "An internal audio route reads the source track post-mute output", both declaring the destination before the source so only the dependency can order them.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fixed in 6993a351. Discovery now applies exactly the emission rules: bypassed devices skipped, bypassed racks and chains skipped, aux-routed chains skipped, and the insert chain skipped entirely when chain power is off. Post-FX stays outside the power gate on both sides.

Two tests cover it, each built so the stale edge would close a cycle against a live send: "A sidechain on an inactive device is not an ordering dependency" and "Chain power gates sidechain dependencies too". Both assert the send survives and no diagnostic is produced.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fixed in 6993a351. `rack.chains.empty()` returns the incoming signal unchanged, so an empty rack is transparent to both audio and MIDI. Test: "An empty rack is transparent", asserting no RackMix or RackFader op is emitted and the track fader reads the chain input directly.

Kept the aux-routed case as it was, per your note that main-out silence there matches the engine.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Agreed, and noted on the epic rather than changed here. Putting bypassed ops in the plan to solve a lifetime problem would cost a block of processing forever; the eviction policy belongs in the state store, keyed on the model (release on deletion, retain across bypass and power toggles).

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fair, the wording claimed more than the algorithm delivers. Reworded in 6993a351 to say what it did rather than what was lost:

    routing cycle: track N compiled ahead of its sources to break it

with a comment recording that the forced track may be merely blocked by a cycle rather than on it, and that each lost connection is reported separately when it arrives too late to connect. Leaving the strongly-connected-component version for when a real session needs it.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Pinned down, and it was a real gap. tracktion_EditNodeBuilder.cpp:1523 inserts the sidechain SendNode after `createPluginNodeForTrack` (which contains the fader) and before both the level meter and TrackMutingNode. So post-fader was right, pre-mute was not.

I had mute folded into the fader value, which would have starved a compressor keyed off a muted track and made that track own meter read silence. 6993a351 makes mute its own stage after the meter, so tracks now expose two ports: `trackSidechainTap_` (post-fader, pre-mute) for sidechains, and `trackRoutedOutput_` (post-mute) for destinations and internal audio routes. Test: "Mute is applied after the meter and the sidechain tap".

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Checked: it cannot happen today. Both sidechain source pickers iterate `TrackManager::getTracks()` (DeviceSlotSidechainControls.cpp:41 and NodeComponent.cpp:1629), and the master lives in `masterTrack_` outside that vector, so it is never offered. Leaving as is; if the picker ever widens, the master has to be hoisted into the ordering rather than appended.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Intentional, yes: both resolve at the RackChainFader in the value layer so toggling never recompiles. Both points carried to the epic. Solo needs cross-chain resolution defined in the value-snapshot spec before the executor lands, and the muted-chain CPU difference is what silence tracking is for rather than something the plan should express.

magda/engine/CMakeLists.txt
lucaromagnoli: Agreed, the granularity is coarser than the intent. It is directory-level because that is the boundary that exists today; #1834 splits the value-type headers into their own directory, and the allowlist should narrow to that when it lands. Noted rather than worked around, since a hand-maintained per-header list would rot faster than it protects.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,668 @@
...
+ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, ChainSignal signal) {
+    if (rack.bypassed)
+        return signal;
+
+    // A rack with no chains is transparent, not a hole. Compiling it to a
+    // zero-input mix would silence the track; the current engine passes audio
+    // and MIDI straight through. (A rack whose chains all route to aux outputs
+    // does render silence on the main output, and that is the engine's
+    // behaviour too, so only the empty case is special.)
+    if (rack.chains.empty())
```
lucaromagnoli: Close, but the empty rack is not fully transparent in the current engine: rack volume/pan map onto RackInstance leftOutDb/rightOutDb (applyRackInstanceState runs regardless of chains), and TE applies those on the wet path before the dry sum (EditNodeBuilder.cpp:1280-1293), so a chainless, non-bypassed rack still applies rack volume/pan to the passthrough. The empty case should still emit the RackFader op over the incoming audio, which also keeps the fader's differ identity stable when the first chain is added. The bypassed case above is genuinely fully transparent (dry path skips the out gains), so that branch is right as is. The empty-rack test's RackFader == 0 assertion flips with this.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: These edges skip the gates emission applies: the audio route only connects under carriesClips plus the monitoring gate, and the MIDI route under carriesClips plus receivesLiveMidiInput. This is the same discovery-vs-emission asymmetry this commit just fixed for sidechain sources, with the same failure: a non-monitoring route can invent a cycle whose breaking costs a real connection (say B input-routes from A while not armed, and a device on A sidechains B). Same for the midiSourceTracks_ insertion in run(): ungated, so a non-monitoring MIDI route makes the source compile ClipMidi and TrackMidiInput ops nothing reads. Suggest gating both spots on the same predicates as emission. The multiOutLink edge shares the property, but since multi-out is unwired and diagnosed anyway I'd leave that one.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,668 @@
...
+            const auto destination = resolveSendDestination(send);
+            if (destination == INVALID_TRACK_ID) {
+                diagnose("track " + std::to_string(track.id) + " send " + std::to_string(slot) +
+                         ": no destination track for aux bus " + std::to_string(send.busIndex));
+                continue;
+            }
+
+            const OpKey key{track.id,          INVALID_RACK_ID, INVALID_CHAIN_ID,
+                            INVALID_DEVICE_ID, OpRole::SendTap, static_cast<int>(slot)};
+            const auto op = addOp(OpKind::SendTap, key, {source}, {SignalKind::Audio});
```
lucaromagnoli, line 733: Spec note for the value layer, no change here: TE couples send audibility to the track's mute state (AuxSendNode.cpp:19-34; with the default shouldProcessAuxSendWhenTrackIsMuted the send survives mute only while the track contents still process). Tapping pre-mute is the correct structure for that behavior, but it means the SendTap's resolved gain has to fold in track mute the same way the TrackMute op does, or a muted track keeps feeding its auxes. Worth capturing in the value-snapshot spec and as a muted-track-with-sends case in the null-diff corpus.

lucaromagnoli (commented):
Second round, on the review-response commit. The mute placement is now verified against TE rather than plausible: EditNodeBuilder.cpp:1046-1053 inserts the sidechain SendNode after the plugin list (fader included), takes the metering tap pre-mute, and applies TrackMutingNode last, which is exactly the fader -> meter -> mute shape the compiler emits, and it also resolves my round-one question about the sidechain tap point (post-fader is correct). The internal track-routing discovery is a real catch beyond the review, and gating dependency discovery by the emission rules is the right fix for the invented-cycle class.

Remaining: one parity gap (empty rack loses rack volume/pan) and one leftover instance of the same discovery-vs-emission asymmetry this commit fixed for sidechains, both inline. Plus one value-layer spec note on sends and mute that needs no code change here.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fair, that was a silent drop. Fixed in dbd2d255 with an explicit diagnostic rather than an edge:

    rack 4: MIDI sidechain from track 2 feeds rack modulation, which the plan does not carry yet

Deliberately no ordering edge, for the same reason the gating fixes exist: rack sidechain feeds the rack own followers and LFOs, so it is a modulation source rather than a signal edge, and the plan emits nothing for it. An edge with no emission counterpart can only invent cycles. The real dependency lands with the modulation slice, and by then there is something for it to connect to. Test: "A rack-level sidechain is reported as unsupported".

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fixed in dbd2d255. `activeAudioInputRoute` and `activeMidiInputRoute` now carry the gate (carriesClips plus monitoring/arm for audio, carriesClips plus receivesLiveMidiInput for MIDI), and ordering, emission and the `midiSourceTracks_` insertion all go through them. A route is a dependency exactly when it is also a connection.

Also dropped the multiOutLink edge. You said to leave it, but by the rule you stated it has no emission counterpart at all, so it can only invent cycles and cannot lose anything by going, multi-out being diagnosed either way.

Your exact scenario is now a test, "An inactive internal route is not an ordering dependency": B input-routes from A while unarmed, a device on A sidechains B, and the assertion is that the sidechain survives with no diagnostic. Plus "An unmonitored MIDI route does not make its source compile MIDI".

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fixed in dbd2d255. Parsing distinguishes four cases instead of two: None (the track does not read that input at all), External (a hardware device or the default master routing), Track, and Malformed. A malformed value is diagnosed and left unconnected rather than falling through:

    track 1: audio input routing track:1-2 does not name a track, input not connected

Tests cover the audio and MIDI input cases, asserting no live input op is emitted and the diagnostic names the offending value.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Good catch, and thanks for the line reference. Confirmed: applyRackInstanceState (RackSyncManager.cpp:51-53) writes rack volume/pan to leftOutDb/rightOutDb unconditionally, and the return channel map carries those gains on the wet path, so a chainless non-bypassed rack does apply them.

dbd2d255 emits the RackFader over the incoming audio for the empty case and skips only the mix. Bypass stays fully transparent as you say, since wetGain 0 / dryGain 1 routes the dry path around the out gains. The test is renamed to "An empty rack passes signal through its fader" and now asserts RackFader == 1 with the track fader reading it.

The differ-identity argument is the better half of this: the fader keeping its key when the first chain is added is worth more than the op it saves.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Same thread as above, fixed in dbd2d255: both edges and the midiSourceTracks_ insertion now share the emission predicates, and the multi-out edge is gone rather than left, since it has no emission counterpart to justify it.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Noted, and added to the epic. Agreed the structure is right and the coupling belongs in the value layer: the SendTap resolved gain has to fold in track mute the same way the TrackMute op does, or a muted track keeps feeding its auxes. Captured with the shouldProcessAuxSendWhenTrackIsMuted default as the reference behaviour, plus a muted-track-with-sends case for the null-diff corpus.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,749 @@
...
+    return parseTrackRoute(track.audioInputDevice);
+}
+
+TrackRoute Compiler::activeMidiInputRoute(const TrackInfo& track) const {
+    if (!carriesClips(track) || !track.receivesLiveMidiInput() || track.midiInputDevice.isEmpty())
+        return {RouteKind::None, INVALID_TRACK_ID};
+    return parseTrackRoute(track.midiInputDevice);
+}
+
+TrackId Compiler::resolveAudioDestination(const TrackInfo& track) const {
```
lucaromagnoli, line 295: The commit message says a malformed value is diagnosed and left unconnected, but that only became true for the input fields: here a Malformed audioOutputDevice collapses into the master fallback with no diagnostic, exactly the silent reinterpretation the input side just stopped doing. Suggest diagnosing it in the output-routing section the same way the missing-destination case is handled there (diagnose, then sum into the master), keeping this function's fallback for ordering purposes.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,749 @@
...
+    // track's meter read silence and would starve any compressor keyed off it.
+
+    auto emitSends = [&](bool preFader, PortRef source) {
+        for (std::size_t slot = 0; slot < track.sends.size(); ++slot) {
+            const auto& send = track.sends[slot];
+            if (send.preFader != preFader)
+                continue;
+
+            const auto destination = resolveSendDestination(send);
+            if (destination == INVALID_TRACK_ID) {
```
lucaromagnoli, line 715: A send whose destTrackId names a track that no longer exists slips past this check: resolveSendDestination returns the id unverified, the tap is emitted into pendingInputs_, nothing ever drains it, and the end-of-run sweep reports it as a connection that arrived after its track was compiled, which claims a routing cycle that does not exist. Ordering is unaffected since indexOf misses. Suggest also requiring findTrack(destination) here and diagnosing a missing destination track, mirroring the audio-output path.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,749 @@
...
+            return consumesMidi(getDevice(element));
+        if (isRack(element))
+            return std::ranges::any_of(getRack(element).chains, [](const ChainInfo& chain) {
+                return elementsConsumeMidi(chain.elements);
+            });
+        return false;
+    });
+}
+
+bool chainConsumesMidi(const TrackInfo& track) {
```
lucaromagnoli, line 74: This still counts consumers the compiler will never emit: a bypassed instrument, an instrument behind chain power off, or one inside a bypassed rack, bypassed chain, or aux-routed chain all keep ClipMidi and TrackMidiInput ops in the plan that nothing reads. Same dead-op class the unmonitored-route fix in this commit eliminated, and the same fix applies: give elementsConsumeMidi and chainConsumesMidi the activity rules emission uses. MIDI sidechains and internal routes stay safe either way, since midiSourceTracks_ covers them independently.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,749 @@
...
+    // --- chain ---
+
+    // Chain power gates the whole insert chain without touching the devices'
+    // own bypass flags. Post-FX and mixer analysis sit outside it.
+    if (track.chain.enabled)
+        signal = emitElements(track.chain.fxChainElements, site, signal);
+
+    for (const auto& element : track.chain.postFxChainElements)
+        signal = emitDevice(element.device, site, signal);
+    for (const auto& element : track.chain.mixerAnalysisElements)
```
lucaromagnoli, line 698: Discovery and emission disagree about this section: emitDevice resolves an audio sidechain for mixer-analysis devices too, but collectSidechainSources never walks mixerAnalysisElements, so such a device would get no ordering edge and its resolution would depend on compile order. As far as I can tell no analysis device sets canSidechain today, so this is unreachable, but it is the one remaining spot where the two walks can drift apart. Either walk the section in collection or skip sidechain resolution for it in emitDevice, whichever matches where analysis devices are headed.

lucaromagnoli (commented):
Third round. Rounds one and two land fully: gated discovery, the empty-rack fader, malformed input routes, and classifying the rack-level sidechain as modulation rather than topology is a good catch beyond the review. I found no remaining parity gaps against the current engine this pass; what is left is four consistency items in the same families as the fixes this branch already made, inline. None are blockers, though the first two produce wrong or misleading diagnostics for reachable model states.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Right, the commit message claimed more than the change delivered. Fixed in 84f1793b: the output-routing section diagnoses a malformed value and then sums into the master, exactly as the missing-destination case beside it does.

    track 1: output routing track: does not name a track, summed into the master instead

`resolveAudioDestination` keeps its plain master fallback, since ordering only needs a destination and has nothing to report. Test: "A malformed output routing is reported", which also asserts the master still receives the track.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Good find, and the misleading diagnostic was the worse half of it. Fixed in 84f1793b: sends now require `findTrack(destination)` and report the real cause.

    track 1 send 0: destination track 99 does not exist, send not connected

Test asserts no SendTap is emitted, that the diagnostic names the missing track, and explicitly that it does not mention a cycle, since blaming a cycle that never existed is what made this worth fixing rather than just tidy.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Agreed, same dead-op class. Fixed in 84f1793b: `elementsConsumeMidi` skips bypassed devices, bypassed racks and inactive chains, and `chainConsumesMidi` puts the insert-chain walk behind chain power. Pulled the aux-routed and bypassed chain test into a shared `chainIsActive` helper so the MIDI walk and the sidechain walk cannot disagree about what active means.

Also extended it to the mixer-analysis section, which the old version skipped entirely.

Your note that sidechains and internal routes stay safe via `midiSourceTracks_` is now a test rather than an argument: a track whose only instrument is bypassed still compiles its MIDI clips when another track reads them.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Fixed in 84f1793b by walking the section in collection, rather than skipping resolution in emitDevice.

Skipping would have encoded a bet that no device in that section ever sidechains, which is a bet about where analysis devices are headed rather than about the compiler. Walking it makes the two traversals structurally identical, which is the invariant worth holding: collection walks exactly what emission emits. The flat-section walk is now one function shared by post-FX and mixer analysis, so a future section cannot be added to one and forgotten in the other.

The test builds the unreachable case deliberately, an analysis device with canSidechain set and its source track declared after it, so it fails without the collection fix.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,797 @@
...
+                        INVALID_DEVICE_ID, OpRole::TrackMidiInput, 0};
+        signal.midi = PortRef{addOp(OpKind::MergeMidi, key, midiSources, {SignalKind::Midi}), 0};
+        trackMidiInput_[track.id] = signal.midi;
+    }
+
+    // --- chain ---
+
+    // Chain power gates the whole insert chain without touching the devices'
+    // own bypass flags. Post-FX and mixer analysis sit outside it.
+    if (track.chain.enabled)
```
lucaromagnoli, line 693: The compiler never reads track.frozen, but freeze is live in the current engine: AudioBridge syncs it to te::AudioTrack individualFreeze, which renders the track, disables its plugins, and plays the render back. A frozen track therefore compiles here to its full live chain, which both diverges from what the engine outputs and defeats the point of freezing (the CPU win). Freeze is structural: the chain is replaced by rendered audio. Minimal handling for this slice is a diagnostic like multi-out's so the gap is named; the real shape later is ClipAudio reading the freeze file with the device ops skipped, which also needs a decision about what the freeze-file source op's key looks like so unfreezing carries state sanely.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -0,0 +1,797 @@
...
+    return found == tracks_.end() ? nullptr : &*found;
+}
+
+bool Compiler::carriesClips(const TrackInfo& track) const {
+    return track.id != master_.id && track.type != TrackType::Aux &&
+           track.type != TrackType::Group && track.type != TrackType::MultiOut;
+}
+
+TrackRoute Compiler::activeAudioInputRoute(const TrackInfo& track) const {
+    const auto monitorsInput = track.recordArmed || track.inputMonitor != InputMonitorMode::Off;
```
lucaromagnoli: Auto is over-included here: it maps to te::InputDevice::MonitorMode::automatic (TrackController.cpp:13-14), which monitors only while armed, so an unarmed track set to Auto emits a live AudioInput op the current engine keeps silent. The cost is not just a dead op: it marks the whole downstream chain Live, which pessimizes the future anticipative executor for a track that can never sound. Since recordArmed is already part of this structural gate, tightening to inputMonitor == In || recordArmed matches the engine's audible set exactly. Same applies to the MIDI gate below via receivesLiveMidiInput, which has the same Auto shape; the compiler may want its own predicate rather than the model helper if you tighten it.

lucaromagnoli (commented):
Fourth round. All four third-round items are addressed the way I would have, and building the mixer-analysis test so it fails without the collection fix is exactly right. This pass I went hunting in the model rather than the compiler, and found two things the plan does not read yet: track freeze and Auto input monitoring. One is a real structural gap, one is a gate that over-emits a Live op. After these I am out of findings.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Confirmed and fixed in db4ea002. TrackController.cpp:13-14 does map Auto to MonitorMode::automatic, so the gate is now armed or monitoring In, for audio and MIDI both.

Took your suggestion of a compiler-local predicate over the model helper. `receivesLiveMidiInput` answers a deliberately broader question for the input-activity light, and narrowing it would have changed what the light does; the compiler wants the audible set, which is a different question that happens to be close. The predicate says so in a comment so nobody merges the two later.

The liveness argument is the part worth having caught: a dead op is cheap, but poisoning the downstream chain Live for a track that cannot sound costs the anticipative executor real work. The test asserts every op stays deterministic for the unarmed Auto case, not just that the input op is absent.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Real gap, and I had not read track.frozen at all. Confirmed the sync: AudioBridge.cpp:379 sets individualFreeze from trackInfo->frozen, with the reverse sync at :1025 on load.

db4ea002 diagnoses it, matching multi-out:

    track 1: freeze is not compiled yet, the live chain is planned instead of the rendered freeze file

Deliberately not half-implemented here. Skipping the device ops without a freeze-file source would produce a third wrong answer (dry clips) rather than the engine output, and as you say the real shape needs the source op key decided first so unfreezing carries state sanely rather than looking like a born node. That belongs with the differ work, where the key question has an owner.

magda/engine/plan/RenderPlan.cpp
```diff
@@ -0,0 +1,260 @@
...
+                              INVALID_OP_ID);
+    std::vector<int> cursor(plan.consumerOffsets.begin(), plan.consumerOffsets.end() - 1);
+    for (std::size_t i = 0; i < numOps; ++i) {
+        for (const auto producer : distinctProducers(plan.ops[i]))
+            plan.consumerEdges[static_cast<std::size_t>(
+                cursor[static_cast<std::size_t>(producer)]++)] = static_cast<OpId>(i);
+    }
+}
+
+std::vector<std::string> validatePlan(const RenderPlan& plan) {
```
lucaromagnoli, line 186: validatePlan could enforce the differ's two core preconditions, cheaply, and every fixture already runs it so the whole corpus would enforce them from day one. First: OpKey uniqueness across the plan. The differ hash-joins old and new plans on the key, so a collision does not fail loudly, it carries the wrong state into the wrong op, which is exactly the concentrated bug class the epic's design notes call out. The compiler produces unique keys today by convention; the validator making it an invariant means a future emission path cannot break it silently. Second: liveness provenance. The check above catches a deterministic op reading a live one, but not the converse: a Live op that has no live inputs and is not an input-source kind. Over-tagging is semantically harmless and so would never be noticed, but it silently shrinks what the anticipative executor may precompute. Only AudioInput and MidiInput should be allowed Live without a live input.

lucaromagnoli (commented):
Fifth round. The freeze and monitoring fixes land as suggested, and splitting the compiler's monitorsInput from TrackInfo::receivesLiveMidiInput with the rationale written down is the right call. I found no further behavioral gaps in the compiler this pass; the one item left is validator hardening, inline. With that, I consider the compiler slice done from my side: parity is verified against the current engine everywhere I could check it (chain gating, rack wet/dry and output gains, mute and sidechain-send placement, post-FX ordering, sidechain tap points, freeze, monitoring), discovery and emission are mechanically in step, and every silent-drop path I could construct now lands in diagnostics.

magda/engine/plan/RenderPlan.cpp
lucaromagnoli: Both added in 90408b7c, and both pass across all 36 existing compiler fixtures unchanged, so the corpus enforces them from here rather than the compiler happening to satisfy them.

Key uniqueness is a std::map<OpKey, OpId> built during the walk, reporting the earlier owner:

    op 1 (Gain T1:clipAudio): has the same key as op 0, so the differ cannot tell them apart

Your framing of why is the reason it belongs in the validator rather than a compiler assertion: a collision does not fail, it carries the wrong state into the wrong op, and that is invisible until someone hears a stale delay line.

Liveness provenance is now checked in both directions off the same lookup, with AudioInput and MidiInput the only kinds allowed Live without reading anything live. The point about over-tagging being semantically harmless is exactly what makes it worth a check: it would never show up as a bug, only as an anticipative executor quietly doing less work than it could.

Both violations are unreachable through the compiler, so the tests build them from hand-written plans instead. That also gives the validator its first direct coverage, rather than only ever being exercised through compiled output.

Issue #1860, opened 2026-07-25
Remote API: permissions, settings, audit log, and conformance tests


lucaromagnoli:
Part of #701. Depends on #1856 and #1858.

## Summary

Harden the Remote API for distribution builds and make exposure visible and controllable from MAGDA.

## Shared scopes

- `read`
- `edit`
- `transport`
- `session`
- `hardware-midi`

Remote access is disabled by default. New clients are read-only unless the user grants more scopes. Hardware MIDI and SysEx require the hardware scope.

## Scope

- [x] Add settings UI for enablement, endpoint status, token rotation, connected clients, and disconnect
- [x] Enforce shared scopes in `RemoteApiService`
- [x] Add per-client audit entries with time, client, transport, operation, request ID, and outcome
- [x] Add payload, rate, concurrency, and queue limits
- [x] Redact secrets in logs and errors
- [x] Add cross-transport schema, error, and permission conformance tests
- [x] Test shutdown, project reload, reconnect, malformed input, and load
- [x] Add user and developer documentation, including the fields intentionally excluded from the
      remote API and why (moved from #1854, whose projections already enforce the exclusions in code)
- [ ] Verify packaging on macOS, Linux, and Windows  

## Acceptance

- The UI clearly shows enabled servers and connected clients.
- Revoking a client or rotating a token takes effect without restart.
- Read-only clients cannot mutate through either transport.
- WebSocket and MCP pass the same operation test vectors.
- Logs support diagnosis without exposing tokens, SysEx, file paths, or plugin state.
- With remote services disabled, no listener is opened and runtime overhead is negligible.

lucaromagnoli:
Shipped in #2056.

One item is deliberately left unticked: **verifying packaging on the three platforms**. CI now builds and tests macOS, Linux and Windows, so the code is covered, but the installed artifact is not — where `magda-mcp` lands, whether the settings page's composed path resolves from a real install, and whether the Clients tab renders (it has never been seen in a running app). Moved to #2059.

Also worth recording against this issue: `hardware-midi` is declared in the scope vocabulary and enforced, but no operation requires it yet — the registry exposes no hardware MIDI surface. It exists now because grants are persisted, and a scope invented later would read as "not granted" on every existing client only if the word already existed when those grants were written.

Issue #1857, opened 2026-07-25
WebSocket API: subscriptions, revisioned events, and backpressure


lucaromagnoli:
Part of #701. Depends on #1855 and #1856.

## Summary

Provide push updates without mirroring every internal notification or flooding the JUCE message queue.

## Initial topics

`project`, `tracks`, `clips`, `devices`, `selection`, `transport`, `session`, and `automation`, plus opt-in high-frequency `meters` and `playhead` topics.

## Scope

- [x] Add a transport-neutral event envelope with topic, revision, event type, and payload
- [x] Add subscribe and unsubscribe methods
- [x] Deliver an initial snapshot followed by revisioned deltas
- [x] Coalesce redundant model notifications
- [x] Add bounded per-client outgoing queues and a slow-client policy
- [x] Support reconnect/resync from a known revision
- [x] Add latest-value-wins sampling limits for meters and playhead
- [x] Add load, ordering, and resync tests

## Acceptance

- Clients observe ordered changes made by MAGDA or another client.
- Missed or unavailable revisions cause an explicit full resync.
- A stalled client cannot grow memory indefinitely or delay other clients.
- Sustained 100 Hz control changes do not enqueue 100 JUCE callbacks per second per client.
- Meter publication reads existing lock-free snapshots and performs no WebSocket work on the audio thread.

lucaromagnoli:
Shipped in #2044. Every scope item is in the tree: the four `subscriptions.*` methods in the shared registry, `isContinuousTopic` and `samplingIntervalMs` for latest-value-wins sampling of meters and playhead, `maxQueuedEventsPerConnection` and the drop-and-resnapshot policy for slow clients, `subscriptions.resync` for reconnect from a known revision, and `test_remote_subscriptions.cpp` covering load, ordering and resync.

Issue #1856, opened 2026-07-25
WebSocket API: authenticated JSON-RPC server and lifecycle


lucaromagnoli:
Part of #701. Depends on #1854 and #1855.

## Summary

Add a WebSocket JSON-RPC 2.0 adapter over `RemoteApiService`. It must not become a second implementation of DAW operations.

## Scope

- [x] Run a dependency spike: validate the vendored CHOC HTTP/WebSocket wrapper and its Boost Beast/Asio packaging, then record the decision
- [x] Bind to `127.0.0.1` by default
- [x] Implement connection lifecycle and JSON-RPC 2.0 framing
- [x] Authenticate connection setup with a MAGDA-generated bearer token
- [x] Validate allowed Origins
- [x] Route requests only through the shared operation registry
- [x] Keep parsing, framing, socket writes, and logging off the message thread
- [x] Add request-size, in-flight request, and rate limits
- [x] Start and stop with Config/MAGDA lifecycle
- [x] Add macOS, Linux, and Windows integration tests

## Acceptance

- A client can authenticate, list tracks, read transport state, and perform an allowed test mutation.
- With the service disabled, MAGDA opens no listener.
- Invalid authentication, invalid Origin, malformed JSON, oversized requests, and unknown methods fail closed.
- Disconnect and shutdown leave no blocked threads or queued writes.
- The adapter has no direct calls to `TrackManager`, `ClipManager`, or other model singletons.

lucaromagnoli:
## Dependency spike: decided, and it is not CHOC

Recorded in [`docs/remote-api-transport.md`](docs/remote-api-transport.md); wired in a6721686 on `feat/websocket-jsonrpc-server`.

**Decision: cpp-httplib, pinned at v0.52.0 via FetchContent.**

### Why not the CHOC wrapper

`choc_HTTPServer.h` accepts every upgrade unconditionally. `upgradeToWebsocket` (L465–473) starts the handshake and *then* calls `upgradedToWebSocket(target)`, which gets the request path and nothing else — no `Origin`, no `Authorization`, and no return value to refuse with. The `decorator` at L365 only stamps headers on a response already going out. `ClientInstance` has no close, so one bad client can only be dropped by tearing down the whole listener. `messageBodySizeLimit` (10 KB), `timeoutSeconds` and `defaultNumThreads` are file-scope constants at L205–207.

[...]

PR #2002, opened 2026-08-03
Remote API: message-thread dispatch, revisions, and undo transactions (closes #1855)

### LINKED ISSUE #1855: Remote API: add safe message-thread dispatch, revisions, and undo transactions

lucaromagnoli:
Implements the execution service between remote transports and `MagdaApi`. Part of #701; closes #1855.

## Dispatch

`RemoteApiService` splits each request three ways:

| Thread | Work |
|---|---|
| Calling thread | operation lookup, JSON-schema validation |
| Message thread | handler, copy results into DTOs |
| Calling thread | serialize the DTO (in the adapter) |

Validation is a pure function of the request, so a malformed or oversized payload is rejected without ever occupying the message thread — a client sending garbage costs the DAW nothing. No `MessageManagerLock` is taken and nothing touches the audio thread; the hop is always an async post.

Handlers run serially. That is a genuine throughput ceiling, inherent to the engine rather than to this design — a worker pool would only queue on the same thread. Deadlines and the bounded idempotency cache are the mitigation; per-client rate and concurrency limits are #1860.

## Handlers on the descriptor

Per the [decision on #1855](https://github.com/Conceptual-Machines/magda-core/issues/1855#issuecomment-5159402577), `OperationDescriptor` now carries its handler, so a transport cannot reach a different implementation than the one declared — there is only one place to look.

All 36 previously declared-but-unimplemented operations are implemented, and the registry constructor asserts every declared operation has one. A declared-but-unimplemented operation is now a startup failure rather than something `system.describe` advertises and no transport can execute.

`devicePathFromJson` moves out of the anonymous namespace so handlers reuse the existing decoder instead of growing a second one; automation targets rebuild through `toChainNodePath`, so a target inside a nested rack survives the round trip.

## Revisions, undo, idempotency

- Monotonic project revisions; only a committed write moves the counter, so a rejected request never invalidates another client's `expectedRevision`.
- `expectedRevision` mismatch returns `Conflict`, checked after the hop since a request can sit queued while the project moves.
- One named undo step per mutating request, via an RAII guard that survives early returns inside a handler.
- Request-id idempotency scoped per client over a bounded FIFO cache. Only successful writes are cached — caching reads would serve state that has since changed.
- New `Timeout` and `Cancelled` error codes complete the mapping the issue asks for.

## Lifetime

Queued work captures a shared cancellation token rather than `this`. After `shutdown()` or `projectReplaced()`, a lambda still in the message queue observes the retired token and completes with `Cancelled` without dereferencing the service. Project replacement also bumps the revision and clears the idempotency cache, since cached responses describe ids from a project that no longer exists.

## Change source

`ChangeSource` lands here rather than in either adapter because **revisions are the coalescing key** — collapsing means "everything since revision N", and splitting them would mean this PR ships revisions and #1857 immediately reopens that code.

#1857 (subscriptions) and #1757 §4 (OSC feedback) need the same collapse over the same model listeners but differ sharply in delivery — ordered per-client queues with resync versus stateless fire-and-forget UDP. So the source is shared and delivery stays private to each adapter. `markChanged` is lock-free, allocation-free, and safe from any thread including the audio thread.

`ModelChangeBridge` feeds the `TrackManager` / `ClipManager` / `AutomationManager` / `SelectionManager` / `ProjectManager` listeners into it. Without it a subscriber only hears about changes a remote client itself made, and the user moving a fader in the MAGDA UI stays invisible — which is the primary OSC feedback case.

## Two calls worth reviewing

**Committed change vs. continuous motion.** `noteModelChanged` bumps the revision; `noteModelActivity` notifies without bumping. Parameter values, macro sweeps, and drag previews fire throughout playback, so bumping for them would leave every client's `expectedRevision` permanently stale whenever the transport is rolling. Ambiguous callbacks — a param change that could be a user edit *or* automation — are treated as motion. The cost is that a write racing an automated parameter change is not rejected; the alternative fails every write during playback, which is worse and far more common.

**`ProjectManager` attachment is conditional on a MessageManager existing.** Its constructor starts an autosave timer, so constructing it headless asserts. A host with no event loop has no project open/close lifecycle to observe anyway, so the bridge skips it there rather than forcing the singleton into existence just to listen to it.

## Tests

45 new cases across three files: dispatch and error mapping, revisions and conflicts, undo grouping and balance, idempotency including per-client scoping and cache eviction, deadlines, shutdown and project replacement, coalescing and latest-value-wins, the model bridge, and the #701 acceptance path (create a track, add a MIDI clip, add notes, list them) through the dispatcher. Two stress cases cover concurrent dispatch from 8 threads (every committed write gets its own revision, none reused or skipped) and concurrent change marks under contention.

`StubUndoApi` now records compound descriptions and depth, which is what makes "one named undo step per mutation" assertable.

Full suite green: 1834 cases, 164,724 assertions.

## Not in this PR

- Nothing instantiates `RemoteApiService` yet — that is #1856 / #1858, matching how the epic is decomposed. This is the library layer they plug into.
- `Transport`, `Meters`, and `Playhead` topics exist but are not fed by model listeners; there is no listener interface for them. That is #1857's opt-in high-frequency work reading the existing lock-free snapshots.


lucaromagnoli (commented):
Inline correctness findings from reviewing the dispatcher, live model bridge, and handlers. The focused `[remote]` suite passes, but the cases below are not covered by the current mocks/tests.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    if (operation.handler == nullptr)
+        return Response::failure(ErrorCode::InternalError,
+                                 "operation " + operation.name + " has no handler", revision);
+
+    const bool isWrite = operation.access == OperationAccess::Write;
+    HandlerResult result;
+    if (isWrite) {
+        const ScopedUndoStep step(api_.undo(), operation.summary);
```
lucaromagnoli: **[P1] This does not make most live mutations undoable.** `UndoManager::endCompoundOperation()` only records a compound when at least one `UndoableCommand` was enqueued, while handlers such as `tracks.create` call the live facade and then `TrackManager::createTrack` directly. The real undo stack therefore remains empty even though `StubUndoApi` records this begin/end pair. Please route these writes through actual commands and add a live test that verifies `canUndo()` and restored state.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,359 @@
...
+        onComplete(execute(*operation, input, context));
+    };
+
+    // Already on the message thread, or there is no message thread to hop to.
+    // The latter is the headless case — a test runner or a console host with no
+    // event loop — where posting would queue work that nothing ever dispatches.
+    // Running inline there is the correct behaviour, not a test accommodation:
+    // the caller *is* the only thread that touches the model.
+    if (juce::MessageManager::existsAndIsCurrentThread() ||
+        juce::MessageManager::getInstanceWithoutCreating() == nullptr)
```
lucaromagnoli: **[P1] The headless path breaks the message-thread serialization guarantee.** `dispatch()` is documented as callable from any thread, so with no `MessageManager` multiple workers run handlers inline and concurrently against message-thread-only model objects. The new stress test already does this and races on `MockProjectApi::info`. Headless mode needs a serialized executor or an explicitly enforced single owner thread.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,359 @@
...
+        return;
+    }
+
+    std::shared_ptr<std::atomic<bool>> token;
+    {
+        const std::lock_guard<std::mutex> lock(tokenMutex_);
+        token = liveToken_;
+    }
+
+    auto job = [this, operation, input, context, token,
```
lucaromagnoli: **[P1] The cancellation token does not make this capture lifetime-safe.** The lambda still holds raw `this`; another thread can destroy the service after the token is observed as live but before/during `execute()`, yielding a use-after-free. Please capture shared execution state, or constrain and synchronize shutdown/destruction with all running jobs.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,359 @@
...
+    if (auto error = validateOperationInput(*operation, input)) {
+        onComplete(Response::failure(*error, revision));
+        return;
+    }
+
+    // Replay a completed retry before queuing, so a client that lost the
+    // response to a network fault does not re-apply the mutation.
+    const auto key = idempotencyKey(context);
+    if (operation->access == OperationAccess::Write && key.isNotEmpty()) {
+        if (auto cached = cachedResponse(key)) {
```
lucaromagnoli, line 183: **[P1] Concurrent retries can both miss this cache and apply the mutation twice.** Both requests can pass this pre-queue lookup before either job finishes; `execute()` never rechecks or reserves the key. With `expectedRevision` the second request conflicts instead of replaying the first response. Please reserve/coalesce in-flight keys or recheck on the serialized execution path before the revision check.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,359 @@
...
+    // Already on the message thread, or there is no message thread to hop to.
+    // The latter is the headless case — a test runner or a console host with no
+    // event loop — where posting would queue work that nothing ever dispatches.
+    // Running inline there is the correct behaviour, not a test accommodation:
+    // the caller *is* the only thread that touches the model.
+    if (juce::MessageManager::existsAndIsCurrentThread() ||
+        juce::MessageManager::getInstanceWithoutCreating() == nullptr)
+        job();
+    else
+        juce::MessageManager::callAsync(std::move(job));
```
lucaromagnoli: **[P2] `callAsync()` can return false, which silently drops `onComplete`.** JUCE returns false once the message loop is quitting or posting fails; the moved job is then destroyed and the documented exactly-once callback guarantee is violated. Please handle the return value and complete the request with `Cancelled`/`InternalError`.

magda/daw/api/remote_model_bridge_live.cpp
```diff
@@ -0,0 +1,202 @@
...
+        SelectionManager::getInstance().removeListener(this);
+        AutomationManager::getInstance().removeListener(this);
+        ClipManager::getInstance().removeListener(this);
+        TrackManager::getInstance().removeListener(this);
+    }
+
+    // ---- TrackManagerListener -------------------------------------------
+
+    void tracksChanged() override {
+        service_.noteModelChanged(Topic::Tracks);
```
lucaromagnoli, line 64: **[P2] This also observes mutations initiated by `RemoteApiService`.** For example, live `tracks.create` synchronously fires `tracksChanged()`, incrementing the revision here, and `execute()` increments it again after the handler succeeds. Multi-field writes can advance it several times. Please suppress bridge revision bumps for service-originated callbacks and let the service publish one revision for the completed request.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,548 @@
...
+}
+
+HandlerResult selectionSet(MagdaApi& api, const juce::var& input, const RequestContext&) {
+    Error error;
+    const auto dto = selectionFromJson(input, error);
+    if (!dto)
+        return HandlerResult::fail(error);
+
+    auto& selection = api.selection();
+    if (dto->trackId)
```
lucaromagnoli, line 540: **[P2] These selection IDs are not checked against the model.** A non-negative but nonexistent track/clip/lane/note target is passed directly to `SelectionManager`, which can accept it and return success. This violates the contract that invalid IDs produce structured errors; please validate each referenced object and the lane/clip/note relationships before mutating selection.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,548 @@
...
+        selection.selectAutomationClip(*dto->automationClipId, *dto->automationLaneId);
+
+    if (dto->noteClipId && !dto->noteIndices.empty()) {
+        std::vector<size_t> indices;
+        indices.reserve(dto->noteIndices.size());
+        for (const auto index : dto->noteIndices)
+            indices.push_back(static_cast<size_t>(index));
+        selection.selectNotes(*dto->noteClipId, indices);
+    } else {
+        selection.clearNoteSelection();
```
lucaromagnoli, line 563: **[P2] An empty selection DTO does not clear the current selection.** `clearNoteSelection()` is a no-op unless the current mode is Note, so an existing track, clip, or automation selection remains even though this operation is described as replacing the selection. The facade needs an overall `clearSelection()` path for the all-null/all-empty DTO.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,548 @@
...
+HandlerResult automationAddPoint(MagdaApi& api, const juce::var& input, const RequestContext&) {
+    const auto laneId = static_cast<AutomationLaneId>(static_cast<int>(input["laneId"]));
+    if (api.automation().getLane(laneId) == nullptr)
+        return notFound("automation lane", laneId);
+    const auto curve = parseCurve(input["curve"].toString());
+    if (!curve)
+        return HandlerResult::fail(ErrorCode::ValidationFailed,
+                                   "unsupported curve: " + input["curve"].toString());
+
+    api.automation().addPoint(laneId, static_cast<double>(input["beatPosition"]),
```
lucaromagnoli: **[P2] The return value must be checked.** The live automation API returns `INVALID_AUTOMATION_POINT_ID` for a clip-based lane, but this handler still returns success and the service advances the revision even though no point was added. Please return a structured conflict/validation error when `addPoint()` fails.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,548 @@
...
+    if (lane == nullptr)
+        return notFound("automation lane", laneId);
+    return HandlerResult::ok(toJson(makeAutomationLaneDto(*lane)));
+}
+
+HandlerResult automationClearLane(MagdaApi& api, const juce::var& input, const RequestContext&) {
+    const auto laneId = static_cast<AutomationLaneId>(static_cast<int>(input["laneId"]));
+    if (api.automation().getLane(laneId) == nullptr)
+        return notFound("automation lane", laneId);
+    api.automation().clearLanePoints(laneId);
```
lucaromagnoli: **[P2] This can also report a committed write for a no-op.** `clearLanePoints()` does nothing for clip-based lanes (and for an already-empty absolute lane), but the handler returns success and the dispatcher advances the revision. The API needs to report whether a mutation occurred, or the handler must reject unsupported lane types before calling it.

lucaromagnoli (commented):
Second review round on `c7ce95b`. The corrective commit addresses the original comments, but the service still has re-entrancy/idempotency issues, undo remains incomplete for several project mutations, and the new live test currently passes while emitting JUCE assertions/leak diagnostics.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,454 @@
...
+        // lock is held across the whole call, so the service cannot be retired
+        // out from under a running handler, and two handlers cannot run at
+        // once even when the caller's thread executes them inline.
+        const std::lock_guard<std::mutex> lock(state->mutex);
+        if (state->service == nullptr) {
+            onComplete(Response::failure(ErrorCode::Cancelled, "request cancelled before execution",
+                                         INITIAL_REVISION));
+            return;
+        }
+        onComplete(state->service->execute(*operation, input, context));
```
lucaromagnoli: **[P1] Invoke the completion after releasing `state->mutex`.** The callback currently runs while the execution gate is held. A valid callback that calls `shutdown()`, destroys the service, calls `projectReplaced()`, or synchronously dispatches the next request will try to acquire this same non-recursive mutex and deadlock. Compute/store the `Response` under the lock, release it, then invoke `onComplete`.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,454 @@
...
+    // nothing would invalidate every other client's expectedRevision.
+    const bool committed = isWrite && result.mutated;
+    if (committed) {
+        revision = revision_.fetch_add(1, std::memory_order_acq_rel) + 1;
+        for (const auto topic : topicsFor(operation.name))
+            changes_.markChanged(topic, revision);
+    }
+
+    auto response = Response::success(result.value, revision);
+    if (committed && key.isNotEmpty())
```
lucaromagnoli: **[P1] Successful no-op writes must still be cached for idempotency.** `HandlerResult::unchanged` returns success but `committed` is false, so its request ID is forgotten. For example, clearing an already-empty lane succeeds; if points are later added, retrying that same request ID will execute again and delete them instead of replaying the original response. Cache every successful write response; only revision/change publication should depend on `mutated`.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,689 @@
...
+// ===========================================================================
+// Project
+// ===========================================================================
+
+HandlerResult projectGet(MagdaApi& api, const juce::var&, const RequestContext&) {
+    return HandlerResult::ok(toJson(makeProjectDto(api.project().getCurrentProjectInfo())));
+}
+
+HandlerResult projectSetTempo(MagdaApi& api, const juce::var& input, const RequestContext&) {
+    api.project().setTempo(static_cast<double>(input["tempo"]));
```
lucaromagnoli, line 219: **[P1] The undo fix is still partial.** `project.setTempo` and `project.setTimeSignature` still mutate the live facade directly, as do rack create/remove/bypass and `automation.createLane`. Their compound scopes therefore close empty, so one Undo cannot reverse these project mutations. Please make every project-mutating handler command-backed and add live undo coverage beyond `tracks.create`.

tests/CMakeLists.txt
```diff
@@ -192,6 +192,10 @@ set(TEST_SOURCES
     # Lua bindings tests (issue #30 — bind MagdaApi)
     test_magda_api_lua_bindings.cpp
     test_remote_api_contract.cpp
+    test_remote_service.cpp
+    test_remote_service_live.cpp
```
lucaromagnoli: **[P2] This live test cannot run cleanly in the non-JUCE test target.** Command execution creates `ProjectManager` through `UndoableMutationScope`, and its constructor starts a JUCE timer. Running `magda_tests '[remote][live]'` now reports timer assertions, a leaked `ShutdownDetector`, and singleton assertions even though Catch says all 9 cases pass. Move this coverage to the JUCE-initialized target or provide correct message-system/singleton setup and teardown.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,454 @@
...
+        if (auto cached = cachedResponse(key))
+            return *cached;
+    }
+
+    // Re-checked after the hop, not only before it: a request can sit in the
+    // queue behind slower work and expire while waiting.
+    if (deadlinePassed(context))
+        return Response::failure(ErrorCode::Timeout, "deadline passed before execution", revision);
+
+    if (context.expectedRevision && *context.expectedRevision != revision) {
```
lucaromagnoli: **[P2] `expectedRevision` is documented to reject stale writes, but this also rejects reads.** A client that includes its revision cursor on `project.get` or `tracks.list` gets `Conflict` after any intervening edit, even though the read is safe and should return current state. Gate this check with `isWrite`.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+    auto job = [state, operation, input, context, onComplete = std::move(onComplete)]() mutable {
+        // The shared state, never a bare `this`, is what makes this safe. The
+        // lock is held across the whole call, so the service cannot be retired
+        // out from under a running handler, and two handlers cannot run at
+        // once even when the caller's thread executes them inline.
+        const std::lock_guard<std::mutex> lock(state->mutex);
+        if (state->service == nullptr) {
+            onComplete(Response::failure(ErrorCode::Cancelled, "request cancelled before execution",
+                                         INITIAL_REVISION));
```
lucaromagnoli: **[P2] A queued cancellation should not reset the response revision to zero.** After shutdown or project replacement at revision N, this emits a failure envelope with revision 0, contradicting `Response`'s contract and potentially moving a client's cursor backwards. Store the cancellation/current revision in `ExecutionState` when retiring it and return that value here.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,689 @@
...
+
+    if (dto->trackId)
+        selection.selectTrack(*dto->trackId);
+
+    if (!dto->clipIds.empty())
+        selection.selectClips({dto->clipIds.begin(), dto->clipIds.end()});
+    else if (dto->clipId)
+        selection.selectClip(*dto->clipId);
+
+    if (dto->automationClipId && dto->automationLaneId)
```
lucaromagnoli, line 551: **[P2] A lane-only automation selection is accepted but ignored.** `SelectionDto` and `makeSelectionDto` support `automationLaneId` without a clip, yet this handler only writes automation selection when both IDs are present. Consequently a `selection.get` -> `selection.set` round trip for a selected lane reports success without restoring it. Add `selectAutomationLane` to the facade or reject this shape in the schema.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,689 @@
...
+        return notFound("automation lane", laneId);
+    return HandlerResult::ok(toJson(makeAutomationLaneDto(*lane)));
+}
+
+HandlerResult automationCreateLane(MagdaApi& api, const juce::var& input, const RequestContext&) {
+    const auto type = parseLaneType(input["type"].toString());
+    if (!type)
+        return HandlerResult::fail(ErrorCode::ValidationFailed,
+                                   "unsupported lane type: " + input["type"].toString());
+    const auto target = toAutomationTarget(input["target"]);
```
lucaromagnoli, line 662: **[P2] This only validates path syntax, not whether the target is addressable.** `toAutomationTarget` does not call `ControlTarget::isValid()` or resolve the referenced track/device/parameter. `AutomationManager::createLane` will therefore create lanes for nonexistent tracks/devices or kind-specific fields left at `-1`, seeding them with fallback values. Validate the target's canonical shape and resolve it against the model before creating the lane.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,689 @@
...
+        runCommand<SetTrackPanCommand>(api, trackId, static_cast<float>(readDouble(input, "pan")));
+    if (has(input, "muted"))
+        runCommand<SetTrackMuteCommand>(api, trackId, readBool(input, "muted"));
+    if (has(input, "soloed"))
+        runCommand<SetTrackSoloCommand>(api, trackId, readBool(input, "soloed"));
+
+    const auto* updated = tracks.getTrack(trackId);
+    if (updated == nullptr)
+        return notFound("track", trackId);
+    return HandlerResult::ok(toJson(makeTrackDto(*updated)));
```
lucaromagnoli: **[P2] A valid empty patch is still reported as a mutation.** The schema requires only `trackId`, so `tracks.update({trackId})` enqueues no command but returns `HandlerResult::ok`, advancing the revision despite changing nothing. Track whether any field actually changed and return `unchanged` when none did (including values already equal to the requested values).

magda/daw/api/remote_model_bridge_live.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    // ---- ProjectManagerListener -----------------------------------------
+
+    void projectOpened(const ProjectInfo&) override {
+        // Everything queued against the outgoing project is now addressed at
+        // state that no longer exists. This is the trigger the service's
+        // cancellation path exists for.
+        service_.projectReplaced();
+        service_.noteModelChanged(Topic::Project);
```
lucaromagnoli: **[P2] Opening or closing a project advances the revision twice.** `projectReplaced()` already increments it to invalidate the outgoing project, then `noteModelChanged()` increments it again just to publish the topic. Have replacement publish `Topic::Project` at its new revision, or mark activity here without another bump.

PR #2006, opened 2026-08-04
i18n: sync translations from Crowdin

lucaromagnoli:
Automated translation sync from Crowdin.


lang/ja.json
```diff
@@ -389,6 +389,7 @@
     "toggle.open_plugin_window_on_drop": "プラグインをチェーンに追加したらウィンドウを開く",
     "toggle.chord_preview_default": "再生時にデフォルトでコードトラックのプレビューを有効にする",
     "toggle.auto_crossfade_default": "新しいオーディオクリップを自動クロスフェード",
+    "toggle.clip_overlap_plays_both": "",
```
lucaromagnoli, line 392: The new key comes down empty in all three locales. Safe to merge as is: StringTable::get treats an empty locale value as untranslated and falls back to English, so the preferences toggle shows the English label in ja/ru/zh until the translations are filled in on Crowdin.

lucaromagnoli (commented):
No deletions anywhere, so this sync is safe from the branch-trampling failure mode. One note inline.

PR #2007, opened 2026-08-04
Feat/1889 plan executor

lucaromagnoli:


magda/engine/exec/PlanExecutor.cpp
```diff
@@ -0,0 +1,373 @@
...
+    }
+    portOffsets_[numOps] = static_cast<int>(portSlots_.size());
+
+    audioSlots_.resize(static_cast<std::size_t>(numAudioSlots));
+    for (auto& buffer : audioSlots_)
+        buffer.setSize(context_.numChannels, context_.maxBlockSize, false, true, false);
+
+    midiSlots_.resize(static_cast<std::size_t>(numMidiSlots));
+    for (auto& buffer : midiSlots_)
+        buffer.ensureSize(kMidiPortCapacityBytes);
```
lucaromagnoli: `ensureSize` only reserves a minimum; it does not impose a capacity. `MidiBuffer::addEvent` grows its backing array when that reservation is exceeded, and fan-in makes this possible even if every individual producer stays below 8192 bytes (for example, merging two full inputs needs roughly twice this space). Sources and devices can also emit more directly. That makes `render`, `addEvents`, or device processing allocate on the callback, violating the executor’s hard RT contract. Please use storage with an enforced bound/drop policy, or precompute a safe per-port bound through the MIDI DAG rather than treating this reservation as allocation-proof.

magda/engine/exec/PlanValues.cpp
```diff
@@ -0,0 +1,372 @@
...
+        report(id, "no track " + std::to_string(key.trackId) + " in the model, rendering silence");
+        value.silent = true;
+        return;
+    }
+
+    // A rack chain that mute or a sibling's solo has taken out of the mix
+    // contributes neither audio nor MIDI, so its ops go silent rather than
+    // taking a zero gain: the current engine simply does not connect the chain,
+    // and a gain cannot silence the MIDI a chain generates.
+    if (key.chainId != INVALID_CHAIN_ID) {
```
lucaromagnoli, line 306: This still does not silence ops nested beneath an inactive outer chain: their keys carry the inner rack/chain IDs, so this lookup sees the active inner chain. Routing both signals through the outer fader fixes audible output, but the nested devices still process, advance state/tails, and publish device-meter levels where the current engine leaves the outer chain disconnected. That also contradicts this layer’s stated “its ops go silent” behavior. The plan/value metadata needs enough ancestry (or a propagated activity flag) to gate nested ops too.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -0,0 +1,373 @@
...
+    auto block = requestedBlock;
+    block.numSamples = std::min(block.numSamples, context_.maxBlockSize);
+    const auto numSamples = block.numSamples;
+    if (numSamples <= 0)
+        return;
+
+    // Values are resolved against a plan; one resolved against a different plan
+    // would apply another op's gain to this one, so unity is the safe reading.
+    static constexpr OpValue kUnity;
+    const auto haveValues = values.ops.size() == plan_->ops.size();
```
lucaromagnoli: This only proves that the value table has the same length, not that it was resolved for this plan. A structural swap can easily replace/reorder ops while preserving the op count; the old snapshot is then accepted and its gains/mute flags are applied by index to unrelated ops. Since plan/value publication is explicitly separate, `PlanValues` needs a plan generation/fingerprint (or equivalent identity) set by `resolvePlanValues` and checked here before any values are used.

lucaromagnoli (commented):
Reviewed with the parity claims checked against source rather than taken on trust. All of these verify correct: the fader math composes to exactly 10^(dB/20) with the position clamp supplying the +6 dB ceiling, matching the engine's setVolumeDb(gainToDecibels(volume)) round trip; send levels match the same round trip via PluginManagerSync; the rack output law matches outputDbForVolumePan operation for operation including both clamps; the pan deadband matches TE's VolumeAndPanPlugin::setPan; and sends-not-coupled-to-mute is right per AuxSendNode with the default shouldProcessAuxSendWhenTrackIsMuted and MAGDA processing muted tracks, which correctly overturns the resolution I recorded on the epic earlier. The chain-fader-as-MIDI-exit change is the right fix for the nested-rack gap.

The audio-thread discipline is right throughout: no allocation, no locks, no hashing in process(), malformed plans refused, unity fallback for mismatched values. Tests pin every formula and the block-size invariance and refused-plan paths are exactly the right invariants for a reference executor.

Three findings inline. Fix the solo one before merge since soloing a group is an everyday action; the other two are small resolver changes with a test each. One nit on block truncation.

magda/engine/exec/PlanValues.cpp
```diff
@@ -0,0 +1,372 @@
...
+    // routed into a muted track is muted too.
+    if (const auto route = parseTrackRoute(track.audioOutputDevice); route.namesTrack())
+        if (const auto* destination = findTrack(route.trackId))
+            if (isMutedIncludingParents(*destination))
+                return true;
+
+    return false;
+}
+
+bool Resolver::isAudible(const TrackInfo& track) const {
```
lucaromagnoli, line 275: Solo does not propagate through the routing graph, so soloing a group renders silence. TE's AudioTrack::isSolo(true) includes indirect solo: parent folders, and recursively the destination track, so a track routed into a soloed destination is soloed, which is what keeps a soloed group's children audible today (children route into the group track). This mirrors the mute half of that pair (isInheritedMute follows the output route) but not the solo half: with a group soloed, anySolo_ is true, the children are not soloed, their TrackMute ops zero, and the group sums silence. Needs an isInheritedSolo walking the same parentId and output-route edges before the otherSoloed check. The solo test covers flat tracks only, so the suite cannot see this; a group-solo case should land with the fix.

magda/engine/exec/PlanValues.cpp
```diff
@@ -0,0 +1,372 @@
...
+        case OpRole::RackChainFader: {
+            const auto* rack = findRack(*track, key.rackId);
+            const auto* chain = rack == nullptr ? nullptr : findChain(*rack, key.chainId);
+            if (chain == nullptr) {
+                report(id, "no chain " + std::to_string(key.chainId) + " in rack " +
+                               std::to_string(key.rackId) + ", rendering silence");
+                value.silent = true;
+                break;
+            }
+            const auto gain = faderGainFromDecibels(chain->volume);
```
lucaromagnoli, line 383: A bypassed chain applies its volume and pan here, but the current engine skips them: the bypassed-chain passthrough wires the rack input pins straight to the output pins (RackSyncManager.cpp:842-847), and the chain's VolumeAndPanPlugin, appended at the end of the chain's plugin sequence (RackSyncManager.cpp:874-879), is bypassed with everything else. The compiler still emits the RackChainFader over the passthrough, which is right for identity, so the cheap fix is here: leave the fader at unity when chain->bypassed. Worth a test asserting a bypassed chain with volume set passes at unity.

magda/engine/exec/PlanValues.cpp
```diff
@@ -0,0 +1,372 @@
...
+}
+
+bool Resolver::isInheritedMute(const TrackInfo& track) const {
+    if (const auto* parent = findTrack(track.parentId))
+        if (isMutedIncludingParents(*parent))
+            return true;
+
+    // Muting a track mutes everything feeding it, so a track whose output is
+    // routed into a muted track is muted too.
+    if (const auto route = parseTrackRoute(track.audioOutputDevice); route.namesTrack())
```
lucaromagnoli: This follows the output route one hop only. TE's isMuted(true) recurses through destinations, so in a chain where A routes into B, B routes into C, and C is muted, A is muted today; here the walk checks the destination's parent mutes but never the destination's own route. Group hierarchies happen to be covered because children carry both parentId and the route, but an explicit multi-hop routing chain misses. Making the route walk recursive under the same kMaxTrackWalk cap closes it.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+void PlanExecutor::process(const PlanValues& values, const BlockInfo& requestedBlock,
+                           juce::AudioBuffer<float>& output) {
+    output.clear();
+    if (plan_ == nullptr)
+        return;
+
+    auto block = requestedBlock;
+    block.numSamples = std::min(block.numSamples, context_.maxBlockSize);
```
lucaromagnoli, line 261: Nit: a block larger than maxBlockSize is silently truncated, leaving the tail of the host buffer at silence with nothing anywhere saying it happened. A jassert here catches host miswiring in debug at zero audio-thread cost. Related nit for later, not this PR: the 8 KB MIDI reservation can still allocate under pathological event density; the comment already scopes it to this executor, just keep the parallel executor from inheriting the pattern.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -0,0 +1,417 @@
...
+#include <set>
+
+namespace magda::engine {
+namespace {
+
+/// Bytes MidiBuffer spends on one event: a sample position, a length, and the
+/// message itself. Sized for the longest short message rather than the
+/// average, so a reservation computed from an event count always holds.
+constexpr int kMidiBytesPerEvent =
+    static_cast<int>(sizeof(std::int32_t) + sizeof(std::uint16_t)) + 3;
```
lucaromagnoli, line 13: This still does not make the reservation allocation-proof: the producer contract limits the number of events, but not their byte lengths. `juce::MidiBuffer` accepts SysEx/variable-length messages, so even one contract-compliant event can exceed the assumed 3-byte payload and grow the buffer on the callback; the post-write event-count assert will still pass. The bound/contract needs to be in encoded bytes (including a maximum message size or explicit SysEx handling), or writes must go through a fixed-capacity sink that can reject/drop before `MidiBuffer` reallocates.

lucaromagnoli (commented):
Second round, on the fix commit. All six fixes verified against the TE source rather than the commit text: the solo walk mirrors AudioTrack::isSolo(true) exactly (ancestors by own flag only, destination asked the whole question, recursively), the mute walk takes isMuted(true)'s parent-before-route precedence, the silenced-racks collection closes the only path the chain-ID check could not reach, the bypassed-chain unity matches the passthrough wiring, the fingerprint covers exactly the fields value application depends on (and identical-structure plans hashing the same is the right call), and the MIDI bound is a correct forward pass over a dependency-ordered graph with the producer cap asserted at every write site. Each fix has a test that pins it.

One small note inline; with that, nothing further from me on this PR.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -0,0 +1,417 @@
...
+#include <algorithm>
+#include <set>
+
+namespace magda::engine {
+namespace {
+
+/// Bytes MidiBuffer spends on one event: a sample position, a length, and the
+/// message itself. Sized for the longest short message rather than the
+/// average, so a reservation computed from an event count always holds.
+constexpr int kMidiBytesPerEvent =
```
lucaromagnoli, line 12: The 9-byte figure holds for short messages, but the debug assert bounds event count, not bytes, so a SysEx event blows the estimate without tripping anything: a hardware controller dumping SysEx through a live MIDI input is the realistic path to an allocation this layer thinks it has excluded. Within this executor's documented lifespan that is acceptable; just make the producer contract on EngineMidiSource/DeviceBlock say short messages only, or extend the assert to bound bytes as well, so a source implementor learns the rule before the callback does.

PR #2008, opened 2026-08-04
Native engine: runtime-state ownership and epoch retirement (#1889)

lucaromagnoli:
Slice 3. Until now the executor held a raw pointer to a plan the caller owned, and the caller owned every device and source alongside it. That works for one plan and falls apart at the first swap, which is the thing this slice is about: a published plan is immutable, so a structural change compiles a new one and swaps it in whole, and the question is who owns what across that seam and where the old epoch's destructors run.

exec/RuntimeStateStore  owns devices and sources, keyed by model ID the way
                        OpKey is. A recompile asks for the same IDs and gets
                        the same instances back, so reordering a chain does
                        not rebuild the instrument in it.
exec/EngineSession      publishes plans and values, renders on the audio
                        thread, and retires the previous epoch.

The lifetime rule is that the audio thread never waits and never destroys. publish() prepares off-thread, swaps, and blocks the publishing thread until the audio thread is out of the block it was in; the retired epoch is handed back there and destroyed there, and only then does the store drop what the new plan no longer names. A plan that fails to prepare is not published at all and the live one keeps playing.

Values travel separately, since they change at mixer speed rather than structural speed. A swap therefore invalidates the values in flight, which the fingerprint check already catches: the executor falls back to unity until the caller publishes values for the new plan.

farbot (MIT) is vendored under third_party/ and carries the swap. Standing rule for the engine holds: nothing real-time-concurrent is hand-rolled. Vendored code is excluded from the formatting hooks so a pinned copy stays byte-identical to what was published.

Verified under ThreadSanitizer with a standalone harness (500 republishes against a continuously rendering thread, 4912 blocks, no reports), and the harness was checked to have teeth by planting a deliberate race and watching TSAN find it. Seven session tests cover the swap, carry-over, retirement on the publishing thread, refused plans and stale values.

Engine suite 71 cases / 430 assertions; full suite 1915 passed, 0 failed.


magda/engine/exec/EngineSession.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    auto prepared = std::make_shared<PreparedRender>();
+    prepared->plan = std::move(plan);
+
+    // Runtime state is realised before the swap and retired after it. Creating
+    // early costs nothing if the plan turns out not to prepare; destroying
+    // early would free something the audio thread is still rendering through.
+    const auto bindings = store_.realise(*prepared->plan);
+    auto messages = prepared->executor.prepare(*prepared->plan, bindings, context);
+    if (!prepared->executor.isPrepared())
```
lucaromagnoli, line 19: A failed prepare leaves every object created by `store_.realise()` resident in the store, even though this plan is never published. Repeated rejected plans with new IDs can therefore instantiate and retain an unbounded number of plugins/sources until some later successful publish happens to retire them (and if no publish succeeds, until session destruction). Please roll back objects introduced for the rejected plan, while preserving anything still named by the current live plan; a regression test should also cover the initial-publish failure case.

magda/engine/exec/RuntimeStateStore.hpp
```diff
@@ -0,0 +1,91 @@
...
+    PlanBindings realise(const RenderPlan& plan);
+
+    /**
+     * @brief Drop everything @p live does not name, and say how much went.
+     *
+     * Call only after the swap has published @p live. Before that, the objects
+     * being dropped are still reachable from the plan on the audio thread, and
+     * this is where their destructors run.
+     */
+    std::size_t retireUnused(const RenderPlan& live);
```
lucaromagnoli: This implements the eviction policy the epic already amended away. Dropping everything the live plan does not name means: chain power off, publish, and every plugin instance on that track is destroyed, because the compiler removes those ops from the plan; power back on and the factory builds fresh instances, losing external plugin state, tails, and load time for an everyday reversible gesture. Same for a single device bypass, and arming off destroys the track's input source. The resolution recorded on the epic after the plan-IR review was retain across bypass and power toggles, release only on model deletion. The fix fits this slice's shape: take the model's live ID sets (every device and track ID that still exists in the model) instead of deriving the named set from the plan, so plan-named means active, model-named-but-unplanned means dormant and retained, and absent from the model means destroyed. The lifetime rule is untouched. A test toggling bypass across two publishes and asserting the instance survives would pin it.

magda/engine/exec/EngineSession.hpp
```diff
@@ -0,0 +1,85 @@
...
+ * The rule the swap exists to keep is that the audio thread never waits and
+ * never destroys. Publishing blocks the thread doing the publishing until the
+ * audio thread has finished the block it is in, and everything the retired
+ * epoch was the last owner of is destroyed there, on that thread. Nothing that
+ * runs on the callback allocates, locks, or frees.
+ */
+
+namespace magda::engine {
+
+class EngineSession {
```
lucaromagnoli, line 34: The single-publisher contract is real but unwritten: farbot's nonRealtimeMutatable requires external synchronization between non-realtime mutators, and livePlan_ and the store are touched unsynchronized in publish(). The intended discipline is clearly one publishing thread; a sentence here saying publish, publishValues, and the accessors share a single non-realtime thread makes misuse a read-the-doc failure instead of a TSAN surprise.

magda/engine/exec/EngineSession.hpp
```diff
@@ -0,0 +1,85 @@
...
+     * PlanExecutor::prepare does. A plan that does not prepare is not
+     * published at all and the previous one keeps playing, which is the only
+     * safe reading of a plan the executor has refused.
+     *
+     * Values are published separately, and a plan swap invalidates the ones in
+     * flight: they were resolved against the old plan, so the executor falls
+     * back to unity until publishValues() catches up. Publish the plan first,
+     * then its values.
+     */
+    std::vector<std::string> publish(std::shared_ptr<const RenderPlan> plan,
```
lucaromagnoli: A refused publish and a published-with-warnings both come back as non-empty messages; the caller can only tell them apart by noticing livePlan() did not change. Either document that here or return the distinction explicitly (a published flag alongside the messages).

lucaromagnoli (commented):
The concurrency and lifetime core is right and well argued: nonRealtimeReplace blocks the publisher until the callback is out of its block, the retired epoch's destructors run on the publishing thread, retireUnused runs only after the swap with the reachability argument written where it matters, and the two-object publish is safe precisely because the fingerprint check from the last slice degrades a torn acquire pair to unity rather than misapplied gains. Vendoring farbot instead of hand-rolling the swap, pinning it with the pre-commit exclusion, and stating the standing rule in the CMake comment is exactly right. realise being insert-only keeps object addresses stable under the node-based maps while the audio thread renders through them.

One real finding on the retirement policy, two documentation-sized ones, inline. Also worth considering: commit the TSAN harness (behind a tag or as an example target) so the 500-republish verification stays runnable, even with the swap-under-load test in the suite. And one observation needing no action: a failed prepare leaves its freshly realised objects in the store until the next successful publish, which is the right call for retry, and stops being dead weight entirely once retirement is model-keyed.

magda/engine/exec/EngineSession.cpp
```diff
@@ -0,0 +1,51 @@
...
+    // The swap. This blocks until the audio thread is out of the block it was
+    // in, then hands the previous epoch back here, where its destructor runs.
+    published_.nonRealtimeReplace(std::move(prepared));
+
+    livePlan_ = published_.nonRealtimeAcquire()->plan;
+    published_.nonRealtimeRelease();
+
+    // Safe only now: before the swap, everything about to be destroyed was
+    // still reachable from the plan the audio thread was rendering.
+    store_.releaseDeleted(modelIds);
```
lucaromagnoli: The doc contract (pass IDs from the same model the plan was compiled from) is the only thing standing between this line and an audio-thread use-after-free, and the epic's own compile pipeline is designed to break that contract: gesture-storm coalescing means a plan compiled from model revision N can be published while the model is at N+2, and the caller will naturally collect IDs from the latest model. Delete a device in N+1 and this call destroys its instance while the plan just swapped in, which still names it, renders through the dangling pointer until the next plan lands. The fix is structural rather than contractual: the keep set should be modelIds unioned with what livePlan_ names, so nothing reachable from the live plan can ever be destroyed regardless of what the caller passes. One walk of the plan's ops at publish time, and the invariant stops depending on a sentence in a header. Test shape: publish a plan naming device X while passing modelIds without X, assert the instance survives, then publish a plan without X and assert it goes.

lucaromagnoli (commented):
Second round, on the fix commit. Model-keyed eviction is implemented the way the epic resolution asks (every section walked, nested racks included, bypass never consulted), the Result flag and the threading contract read right, the refused-first-publish and bypass-survival tests pin the behavior, and moving the TSAN harness into the tree with the invocation in the header and the planted-race sanity note is exactly what I hoped for. One new finding the model-keyed change opens up, inline; it is the last thing I have on this PR.

magda/engine/exec/EngineSession.cpp
lucaromagnoli: This makes audio-thread safety depend on `modelIds` perfectly matching the plan. If the caller passes a stale/incomplete set, `releaseDeleted()` can erase a device/source that the plan just published still points to; the audio thread may already be rendering that new epoch, so this becomes a use-after-free rather than merely premature eviction. The live plan’s IDs need to be retained unconditionally (for example, release against the union of plan-named IDs and model IDs), with model IDs only extending retention to bypassed/dormant state. A mismatched-ID regression test would pin that lifetime invariant.

PR #2009, opened 2026-08-05
Native engine: latency compensation and buffer assignment (#1889)

### LINKED ISSUE #2013: Engine core slice 4: latency compensation and buffer assignment
Slice 4 of #1889.

Latency compensation, and the buffer arena it decides.

- `OpKind::Delay`, one per input edge of a fan-in, keyed to the op it feeds. Where compensation goes is topology and lives in the plan.
- How many samples each holds is not in the plan: latency belongs to a loaded instance, so it is resolved when the plan is prepared against its bindings. A plugin changing what it reports re-prepares the same plan rather than recompiling a new one.
- The rule is the current engine's, applied wherever paths meet: every input of a fan-in is delayed up to the longest of them and nothing is pulled early. That covers sends, sidechains, group tracks and parallel rack chains without any of them being a special case.
- Buffer assignment replaces one buffer per port with an arena. Ports share only where no schedule can want both at once, tested against transitive dependency rather than the order the ops sit in, because the parallel executor is free to run unrelated ops in either order.
- A delay that turns out to hold no samples is elided, so compensation is uniform in the plan and free where there is none to do.

Implemented in #2009.

lucaromagnoli:
Two prepare-time passes over the compiled plan, and one new op.

Compensation is a Delay op on every input edge of a fan-in, keyed to the op it feeds. How many samples each one holds is not in the plan: latency belongs to a loaded instance, which the compiler has never seen, so it is resolved when the plan is prepared against its bindings. A plugin that changes what it reports therefore re-prepares the same plan instead of recompiling a new one, and the plan, its fingerprint and everything keyed on an OpKey come through untouched.

The rule is the current engine's, applied wherever paths meet: every input of a fan-in is delayed up to the longest of them and nothing is pulled early. That covers sends landing on another track, sidechains meeting the chain they key, groups summing children, and a rack's parallel chains, all without any of them being a special case. MIDI is aligned against the audio it travels with, which that engine gets for free by carrying both in one stream.

Buffer assignment replaces one buffer per port with an arena. Ports share only where no schedule can want both at once, tested against transitive dependency rather than the order the ops sit in: the reference executor's straight walk is one of the orders the parallel executor may pick, and a slot reused on the strength of that walk is a race in the others. An op may write over an input it is the only reader of, which keeps a track's whole chain in one buffer.

A delay that turns out to hold no samples is elided: the op does not run and its output port is its input's. That is what lets compensation be uniform in the plan without charging for it in the common case where there is none to do.

Closes #2013.

### DIFF
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -30,6 +30,13 @@ void copyWithGain(juce::dsp::AudioBlock<float> destination, juce::dsp::AudioBloc
             channelGain(value, channel), numSamples);
 }
 
+void applyGain(juce::dsp::AudioBlock<float> block, const OpValue& value, int numSamples) {
+    for (std::size_t channel = 0; channel < block.getNumChannels(); ++channel)
+        juce::FloatVectorOperations::multiply(block.getChannelPointer(channel),
+                                              channelGain(value, static_cast<int>(channel)),
+                                              numSamples);
+}
+
 float peakOf(juce::dsp::AudioBlock<float> block, int numSamples) {
     float peak = 0.0f;
     for (std::size_t channel = 0; channel < block.getNumChannels(); ++channel) {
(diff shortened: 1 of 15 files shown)


magda/engine/exec/PlanExecutor.cpp
```diff
@@ -212,16 +245,134 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+    for (std::size_t i = 0; i < numOps; ++i) {
+        const auto samples = latency.delaySamples[i];
+        if (audioDelayForOp_[i] >= 0)
+            audioDelays_[static_cast<std::size_t>(audioDelayForOp_[i])].prepare(
+                context_.numChannels, samples, context_.maxBlockSize);
+
+        if (midiDelayForOp_[i] >= 0) {
+            // Every block the delay spans is holding events, plus the one being
+            // filled and the one being drained.
+            const auto blocks = samples / context_.maxBlockSize + 2;
```
lucaromagnoli, line 384: This bound only holds when every callback is exactly `maxBlockSize`. `process()` supports smaller/variable blocks, though: with an 80-sample delay and 1-sample callbacks, `pending_` can retain roughly 80 producer blocks, while this reserves only 3. That makes `scratch_.addEvent()` grow on the audio thread. The related `2 * carried` delay-output bound above also fails if many small input blocks become due during a later large block. Capacity needs a byte/event-rate bound expressed per sample (or a minimum callback size/fixed-block contract), and both pending storage and the output-port bound must use the worst case across valid block-size sequences; add a variable-block MIDI regression alongside the audio one.

lucaromagnoli, line 384: This sizing assumes blocks are always maxBlockSize, and RenderContext explicitly promises shorter blocks work. The producer cap the bound is built from is per block, so a host delivering 64-sample blocks against a 512-prepared plan fits eight times as many blocks inside the same delay window, and a dense producer can legally overflow the reservation, making pending_ grow on the callback. That is the same reservation-vs-reality class the flat-cap fix eliminated, reintroduced through a block-size assumption instead of a fan-in one. Either size the window from a documented minimum block size, or bound pending_ by count with a deterministic drop and a debug assert. Related test gap: the block-size invariance case uses a latency-free plan, so nothing currently runs delay lines at short blocks; extending it to a latent plan with MIDI would exercise both the audio ring at odd sizes and this capacity edge.

lucaromagnoli (commented):
Reviewed both passes and the executor changes in depth. Verified correct: resolvePlanLatency leans on exactly the invariants the new validator rules enforce, so the single forward pass reading each delay off its consumer is well-founded; assignBuffers' sharing criterion is transitive dependency rather than walk order, and the in-place path that skips the subset check is still safe because every older slot user is an ancestor of the previous writer, which is an ancestor of the in-place taker; elision aliasing is applied consistently in the reader lists, keep-alive sets, and the final remap; the executor's in-place cases preserve compiled summing order so the bit-identical claim survives; AudioDelayLine writes before it reads, which is what makes buffer sharing across a delay sound; and running delays while the value layer says silent is the right call, with the reasoning written at the exception. The prepare-time seam (delay placement is topology, sample counts are instance properties, latency change re-prepares without recompiling) is the standout decision and it is tested.

One real finding inline. One forward-looking note, no action here: this slice creates the executor's first real cross-block state, the delay-line contents, and every publish re-prepares them fresh, so a structural edit mid-playback flushes in-flight latent audio. That is the differ's state-carry problem arriving on schedule, and delay lines are already on the epic's carried-state list.

magda/engine/exec/PlanExecutor.hpp
```diff
@@ -24,16 +25,68 @@
...
+        return overflowed_;
+    }
+
+  private:
+    juce::MidiBuffer pending_, scratch_;
+    int delay_ = 0;
+    /// What prepare() reserved, kept so process() can tell when the budget the
+    /// reservation was computed from has been exceeded.
+    int capacity_ = 0;
+    bool overflowed_ = false;
```
lucaromagnoli, line 73: This flag is written on the audio thread and read from whatever thread calls midiDelayOverflows(), with nothing synchronizing the two: a plain bool racing across threads is undefined behavior in the letter of the standard, and it is exactly the kind of thing the swap-stress harness would report if it polled the accessor while rendering. Since the overflow flag is now the enforcement point of the span budget (the per-write asserts can only see one callback's worth), it is worth making honest: std::atomic<bool> with relaxed store and load costs nothing on the callback and makes the accessor safe to poll live, which a session-level diagnostic will eventually want to do.

lucaromagnoli (commented):
Second round, on the fix commit. Restating the budget over a span of maxBlockSize samples is the right fix and I verified it makes every reservation correct without resizing any of them: a callback covers at most one span, so the per-block reservations were already within budget, and the delay's (samples / maxBlockSize + 2) spans genuinely bounds the in-flight window (length delay + maxBlockSize overlaps at most that many spans, however misaligned). The dense-traffic test at 1, 3, 5, and 11-sample callbacks with the port at its exact budget is precisely the test I asked for, and asserting on the reservation as well as note positions gives it teeth. Choosing detect-and-report over a deterministic drop is defensible: correctness is preserved, and the overflow is visible instead of a silent allocation.

One consequence of the redefinition worth stating in the contract rather than leaving implied: the per-write asserts can only check a callback's worth, which for short callbacks is looser than the span budget, so the span contract is enforced downstream by the overflow flag rather than at the write site. That is fine, but it means the flag is the enforcement, which makes the one finding inline matter slightly more than a diagnostics nit would.

With that, this PR is done from my side.

PR #2020, opened 2026-08-05
Feat/1889 plan differ

lucaromagnoli:


magda/engine/exec/EngineSession.cpp
```diff
@@ -14,17 +14,25 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
...
     // it created stays in the store, where the next attempt reuses it and the
     // model, not the plan, decides when it goes.
+    //
+    // The epoch still rendering is handed over so the new one can take on what
+    // the differ says survived, rather than starting every delay line empty.
+    // Read only: it is live until the swap below, and it is this thread's own
+    // handle on it that keeps it alive long enough to be read at all.
     const auto bindings = store_.realise(*prepared->plan);
-    auto messages = prepared->executor.prepare(*prepared->plan, bindings, context);
+    auto messages = prepared->executor.prepare(*prepared->plan, bindings, context,
```
lucaromagnoli, line 24: The executor-state handoff is safe, but the reused runtime objects are still mutated before the handoff: `store_.realise()` returns the same device/source instances owned by the live epoch, and this `prepare()` calls `prepare(context)` on each of them while the audio thread can still be inside their old `process()`/`render()`. Real plugin/source prepare paths reset and resize internal state, so a structural publish can race the callback and also erase exactly the tails/state this slice is preserving. The swap only waits after this call, which is too late. Reused runtime objects need to skip prepare when their configuration is unchanged, or their reconfiguration must happen after quiescing the old epoch (with a replacement instance when preparation must remain off the callback). A blocking test with `prepare()` mutating state while `process()` is held open would expose the race; the current ledger devices have inert prepare methods.

lucaromagnoli (commented):
Reviewed the differ, the adoption path, and the session changes in depth. The core arguments all verify: old and new epochs never render concurrently (the swap primitive guarantees the audio thread is out of the old epoch before the new one is acquirable), so shared delay lines have an owner to outlive and no concurrent use to guard, exactly as the comment claims; prepare() only reads the predecessor's configuration and copies shared pointers, all of which are constant after that executor's own prepare, so reading a live executor is sound; the two-step release in publish() preserves the destructor-thread rule (the RealtimeObject drops its reference inside the replace, live_ drops the last one after, both on the publishing thread); the hasConfiguration gates include the MIDI capacity, so a port whose budget widened rebuilds instead of adopting short storage; and the locality rule (carry unless the immediate producer's key changed) is the incumbent's adopt-by-ID mechanism with the same non-recursive reach, with both the positive and negative cases tested. The atomic overflow fix from the last round also landed here.

One design-level finding inline, promoted by this slice's own success. Nothing else.

magda/engine/exec/EngineSession.cpp
```diff
@@ -14,17 +14,25 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
...
+                                               live_ == nullptr ? nullptr : &live_->executor);
     if (!prepared->executor.isPrepared())
         return {false, std::move(messages)};
 
     // The swap. This blocks until the audio thread is out of the block it was
     // in, then hands the previous epoch back here, where its destructor runs.
-    published_.nonRealtimeReplace(std::move(prepared));
+    published_.nonRealtimeReplace(prepared);
```
lucaromagnoli, line 54: Now that delay lines survive the swap, the unity-values window becomes the dominant swap artifact. Between this line and the publishValues() that follows, the audio thread can take a block or two with the new plan and the old values table, and the fingerprint mismatch falls back to unity: a fader sitting at -20 dB pops to 0 dB for those blocks, on every structural edit, which is louder than the rebuild click this whole slice exists to kill. Unity was a reasonable safe reading back when a swap flushed everything anyway; it is the wrong one now that swaps are otherwise seamless. The fix fits the epoch shape: let PreparedRender carry the initial PlanValues resolved against the new plan (publish takes them, or resolves them before the swap), so the executor never renders a plan without values that belong to it, and publishValues stays what it is, the mixer-speed update path. A test rendering between publish and publishValues with a fader well below unity would pin it.

magda/engine/exec/EngineSession.cpp
```diff
@@ -4,27 +4,36 @@ namespace magda::engine {
...
 EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> plan,
                                              const RenderContext& context,
-                                             const RuntimeStateIds& modelIds) {
+                                             const RuntimeStateIds& modelIds, PlanValues values) {
     if (plan == nullptr)
         return {false, {"no plan to publish"}};
 
     auto prepared = std::make_shared<PreparedRender>();
     prepared->plan = std::move(plan);
+    prepared->values = std::move(values);
```
lucaromagnoli, line 13: Nothing verifies that these values were resolved against this plan. If a caller pairs a plan with a table resolved against a different one (the coalescing pipeline shuffling revisions is the realistic path), the epoch's fallback table has a mismatched fingerprint and the executor's own internal check quietly degrades to unity, the exact behavior this commit exists to remove, now reachable only through a caller error nothing reports. One comparison at the top, values.planFingerprint == planFingerprint(*plan), refusing or at least reporting the mismatch, turns that silent regression into a diagnostic. Same structural-over-contractual principle as the modelIds union: the invariant an epoch never renders without values resolved for it is worth enforcing where the epoch is assembled, not just documenting in the parameter comment.

lucaromagnoli (commented):
Second round, on the fix commit. Both fixes verify. Values riding the epoch closes the unity window exactly as intended: the published table is preferred when its fingerprint matches the rendering executor, the epoch's own table is the floor otherwise, both reads are of post-publish-immutable data, and a torn acquire pair degrades to the epoch's own values, which is now a correct reading rather than unity. Moving preparation into the store's create moment is the right generalization and honestly the more important catch: the executor preparing shared instances on every publish was both a race against the live callback and the loss of the very state the differ preserves, and preparing at creation is the one moment nothing else can reach the object. The context-change path being documented as a stopped-device operation is the honest framing of something the engine cannot enforce. The prepare-counting device and the below-unity fader test pin both behaviors from the direction they would have broken.

One small guard left, inline. After that I have nothing further on this PR, and slice 5 closes out what looks like a genuinely solid run: the differ, the state store, and the epoch machinery now enforce their invariants structurally rather than by convention, which is where this epic said the bugs would concentrate.

magda/engine/exec/EngineSession.cpp
```diff
@@ -46,8 +55,16 @@ void EngineSession::process(const BlockInfo& block, juce::AudioBuffer<float>& ou
...
     }
 
+    // Values and plans travel separately, so the ones in flight can belong to
+    // the plan this replaced. The epoch's own are what it renders with until
+    // newer ones for this plan arrive, which is the difference between a fader
+    // holding its position through a structural edit and jumping to unity for
+    // a block.
     PublishedValues::ScopedAccess<farbot::ThreadType::realtime> values(values_);
-    (*render)->executor.process(*values, block, output);
+    const auto& table = values->planFingerprint == (*render)->executor.planFingerprint()
```
lucaromagnoli: Fingerprint equality proves compatibility, not freshness. If plan P has global values V1 in `values_`, then P is republished with bundled V2 (for example a latency re-prepare of the same plan, or any recompile whose structure is unchanged), both tables have the same fingerprint and this branch keeps selecting the older V1, ignoring the epoch’s V2 until another `publishValues(V2)` happens. That contradicts the new contract that the bundled table is used until something *newer* arrives, and can leave the session indefinitely at stale fader/mute values. The selection needs an ordering/generation as well as compatibility, or publish must replace/invalidate the global table as part of publishing the epoch. A regression should publish V1 globally, republish the same fingerprint with a visibly different bundled V2, render before any later value update, and expect V2.

lucaromagnoli (commented):
Fourth look, no findings. The third-round commit holds up under scrutiny: the refusal check sits where the fingerprint is already computed and its side effects (instances realised before the refusal) are the documented retry semantics, not a leak. The staleness fix's ordering is right in both fingerprint cases: replacing the in-flight table before the plan swap means a structurally-different plan renders the window on the old epoch's own floor, and a same-structure republish applies the new table one block early, which is exactly publishValues semantics anyway. The newest-wins argument (anything still matching after publish arrived later than the epoch) is sound under the single-publisher contract, since a publishValues from before the publish is replaced by it in program order. Both tests fail without their fixes and test the behavior from the direction it would have broken. Done from my side; this closes slice 5 cleanly.

magda/engine/exec/EngineSession.cpp
```diff
@@ -4,27 +4,60 @@ namespace magda::engine {
...
         return {false, std::move(messages)};
 
+    // Values resolved against a different plan are refused rather than
+    // published, because the executor would quietly decline to apply them and
+    // render at unity: the caller would have asked for a fader and got 0 dB,
+    // with nothing anywhere saying why. The check is free here, where prepare
+    // has just computed the fingerprint, and it makes "an epoch never renders
+    // without values resolved for it" something the epoch enforces where it is
+    // assembled rather than something its parameter comment asks for.
+    if (prepared->values.planFingerprint != prepared->executor.planFingerprint()) {
```
lucaromagnoli: This needs to mirror the executor’s full applicability check, not only its fingerprint half. `PlanExecutor::process()` also requires `values.ops.size() == plan.ops.size()`; a truncated/partially assembled table can retain the correct fingerprint, pass this guard, and then make the newly published epoch fall back to unity anyway. Please refuse a size mismatch here too (and cover it by removing one entry from an otherwise correctly resolved table), so the “published epoch always has applicable values” invariant is actually complete.

lucaromagnoli (commented):
Checked the consolidation commit. appliesValues as the single definition is right, and the half it adds (op count, catching a table assembled partway) is a hole my suggested guard had: I asked for the fingerprint comparison and the fingerprint alone was not the executor's whole question. All three callers now ask the same predicate, the audio-thread call reads only post-prepare-constant state, and the epoch's own fallback table is always applicable by construction since publish checked it. The truncated-table test covers the new half. Nothing further; still done from my side.

Issue #2015, opened 2026-08-05
Engine core slice 6: transport and tempo snapshots


lucaromagnoli:
Slice 6 of #1889.

Transport and tempo, as snapshots rather than as engine state.

- A native TempoMap: beats are the canonical domain, and the executor's sample clock is the one place a musical position becomes a sample position.
- Loop, count-in, click.
- The transport cursor currently arrives from the caller (`BlockInfo::timelineSample`). This slice replaces it with a tempo-snapshot cursor reading beats, published the way values are.
- Playhead jumps, and what an op is allowed to assume about continuity across one.

PR #2021, opened 2026-08-05
Native engine: the transport, in beats (#2015)

lucaromagnoli:
Slice 6 of #1889. BlockInfo::timelineSample is gone: an op is handed the stretch of timeline its block covers, in beats, and EngineSession takes a sample count rather than a cursor.

A native TempoMap bakes the tempo track into constant-tempo sections, so beats and seconds convert with a binary search and a multiply. A beat is a quarter note whatever the signature says; the signature groups beats into bars and says what the metronome ticks on. Consecutive tempo changes ramp, shaped by the same tension the model's curve lane stores and evaluated through the same maths, so a ramp is the curve the user drew rather than a second opinion about it.

The clock holds an anchor and derives the cursor from the samples since, so it cannot drift and a block cut in half lands exactly where the whole one would have. Where a callback covers two stretches of timeline it is cut at the wrap and rendered as two blocks, which is what keeps every op from having to know a loop can wrap mid-buffer. The loop catches a cursor that reaches its end and leaves alone one that was put down beyond it: a locate is honoured as asked.

Transport, loop, count-in and metronome are published the way values are, with a generation counter that makes the play request a request rather than a description. A tempo edit re-anchors on the beat and is not a jump; starting, locating and wrapping are, and what that licenses is re-deriving what came from the cursor, never resetting what came from time.

The metronome is summed after the plan rather than compiled into it: it is never recorded, never routed, and not the master fader's to attenuate.

### DIFF
--- a/magda/engine/transport/TempoMap.cpp
+++ b/magda/engine/transport/TempoMap.cpp
@@ -1,0 +1,293 @@
+#include "transport/TempoMap.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <cstring>
+#include <limits>
+
+#include "core/CurveMath.hpp"
+#include "core/TempoUtils.hpp"
+
+namespace magda::engine {
+namespace {
+
+/// Beats are compared, never accumulated, so this only has to absorb the error
+/// in one multiply. A tick computed a billionth of a beat early is on the beat.
+constexpr double kBeatEpsilon = 1.0e-9;
+
+/// How finely a tempo ramp is cut into constant-tempo sections. Four to the
+/// beat, capped, which is what the incumbent uses: at 120 bpm a section is
+/// 125 ms of a curve that is already smooth, and the cap keeps a ramp across a
+/// hundred bars from baking a hundred thousand sections nobody can hear.
+constexpr double kRampSectionsPerBeat = 4.0;
+constexpr double kMaxRampSections = 100.0;
+
+/// The last section has no end. It is given a length rather than an infinity so
+/// that its start time stays a finite number; nothing reads past it, because
+/// every lookup past the last section start resolves to it.
+constexpr double kTailBeats = 1.0e6;
+
+/// Where a ramp has got to, at a beat between two changes.
+///
+/// Evaluated through the model's own curve maths, so a tempo ramp is the curve
+/// the tempo lane draws rather than a second opinion about it.
+double rampedBpm(double beat, const TempoChange& from, const TempoChange& to) {
+    const auto span = to.startBeat - from.startBeat;
+    if (span <= 0.0)
+        return to.bpm;
+
+    const auto position = std::clamp((beat - from.startBeat) / span, 0.0, 1.0);
+    return static_cast<double>(
+        curvemath::evalSegment(static_cast<float>(from.bpm), static_cast<float>(to.bpm), 0.0f,
+                               from.tension, false, static_cast<float>(position)));
+}
+
+/// Bar length in beats. A beat is a quarter note, so a bar of x/y is x quarter
+/// notes scaled by what the denominator says a beat of the signature is worth.
+double barBeatsOf(int numerator, int denominator) {
+    return 4.0 * numerator / denominator;
+}
+
+/// The note value the metronome ticks on, in beats.
+double tickBeatsOf(int denominator) {
+    return 4.0 / denominator;
+}
+
+std::uint64_t mixInto(std::uint64_t hash, std::uint64_t value) {
+    // FNV-1a, as everywhere else in the engine that fingerprints a structure.
+    for (int byte = 0; byte < 8; ++byte) {
+        hash ^= (value >> (byte * 8)) & 0xff;
+        hash *= 1099511628211ULL;
+    }
+    return hash;
+}
+
+std::uint64_t mixDouble(std::uint64_t hash, double value) {
+    std::uint64_t bits = 0;
+    static_assert(sizeof(bits) == sizeof(value));
+    std::memcpy(&bits, &value, sizeof(bits));
+    return mixInto(hash, bits);
+}
+
+}  // namespace
+
+TempoMap::TempoMap() : TempoMap({}, {}) {}
+
+TempoMap::TempoMap(std::vector<TempoChange> tempos, std::vector<TimeSignatureChange> signatures) {
+    if (tempos.empty())
+        tempos.push_back({});
+    if (signatures.empty())
+        signatures.push_back({});
+
+    // Negative positions are clamped rather than refused: the timeline runs
+    // before beat zero (count-in does), but nothing there is editable, so a
+    // change that landed there is a change at the beginning.
+    for (auto& tempo : tempos) {
+        tempo.startBeat = std::max(0.0, tempo.startBeat);
+        tempo.bpm = clampBpm(tempo.bpm);
+        tempo.tension = std::clamp(tempo.tension, -1.0f, 1.0f);
+    }
+
+    for (auto& signature : signatures) {
+        signature.startBeat = std::max(0.0, signature.startBeat);
+        signature.numerator = clampTimeSignatureValue(signature.numerator);
+        signature.denominator = clampTimeSignatureValue(signature.denominator);
+    }
+
+    // Stable, so two changes at the same beat keep the order they arrived in:
+    // that pair is how a step is written, and which of them wins is the whole
+    // meaning of it.
+    std::stable_sort(tempos.begin(), tempos.end(),
+                     [](const auto& a, const auto& b) { return a.startBeat < b.startBeat; });
+    std::stable_sort(signatures.begin(), signatures.end(),
+                     [](const auto& a, const auto& b) { return a.startBeat < b.startBeat; });
+
+    // What precedes the first change is that change's own value, held. For the
+    // tempo that means an extra change at the beginning rather than moving the
+    // first one: moving it would drag the start of its ramp back with it.
+    if (tempos.front().startBeat > 0.0)
+        tempos.insert(tempos.begin(), TempoChange{0.0, tempos.front().bpm, 0.0f});
+    signatures.front().startBeat = 0.0;
+
+    fingerprint_ = 14695981039346656037ULL;
+    fingerprint_ = mixInto(fingerprint_, tempos.size());
+    for (const auto& tempo : tempos) {
+        fingerprint_ = mixDouble(fingerprint_, tempo.startBeat);
+        fingerprint_ = mixDouble(fingerprint_, tempo.bpm);
+        fingerprint_ = mixDouble(fingerprint_, tempo.tension);
+    }
+
+    fingerprint_ = mixInto(fingerprint_, signatures.size());
+    for (const auto& signature : signatures) {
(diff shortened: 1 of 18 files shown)


magda/engine/transport/TransportState.hpp
```diff
@@ -0,0 +1,91 @@
...
+ * clock applies a snapshot whose generation it has not seen and ignores the
+ * position in every other one, so republishing the same transport because the
+ * tempo changed does not drag the cursor back to where playback started.
+ * Whoever publishes bumps it for every play, stop and locate, and for nothing
+ * else.
+ */
+struct TransportRequest {
+    std::uint64_t generation = 0;
+    bool playing = false;
+    double positionBeat = 0.0;
```
lucaromagnoli, line 83: Stop-in-place is not expressible, and it is the most common transport action there is. Every new generation anchors to positionBeat, so a stop request must echo a position the publisher read from positionBeats(), which is stale by up to a callback by the time the clock applies it: the cursor visibly steps back a few milliseconds on every stop, and the position delta also breaks continuity, so sources re-derive for a jump nobody made. Suggest letting the request say keep the cursor where it is: a NaN sentinel or a bool alongside positionBeat, applied as playing-state change only, no re-anchor, no continuity break. The locate-while-stopped and stop-and-return cases keep the current shape; it is only stop-where-you-are that needs to not be a round trip through a stale read.

lucaromagnoli (commented):
Reviewed the clock, the map, the click, and the session integration in depth. The load-bearing claims all verify. The anchor design genuinely cannot drift: the cursor is a function of (anchorSeconds, samplesSince, map), segment beats are derived from the same anchor so consecutive segments meet exactly, and the block-cut invariance that buys is tested. The loop semantics are right at every edge I could construct: a cursor exactly at the loop end wraps (the sample epsilon keeps the just-anchored position from becoming a boundary one sample ahead of itself), one located beyond it plays through as asked, the count-in never wraps, the too-short-loop escape hatch renders straight through and counts it, and at most one overflow increment per callback. The tick walk cannot miss or double a tick across segment boundaries because adjacent segments share the same derived beat, tickAtOrAfter is inclusive at the start and the walk exclusive at the end, and nextBeat strictly advances even across a signature cut. The step spelling (ramp to the first change at a beat, jump to the last) matches the stable sort plus last-wins loop exactly as the comment claims. Sharing the click across epochs of the same context extends the delay-line ownership argument correctly, and the metronome sitting outside the plan is the right call for all three stated reasons. The bake reusing curvemath so a ramp is the drawn curve is exactly the kind of single-source-of-truth this epic keeps getting right, and the ramp's left-endpoint integration being internally consistent (every consumer reads the same baked map) makes the approximation honest.

One API gap inline; it is the only finding. Nits not worth threads: at extreme signature-times-tempo combinations (16th-note ticks past ~300 bpm) a new tick cuts the previous click's decay, which the 40 ms envelope makes inaudible in practice; and the first real transport request must carry a generation above zero, which the bump-on-every-action contract implies but a first-time integrator could miss.

magda/engine/transport/TransportClock.cpp
```diff
@@ -0,0 +1,183 @@
...
+            samples = std::min(samples, untilLoopEnd);
+
+        // Both boundaries were just moved past if they were behind, so a whole
+        // segment is left. What is not left is room to keep cutting: past the
+        // last slot, and for a loop too short to have a sample in it, the rest
+        // of the callback plays straight through and the overflow is counted.
+        if (samples < 1 || segmentCount_ + 1 == kMaxSegmentsPerBlock) {
+            if (samples < remaining)
+                loopWrapOverflows_.fetch_add(1, std::memory_order_relaxed);
+            samples = remaining;
```
lucaromagnoli, line 180: After this fallback renders the remainder straight through, `samplesSinceAnchor_` finishes beyond `loop.endBeat`. On the next callback, the guard above treats that cursor like an intentional locate beyond the loop and refuses to wrap it, so one segment-budget overflow disables looping permanently rather than only for the rest of this callback. The overflow path needs to preserve enough state to resume/wrap at the next callback (and a regression should advance twice after overflowing).

magda/engine/transport/TempoMap.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+        const auto isLast = i + 1 == changeBeats.size();
+        const auto segmentEnd = isLast ? changeBeat + kTailBeats : changeBeats[i + 1];
+        const auto* rampTarget = tempoIndex < tempos.size() ? &tempos[tempoIndex] : nullptr;
+        const auto ramping =
+            rampTarget != nullptr && std::abs(rampTarget->bpm - currentTempo.bpm) > 1.0e-9;
+
+        auto subdivisions = 1;
+        if (ramping)
+            subdivisions = static_cast<int>(std::clamp(
```
lucaromagnoli, line 171: A time-signature-only edit can change beat timing here. `changeBeats` contains signature boundaries as well as tempo boundaries, and this applies the 100-section cap independently to every resulting segment. For example, a long 0→1000-beat tempo ramp gets 100 sections, but adding a signature change at beat 500 gives each half 100 sections, changing `beatToTime()` even though the tempo track is identical. That breaks the stated invariant that signatures only group beats. Please derive the ramp subdivision grid/budget from tempo-change spans independently of signature cuts, then attach signature metadata to those sections.

magda/engine/transport/TransportState.hpp
lucaromagnoli: Fixed in be2899ce. Right on both counts: the position a stop can name is one read from `positionBeats()` on the message thread, so it is a callback out of date by the time the clock applies it, and the delta then reads as a jump on top of the visible step back.

Went with the bool rather than the NaN sentinel: `TransportRequest::locate`, false meaning the request is about the play state and nothing else. The anchor is left untouched entirely in that case rather than re-derived from the position it already has, so there is no beat to time to beat round trip either. A roll-in still moves the cursor when `locate` is false, because starting a count-in from where you are is a real request, and that move is correctly a jump.

Locate-while-stopped and stop-and-return are unchanged. New test: "Stopping does not have to say where", plus a section for resuming from the same place. Confirmed to fail against the previous commit, where the cursor lands back at beat 0.

magda/engine/transport/TransportClock.cpp
lucaromagnoli: Fixed in be2899ce. Confirmed the permanence: the fallback leaves `samplesSinceAnchor_` past `loop.endBeat`, and the guard above cannot tell that cursor from one located beyond the loop on purpose, so it refuses to wrap it for ever after. One short-loop callback and looping is gone for the rest of the session.

The overflow path now re-anchors to `loop.startBeat` after emitting the straight-through segment, so the next callback starts inside the loop and wraps normally. Gated on the same condition as the counter, so a callback that used its last segment slot without losing a boundary does not re-anchor for nothing. Where the cursor "is" after rendering a loop shorter than a sample through 512 samples of timeline is a fiction either way; the top of the loop is the useful fiction, and the count still says the rendering was not honest.

Regression added as a section on the existing short-loop test: it advances twice and requires the second callback to still be cut and to start at the loop start. Confirmed to fail against the previous commit.

magda/engine/transport/TempoMap.cpp
lucaromagnoli: Fixed in be2899ce, and it was worse than the cap: under a left-endpoint bake, *any* extra boundary changes `beatToTime`, because the half after the cut gets the tempo re-evaluated at the cut rather than the one its section started with. The cap only made it reachable with a realistic tempo track.

So rather than budgeting the subdivision per tempo span and cutting afterwards, the bake now builds the tempo grid from the tempo track alone (ramp subdivision and all) and merges the signature boundaries into it as a second pass. A section takes its tempo from the grid step it falls in, never from its own start beat, so a signature boundary splits a section into two that carry the same tempo and the same total duration. Signature metadata is attached during the merge, as suggested.

That makes the invariant structural rather than a claim in a comment: nothing about a signature can reach a section tempo. New test: "A signature groups beats and never moves one", a 1000-beat ramp where the cap binds, compared with and without a 7/8 change at beat 500 across eight positions, plus a check that the signature did do its own job. Confirmed to fail against the previous commit.

lucaromagnoli (commented):
Checked the response commit; all three fixes hold and I have nothing further. The locate flag is a cleaner shape than the sentinel I suggested, and target-equals-wasAt gives count-in-from-here without a stale echo as a bonus. The overflow re-anchor is guarded consistently with the wrap path and its reasoning is written exactly where the trap was: the cursor lands at the loop top, the next callback wraps again, and one degenerate callback no longer disables looping forever. The signature-independent tempo grid is the best fix of the three because it is exact, not approximate: the time accumulation is a piecewise-constant sum, and refining its boundaries with signature splits cannot change the total, so beat-to-time invariance under signature edits is structural. The step spelling survives the rewrite (the skipped coincident entries still ramp to the first and hold the last), and the coincident-signature region keeps its counted bar. Three failing-first tests pin all of it. Done from my side.

magda/engine/transport/TransportClock.cpp
```diff
@@ -0,0 +1,205 @@
...
+            beatAfter(tempo, samplesSinceAnchor_) <= snapshot.loop.endBeat + kBeatEpsilon) {
+            anchorTo(tempo, snapshot.loop.startBeat);
+            continuous_ = false;
+            untilLoopEnd = samplesUntil(tempo, snapshot.loop.endBeat);
+        }
+
+        auto samples = static_cast<std::int64_t>(remaining);
+        if (countingIn_)
+            samples = std::min(samples, samplesUntil(tempo, countInUntilBeat_));
+        if (untilLoopEnd > 0)
```
lucaromagnoli: The zero-sample case still bypasses the overflow recovery. For a valid loop shorter than `kSampleEpsilon / sampleRate`, `samplesUntil(loop.endBeat)` returns 0; the wrap above re-anchors and recomputes the same 0, but this `> 0` check leaves `samples` equal to `remaining`. Consequently `samples < 1` never fires, `overflowed` stays false, and the cursor again finishes beyond the loop where later callbacks treat it as an intentional locate. This is exactly the “loop too short to have a sample in it” case the fallback comment promises to handle. Please route a zero result into the overflow branch and add a regression using a sub-0.01-sample loop.

lucaromagnoli: Fixed in 1fc1894d. Confirmed the trace exactly: with a loop of 0.005 samples the wrap re-anchors and `samplesUntil` comes back zero again, the `> 0` guard skips the clamp, `samples` stays at `remaining`, so `samples < 1` never fires and nothing is counted. Worse than the case I fixed last round, because the previous one at least reported itself: this one ends looping silently, with the overflow count still reading zero.

The clamp is now conditioned on being inside the loop rather than on the sign of the count, which is what it always meant. From outside the loop the count is negative and clamping on it would cut a callback that is not looping at all; from inside it can legitimately be zero, and that zero is exactly the signal the fallback wants. So the sign test was standing in for a question about where the cursor is, and asking that question directly makes both cases fall out.

Regression added as a section on the short-loop test: two callbacks against a sub-hundredth-of-a-sample loop, requiring one overflow counted per callback and the second callback still starting at the loop start. Against the previous commit it fails on all three assertions, with the second callback starting at beat 0.023, past the end and unreachable by any later wrap.

PR #2022, opened 2026-08-05
Native engine: getting audio off a disk in time (#2016)

### LINKED ISSUE #2016: Engine core slice 7: disk reader and prefetch
Slice 7 of #1889.

Getting audio off disk without the callback waiting for it.

- Prefetch on a dedicated thread, per-reader FIFOs.
- Underrun behaviour, and what a reader that has not caught up renders.
- Seek and loop wrap.
- The `EngineAudioSource` binding is already the seam; this is the implementation behind it for arrangement clips.

Related to #1890, which owns clip playback, stretch and warp above this layer.

lucaromagnoli:
Slice 7 of #1889. Closes #2016.

A stream is a pool of chunks and two lock-free fifos: the prefetch thread reads whole chunks ahead of playback and pushes them, the callback pops them, copies out and hands them back. Nothing on the audio thread allocates, locks or waits for a disk.

## The shape

```
io/AudioFileReader       random access to a file's samples, and a JUCE one
io/PrefetchStream        one file read ahead of the callback that plays it
io/PrefetchThread        the one thread allowed to wait for a disk
io/StreamingAudioSource  one placed file, on the EngineAudioSource seam
```

Chunks rather than a ring of samples, because a chunk can be stamped. A seek makes everything already read wrong, and the stamp is how the callback knows without the reader having to reach into a buffer the callback is holding: audio read for a position since abandoned is recognised on the way out and dropped. The same epoch discipline the plan swap uses, for the same reason.

## Decisions worth a look

**Silence is what a reader that has not caught up renders**, and it counts every block it could not fill. Silence nobody counted is indistinguishable from a gap in the material. The end of the file is not an underrun: there is nothing late about silence past the last sample.

**An underrun is a gap, never a delay.** The cursor moves whether or not the samples arrived, so the block after one plays what that block asked for. A stream that resumed where it ran dry would be late for the rest of the take.

**A loop wrap needs nothing said about it.** The source derives its position in the file from the block it was handed rather than from what it played last, so a wrap, a locate and a clip starting are one path: the stream sees a position it was not expecting, and that is what a seek is. `BlockInfo::continuous` is not read here at all.

**The thread polls; the callback never wakes it.** Waking a thread takes a lock. A stream that has just been seeked is found on the next round, five milliseconds later at worst, which is inside the block or two the underrun count already accounts for.

**BlockInfo gained seconds alongside beats**, the one change to the slice 6 contract. A file's samples are counted in time, and a source that converted beats itself would need a tempo map on the audio thread, of which there would then be two. The clock already has the seconds and derives the beats from them, so it publishes both faces of one instant: beats say where on the timeline something is, seconds say how much audio has gone by. `sampleForTime` sits beside `sampleForBeat` and clamps differently on purpose, which the header explains.

**Rate conversion and warp are not here.** The stream delivers the file's own samples; mapping them to another rate is #1890's question and the same one stretch asks. `ClipPlacement` plus `sourceSampleAt` is the seam between the two layers.

## Testing

`tests/test_prefetch_stream.cpp` and `tests/test_streaming_audio_source.cpp`: sequential playback across chunk boundaries, blocks of different lengths, a pool small enough to have to be recycled, underruns and what they render, the gap that does not become a delay, the end of the file, seeks in both directions, a clip starting and ending inside a block, forty blocks in step with the timeline, a wrap with and without a cue, and the prefetch thread reading for a registered stream.

The reader fixtures compute their samples rather than reading a file off a disk, so a sample says where it came from and every path is deterministic.

Confirmed to fail without their fix: dropping the generation check makes the seek tests play the old position's audio (two tests, three assertions).

ThreadSanitizer over `[io]` is clean, and the harness was shown able to fail first: a non-atomic counter touched from both threads was reported at `PrefetchStream.cpp:98`, then reverted and re-run clean.

Full suite: 1974 passed.

### DIFF
--- a/magda/engine/io/AudioFileReader.cpp
+++ b/magda/engine/io/AudioFileReader.cpp
@@ -34,8 +34,20 @@ int JuceAudioFileReader::read(juce::AudioBuffer<float>& destination, int destina
     if (available < numSamples)
         destination.clear(destinationOffset + available, numSamples - available);
 
-    if (available > 0)
-        reader_->read(&destination, destinationOffset, available, startSample, true, true);
+    if (available <= 0)
+        return 0;
+
+    // Whether the decode worked is the only thing that says the destination now
+    // holds audio rather than whatever was in it. A reader can fail inside the
+    // length it advertises: a file truncated under us, a stream that went away
+    // mid-frame. Reporting the samples as read anyway would publish undefined
+    // buffer contents as material, and the stream would have no reason to
+    // suspect anything, because a full chunk is exactly what success looks
+    // like.
+    if (!reader_->read(&destination, destinationOffset, available, startSample, true, true)) {
+        destination.clear(destinationOffset, numSamples);
+        return 0;
+    }
 
     return available;
 }
--- a/magda/engine/io/AudioFileReader.cpp
+++ b/magda/engine/io/AudioFileReader.cpp
@@ -1,0 +1,43 @@
+#include "io/AudioFileReader.hpp"
+
+#include <algorithm>
+
+namespace magda::engine {
+
+JuceAudioFileReader::JuceAudioFileReader(std::unique_ptr<juce::AudioFormatReader> reader)
+    : reader_(std::move(reader)) {}
+
+std::int64_t JuceAudioFileReader::lengthInSamples() const {
+    return reader_ != nullptr ? reader_->lengthInSamples : 0;
+}
+
+double JuceAudioFileReader::sampleRate() const {
+    return reader_ != nullptr ? reader_->sampleRate : 0.0;
+}
+
+int JuceAudioFileReader::numChannels() const {
+    return reader_ != nullptr ? static_cast<int>(reader_->numChannels) : 0;
+}
+
+int JuceAudioFileReader::read(juce::AudioBuffer<float>& destination, int destinationOffset,
+                              std::int64_t startSample, int numSamples) {
+    if (reader_ == nullptr || numSamples <= 0)
+        return 0;
+
+    // What the file actually has from here. JUCE would pad the rest with
+    // silence itself, but it would not say how much it padded, and the
+    // difference between a short read and a late one is the difference between
+    // the end of the file and a glitch.
+    const auto available = static_cast<int>(
+        std::clamp<std::int64_t>(reader_->lengthInSamples - startSample, 0, numSamples));
+
+    if (available < numSamples)
+        destination.clear(destinationOffset + available, numSamples - available);
+
+    if (available > 0)
+        reader_->read(&destination, destinationOffset, available, startSample, true, true);
+
+    return available;
+}
+
+}  // namespace magda::engine
--- a/magda/engine/io/PrefetchStream.cpp
+++ b/magda/engine/io/PrefetchStream.cpp
@@ -1,6 +1,7 @@
 #include "io/PrefetchStream.hpp"
 
 #include <algorithm>
+#include <utility>
 
 namespace magda::engine {
 
@@ -42,11 +43,24 @@ void PrefetchStream::seek(std::int64_t sourceStart) {
 }
 
 void PrefetchStream::applyPendingCue() {
+    // Whether anything played out of this stream since the last time it was
+    // asked. Read here rather than in read(), so that what it answers is a
+    // question about the block that just went by.
+    const auto wasSounding = std::exchange(readSinceCueCheck_, false);
+
     farbot::RealtimeObject<SeekRequest, farbot::RealtimeObjectOptions::nonRealtimeMutatable>::
         ScopedAccess<farbot::ThreadType::realtime>
             cue(cue_);
 
     if (cue->generation == appliedCue_)
+        return;
+
+    // A stream that is sounding keeps sounding. Taking the cue up would abandon
+    // the generation the material still playing was read for and hand its pool
+    // back, so the cue would cost a gap and then be overridden by the very next
+    // read anyway, which asks for where playback actually is. It waits instead,
+    // and is taken up when the stream falls silent.
+    if (wasSounding)
         return;
 
     appliedCue_ = cue->generation;
@@ -126,9 +140,11 @@ int PrefetchStream::read(std::int64_t sourceStart, juce::dsp::AudioBlock<float> 
     if (numSamples <= 0)
         return 0;
 
-    // Anything cued since the last block, before deciding whether this one is
-    // a seek: a read that lands exactly where the cue pointed is then not one.
-    applyPendingCue();
+    // Reading is what sounding means. Taking a cue up is not done here: it
+    // happens once at the top of a block, through applyPendingCue, and doing it
+    // again inside the read would see a stream that has not read *yet this
+    // block* and mistake it for one that is not playing at all.
+    readSinceCueCheck_ = true;
 
     if (sourceStart != nextSample_)
         requestSeek(sourceStart);
--- a/magda/engine/io/PrefetchStream.cpp
+++ b/magda/engine/io/PrefetchStream.cpp
@@ -35,7 +35,22 @@ PrefetchStream::PrefetchStream(std::unique_ptr<AudioFileReader> reader,
 }
 
 void PrefetchStream::seek(std::int64_t sourceStart) {
-    requestSeek(sourceStart);
+    // Publish and return. Everything the callback owns is left alone, which is
+    // what makes this safe to call while the stream is playing rather than
+    // safe only if nobody forgets it is not.
+    cue_.nonRealtimeReplace(SeekRequest{sourceStart, ++cueGeneration_});
+}
+
+void PrefetchStream::applyPendingCue() {
+    farbot::RealtimeObject<SeekRequest, farbot::RealtimeObjectOptions::nonRealtimeMutatable>::
+        ScopedAccess<farbot::ThreadType::realtime>
+            cue(cue_);
+
+    if (cue->generation == appliedCue_)
+        return;
+
+    appliedCue_ = cue->generation;
+    requestSeek(cue->sourceStart);
 }
 
 void PrefetchStream::requestSeek(std::int64_t sourceStart) {
@@ -44,6 +59,19 @@ void PrefetchStream::requestSeek(std::int64_t sourceStart) {
 
     // Whatever the callback was reading came from somewhere else now.
     releaseCurrent();
+
+    // And so is everything waiting behind it. Handing the pool back here rather
+    // than as the callback works through it is what lets the reader start
+    // refilling immediately: a stream that was cued but is not being read yet
+    // is the whole point of cueing, and it has no callback coming to recycle
+    // chunks for it. Every chunk in the fifo was read for the generation this
+    // call just left behind, so all of them go.
+    Chunk* stale = nullptr;
+    while (filled_.pop(stale)) {
+        const auto returned = spent_.push(std::move(stale));
+        jassert(returned);
+        juce::ignoreUnused(returned);
+    }
 
     farbot::RealtimeObject<SeekRequest, farbot::RealtimeObjectOptions::realtimeMutatable>::
         ScopedAccess<farbot::ThreadType::realtime>
@@ -97,6 +125,10 @@ int PrefetchStream::read(std::int64_t sourceStart, juce::dsp::AudioBlock<float> 
                          int numSamples) {
     if (numSamples <= 0)
         return 0;
+
+    // Anything cued since the last block, before deciding whether this one is
+    // a seek: a read that lands exactly where the cue pointed is then not one.
+    applyPendingCue();
 
     if (sourceStart != nextSample_)
         requestSeek(sourceStart);
--- a/magda/engine/io/PrefetchStream.cpp
+++ b/magda/engine/io/PrefetchStream.cpp
@@ -1,0 +1,202 @@
+#include "io/PrefetchStream.hpp"
+
+#include <algorithm>
+
+namespace magda::engine {
+
+PrefetchStream::PrefetchStream(std::unique_ptr<AudioFileReader> reader,
+                               const RenderContext& context, const PrefetchSettings& settings)
+    : reader_(std::move(reader)),
+      chunkSamples_(std::max(1, settings.chunkSamples)),
+      // Room for every chunk at once, rounded up to what the fifo insists on:
+      // one sized to the pool exactly could not hold all of it, and farbot
+      // indexes by mask rather than by modulo.
+      filled_(juce::nextPowerOfTwo(std::max(2, settings.chunkCount) + 1)),
+      spent_(juce::nextPowerOfTwo(std::max(2, settings.chunkCount) + 1)) {
+    if (reader_ != nullptr) {
+        length_ = reader_->lengthInSamples();
+        sourceSampleRate_ = reader_->sampleRate();
+    }
+
+    numChannels_ = std::max(1, context.numChannels);
+
+    // A chunk has to be able to answer a whole callback, otherwise a block
+    // longer than a chunk would underrun however far ahead the reader got.
+    chunkSamples_ = std::max(chunkSamples_, context.maxBlockSize);
+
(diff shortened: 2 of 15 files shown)


lucaromagnoli (commented):
Reviewed the stream, the source, the thread, and the BlockInfo change in depth. The core verifies: both fifos are genuinely single-producer single-consumer in steady state (audio thread owns spent_'s producer side and filled_'s consumer side, prefetch thread the reverse), the generation stamp is applied and checked everywhere a stale chunk could leak through, the fifo sizing (next power of two above count plus one) actually holds the whole pool, and the stall latch stops a dead reader from spinning the thread without wedging the stream, since a seek clears it. The underrun accounting is right at the edges: end-of-file silence is excluded by the min against length, a fully-past-the-end read counts nothing, and the gap-not-delay rule (cursor advances regardless) is the correct policy and correctly implemented, with catch-up after a stall costing only disk-speed reading of the skipped material, which is orders of magnitude faster than realtime.

The subtle hinge of the whole stateless-position design also holds: deriving sourceStart from the timeline every block does not cause spurious seeks at matched rates, because the fractional part of (blockStartSeconds minus placementStart) times rate is constant across blocks within one anchor, so consecutive llrounds advance by exactly the consumed count. Re-anchors are legitimate seeks. sampleForTime clamping half-open where sampleForBeat clamps closed is properly reasoned and the header says why. The TSAN run with a planted-race check and the computed fixtures are the right verification posture.

Two findings inline: the documented rate-mismatch behavior is not what the code does, and the seek contract can be made structural. Nothing else.

magda/engine/io/StreamingAudioSource.hpp
```diff
@@ -0,0 +1,83 @@
...
+     * The rest is silence: a source fills its whole block, and a clip that
+     * starts a hundred samples in has a hundred samples of nothing in front of
+     * it rather than a hundred samples of whatever the buffer held.
+     *
+     * A block the transport is not moving through renders nothing at all. A
+     * stopped timeline is not a held sample, it is no material.
+     *
+     * Rate conversion is not done here. The stream delivers the file's own
+     * samples, and a file recorded at another rate than the device runs at
+     * plays at the wrong pitch until the clip layer resamples it; that layer
```
lucaromagnoli: This claim does not match what the code does. Wrong-pitch playback would require the stream position to advance sequentially; instead sourceSampleAt advances by count times fileRate over deviceRate file samples per block while read() advanced nextSample_ by count, so at mismatched rates every block's sourceStart misses the stream's expectation and triggers a seek: a generation bump, the prefetched chunks dropped, and the reader restarted, every block. The realistic interim behavior for a 48k file in a 44.1k session is a per-block seek storm and persistent underruns, not wrong pitch. Either make the fallback genuinely sequential so it degrades the way this comment promises (derive the advance from consumed samples when rates differ, accepting drift until #1890 resamples), or state the actual behavior here so the clip layer inherits a documented contract. Either way a test with a mismatched-rate fixture would pin whichever behavior is chosen, and today there is none.

magda/engine/io/PrefetchStream.hpp
```diff
@@ -0,0 +1,174 @@
...
+     * @brief Read ahead. On the prefetch thread.
+     *
+     * Returns true if it did anything, so a caller servicing several streams
+     * can tell a busy round from an idle one and sleep on the difference.
+     */
+    bool fill();
+
+    /// Where the stream will be pointed next. Off the audio thread, before it
+    /// is playing: the callback is what moves the cursor once it is.
+    void seek(std::int64_t sourceStart);
```
lucaromagnoli, line 110: The before-it-is-playing contract is real but only a sentence: seek() from another thread writes nextSample_ and generation_, pushes to spent_ (a second producer on an SPSC fifo), and takes the realtime side of the request object, all of which race the callback if anyone ever cues a stream that is already rendering. The cueing story this API invites (sourceSampleAt is public precisely so a scheduler can point the stream early) makes that misuse likely eventually. The structural fix is small: let seek() write only a pending cue (an atomic pair, or a second request object oriented non-realtime to realtime) that read() folds in at block start on the audio thread, so the cross-thread path never touches audio-thread state and the contract stops needing to be remembered. Same principle as the modelIds union and the publish fingerprint guard in earlier slices.

magda/engine/io/StreamingAudioSource.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+void StreamingAudioSource::prepare(const RenderContext& context) {
+    sampleRate_ = context.sampleRate;
+}
+
+std::int64_t StreamingAudioSource::sourceSampleAt(double seconds) const {
+    // Counted in the file's own samples, from the moment its first sample
+    // plays. The rate is the file's rather than the device's, so a locate lands
+    // on the right sample of the material even when the two differ.
+    const auto rate = stream_.sampleRate() > 0.0 ? stream_.sampleRate() : sampleRate_;
```
lucaromagnoli: Using the file rate to derive every block start is incompatible with copying one source sample per output sample below. With a 48 kHz file on a 44.1 kHz device, a 441-sample block reads source samples 0..440, but the next block derives `sourceStart == 480`, so 39 samples are skipped and the stream sees a seek on every callback. This is not merely the documented wrong-pitch playback while resampling is deferred; it is a discontinuity/underrun every block. Until rate conversion exists, either advance the raw file at the device rate (continuous wrong pitch) or reject mismatched rates explicitly; otherwise `sourceSampleAt()` and the number consumed cannot agree. Please add a mismatched-rate multi-block regression.

magda/engine/io/AudioFileReader.cpp
```diff
@@ -0,0 +1,43 @@
...
+    // difference between a short read and a late one is the difference between
+    // the end of the file and a glitch.
+    const auto available = static_cast<int>(
+        std::clamp<std::int64_t>(reader_->lengthInSamples - startSample, 0, numSamples));
+
+    if (available < numSamples)
+        destination.clear(destinationOffset + available, numSamples - available);
+
+    if (available > 0)
+        reader_->read(&destination, destinationOffset, available, startSample, true, true);
```
lucaromagnoli: The return value from JUCE is the only indication that decoding actually succeeded, but it is ignored here and `available` is returned unconditionally. If the file is truncated/unavailable or the decoder fails inside the advertised length, `PrefetchStream::fill()` sees a full successful chunk, advances past it, and never enters its documented stalled path; stale/undefined destination contents can then be published as valid audio with no underrun signal. Please clear the attempted range and return 0 (or otherwise propagate failure) when `reader_->read(...)` returns false, with a failing-reader regression.

magda/engine/io/StreamingAudioSource.cpp
lucaromagnoli: Fixed in 3d0b3e06, and you are right about the mechanism: the mismatch is between how the position is derived and how much is consumed, not the pitch. Traced it at 48k in a 44.1k session, and it is a seek every callback, so the interim behaviour was silence throughout rather than anything audible.

Took the first option: positions are derived at the device rate, so the derived start advances by exactly the samples the block consumed and the two agree. What a mismatched file costs is then pitch and length, which is what the header claimed all along and now describes accurately. Rejecting mismatched rates would have been silence with nothing to explain it, since this layer has no diagnostic channel to say why.

Regression added: a 48k reader in a 44.1k context, twenty blocks, requiring every sample of the file in order with none skipped between blocks and no seeks at all. Against the previous commit it fails on the first block boundary.

magda/engine/io/PrefetchStream.hpp
lucaromagnoli: Fixed in 3d0b3e06, and the misuse you predicted is the one the API was built to invite, which made keeping it a sentence indefensible.

Took the second-object shape: `seek()` publishes into a non-realtime-to-realtime object of its own and returns, and `applyPendingCue()` folds it in on the audio thread at the top of the block, where every other seek already happens. One writer each way, so neither side writes what the other reads, and nothing cross-thread touches the cursor, the generation or the fifos any more.

The source calls `applyPendingCue()` every block whether or not the clip is sounding. Without that the cue is useless in the case it exists for: a clip that has not started is not being read, so a fold-in that only happened inside `read()` would arrive when the material was already due.

Writing it turned up a gap underneath the one you found. Stale chunks were recycled only as the callback read through them, so a stream that was cued but not yet being read had no way to get its pool back and the reader could never refill: the whole pool sat in the filled fifo carrying an abandoned generation. A seek now hands the entire fifo back where it happens, which is legal because the callback is the only consumer and every chunk in there is stale by construction. That also lets the reader start moving before the next callback instead of after it.

Regression: a clip placed at five seconds, cued while the transport rolls, four silent blocks before it starts, then its first block playing with the material already in memory and no underrun. Against the previous commit the pool is never returned and that block underruns.

magda/engine/io/AudioFileReader.cpp
lucaromagnoli: Fixed in 3d0b3e06. Checked the JUCE contract before touching it: the AudioBuffer overload does return bool, and reading past the end of the stream is explicitly not an error, so the return value means exactly what you said it does and nothing else was carrying that information.

A failed decode now clears the range it was given and returns zero. Clearing matters as much as the count: the chunk buffer is reused across reads, so leaving it alone would hand the previous position back as though it were this one, which is worse than the undefined contents the failure implies. Returning zero puts it on the documented stalled path, so the stream stops asking rather than spinning at full tilt against a file that is not going to answer.

Regression: a `juce::AudioFormatReader` whose `readSamples` returns false, over a destination pre-filled with a non-zero value, requiring nothing read and the buffer cleared. Against the previous commit it reports a full successful read and hands back the pre-filled contents.

magda/engine/io/StreamingAudioSource.hpp
lucaromagnoli: Same fix as the thread on `sourceSampleAt` (3d0b3e06). The comment is now what the code does: positions derived at the device rate, so playback of a mismatched file is continuous and pitched wrong rather than seeking every block, and the header says both that and why the derivation rate is not a slip. Pinned by the twenty-block 48k regression.

lucaromagnoli (commented):
Checked the response commit; everything holds and I have nothing further. The cue object is the right structure: one writer per direction, the audio thread folds it in at block start, and the apply-before-compare ordering means a read landing where the cue pointed is not a second seek, while a cue taken up early lets the reader prefetch so the first covered block plays with no seek at all, which is the entire point of cueing, now actually achievable. The fifo drain inside requestSeek stays on the audio thread's own sides of both fifos, so SPSC is preserved, and it closes the recycling gap that cued-but-unread streams had, which was a better catch than either of my findings. The device-rate derivation makes consumption and position agree so a mismatched file now genuinely degrades to pitch rather than to a seek storm, with the header saying what actually happens, and the failed-decode fix stops a reader error from being published as material. Regressions confirmed failing-first across all four. Done from my side; this closes slice 7 cleanly.

magda/engine/io/PrefetchStream.cpp
```diff
@@ -0,0 +1,234 @@
...
+void PrefetchStream::applyPendingCue() {
+    farbot::RealtimeObject<SeekRequest, farbot::RealtimeObjectOptions::nonRealtimeMutatable>::
+        ScopedAccess<farbot::ThreadType::realtime>
+            cue(cue_);
+
+    if (cue->generation == appliedCue_)
+        return;
+
+    appliedCue_ = cue->generation;
+    requestSeek(cue->sourceStart);
```
lucaromagnoli, line 67: Applying an advance cue immediately calls `requestSeek()`, which releases the current chunk and drains every chunk buffered for the position still playing. That makes the new inactive-clip test pass, but breaks the other advertised cue case: cueing a loop return while this stream is still rendering the current loop iteration silences playback from the next callback until the wrap, because the current generation has been abandoned early. The existing wrap test hides this by applying the cue, filling, and rendering only the wrapped block—there is no intervening current-position block. Please add a regression that cues the loop top while sequential playback is still active, requires the blocks before the wrap to remain continuous, then requires the wrapped block prefetched. Supporting that requires keeping future-position prefetch separate from the active generation (or narrowing/removing the advance-cue contract); an eager seek cannot provide both.

lucaromagnoli: Fixed in a6b4f5e8, and you called the trade correctly: an eager seek cannot serve both, so I narrowed the contract rather than half-building the other half.

A cue is now taken up only when the stream is not sounding, and waits otherwise. Cueing a stream mid-playback costs nothing and changes nothing; the cue is taken up when the stream falls silent. The advance case that motivated the last commit, a clip that has not started, is unaffected, because nothing is reading it.

Writing that turned up why it has to be a block-boundary question. Taking cues up was happening in `read()` as well as at the top of the block, and the second call sees a stream that has not reached its read *yet this block* and cannot tell it from one that is silent, so it applied the cue anyway and reintroduced exactly the gap. Taking cues up now happens once, in `applyPendingCue`, which is the call the source already makes every block; `read()` only records that the stream sounded.

On your regression: the first two requirements are now pinned. A cue published while sequential playback is running leaves every following block sample for sample what it would have been, with the underrun count unchanged, and there is a section for it being taken up once the clip stops. The third is not, and I have not pretended otherwise: the wrapped block is not prefetched, because the reader was never sent there. Reading the top of a loop while the end of it is still playing is a second fill cursor and a second pool, and I would rather that landed with whoever owns clip looping (#1890) than be built halfway underneath them. The header now states it as a limitation instead of implying a seamless return, and a wrap costs the block it happens in with the underrun count saying so.

I also deleted the test that asserted a cued wrap costs nothing. It was asserting the undeliverable property, and it only passed because it applied the cue, filled, and rendered the wrapped block with no intervening block at the current position, which you spotted.

magda/engine/io/PrefetchStream.cpp
```diff
@@ -0,0 +1,250 @@
...
+int PrefetchStream::read(std::int64_t sourceStart, juce::dsp::AudioBlock<float> destination,
+                         int numSamples) {
+    if (numSamples <= 0)
+        return 0;
+
+    // Reading is what sounding means. Taking a cue up is not done here: it
+    // happens once at the top of a block, through applyPendingCue, and doing it
+    // again inside the read would see a stream that has not read *yet this
+    // block* and mistake it for one that is not playing at all.
+    readSinceCueCheck_ = true;
```
lucaromagnoli, line 160: This marks the stream as sounding for every positive-size request, including one wholly past `length_`. If a placement extends beyond the file (or its source offset has already exhausted it), `StreamingAudioSource` keeps calling `read()` each block, each call returns terminal silence without an underrun, but `wasSounding` remains true forever and a pending cue can never be taken up. That contradicts the state comment (“anything played out”) and prevents cueing the stream for its next use precisely while it is silent. Please treat a request as active only while its source range can still contain file samples (while retaining attempted reads inside the file as active during an underrun), and add a regression that publishes a cue while the placement is rolling past EOF.

lucaromagnoli: Fixed in a4596aa7. Confirmed the trap: a placement outstaying its material keeps the source asking every block, each ask returns terminal silence, and the flag latches on, so the stream looks busy for ever at exactly the moment pointing it somewhere else is worth doing.

Went one step past the narrow fix, because writing the regression showed the narrow one is not enough on its own. Marking those asks as not-sounding lets the cue be taken up, and then the same block reads a position past the end, misses the cursor, and seeks straight over it, so the cue is lost a moment after being applied. The asks were doing damage on the way past regardless of what the flag said.

A request that cannot contain file samples now returns silence and touches nothing at all: no underrun, no cursor move, no generation bump, no drained pool. A stream whose material has run out has not gone anywhere, so there is nothing to seek to and nothing to invalidate. That covers your point and closes the override behind it.

A short read from inside the file still counts as sounding, as you asked: that stream is playing and merely starved, which is not the same thing as having nothing to play.

Regression: five thousand samples of material under a second-long placement, played well past the end, a cue published while it rolls, two more silent blocks, then playback returning to the top with the material already in memory and no underrun anywhere. Against the previous commit the cue never survives and that last block plays silence.

Issue #1896, opened 2026-07-27
Null-diff validation harness and project migration


lucaromagnoli:
Part of #1882. Starts alongside the engine core, not after it.

The verification and migration rig that everything else converges against.

- Null-diff harness: render the same .mgd through TE and the native engine, assert near-null difference; corpus of real projects covering stretch, warp, racks, modifiers, sends, automation.
- Plan golden tests in CI (compiled-plan snapshots for model fixtures).
- DAWproject round-trip as a semantic cross-check.
- Migrators: internal device pluginState v1, paramIndex mapping where ordering changes, pinned enum values. Old projects and presets must load bit-identically or with a documented, tested migration.

lucaromagnoli:
I would define a tiered comparison oracle rather than requiring near-null output universally:

- Bit-exact or tight epsilon for deterministic internal DSP and simple routing.
- Latency-aligned numerical comparison for PDC, sends, bypass, and graph transitions.
- Time/frequency-domain tolerances for stretch, warp, and resampling.
- State, latency, event-order, tail, and no-discontinuity invariants for nondeterministic external plugins.
- Scripted interactive checks for launcher, monitoring, hardware input, and recording, where offline nulling is not representative.

Plan goldens should use a canonical dump that excludes addresses, allocation order, and other unstable implementation details.

The corpus should also record sample rate, block size, channel layout, random seeds, plugin versions, and expected determinism class. Otherwise a “null failure” will often say more about the fixture environment than the engine.

For the differ specifically, property tests should generate random structural edit sequences and assert state compatibility decisions, correct retirement, PDC alignment, and silence/discontinuity bounds outside the changed graph region.

lucaromagnoli:
Scope addition: alongside null-diff sample parity, a performance suite measuring CPU per block on the project corpus, native vs TE at matched buffer sizes, tracked in CI. This gates cutover (#1897).

lucaromagnoli:
Widening the parity envelope: the suite gates cutover on the incumbent's full envelope, not CPU alone. Alongside null-diff sample parity and CPU per block: dropout rate under a stress corpus (small buffers, heavy sessions, structural edits during playback), round-trip and PDC-reported latency at matched settings, memory footprint per session, and project load time. Anything the current engine delivers that users depend on is a parity requirement by default; exclusions need an explicit argument on this issue.

lucaromagnoli:
[...]

PR #2023, opened 2026-08-05
Native engine: offline render and tap ops (#2017)

### LINKED ISSUE #2017: Engine core slice 8: offline render and tap ops
Slice 8 of #1889.

Rendering a plan faster than real time, and the taps the UI reads.

- Offline plan rendering: the same plan, driven by a caller that is not an audio device, for bounce, freeze and the null-diff harness.
- Native meter and tap ops for master, track and device. `Meter` ops exist and their levels are a plain vector today; publishing them to the UI thread without tearing is this slice.
- Analysis taps (oscilloscope, spectrum) read through the same mechanism.

Feeds #1896, which cannot render both engines over the same fixture without this.

lucaromagnoli:
Slice 8 of #1889. Closes #2017.

## Offline plan rendering

`exec/OfflineRender` drives the executor playback drives, over a beat range, through its own `TransportClock`. Loop, count-in and metronome are all off, and named as off rather than left to defaults: a loop would print the range twice into a bounce that asked for it once, and a click printed into a bounce is the oldest bug in audio software.

The tail renders with the transport stopped, which the engine already defines as the graph running while the timeline stands still, so reverbs ring out and nothing after the range starts playing. It is exactly as long as it was asked for. Stopping at the first silent block would be shorter and would sometimes be wrong, because a gated or tremolo'd tail goes fully silent in the middle; trimming is the caller's, since only it has the samples and the intent.

Values that do not belong to the prepared plan are refused rather than rendered, for the reason a publish refuses them: the executor's answer to values it cannot apply is unity, so the alternative is a bounce with every fader at 0 dB and nothing saying why.

## The offline source

`io/FileAudioSource` reads straight through `AudioFileReader` and is allowed to wait, because there is no callback to starve: no stream, no thread, no chunk pool, no underrun to count.

The placement mapping moves out of `StreamingAudioSource` into `io/ClipPlacement`, so live and offline resolve file positions through one function. A bounce that worked out its own could disagree with what was heard by a sample, and the only way anyone finds out is by listening to the render.

## Taps

`tap/LevelTap` is what a Meter op publishes through: stereo peak, bound by `OpKey` in `PlanBindings`, owned by `RuntimeStateStore` and carried across plan swaps, so a structural edit elsewhere in the project does not reset a meter that was reading. A read takes the peak and starts again, so a meter reports the loudest thing since it was last looked at rather than whichever block the UI happened to land on. `PlanExecutor::meterLevels()` is gone.

`tap/SampleRing` is the same seam for analysis (oscilloscope, spectrum): the most recent N samples, single producer, single consumer. It downmixes straight into the ring, so unlike the incumbent's `AudioTapBuffer` it has no maximum block size and cannot silently drop a block when a host changes its buffer size.

A Meter op with nothing bound is a meter nobody is reading. It is not reported as unbound, because a plan has a meter at every track and every device slot, and every offline render there is would pay for those diagnostics.

## Tests

`test_offline_render.cpp`, `test_file_audio_source.cpp`, `test_engine_taps.cpp`, plus meter lifetime cases in `test_engine_session.cpp`.

The load-bearing one is block-size independence: the same range renders sample-identical at 17, 64, 512 and 4096. #1896 cannot compare two engines over one fixture unless a difference in the output means a difference in what they render rather than in how the comparison was pulled.

Both cross-checking tests were confirmed to fail without their fix. Offsetting the offline source's mapping by one sample fails the agreement test at sample 441, the clip start; leaving the final block uncut fails block-size independence on length. The agreement test was vacuous when first written (the range never reached the clip, so it compared two runs of silence), which is what that mutation found; it now starts before the clip and asserts the comparison is over material.

Meters are store-owned objects, so `runtimeObjectCount()` counts them. One pre-existing session assertion moved from 1 to 3 and now names what survives instead of only counting.

Full Catch2 suite green: 2012 passed, 0 failed, 13 skipped. `magda_juce_tests` builds clean.

### DIFF
--- a/magda/engine/exec/OfflineRender.cpp
+++ b/magda/engine/exec/OfflineRender.cpp
@@ -1,0 +1,128 @@
+#include "exec/OfflineRender.hpp"
+
+#include <algorithm>
+#include <cmath>
+
+#include "transport/TransportState.hpp"
+
+namespace magda::engine {
+namespace {
+
+std::int64_t samplesBetween(double startSeconds, double endSeconds, double sampleRate) {
+    if (endSeconds <= startSeconds)
+        return 0;
+    return static_cast<std::int64_t>(std::llround((endSeconds - startSeconds) * sampleRate));
+}
+
+}  // namespace
+
+OfflineRenderResult renderOffline(PlanExecutor& executor, const PlanValues& values,
+                                  const RenderContext& context, const TempoMap& tempo,
+                                  const OfflineRenderRequest& request, OfflineRenderSink& sink,
+                                  const std::function<bool()>& shouldContinue) {
+    OfflineRenderResult result;
+
+    if (!executor.isPrepared() || !executor.appliesValues(values)) {
+        result.refused = true;
+        return result;
+    }
+
+    // Held to what the plan was prepared for, the way a callback is: the
+    // executor asserts on a longer block, and here the same number also says
+    // how much buffer there is to render into.
+    const auto blockSize = std::clamp(request.blockSize, 1, context.maxBlockSize);
+
+    // How long the range is, asked once. Deriving the length up front rather
+    // than watching the cursor cross the end is what keeps the answer the same
+    // at every block size: the last block is cut to land exactly on it, and no
+    // block's length depends on how the ones before it were cut.
+    const auto rangeSamples = samplesBetween(tempo.beatToTime(request.startBeat),
+                                             tempo.beatToTime(request.endBeat), context.sampleRate);
+    const auto tailSamples =
+        request.tailSeconds > 0.0
+            ? static_cast<std::int64_t>(std::llround(request.tailSeconds * context.sampleRate))
+            : 0;
+
+    juce::AudioBuffer<float> buffer(context.numChannels, blockSize);
+
+    // The clock playback uses, driven by a request nobody typed. It is what
+    // makes the cursor a function of the sample count rather than a sum of
+    // per-block steps, which is the whole of why this renders the same at any
+    // block size.
+    TransportClock clock;
+
+    TransportSnapshot snapshot;
+    snapshot.tempo = tempo;
+
+    // Both left off, and named here rather than left to the defaults, because
+    // both would be wrong in a way that is hard to hear until it has shipped. A
+    // loop would print the range twice into a bounce that asked for it once,
+    // and the metronome is not the graph's to begin with: it is summed after
+    // the plan during playback and is not summed at all here.
+    snapshot.loop = {};
+    snapshot.click = {};
+
+    // Generation one, because the clock starts at zero and applies the first
+    // request it has not seen. A count-in is a monitoring aid and would print
+    // the beats before the range into the render.
+    snapshot.request.generation = 1;
+    snapshot.request.playing = true;
+    snapshot.request.locate = true;
+    snapshot.request.positionBeat = request.startBeat;
+    snapshot.request.countInBeats = 0.0;
+
+    const auto renderSpan = [&](std::int64_t samples) {
+        while (samples > 0) {
+            if (shouldContinue && !shouldContinue()) {
+                result.cancelled = true;
+                return;
+            }
+
+            const auto numSamples = static_cast<int>(std::min<std::int64_t>(samples, blockSize));
+
+            // Cleared here rather than by the executor, exactly as a callback
+            // clears it: the Output op sums into what it is handed, so what a
+            // plan with nothing routed to the output renders is silence only if
+            // silence is what was already there.
+            buffer.clear();
+
+            for (const auto& segment : clock.advance(snapshot, context.sampleRate, numSamples)) {
+                juce::AudioBuffer<float> piece(buffer.getArrayOfWritePointers(),
+                                               buffer.getNumChannels(), segment.startSample,
+                                               segment.block.numSamples);
+                executor.process(values, segment.block, piece);
+            }
+
+            sink.write(buffer, numSamples);
+
+            result.samplesRendered += numSamples;
+            ++result.blocksRendered;
+            samples -= numSamples;
+        }
+    };
+
+    renderSpan(rangeSamples);
+
+    if (result.cancelled || tailSamples <= 0)
+        return result;
+
+    // The tail, with the transport stopped where the range ended. A stopped
+    // block is already defined as one the graph runs through without the
+    // timeline moving, so reverbs and delays ring out and nothing that comes
+    // after the range starts playing.
+    //
+    // It is exactly as long as it was asked for. Stopping at the first silent
+    // block would be shorter and would sometimes be wrong: a gated or tremolo'd
+    // tail goes fully silent in the middle and would be cut off there. The
+    // caller has the samples and can trim what it does not want, which is a
+    // decision it can make and this cannot.
+    snapshot.request.generation = 2;
+    snapshot.request.playing = false;
+    snapshot.request.locate = false;
+
+    renderSpan(tailSamples);
+
+    return result;
+}
+
+}  // namespace magda::engine
--- a/magda/engine/exec/OfflineRender.cpp
+++ b/magda/engine/exec/OfflineRender.cpp
@@ -22,7 +22,16 @@ OfflineRenderResult renderOffline(PlanExecutor& executor, const PlanValues& valu
                                   const std::function<bool()>& shouldContinue) {
     OfflineRenderResult result;
 
-    if (!executor.isPrepared() || !executor.appliesValues(values)) {
+    // The context is refused on the same grounds the values are, and for the
+    // same reason: what the executor does with either when they do not belong
+    // to it is render something plausible and say nothing. Blocks are sized and
+    // the clock is advanced from the argument, while process() renders at most
+    // what it was prepared for, so a longer maxBlockSize prints the head of
+    // every block and silence for the rest. A sample rate that disagrees puts
+    // the whole render at the wrong speed, and a channel count that does at the
+    // wrong width.
+    if (!executor.isPrepared() || !executor.appliesValues(values) ||
+        !(executor.preparedContext() == context)) {
         result.refused = true;
         return result;
     }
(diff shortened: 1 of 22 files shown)


magda/engine/tap/LevelTap.hpp
```diff
@@ -0,0 +1,122 @@
...
+            const auto peak = std::max(std::abs(range.getStart()), std::abs(range.getEnd()));
+            accumulate(channel, peak);
+        }
+
+        // A mono block on a stereo tap reads as the same level on both, which is
+        // what it sounds like. Nothing in the engine produces one today; a tap
+        // that reported silence on the right if something did would be a meter
+        // lying about audible signal.
+        if (channels == 1)
+            accumulate(1, peaks_[0].load(std::memory_order_relaxed));
```
lucaromagnoli: [P2] Preserve the computed mono peak across a concurrent read

Loading channel 0 again here can lose the right-channel copy when `read()` interleaves after `accumulate(0, peak)`: the reader exchanges channel 0 to zero, this load observes zero, and the mono transient is returned as left-only and never reaches channel 1. Keep the block peak computed above and accumulate that same value into both slots for a mono block.

magda/engine/tap/SampleRing.hpp
```diff
@@ -0,0 +1,129 @@
...
+     * whatever the allocation held. Returns the running count of samples ever
+     * written, which is how a caller tells that nothing new has arrived since
+     * it last asked without comparing the audio itself.
+     */
+    std::size_t readLatest(float* destination, int numSamples) const {
+        const auto position = writePosition_.load(std::memory_order_acquire);
+
+        for (auto i = 0; i < numSamples; ++i) {
+            const auto index = static_cast<long long>(position) - numSamples + i;
+            destination[i] = index < 0 ? 0.0f : samples_[static_cast<std::size_t>(index) & mask_];
```
lucaromagnoli: [P1] Avoid concurrent non-atomic access to the ring slots

`readLatest()` reads `samples_` while the audio thread can overwrite the same wrapped slots. The release/acquire counter only orders writes completed before the published position; it does not make a concurrent later overwrite of a slot safe. Once the writer laps a stalled reader (a case the class explicitly allows), these plain `float` reads/writes form a C++ data race, so the behavior is undefined rather than merely a bounded visual tear. Please use a synchronization scheme that prevents overlapping slot access (or atomic/versioned slots) before exposing this as the cross-thread tap primitive.

magda/engine/tap/SampleRing.hpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// Audio thread. Append mono samples.
+    void write(const float* samples, int numSamples) {
+        if (numSamples <= 0)
+            return;
+
+        auto position = writePosition_.load(std::memory_order_relaxed);
+        for (auto i = 0; i < numSamples; ++i)
+            samples_[(position + static_cast<std::size_t>(i)) & mask_] = samples[i];
```
lucaromagnoli: This is the one place in the engine where a cross-thread data race is by design rather than by accident: the reader copies samples_ non-atomically while this store may be lapping it, and the class doc honestly bounds the effect. But bounded is not defined. In the letter of the standard the lap case is UB, and more practically it is exactly what the TSAN harness this engine adopted last slice exists to catch: the first analysis device wired to a ring hands that harness a genuine positive, and the choice then is a suppression that erodes the tool or a rewrite under pressure. The fix costs nothing here: both write loops are already scalar (the masked index prevents vectorization anyway), so storing through std::atomic<float> with relaxed ordering compiles to the plain loads and stores this already does on every target, and the documented tear becomes defined behavior with the same bound. Worth doing now while the class has no consumers, along with a threaded test mirroring the LevelTap one, since the SPSC claim is currently untested across threads.

magda/engine/io/FileAudioSource.hpp
```diff
@@ -0,0 +1,74 @@
...
+ * Never on the audio thread. It reads a file where it stands.
+ */
+
+namespace magda::engine {
+
+class FileAudioSource final : public EngineAudioSource {
+  public:
+    /// The reader is owned elsewhere, so the same file can back an offline
+    /// render and whatever else the host has open on it.
+    FileAudioSource(AudioFileReader& reader, const ClipPlacement& placement);
```
lucaromagnoli, line 39: The non-owning reference plus this comment invites a sharing the reader cannot survive. JuceAudioFileReader::read is stateful underneath (a seek and a read on one input stream), so two consumers reading one instance concurrently race inside JUCE, and the realistic path there is not exotic: stem export is one render per stem, naturally parallelized, all over the same files, and this line reads as license to hand them one reader. AudioFileReader's own contract does not help, because it still says reads happen on the prefetch thread, which this class has just made untrue. Two sentences close it structurally where it can be misread: here, that the same file backing two consumers is two readers over it; and in AudioFileReader.hpp, replacing the stale prefetch-thread line with the actual rule, one consumer per reader, any thread that is not the audio thread. The alternative of locking inside JuceAudioFileReader would serialize the prefetch thread against a bounce, which is the wrong trade.

magda/engine/tap/LevelTap.hpp
lucaromagnoli: Re-loading peaks_[0] here instead of using the peak just computed opens a window the CAS in accumulate was built to close: a read() landing between the channel-0 accumulate and this load exchanges the slot to zero, and the mono block's peak then never reaches channel 1 at all, so the two channels of a mono signal disagree about a transient one of them reported. The fix is strictly simpler than the current code: keep the block's own peak from the loop in a local and accumulate that into channel 1. Nothing produces a mono block today, as the comment says, which is exactly the argument for making the path correct while nothing depends on it.

lucaromagnoli (commented):
Reviewed the renderer, the offline source, the shared placement, both taps and the store integration in depth. The load-bearing claims verify. Block-size independence genuinely holds by construction: the range length is derived once from the tempo map, the clock's cursor is a function of the sample count from one anchor rather than a sum of per-block steps, and the offline source derives its file position from the timeline the same way the streaming one does, through the one shared function, which is the whole point of extracting it. The tail is correct at both ends: generation two with locate false is exactly the stop-in-place the clock grew last slice, so the cursor stands where the range ended, sources see playing false and render nothing, devices keep processing and ring out. The refusal path mirrors publish through the same appliesValues predicate, so an executor can never be driven offline with values it would render at unity. The half-open convention survives the sample grid: every rendered sample sits strictly inside the range, so a clip starting exactly at the range end contributes nothing, at any rounding.

The store side is right too. Meters ride the same keep-set rules as everything else, including the live-plan union that closed the coalescing use-after-free, so a tap the audio thread is writing through cannot be evicted; keyed by whole OpKey because a device and its meter share a DeviceId, retained by the model IDs the key names, asked of the factory again on decline so a mixer that opens later gets its meters at the next publish. The executor resolving the map to a flat per-op pointer table at prepare keeps the audio thread free of lookups, and writing taps on silent blocks so that only reading clears them is the correct metering policy, stated and tested. The agreement test is the strongest one in the PR, and finding it vacuous by mutation and moving the range onto the clip is exactly how that class of test earns its keep.

Two findings inline, one nit. Body-level nit not worth a thread: LevelTap::clear() has no caller in the engine, and the scenario its comment gives (a removed device's meter falling to silence) is actually handled by the tap being destroyed with the device; either a future slice wires it or the comment should say what it is really for.

magda/engine/tap/SampleRing.hpp
```diff
@@ -0,0 +1,144 @@
...
+     * filled once, so a display opened a moment ago draws silence rather than
+     * whatever the allocation held. Returns the running count of samples ever
+     * written, which is how a caller tells that nothing new has arrived since
+     * it last asked without comparing the audio itself.
+     */
+    std::size_t readLatest(float* destination, int numSamples) const {
+        const auto position = writePosition_.load(std::memory_order_acquire);
+
+        for (auto i = 0; i < numSamples; ++i) {
+            const auto index = static_cast<long long>(position) - numSamples + i;
```
lucaromagnoli, line 132: [P2] Define or enforce the maximum read size

Nothing constrains `numSamples` to `capacity_`, so after the ring has filled, asking for more than its capacity masks multiple logical indices onto the same slots and returns repeated copies as though they were older history. For example, reading 2048 samples from a 1024-slot ring duplicates the current ring once, which produces a misleading waveform/FFT rather than padding the unavailable prefix or rejecting the request. Please clamp/zero-pad here or make `numSamples <= capacity()` an enforced contract.

magda/engine/exec/OfflineRender.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+            sink.write(buffer, numSamples);
+
+            result.samplesRendered += numSamples;
+            ++result.blocksRendered;
+            samples -= numSamples;
+        }
+    };
+
+    renderSpan(rangeSamples);
```
lucaromagnoli, line 123: [P2] Apply the initial locate before rendering an empty range tail

When the beat range rounds to zero samples but `tailSeconds` is positive, `renderSpan(rangeSamples)` never calls `clock.advance()`, so generation 1 and its locate to `startBeat` are never applied. Generation 2 then stops with `locate = false` from the clock default at beat 0, rather than standing where this range ended as the API promises. A transport-aware effect therefore receives the tail at the wrong timeline position for requests such as `{4.0, 4.0, 0.5, ...}`. Prime the clock with the locate even when the range has no samples (or explicitly locate the stop in this case).

magda/engine/exec/OfflineRender.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+}  // namespace
+
+OfflineRenderResult renderOffline(PlanExecutor& executor, const PlanValues& values,
+                                  const RenderContext& context, const TempoMap& tempo,
+                                  const OfflineRenderRequest& request, OfflineRenderSink& sink,
+                                  const std::function<bool()>& shouldContinue) {
+    OfflineRenderResult result;
+
+    if (!executor.isPrepared() || !executor.appliesValues(values)) {
```
lucaromagnoli: [P1] Refuse a context the executor was not prepared for

This validates the values/plan pairing but not the independently supplied `context`. The renderer sizes its buffer and advances its clock from this argument, while `PlanExecutor::process()` clamps against its own prepared context. In release, an executor prepared for 512 samples paired with a context whose max is 4096 can therefore advance and write 4096-sample blocks while the executor renders only the first 512, silently inserting gaps into the bounce; sample-rate and channel mismatches are similarly inconsistent. Please expose/check the executor’s prepared context here and refuse a mismatch, just as mismatched values are refused.

tests/test_engine_taps.cpp
```diff
@@ -0,0 +1,274 @@
...
+    while (writing) {
+        const auto position = ring.readLatest(read.data(), static_cast<int>(read.size()));
+        REQUIRE(position >= lastPosition);
+        lastPosition = position;
+        ++reads;
+    }
+
+    writer.join();
+
+    CHECK(reads > 0);
```
lucaromagnoli, line 246: [P2] Do not require the reader thread to win the scheduling race

The writer can complete all 20,000 iterations before the creating thread first evaluates `while (writing)` (especially on a loaded or single-core CI worker). In that valid schedule `reads` remains zero and this test fails even though the ring is correct. Use a start barrier/handshake that guarantees an overlapping read, or remove this scheduling-dependent assertion; the final settled-state check still covers completion.

PR #2024, opened 2026-08-05
Native engine: the parallel executor (#2018)

### LINKED ISSUE #2018: Engine core slice 9: parallel executor
Slice 9 of #1889.

The executor that actually ships.

The plan already carries everything this needs, baked at compile time by `bakeScheduling`: dependency counts, consumer edges in CSR form, and the initially-ready set. Nothing in the plan format changes for it.

- Work-stealing or ready-queue scheduling over the baked constants, no graph walking on the callback.
- Bit-identical output against the reference executor for every fixture, at every thread count. Summing order is compiled, never scheduling order, which is what makes that possible.
- Buffer assignment (#2009) is already computed against the dependency DAG rather than the linear op order, so slot reuse holds under any valid schedule.
- Must match or beat TE CPU on the benchmark corpus before cutover, per #1882.

Tracked separately because it is a cutover blocker named in the epic's architecture with no issue of its own: #1898 is the anticipative executor, which is a different thing and lands after cutover.

lucaromagnoli:
Closes #2018. Slice 9 of #1889.

The executor that ships. It renders a plan across a pool of threads and comes out bit-identical to the reference executor at every thread count.

## How it stays identical

It shares everything except the loop. `PlanExecutor` keeps prepare, the buffer arena and the body of every op, and grew two seams for this: `beginBlock`, which settles what a block is, and `renderOp`, which renders one. The parallel executor calls those and adds a schedule.

- Anything that sums sums in compiled order, whatever order its inputs finished in.
- Buffer assignment (#2009) was already computed over the dependency DAG rather than the op list, so a slot reused under the reference executor's straight walk is safe under every other valid schedule.
- Output ops all add into one buffer, so a schedule deciding which went first would be a schedule deciding the arithmetic. They are left for whoever drove the block, in plan order, once the graph has drained.

## The schedule

Nothing walks the graph on the callback. The plan already carries dependency counts, reversed edges and the initially-ready set from `bakeScheduling`; a block copies the counts, seeds the ready set and runs. A plan that arrives without those constants is refused rather than baked here, where it would be computed twice.

The ready set is a lock-free stack threaded through a per-op link array, with a tag that moves on every push and pop and never resets across blocks: without it a thread that read the top, stalled through the end of a block and came back could find the same op there and mistake it for the one it was holding. A thread also carries straight on with the first consumer it releases rather than pushing it, so a linear chain never touches the stack.

`RenderThreadPool` owns the threads and outlives plans. Realtime priority where the OS grants it, the highest ordinary priority where it does not, because a worker that never started is one the audio thread waits on for nothing. Retiring an epoch waits out any worker still inside it, on the thread that is allowed to wait.

## A race TSan found

Making a `juce::dsp::AudioBlock` out of a `juce::AudioBuffer` writes the buffer's `isClear` flag, so two ops reading the same port were two threads writing one byte: benign in value, a data race in fact. Ports are read through cached channel pointers now, and nothing on the render path touches an `AudioBuffer`.

## Session

`EngineSession` renders through this executor, with the pool optional. A null pool is not a degraded mode, it is the same executor with one thread, which is what makes "identical at every thread count" a claim about one code path rather than two.

## Verification

- Seven scenes (one chain, eight tracks, sends into an aux, a rack's chains, latency, MIDI, a group track) at 1, 2, 3, 4 and 8 threads, compared with `==` against the reference executor. Also with ragged block lengths, and across a plan swap that carries delay lines.
- A rendezvous test that two ops on independent branches really are inside `process()` at the same time, since everything else would pass on a pool whose workers never woke.
- Confirmed to fail without their fix: a pool with no workers takes the concurrency test from two devices at once to one; seeding the ready set with every op rather than the ready ones breaks 55 assertions across the comparison scenes.
- The whole engine suite is clean under ThreadSanitizer. `make test-tsan` builds and runs the Catch2 tests under it, with `TEST=<filter>` to narrow.
- Scaling, debug build, 200 blocks over 16 tracks of 4 busy devices: 315 ms on one thread, 167 on two, 96 on four, 76 on eight. Hidden `[bench]` test.

## Not in this slice

"Match or beat TE CPU on the benchmark corpus" is a cutover gate and there is no corpus in the repo yet; it arrives with the null-diff harness (#1896). The `[bench]` test is the measuring stick for it when it does.

### DIFF
--- a/magda/engine/exec/ParallelPlanExecutor.cpp
+++ b/magda/engine/exec/ParallelPlanExecutor.cpp
@@ -1,0 +1,231 @@
+#include "exec/ParallelPlanExecutor.hpp"
+
+namespace magda::engine {
+namespace {
+
+/// Failed pops before a thread stops asking and lets the scheduler have the
+/// core. The block is not over (something is still running, or nothing would be
+/// waiting), so leaving is not an option; what this decides is only how hard a
+/// thread with nothing to do leans on the ready set while it waits.
+constexpr int kSpinsBeforeYield = 200;
+
+constexpr std::uint64_t packReady(OpId op, std::uint32_t tag) {
+    return (static_cast<std::uint64_t>(tag) << 32) | static_cast<std::uint32_t>(op);
+}
+
+constexpr OpId readyOp(std::uint64_t packed) {
+    return static_cast<OpId>(static_cast<std::int32_t>(packed & 0xffffffffULL));
+}
+
+constexpr std::uint32_t readyTag(std::uint64_t packed) {
+    return static_cast<std::uint32_t>(packed >> 32);
+}
+
+static_assert(packReady(INVALID_OP_ID, 0) == kEmptyReadyStack,
+              "an executor that has not started a block must not look like one with op 0 ready");
+
+/// Whether the plan carries a schedule this executor can run: the counts, the
+/// reversed edges and the ready set, all agreeing about the same op count.
+/// bakeScheduling produces exactly this, and nothing else does.
+bool hasSchedule(const RenderPlan& plan) {
+    const auto numOps = plan.ops.size();
+    return plan.dependencyCounts.size() == numOps && plan.consumerOffsets.size() == numOps + 1 &&
+           plan.consumerEdges.size() == static_cast<std::size_t>(plan.consumerOffsets.empty()
+                                                                     ? 0
+                                                                     : plan.consumerOffsets.back());
+}
+
+}  // namespace
+
+void ParallelPlanExecutor::letGoOfPool() {
+    // A worker can still be inside the last block's takeWork(): render() returns
+    // as soon as the block is finished, not as soon as everyone has noticed. It
+    // is inside this object while it is, so nothing here may be freed or resized
+    // until the pool says otherwise.
+    //
+    // Skipped for a job the pool has never been given, which is every executor
+    // that has not rendered yet. Letting go waits out whatever is in a job right
+    // now, and making a publish wait a block for an executor no thread could
+    // possibly be inside would be a stall bought for nothing.
+    if (pool_ != nullptr && handedToPool_.load(std::memory_order_relaxed))
+        pool_->release(*this);
+    handedToPool_.store(false, std::memory_order_relaxed);
+}
+
+ParallelPlanExecutor::~ParallelPlanExecutor() {
+    letGoOfPool();
+}
+
+std::vector<std::string> ParallelPlanExecutor::prepare(const RenderPlan& plan,
+                                                       const PlanBindings& bindings,
+                                                       const RenderContext& context,
+                                                       const ParallelPlanExecutor* previous) {
+    // Everything below reallocates what a worker still finishing the last block
+    // would read, so this comes first, the same as at destruction.
+    letGoOfPool();
+
+    plan_ = nullptr;
+    outputOps_.clear();
+
+    auto messages =
+        core_.prepare(plan, bindings, context, previous != nullptr ? &previous->core_ : nullptr);
+    if (!core_.isPrepared())
+        return messages;
+
+    if (!hasSchedule(plan)) {
+        // Refused rather than worked around. Baking it here would put the one
+        // thing every block depends on somewhere it is computed twice, and a
+        // plan without it did not come from the compiler.
+        core_.reset();
+        messages.push_back(
+            "plan carries no baked scheduling: dependency counts and consumer edges are missing "
+            "or do not match its ops, so it cannot be scheduled");
+        return messages;
+    }
+
+    const auto numOps = plan.ops.size();
+    pending_ = std::vector<std::atomic<std::uint16_t>>(numOps);
+    nextReady_ = std::vector<std::atomic<OpId>>(numOps);
+
+    for (std::size_t i = 0; i < numOps; ++i)
+        if (plan.ops[i].kind == OpKind::Output)
+            outputOps_.push_back(static_cast<OpId>(i));
+
+    plan_ = &plan;
+    return messages;
+}
+
+void ParallelPlanExecutor::push(OpId op) {
+    const auto index = static_cast<std::size_t>(op);
+    auto top = readyTop_.load(std::memory_order_relaxed);
+    std::uint64_t pushed = 0;
+
+    do {
+        nextReady_[index].store(readyOp(top), std::memory_order_relaxed);
+        pushed = packReady(op, readyTag(top) + 1);
+    } while (!readyTop_.compare_exchange_weak(top, pushed, std::memory_order_release,
+                                              std::memory_order_relaxed));
+}
+
+OpId ParallelPlanExecutor::pop() {
+    auto top = readyTop_.load(std::memory_order_acquire);
+
+    for (;;) {
+        const auto op = readyOp(top);
+        if (op == INVALID_OP_ID)
+            return INVALID_OP_ID;
+
+        // Read under the tag: if the top is still the word this came from, the
+        // link is the one whoever pushed it wrote, and if it is not, the swap
+        // below fails and none of this counted.
+        const auto next = nextReady_[static_cast<std::size_t>(op)].load(std::memory_order_relaxed);
+        if (readyTop_.compare_exchange_weak(top, packReady(next, readyTag(top) + 1),
+                                            std::memory_order_acquire, std::memory_order_acquire))
+            return op;
+    }
+}
+
+OpId ParallelPlanExecutor::runOp(OpId op) {
+    // Output ops are the exception the whole schedule is written around: they
+    // add into a buffer they all share, so they are left for the thread that
+    // drove the block to run in plan order. Counted here anyway, because what
+    // waits on what is the plan's business and not this decision's.
+    if (plan_->ops[static_cast<std::size_t>(op)].kind != OpKind::Output)
+        core_.renderOp(op, valueOf(op), block_, *output_);
+
+    OpId carryOn = INVALID_OP_ID;
+
+    const auto first = plan_->consumerOffsets[static_cast<std::size_t>(op)];
+    const auto last = plan_->consumerOffsets[static_cast<std::size_t>(op) + 1];
+    for (auto edge = first; edge < last; ++edge) {
+        const auto consumer = plan_->consumerEdges[static_cast<std::size_t>(edge)];
+
+        // Acquire-release, not relaxed: this is where a consumer inherits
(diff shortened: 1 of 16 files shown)


magda/engine/exec/ParallelPlanExecutor.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+static_assert(packReady(INVALID_OP_ID, 0) == kEmptyReadyStack,
+              "an executor that has not started a block must not look like one with op 0 ready");
+
+/// Whether the plan carries a schedule this executor can run: the counts, the
+/// reversed edges and the ready set, all agreeing about the same op count.
+/// bakeScheduling produces exactly this, and nothing else does.
+bool hasSchedule(const RenderPlan& plan) {
+    const auto numOps = plan.ops.size();
+    return plan.dependencyCounts.size() == numOps && plan.consumerOffsets.size() == numOps + 1 &&
```
lucaromagnoli: [P1] Validate the schedule contents before accepting the plan

This only checks container sizes, so a schedule that is present but inconsistent still passes preparation. For example, an out-of-range entry in `initialReadyOps` indexes past `nextReady_` in `push()`, non-monotonic/out-of-range offsets can walk outside `consumerEdges`, and a dependency count that does not match the reversed edges leaves `remaining_` nonzero with no ready work, spinning the audio callback forever. `validatePlan()` does not inspect these baked arrays, and their contents are not covered by the plan fingerprint. Please compare all four constants with the topology (or otherwise fully validate IDs, CSR offsets/edges, counts, and the initial-ready set) off the audio thread before setting `plan_`.

magda/engine/exec/RenderThreadPool.cpp
```diff
@@ -0,0 +1,105 @@
...
+void RenderThreadPool::release(Job& job) {
+    // Only this job: another one may have been published since, and taking that
+    // one down with it would silence an epoch that is rendering.
+    auto* expected = &job;
+    job_.compare_exchange_strong(expected, nullptr, std::memory_order_seq_cst);
+
+    // Anything inside a job now is either this one, which has to be waited out,
+    // or a newer one, which is a block and will end. Either way this is a wait
+    // measured in one callback, on a thread that is allowed to wait.
+    while (inJob_.load(std::memory_order_seq_cst) != 0)
```
lucaromagnoli: [P1] Do not wait for workers belonging only to newer jobs

`inJob_` is pool-global, so after the CAS loses to a newer `job_`, this loop waits for workers rendering that newer epoch too. With back-to-back callbacks there is no guarantee it ever observes zero: W1 can enter the next block while W2 is still leaving the previous one, then W2 can enter the following block while W1 leaves, indefinitely. The publisher can therefore spin forever even though every worker that could have loaded the retired `job` has already gone. Track entrants by job/generation (or use an epoch-specific active count/acknowledgement) so `release(job)` waits only for workers that may still dereference that job and cannot be extended by future renders.

magda/engine/exec/ParallelPlanExecutor.cpp
lucaromagnoli: This checks three of the four constants bakeScheduling produces, and only their sizes. initialReadyOps is not looked at, and it is the one whose inconsistency does not misrender but hangs: a zero-dependency op missing from the seed list is never pushed and never decremented, because nothing produces into it, so remaining_ never reaches zero and takeWork() spins the audio thread forever, workers included. The size checks have the same gap in a milder form: matching lengths prove the arrays exist, not that they describe this plan. Counts too high hang the same way; counts too low release an op before its producers finished, which is a data race on a shared slot and wrong audio, the exact things the schedule exists to prevent. All four constants are pure functions of op.inputs and this runs off the audio thread, so the structural fix is cheap: recompute them here (or bake into a scratch plan and compare), and refuse on any disagreement, making "did not come from the compiler" the thing actually verified rather than inferred from array lengths. Telling detail: the existing refusal test clears dependencyCounts and passes, while a test for the initialReadyOps hole cannot be written today because it would hang the test binary, which is the whole argument for the check.

magda/engine/exec/RenderThreadPool.cpp
lucaromagnoli: The comment above says this wait is measured in one callback, and for the newer-job case that is not quite what the code does: inJob_ counts workers inside any job, so release() must catch a window in which no worker is inside the new epoch either, and every callback reopens the occupancy. Under sustained rendering the window is the gap between callbacks, which polling at yield granularity will catch in practice, so this is a latency wrinkle on the publishing thread rather than a hazard; but it does mean a heavily loaded session pays an unbounded-in-principle number of blocks to retire an epoch. If it ever shows up in publish latency, the shape that fixes it is counting per job rather than per pool (the job carries its own inside count, the pool's takeWork increments the one it entered), after which release() waits only for the job it is releasing and the comment becomes true as written. At minimum the comment should say what is actually waited for.

lucaromagnoli (commented):
Reviewed the scheduler, the pool, the seam refactor and the session integration in depth, with most of the time on the memory ordering, since a lock-free scheduler is only as good as its happens-before graph. That graph is sound. The last producer to decrement a consumer has seen every other producer's writes because the acq_rel decrements on one counter form a release sequence, and its push (or hand-over on the same thread) passes all of it to whoever takes the op; the driving thread leaving on remaining_ reaching zero synchronizes with every decrement in the chain for the same reason, so the output ops it then runs see every worker's writes. Block state is published through the seed pushes, and an op is only ever reached through an acquire on the stack or on the same thread, so nothing reads block_ or pending_ before the release that made them current. The pool handshake is a correct Dekker: a worker announces before it reads the job, release() nulls before it reads the count, both seq_cst, so a worker on its way in is either seen or sees null. The tag argument holds, including the deliberately unsettling cases: a worker stalled since the last block that pops a stale word fails the CAS on the moved tag, and a worker that wakes late and renders the whole of the next block before render() was even called is doing correct work with correctly published data. The isClear fix is the right shape, one flat pointer table taken at prepare rather than a per-call workaround, and the silence row falling out of the same array is tidy.

The identity argument is airtight because it is structural: renderOp is a pure extraction (I diffed the bodies), sums are compiled order with the schedule unable to reorder them, and output ops run on the driving thread in plan order. The test posture matches the claim: scenes as recipes so state starts equal, == with no tolerance, the rendezvous test that a sleeping pool cannot pass, determinism across repeated runs, and the swap test exercising release() with a live pool. Refusal symmetry through the session verifies too: an unschedulable plan leaves isPrepared false and the previous epoch playing.

Two findings inline. The first is the one I care about: the refusal path this PR builds its safety story on checks three of the four baked constants, and the unchecked one is the one whose absence hangs the callback rather than misrendering.

lucaromagnoli (commented):
Second round, on the fix commit. Both land, and both got the structural treatment rather than the minimal one.

carriesSchedule is the right generalization: recompute-and-compare through the same scheduleOf that bakes, so the check cannot drift from the producer, and plain vector equality quietly subsumes every range check too, since an out-of-range seed or edge can never equal what the topology implies. The refusal test growing a control section ("nothing wrong with it") and a REQUIRE_FALSE(carriesSchedule) before each prepare is exactly the discipline that keeps eight mutation sections honest: a section that meant to break something and did not fails loudly instead of agreeing quietly. And the section the old size check could never catch, the missing seed op, now runs instead of hanging the binary, which is the whole finding closed in one line of test.

The per-job count survives the same scrutiny the original handshake got. The load-bearing property is that a worker holds arriving_ from before its job_ load until after its workersInside increment, so the two waits in release(), in that order, cannot miss a worker: one that read the pointer before the CAS is still pinned in the arrival window when the first wait runs, and its increment is therefore visible to the second. The arrival window holding nothing but a load and an increment is what makes the first wait scheduling-bounded rather than work-bounded, which is the distinction the old pool-wide count could not make. The last thing a worker touches is the fetch_sub itself, so the epoch destructor running the moment the count reaches zero is the shared_ptr control-block pattern and safe. The driving thread staying uncounted is right, since the caller already owes render() having returned, and counting it would make release() deadlock against the contract that excuses it.

The churn test is the shape the finding described, and its one subtle claim holds: by the time rendered catches up to published, the render thread has replaced its shared_ptr under the lock and dropped the old epoch before processing the new one, so retiring.reset() really is the last reference and the wait really runs on the publishing thread, while the pool is busy.

Nothing further. Done from my side; slice 9 closes with the executor's safety story verified end to end rather than inferred, which is where this epic keeps insisting the bugs would have hidden.

magda/engine/exec/RenderThreadPool.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // Two waits, and they are for different things. The first is for workers
+    // that read the job pointer before it was taken away and have not yet
+    // counted themselves against it; nothing but a load and an increment lives
+    // in that window, so it drains as fast as those threads get a core. The
+    // second is for workers actually inside this job, which is bounded by the
+    // block they are finishing.
+    //
+    // In this order, because a worker still arriving is one that has not
+    // reached the second count yet.
+    while (arriving_.load(std::memory_order_seq_cst) != 0)
```
lucaromagnoli: [P2] Make the arrival drain immune to later jobs too

The per-job count fixes the long wait, but this first wait is still a pool-global zero-crossing that future callbacks can extend. A valid schedule can keep `arriving_` nonzero by alternating workers (W1 increments, W2 increments before W1 decrements, W1 enters the next latched notification before W2 decrements, and so on), so retirement still has no finite bound under sustained rendering—it just depends on catching a much shorter window now. A monotonic started/finished ticket pair would let `release()` snapshot the arrivals begun before its CAS and wait only through that ticket; workers arriving afterwards cannot have loaded the retired pointer and would not move the target.

lucaromagnoli (commented):
Third pass on ffdce48: the per-worker arrival marks close the remaining retirement-liveness hole.\n\nThe ordering holds in both cases at each worker slot. If release() observes an odd mark, that worker announced before loading job_; waiting for that exact mark to change carries it through the old job workersInside increment before the final per-job drain. If release() observes an even mark, that observation is after the CAS in the sequentially-consistent order, so the worker next announcement and subsequent job_ load cannot acquire the retired pointer. Scanning workers one at a time can wait for a newer arrival, but only the single arrival whose odd mark was sampled; later callbacks cannot renew that target.\n\nThe newer-job case is sound too: a failed CAS means another job is already published, and any worker counted against the retired job is still caught either by its sampled arrival or by the retired job own count. Counter wrap does not change odd/even window state, and equality on the sampled odd value remains conservative.\n\nNo further findings in this update. Static review only, as requested.

PR #2025, opened 2026-08-05
Remote API: authenticated JSON-RPC 2.0 over WebSocket (#1856)

### LINKED ISSUE #1856: WebSocket API: authenticated JSON-RPC server and lifecycle

lucaromagnoli:
Closes #1856. Builds on #1855's dispatcher; nothing here reimplements an operation.

## The dependency spike said no to CHOC

The issue named `choc_HTTPServer.h` as the thing to evaluate. It cannot refuse a connection: `upgradeToWebsocket` starts the handshake and *then* calls `upgradedToWebSocket(target)`, which gets the request path and nothing else — no `Origin`, no `Authorization`, no return value to refuse with. `ClientInstance` has no close, so one bad client can only be dropped by tearing down the whole listener. Two acceptance criteria here are about failing closed, so it would need a patched copy of the header **plus** the first Boost in this tree — brew on macOS, `libboost-dev` on Linux, a vcpkg port on Windows, and a README prerequisite change on all three.

websocketpp is dormant (last release April 2020, one commit since) and still needs Asio. Hand-rolling on `juce::StreamingSocket` was the fallback until it turned out the codebase's networking is all `juce::URL` — an HTTP client that cannot serve — so the consistency would have been with the namespace, not with anything structural, at a price of ~500 lines of framing plus a client implementation for the tests.

**cpp-httplib** was already in the tree, vendored by llama.cpp. MIT, no required dependencies, and its upgrade path consults the pre-routing handler before sending the 101 exactly so authentication can answer 401 instead. It ships a WebSocket client too, so the tests drive real sockets with no test-only dependency. Pinned at `v0.52.0` via FetchContent rather than reached for inside llama.cpp's vendor directory, which Dependabot bumps. Full reasoning in `docs/remote-api-transport.md`.

## What it does

`RemoteWebSocketServer` parses frames, decides who may connect, and hands everything to `RemoteApiService::dispatch`. It knows no operation names and reaches for no manager — the "no direct calls to `TrackManager`/`ClipManager`" criterion holds structurally rather than by discipline.

Refusal happens before the protocol switches: bad token → 401, unlisted browser Origin → 403 — in each case a failed HTTP request and no socket. The connection cap is exact but its *refusal* is best-effort: usually a 503 before the handshake, though a client winning the race for the last slot sees the handshake succeed and a `PolicyViolation` close immediately after. The cap is claimed on the connection's own thread rather than reserved ahead of it, because a reservation has to be handed back along every path cpp-httplib can take to never reach a handler, and one missed path strands a slot for the life of the process. A request with no `Origin` at all is a native client, not an unrecognised page, and is allowed. Token comparison is constant-time.

The message thread does no I/O. Listener thread, a thread per connection, and a writer per connection draining an outbox. The writer is the point: a dispatch completion arrives on the message thread, where writing the reply would put a socket write on the UI's thread, so it appends bytes and returns. Completions hold a `shared_ptr` to per-connection state, so one outliving its socket drops its payload rather than reaching for a returned stack frame.

`RemoteApiHost` owns the dispatcher, model bridge and transport as one lifetime — the first thing to instantiate `RemoteApiService`. Off by default in config. The bearer token is generated per run from `std::random_device`, published to `remote-api.json` in the data dir with the port beside it, and deleted on shutdown; never persisted to config, which gets copied around and pasted into bug reports.

## Wire format

    → {"jsonrpc":"2.0","id":7,"method":"tracks.rename","params":{...},"meta":{"expectedRevision":42}}
    ← {"jsonrpc":"2.0","id":7,"result":{...},"meta":{"revision":43,"apiVersion":"..."}}

`meta` is a sibling of `params` rather than inside it because operation schemas are `additionalProperties:false` and would reject `expectedRevision` as unknown. Replies mirror that: `result` is the operation's output verbatim. Folding meta into the result breaks on the first operation that doesn't return one — `tracks.list` answers with a bare array — and wrapping only those would make the reply shape depend on which operation was called. Failures put the revision in `error.data`, so a client that lost an optimistic-concurrency race learns the real revision in the message that rejects its write.

## Tests

21 Catch2 cases over real loopback sockets (auth, Origin, JSON-RPC error mapping, connection cap, oversized frame, shutdown, token lifecycle) and 4 `juce::UnitTest` cases against the live facade, where a create over the wire produces a track `TrackManager` can be asked about by id. The live ones run each request on a worker thread while pumping the message loop — dispatch hops to the message thread, so a blocking read on it would wait for a handler that cannot run.

Suite: 2050 passed, 13 skipped, 0 failed. The trailing `juce_Timer.cpp:99` assertion is pre-existing and reproduces with all remote tests excluded.

Two bugs the tests caught, both worth naming:

- `bind_to_any_port(host, socket_flags)` takes socket flags second, not a port — a configured port was silently ignored and passed as flags.
- The token file came out `0644`. `replaceWithText` writes a temp file and renames it over the target, so the new inode arrives with default permissions and silently undoes the `chmod`. Now written through a `FileOutputStream` into the already-restricted file.

## Notes for review

- The frame cap is a compile-time constant because it governs how much is read into memory before a handler sees anything; enforcing it only in the handler would mean buffering 16 MB of a frame we intend to reject.
- cpp-httplib's `*_IF_AVAILABLE` options default ON and link whatever OpenSSL/zlib/Brotli/zstd is installed. Forced off, so builds don't differ per machine.
- WebSocket support is newer in cpp-httplib than in Beast. Mitigated by the threat model — loopback, token-gated, hard frame cap — but it is the one axis where the chosen option is weaker than the rejected one.

### DIFF
--- a/magda/daw/api/remote_api_host.cpp
+++ b/magda/daw/api/remote_api_host.cpp
@@ -8,8 +8,14 @@
 #include "remote_service.hpp"
 #include "remote_websocket_server.hpp"
 
-#if !JUCE_WINDOWS
+#if JUCE_WINDOWS
+    #include <windows.h>
+#else
     #include <sys/stat.h>
+    #include <unistd.h>
+
+    #include <cerrno>
+    #include <csignal>
 #endif
 
 namespace magda {
@@ -17,7 +23,74 @@ namespace remote {
 
 namespace {
 
-constexpr const char* kTokenFileName = "remote-api.json";
+/**
+ * Discovery records are per process, not per installation.
+ *
+ * MAGDA does not set `moreThanOneInstanceAllowed()` to false, so two instances
+ * are a supported thing to run. Sharing one file between them means the second
+ * to start hides the first, and the first to stop deletes the record of the one
+ * still running. With a fixed port it is worse, because cpp-httplib sets
+ * SO_REUSEPORT and both listeners bind: a client then holds one instance's token
+ * and may be routed to the other, which does not accept it.
+ *
+ * Naming the record after the process keeps each instance's file its own, and
+ * the pid inside is what makes cleanup safe — an instance deletes only its own,
+ * and a record whose process no longer exists is debris anyone may collect.
+ */
+constexpr const char* kTokenFilePrefix = "remote-api-";
+constexpr const char* kTokenFileSuffix = ".json";
+constexpr const char* kTokenFilePattern = "remote-api-*.json";
+
+juce::int64 currentProcessId() {
+#if JUCE_WINDOWS
+    return static_cast<juce::int64>(GetCurrentProcessId());
+#else
+    return static_cast<juce::int64>(getpid());
+#endif
+}
+
+bool isProcessAlive(juce::int64 processId) {
+#if JUCE_WINDOWS
+    auto* handle =
+        OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, static_cast<DWORD>(processId));
+    if (handle == nullptr)
+        return false;
+    DWORD exitCode = 0;
+    const auto running = GetExitCodeProcess(handle, &exitCode) != 0 && exitCode == STILL_ACTIVE;
+    CloseHandle(handle);
+    return running;
+#else
+    // Signal 0 runs the existence and permission checks without delivering
+    // anything. EPERM means the process is there and belongs to someone else,
+    // which still counts as alive.
+    return kill(static_cast<pid_t>(processId), 0) == 0 || errno == EPERM;
+#endif
+}
+
+/**
+ * @brief Delete records left by instances that are no longer running.
+ *
+ * A crashed run cannot clean up after itself, and with per-process records
+ * nobody else was going to. Ours is left alone here — `start()` handles that —
+ * and so is every record whose process still answers, because that one belongs
+ * to a live instance.
+ *
+ * A pid can in principle be recycled onto an unrelated process, which would keep
+ * one dead record alive a while longer. The cost of that is a file, and the next
+ * sweep after the recycled process exits takes it.
+ */
+void sweepAbandonedRecords(const juce::File& directory, juce::int64 ownProcessId) {
+    for (const auto& record :
+         directory.findChildFiles(juce::File::findFiles, false, kTokenFilePattern)) {
+        const auto owner = record.getFileNameWithoutExtension()
+                               .fromLastOccurrenceOf("-", false, false)
+                               .getLargeIntValue();
+        if (owner <= 0 || owner == ownProcessId)
+            continue;
+        if (!isProcessAlive(owner))
+            record.deleteFile();
+    }
+}
 
 /**
  * @brief 256 bits of entropy, hex encoded.
@@ -60,6 +133,9 @@ bool writeTokenFile(const juce::File& file, const juce::String& token, int port)
     payload->setProperty("port", port);
     payload->setProperty("token", token);
     payload->setProperty("url", "ws://127.0.0.1:" + juce::String(port) + "/rpc");
+    // Whose listener this is. A reader can tell two live instances apart, and
+    // cleanup can tell an abandoned record from someone else's live one.
+    payload->setProperty("pid", currentProcessId());
 
     // Write into the file we just restricted, rather than replaceWithText: that
     // writes a temp file and moves it over the target, so the new inode arrives
@@ -88,14 +164,18 @@ bool RemoteApiHost::start() {
     if (server_ != nullptr && server_->isRunning())
         return true;
 
-    // Anything here predates this listener, and a token file without a listener
-    // behind it is worse than no file: it hands a client a port and a credential
-    // that a crashed previous run left behind. Clearing it before every early
-    // return — disabled in config, no token, a port already taken — keeps the
-    // rule that the file exists only while something is answering on it. The
-    // already-running check comes first so a second call cannot delete a file
-    // that is still good.
+    // Our own record predates this listener, and a record without a listener
+    // behind it is worse than none: it hands a client a port and a credential
+    // that a crashed previous run left behind, and if the OS has reissued that
+    // port it points them at a stranger. Clearing it before every early return —
+    // disabled in config, no token, a port already taken — keeps the rule that a
+    // record exists only while something is answering on it. The already-running
+    // check comes first so a second call cannot delete a record still in use.
     tokenFile().deleteFile();
+
+    // Other instances' records are theirs to remove; only the ones whose process
+    // has gone are ours to sweep.
+    sweepAbandonedRecords(paths::dataDir(), currentProcessId());
 
     auto& config = Config::getInstance();
     if (!config.getRemoteApiEnabled())
@@ -159,7 +239,8 @@ int RemoteApiHost::boundPort() const {
 }
 
 juce::File RemoteApiHost::tokenFile() const {
-    return paths::dataDir().getChildFile(kTokenFileName);
+    return paths::dataDir().getChildFile(juce::String(kTokenFilePrefix) +
+                                         juce::String(currentProcessId()) + kTokenFileSuffix);
 }
 
 RemoteApiService& RemoteApiHost::service() {
--- a/magda/daw/api/remote_api_host.cpp
+++ b/magda/daw/api/remote_api_host.cpp
@@ -85,12 +85,21 @@ RemoteApiHost::~RemoteApiHost() {
 }
 
 bool RemoteApiHost::start() {
+    if (server_ != nullptr && server_->isRunning())
+        return true;
+
+    // Anything here predates this listener, and a token file without a listener
+    // behind it is worse than no file: it hands a client a port and a credential
+    // that a crashed previous run left behind. Clearing it before every early
+    // return — disabled in config, no token, a port already taken — keeps the
+    // rule that the file exists only while something is answering on it. The
+    // already-running check comes first so a second call cannot delete a file
+    // that is still good.
+    tokenFile().deleteFile();
+
     auto& config = Config::getInstance();
     if (!config.getRemoteApiEnabled())
         return false;
-
-    if (server_ != nullptr && server_->isRunning())
-        return true;
 
     token_ = generateToken();
     if (token_.isEmpty()) {
--- a/magda/daw/api/remote_api_host.cpp
+++ b/magda/daw/api/remote_api_host.cpp
@@ -1,0 +1,161 @@
+#include "remote_api_host.hpp"
+
+#include <random>
+
+#include "AppPaths.hpp"
+#include "Config.hpp"
+#include "remote_model_bridge.hpp"
+#include "remote_service.hpp"
+#include "remote_websocket_server.hpp"
+
+#if !JUCE_WINDOWS
+    #include <sys/stat.h>
+#endif
+
+namespace magda {
+namespace remote {
+
+namespace {
+
+constexpr const char* kTokenFileName = "remote-api.json";
+
+/**
+ * @brief 256 bits of entropy, hex encoded.
+ *
+ * `std::random_device` rather than `juce::Random`, because this is a credential:
+ * JUCE's generator is a seeded PRNG whose output is predictable to anyone who
+ * can guess the seed, while `random_device` is backed by the OS entropy source
+ * on every platform MAGDA ships to.
+ */
+juce::String generateToken() {
+    std::random_device entropy;
+    juce::String token;
+    for (int i = 0; i < 8; ++i)
+        token += juce::String::toHexString(static_cast<int>(entropy())).paddedLeft('0', 8);
+    return token;
+}
+
+/**
+ * @brief Publish the token where a local client can find it, and only it.
+ *
+ * Created empty and restricted *before* anything secret goes in, so the token
+ * is never briefly world-readable. On Windows the file inherits the ACL of the
+ * per-user app data directory, which is already owner-only.
+ */
+bool writeTokenFile(const juce::File& file, const juce::String& token, int port) {
+    file.getParentDirectory().createDirectory();
+    file.deleteFile();
+    if (!file.create().wasOk())
+        return false;
+
+#if !JUCE_WINDOWS
+    if (chmod(file.getFullPathName().toRawUTF8(), S_IRUSR | S_IWUSR) != 0) {
+        // Refuse rather than leave a readable credential behind.
+        file.deleteFile();
+        return false;
+    }
+#endif
+
+    auto* payload = new juce::DynamicObject();
+    payload->setProperty("port", port);
+    payload->setProperty("token", token);
+    payload->setProperty("url", "ws://127.0.0.1:" + juce::String(port) + "/rpc");
+
+    // Write into the file we just restricted, rather than replaceWithText: that
+    // writes a temp file and moves it over the target, so the new inode arrives
+    // with default permissions and the chmod above is silently undone.
+    juce::FileOutputStream stream(file);
+    if (!stream.openedOk() ||
+        !stream.writeText(juce::JSON::toString(juce::var(payload), false), false, false, nullptr)) {
+        file.deleteFile();
+        return false;
+    }
+    stream.flush();
+    return true;
+}
+
+}  // namespace
+
+RemoteApiHost::RemoteApiHost(MagdaApi& api)
+    : service_(std::make_unique<RemoteApiService>(api)),
+      bridge_(std::make_unique<ModelChangeBridge>(*service_)) {}
+
+RemoteApiHost::~RemoteApiHost() {
+    stop();
+}
+
+bool RemoteApiHost::start() {
+    auto& config = Config::getInstance();
+    if (!config.getRemoteApiEnabled())
+        return false;
+
+    if (server_ != nullptr && server_->isRunning())
+        return true;
+
+    token_ = generateToken();
+    if (token_.isEmpty()) {
+        DBG("RemoteApiHost: could not generate a token; not starting");
+        return false;
+    }
+
+    RemoteWebSocketServer::Options options;
+    options.bearerToken = token_;
+    options.port = config.getRemoteApiPort();
+    for (const auto& origin : config.getRemoteApiAllowedOrigins())
+        options.allowedOrigins.push_back(juce::String::fromUTF8(origin.c_str()));
+
+    server_ = std::make_unique<RemoteWebSocketServer>(*service_, options);
+    if (!server_->start()) {
+        server_.reset();
+        token_ = {};
+        return false;
+    }
+
+    // A running listener whose token nobody can read is useless and still a
+    // listener, so a failure to publish takes the server down with it.
+    if (!writeTokenFile(tokenFile(), token_, server_->boundPort())) {
+        DBG("RemoteApiHost: could not write " + tokenFile().getFullPathName());
+        server_->stop();
+        server_.reset();
+        token_ = {};
+        return false;
+    }
+
+    juce::Logger::writeToLog(
+        "Remote API listening on ws://127.0.0.1:" + juce::String(server_->boundPort()) + "/rpc");
+    return true;
+}
+
+void RemoteApiHost::stop() {
+    if (server_ != nullptr) {
+        server_->stop();
+        server_.reset();
+    }
+    // The token dies with the run that generated it; leaving the file behind
+    // would advertise a port that is no longer listening and a credential that
+    // no longer works.
+    tokenFile().deleteFile();
+    token_ = {};
+
+    if (service_ != nullptr)
+        service_->shutdown();
+}
+
+bool RemoteApiHost::isRunning() const {
+    return server_ != nullptr && server_->isRunning();
+}
+
+int RemoteApiHost::boundPort() const {
+    return server_ != nullptr ? server_->boundPort() : 0;
+}
+
+juce::File RemoteApiHost::tokenFile() const {
+    return paths::dataDir().getChildFile(kTokenFileName);
+}
+
+RemoteApiService& RemoteApiHost::service() {
+    return *service_;
+}
+
+}  // namespace remote
+}  // namespace magda
--- a/magda/daw/api/remote_websocket_server.cpp
+++ b/magda/daw/api/remote_websocket_server.cpp
@@ -114,12 +114,18 @@ std::string errorReply(const juce::var& id, int code, const juce::String& messag
 /**
  * A dispatcher response as a JSON-RPC reply.
  *
- * The MAGDA envelope and the JSON-RPC envelope say overlapping things, so
- * rather than nest one inside the other the operation's `result` becomes the
- * JSON-RPC `result` with `revision` and `apiVersion` folded in beside it. On
- * failure the revision goes into `error.data`, where it matters most: a client
- * that lost an optimistic-concurrency race learns what the revision actually is
- * in the same message that rejects its write.
+ * `result` is whatever the operation produced, untouched. The revision and API
+ * version go in a `meta` sibling — the mirror of how a request carries its own
+ * meta beside `params`, and the only arrangement that works for every
+ * operation: `tracks.list` answers with a bare array, which has nowhere to hang
+ * a revision, so folding meta into the result would mean wrapping some
+ * operations' output and not others'. A client would then have to know which
+ * shape each operation returns before it could read either one.
+ *
+ * On failure the revision goes into `error.data` instead, where JSON-RPC keeps
+ * application detail, and where it matters most: a client that lost an
+ * optimistic-concurrency race learns the real revision in the message that
+ * rejects its write.
  */
 std::string replyFor(const juce::var& id, const Response& response) {
     if (!response.ok) {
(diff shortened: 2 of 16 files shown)


lucaromagnoli (commented):
I found five correctness and lifecycle issues worth addressing before merge.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,541 @@
...
+        }
+    }
+
+    /// Close every live client so their reading threads leave `read()`. Stopping
+    /// the acceptor alone would leave connected clients running indefinitely.
+    void closeLiveConnections() {
+        const std::lock_guard<std::mutex> lock(liveMutex);
+        for (const auto& connection : live) {
+            connection->markClosed();
+            connection->socket.close(httplib::ws::CloseStatus::GoingAway, "server shutting down");
```
lucaromagnoli: **[P1] Shutdown does not interrupt the connection reader.** This calls `WebSocket::close()` from the stopping thread while the handler thread can already be blocked in `socket.read()` at line 430. In cpp-httplib 0.52, `close()` sets `closed_` and then performs another `read_websocket_frame`; it does not shut down the underlying socket, so the original reader remains blocked and the two threads can read the same stream concurrently. The focused lifecycle test takes about 30 seconds with the default client heartbeat; with heartbeat disabled it can wait for the 300-second WebSocket read timeout. Please use a shutdown/cancellation path that actually wakes the owning reader before joining it.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,541 @@
...
+
+    /// Called from the dispatch completion on the message thread. Nothing here
+    /// touches the socket — appending to a deque is all the message thread does.
+    void complete(std::string payload) {
+        {
+            const std::lock_guard<std::mutex> lock(mutex);
+            --inFlight;
+            if (closed)
+                return;
+            outbox.push_back(std::move(payload));
```
lucaromagnoli: **[P2] The response queue is unbounded and no longer covered by the in-flight limit.** `complete()` decrements `inFlight` before the payload is written, then leaves the response in an unbounded deque. A client that continues sending while not reading can therefore keep freeing request slots and accumulate responses indefinitely. Early malformed/invalid requests also return before `admit()` and bypass the rate limit entirely. Please bound queued output (or keep the request accounted for until write/drop) and apply admission to every incoming frame.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,541 @@
...
+        if (request.has_header("Origin")) {
+            const juce::String origin(request.get_header_value("Origin"));
+            const auto& permitted = options.allowedOrigins;
+            if (std::find(permitted.begin(), permitted.end(), origin) == permitted.end()) {
+                response.status = httplib::StatusCode::Forbidden_403;
+                return httplib::Server::HandlerResponse::Handled;
+            }
+        }
+
+        if (connectionCount() >= options.maxConnections) {
```
lucaromagnoli, line 477: **[P2] This connection-cap check races with registration.** The count is read under `liveMutex`, but the lock is released here and the connection is only added to `live` after the handshake in `runConnection()`. Concurrent upgrade workers can all observe spare capacity and exceed `maxConnections`. Admission needs an atomic reservation spanning this check and the eventual connection registration/failure.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,541 @@
...
+        // that is not operation input has to arrive beside params rather than
+        // inside it.
+        const auto meta = parsed["meta"];
+        auto deadlineMs = options.defaultDeadlineMs;
+        if (meta.getDynamicObject() != nullptr) {
+            if (const auto expected = meta["expectedRevision"]; !expected.isVoid())
+                context.expectedRevision =
+                    static_cast<Revision>(static_cast<juce::int64>(expected));
+            if (const auto requested = meta["deadlineMs"]; !requested.isVoid())
+                deadlineMs = std::min(deadlineMs, static_cast<int>(requested));
```
lucaromagnoli: **[P2] A negative `meta.deadlineMs` disables the server deadline.** For example, `-1` wins this `min`, then the `deadlineMs > 0` check below leaves `context.deadline` unset. That lets a client ask for more than the documented maximum. Please validate the field's numeric type/range and reject or safely clamp non-positive values.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,541 @@
...
+            refusal.isNotEmpty()) {
+            connection->send(errorReply(id, kTooManyRequests, refusal));
+            return;
+        }
+
+        RequestContext context;
+        context.clientId = "ws:" + juce::String(connection->id);
+        // Scoped by connection so two clients reusing id 1 do not collide in the
+        // dispatcher's idempotency cache.
+        context.requestId = context.clientId + ":" + id.toString();
```
lucaromagnoli: **[P2] Distinct valid JSON-RPC IDs collide here.** Numeric `1` and string `"1"` are different identifiers under JSON-RPC, but both become the same `toString()` cache key. A later write using the other ID can silently replay the first write's cached response instead of executing. Preserve the ID type in the idempotency key (and reject unsupported ID shapes).

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed, and the race was the worse half. `close()` at `httplib.h:21164` writes the Close frame and then loops on `read_websocket_frame` itself waiting for the peer echo, so the stopping thread joined the reader inside one `SocketStream` — which buffers internally. Timing matched exactly: the lifecycle test was **30.089s**.

No way to interrupt a blocked reader exists in 0.52. No socket handle on `Request` or `ws::WebSocket`, `set_socket_options` only reaches the listening socket (`httplib.h:12626`), and `read()` latches `closed_` on any failure so a short timeout cannot be resumed. So the reader has to notice by itself: read timeout is now the cancellation quantum at 3s, with the server pinging every 1s so a client sitting in `read()` answers automatically and keeps frames arriving well inside it. `stop()` only marks connections now — the owning thread is the only one that touches a socket.

Deliberately no `max_missed_pongs`: that arms httplib's own heartbeat thread to call `close()`, which reads, reintroducing the same race from a thread we do not control.

Cost is that a client which neither sends nor reads is indistinguishable from a dead one and gets dropped. Now documented in the header as a client requirement. Suite went 27s → 3.1s as a side effect, since the tests had been stopping servers with clients still attached.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Both correct. The slot now lasts until the bytes are written rather than until the handler returns, so a client that stops reading stops being admitted instead of freeing slots while the outbox grows; `maxQueuedRepliesPerConnection` (32) is the hard cap behind it, and anything past it is dropped rather than held on behalf of someone not listening.

Admission moved ahead of parsing. It was after it, which left the cheapest thing a client can send as the one thing no limit applied to. Every inbound frame is admitted now, including one too malformed to have a method, and the slot is released by the same path that answers it. Covered by "Garbage is rate limited like anything else".

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Right — a check before the handshake against a registration after it. Now an atomic claim at admission, released when the connection thread returns, so concurrent upgrades contend on the counter rather than all reading an empty roster.

One thing your description surfaced: the claim has to be conditional on the request actually being an upgrade to `/rpc`, or a plain GET would take a slot it never gives back. `isUpgradeToEndpoint()` gates it. Test spawns 8 simultaneous connects against a cap of 2.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed: `std::min(10000, -1)` is `-1`, and the `> 0` guard below then read that as "no deadline", so the way to escape the bound was to ask for less than none of it. Both `meta` fields are type-checked now and `deadlineMs` must be positive — rejected rather than clamped, since a negative value is a client bug worth reporting rather than silently reinterpreting. The deadline is always set now, so there is no path that leaves it unset.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed. Key now carries the type (`n:1` vs `s:1`) and ids of any other shape are refused with `-32600`.

Worth noting the blast radius was wider than the same operation: `RemoteApiService::idempotencyKey` is `clientId + requestId` with no method in it (`remote_service.cpp:441`), so a replayed response need not even have come from the same call. The test uses two different `project.setTempo` values under id `1` and `"1"` and asserts the second actually took effect.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,703 @@
...
+                    if (connection->closed)
+                        break;
+                    payload = std::move(connection->outbox.front());
+                    connection->outbox.pop_front();
+                }
+                const auto sent = connection->socket.send(payload);
+                // The slot belongs to the request until its bytes are gone, so a
+                // client that stops reading stops being admitted instead of
+                // queueing more work behind a stalled write.
+                connection->release();
```
lucaromagnoli: **[P1] This releases slots that the queued reply may not own.** `refuse()` enqueues an error and releases its admitted slot immediately, then this writer path releases it again after sending. Conversely, the rate/in-flight rejection at lines 437–442 enqueues a reply without ever acquiring a slot, but this path still releases one. With pipelined requests either case can decrement `inFlight` for an unrelated request and reopen admission while earlier work is outstanding, defeating the hard limit. Please carry slot ownership with each queued reply and release exactly once, on write or drop.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,703 @@
...
+                response.status = httplib::StatusCode::Forbidden_403;
+                return httplib::Server::HandlerResponse::Handled;
+            }
+        }
+
+        if (isUpgradeToEndpoint(request)) {
+            // Claim the slot here rather than testing a count: between a test and
+            // the registration that follows the handshake, every concurrent
+            // upgrade would see the same spare capacity.
+            if (admitted.fetch_add(1) >= options.maxConnections) {
```
lucaromagnoli: **[P1] A reservation can be claimed without ever reaching `ReservationGuard`.** The global pre-routing handler also runs for ordinary HTTP routing. `isUpgradeToEndpoint()` only checks path, `Upgrade`, and a nonempty key, while cpp-httplib additionally requires GET, `Connection: Upgrade`, a valid 24-character key, and version 13. An authenticated malformed upgrade can therefore increment `admitted`, fall through to a 404, and leak the slot permanently; repeating it makes every valid client receive 503 until restart. A failed 101 write also skips the guard. The reservation needs a cleanup path for every outcome where the WebSocket handler is not entered.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,703 @@
...
+                context.expectedRevision =
+                    static_cast<Revision>(static_cast<juce::int64>(expected));
+            }
+
+            // A client may ask for less time than the server allows, never for
+            // more — and never for none. Taking the minimum without checking the
+            // sign lets -1 win it, after which a non-positive deadline is read as
+            // "no deadline" and the request outlives every bound there is.
+            if (const auto requested = meta["deadlineMs"]; !requested.isVoid()) {
+                if (!isNumber(requested) || static_cast<int>(requested) <= 0) {
```
lucaromagnoli: **[P2] Type-checking as `double` is not enough before this cast.** JSON can supply fractional or values outside `int` range; fractional deadlines are silently truncated and an out-of-range floating-to-integer conversion is undefined behavior. `expectedRevision` has the same issue when converting to `int64`. Please validate that both values are finite, integral, non-negative/positive as appropriate, and within the destination type's range before casting.

lucaromagnoli (commented):
Second-pass review found three remaining resource-accounting and input-validation issues.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Both halves confirmed. `refuse()` released when it queued and the writer released again after sending; the rate-limit path queued a reply that had never acquired anything and the writer released for it anyway. Net effect exactly as described — `inFlight` below the truth, readmitting while earlier work was outstanding, so the hard limit added in the previous round was not one.

Ownership travels with the payload now: `Outgoing { payload, ownsSlot }`. A dispatch reply owns its slot, a refusal for an admitted frame owns that same one, a rate-limit rejection owns none. Release happens exactly once, on write or on drop — including the drop when the queue is at its cap or the connection is closing.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed, and worse than a slow leak: ten malformed upgrades against a cap of 2 and every honest client gets 503 until restart. `is_websocket_upgrade` (`httplib.h:5172`) also demands GET, `Connection: upgrade`, a 24-character base64 key and version 13 — my check tested none of those.

I did not restate that predicate. A copy of upstream validation drifts the next time the pin moves, and the failed-101 path you mention falls outside it anyway. The cap moved instead to where a guard cannot be skipped: `registerConnection()` tests and inserts under one lock, on the connection own thread. The only way to hold a slot is to be a connection, and the only way to reach the claim is through the function that also removes it — nothing to leak, no predicate to keep in sync.

**One behavioural tradeoff worth your call.** Nothing is reserved before the handshake now, so the 503 is best-effort: a client that wins the race for the last slot can see its handshake succeed and be closed immediately after with `PolicyViolation`. The bound is exact — `live` never exceeds `maxConnections` — but a racer learns after the 101 rather than instead of it. I took an exact bound that cannot be leaked over a nicer error in a rare race. Happy to switch to reservations with an expiry sweep if you would rather keep the pre-handshake 503; that costs a grace period to tune.

Test sends 10 short-key upgrades and then requires a real client to still connect and round-trip.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Right on both fields. `asIntegerInRange()` proves the value finite, integral and within the destination range before anything is cast, comparing against the bounds as doubles so the undefined conversion never executes. `expectedRevision` takes `[0, INT64_MAX]`, `deadlineMs` takes `[1, INT_MAX]`.

Covered by "Numbers in meta must be whole and in range": `250.5`, `1e18`, `-1` and `2.5` all come back `-32600`, and a whole in-range value still passes.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,770 @@
...
+
+    if (value.isInt() || value.isInt64()) {
+        result = static_cast<juce::int64>(value);
+    } else if (value.isDouble()) {
+        const auto number = static_cast<double>(value);
+        if (!std::isfinite(number) || number != std::floor(number))
+            return std::nullopt;
+        // Compared as doubles, before any conversion: the bounds are exactly
+        // representable and the cast that would be undefined never happens.
+        if (number < static_cast<double>(lowest) || number > static_cast<double>(highest))
```
lucaromagnoli: **[P1] The `int64` upper bound is not exactly representable as a double.** `static_cast<double>(INT64_MAX)` rounds up to `2^63`, so a JSON double equal to `2^63` passes this `>` comparison and reaches the out-of-range `static_cast<int64>`, which is still undefined behavior. The current tests only exercise the narrower `int` deadline bound. Please compare against an exclusive `2^63` limit (or restrict double inputs to the exactly representable integer range) before casting, and add an `expectedRevision` case at that boundary.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,770 @@
...
+#define CPPHTTPLIB_WEBSOCKET_MAX_PAYLOAD_LENGTH (256 * 1024)
+
+// How long a reader may sit in read() before it comes up for air, and therefore
+// how long shutdown can take. cpp-httplib exposes no way to interrupt a blocked
+// reader — there is no socket handle on the request or the socket, and
+// set_socket_options only reaches the listening socket — so this timeout is the
+// cancellation quantum. It is short because the server pings every second (see
+// kPingIntervalSeconds), which keeps frames arriving from any client that is
+// reading, so the timeout only expires on a client that has genuinely stopped.
+#define CPPHTTPLIB_WEBSOCKET_READ_TIMEOUT_SECOND 3
```
lucaromagnoli: **[P2] cpp-httplib's configuration macros differ across translation units.** This header-only library is compiled here with a 256 KiB payload cap and 3-second WebSocket timeout, `test_remote_websocket_server.cpp`/the live test define only the payload cap, and `test_remote_api_host.cpp` defines neither. That gives one executable differing inline definitions—an ODR violation whose selected limits can depend on optimizer/link order/platform. Please move these definitions to target-wide compile definitions or a shared configuration header included before every `httplib.h`.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,770 @@
...
+
+        juce::var parsed;
+        if (juce::JSON::parse(juce::String(message), parsed).failed() ||
+            parsed.getDynamicObject() == nullptr) {
+            refuse(connection, {}, kParseError, "malformed JSON");
+            return;
+        }
+
+        const auto id = parsed["id"];
+        const auto method = parsed["method"].toString();
```
lucaromagnoli: **[P2] JSON-RPC requires `method` to be a string, but `toString()` accepts other shapes.** For example, `"method":123` becomes `"123"` and is dispatched as an unknown operation, producing `-32601`; it should be rejected as an invalid request (`-32600`). Please check `isString()` before conversion and cover a non-string method.

magda/daw/api/remote_websocket_server.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+        /// Browser origins permitted to upgrade. Empty means no browser may
+        /// connect; a native client sends no `Origin` at all and is unaffected.
+        /// Absent from a request is not the same as unrecognised — the first is
+        /// a non-browser client, the second is a page we did not authorise.
+        std::vector<juce::String> allowedOrigins;
+
+        /// Connections are a thread each, so this is a real resource bound
+        /// rather than a policy knob. Further connections are refused at the
+        /// handshake with 503.
```
lucaromagnoli: This no longer describes the concurrent-race behavior. The exact cap is now enforced after the 101, so a racer may see `connect()` succeed and then receive a `PolicyViolation` close instead of being refused at the handshake with 503. The leak-free post-handshake enforcement is a reasonable tradeoff, but this public contract—and the PR description's unconditional `over the connection cap → 503` claim—should document that 503 is best-effort.

lucaromagnoli (commented):
Third-pass review found three remaining correctness issues and one public-contract mismatch.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed, and I checked it rather than reasoning about it — `(double)INT64_MAX` prints as `9223372036854775808.0`, exactly `2^63`, and `2^63 > (double)INT64_MAX` is false. So the value is equal to the bound rather than above it, sails through the `>`, and reaches the conversion the check exists to prevent.

The double path now proves the value fits an int64 against `2^63` **exclusive** before converting, then range-checks as an integer afterwards. `-2^63` needs no special handling — exactly representable and a valid int64.

Added the boundary case on `expectedRevision`, which is the field that reaches for the full int64 range. I used `-1e19` rather than `-9223372036854775809` for the low side: an integer literal below `INT64_MIN` tests juce::JSON's parser more than it tests this function, and it was returning success for a reason on the parser's side rather than mine.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Good catch, and it was worse than inconsistent — the server TU had the cap and the timeout, the transport tests had only the cap, the host test had neither, and all three link into one executable. Header-only plus macro configuration means each unit gets its own inline definitions, resolved by whichever the linker picks.

Both values are now compile definitions on the `httplib` target in the root CMakeLists, so every consumer that links it gets them by construction rather than by remembering to `#define` first. All four local definitions are gone; the .cpp keeps a comment saying where they live and why they cannot be set there.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed. `"method":123` became the operation name `"123"` and came back `-32601`, which tells a client it asked for something that does not exist when what it did was send something malformed. Now `isString()` before conversion, `-32600` otherwise.

New test covers a number, a bool and an array, and asserts that a *string* naming nothing still gets `-32601` — the two cases stay distinguishable.

magda/daw/api/remote_websocket_server.hpp
lucaromagnoli: Fair — I put that tradeoff in a review reply and not in the contract, which is the wrong place for it.

`Options::maxConnections` now says the bound is exact and that *how* an over-cap client is told is best-effort: usually a 503 before the handshake, but a client winning the race for the last slot sees the handshake succeed and a `PolicyViolation` close immediately after. It also records why reserving ahead of the handshake is the worse trade, so the next person does not re-derive it.

PR description updated to match.

magda/daw/api/remote_api_host.cpp
```diff
@@ -0,0 +1,161 @@
...
+    : service_(std::make_unique<RemoteApiService>(api)),
+      bridge_(std::make_unique<ModelChangeBridge>(*service_)) {}
+
+RemoteApiHost::~RemoteApiHost() {
+    stop();
+}
+
+bool RemoteApiHost::start() {
+    auto& config = Config::getInstance();
+    if (!config.getRemoteApiEnabled())
```
lucaromagnoli, line 181: [P2] Clear stale discovery credentials before early returns

If a previous MAGDA process crashes, `remote-api.json` survives. A subsequent run with the remote API disabled returns here without deleting it, so clients see an obsolete port and token for the entire session even though no listener exists; the same happens on later startup failures. Please remove any pre-existing token file at the beginning of `start()` and cover this by pre-creating the file before the disabled-start test.

lucaromagnoli: Confirmed, and it is the nastier version of a stale file: the port is not just dead, it may have been handed to some other process by then — at which point the file points a client at a stranger and the client presents a bearer token to reach it.

Cleared before every early return now, not only the disabled one, so the file existing means a listener exists. The already-running check moved above it, since a second `start()` must not delete a file that is still good.

Test pre-creates a stale file before the disabled start, as you suggested.

**On covering the failure paths the same way** — I tried the obvious "hold the port, then bind it again" and it does not fail: cpp-httplib sets `SO_REUSEPORT` where it exists (`httplib.h:10050`), so a second listener on the same port succeeds on macOS and Linux. Privileged ports would pass or fail depending on who CI runs as. I dropped that test rather than ship a flaky one; the clearing is one statement ahead of all the early returns, so the disabled case exercises it.

That does surface something worth its own issue, though: with a fixed `remoteApi.port`, two MAGDA instances will both bind it and the OS will split connections between them, while the second overwrites the first token file. A client then reads one token and may land on the instance that does not accept it. Out of scope here, but I can open it if you want.

magda/daw/api/remote_api_host.cpp
```diff
@@ -0,0 +1,170 @@
...
+bool RemoteApiHost::isRunning() const {
+    return server_ != nullptr && server_->isRunning();
+}
+
+int RemoteApiHost::boundPort() const {
+    return server_ != nullptr ? server_->boundPort() : 0;
+}
+
+juce::File RemoteApiHost::tokenFile() const {
+    return paths::dataDir().getChildFile(kTokenFileName);
```
lucaromagnoli: [P2] Do not share one discovery record across allowed app instances

MAGDA does not override `JUCEApplication::moreThanOneInstanceAllowed()`, whose default is `true`, so two supported app instances both publish this same `remote-api.json`. The later one overwrites the first, and when either instance stops its unconditional `deleteFile()` removes the remaining instance’s live discovery record. With a configured fixed port this is worse: cpp-httplib enables `SO_REUSEPORT`, so both listeners may bind and a client holding one instance’s token can be routed to the other. Please either make MAGDA explicitly single-instance or use per-instance discovery records with ownership-aware cleanup.

lucaromagnoli: Confirmed on all three counts, and it is the follow-up I raised on the previous thread — you are right that it belongs in this PR rather than after it, since this PR is what introduced the record.

Gone with per-instance records: `remote-api-<pid>.json`, with the pid inside it too. Each instance publishes and removes only its own, so the second to start no longer hides the first and the first to stop no longer strands the one still running. A client finding several records is looking at several running instances and picks one; each is valid for the port it names.

That removes the only thing that used to clean up after a crash, since no later run owns the old name. `start()` now sweeps records whose process is gone — signal 0 on POSIX, `OpenProcess` on Windows — and leaves any whose process still answers, because that one is live. A recycled pid keeps one dead record around longer than it deserves; the cost is a file, and the next sweep after that process exits takes it.

Three tests: another instance's record (pid 1, always alive) survives both start and stop; a record for a pid that cannot exist is collected; the record is named after the process that owns it.

**I did not take the single-instance route**, because `moreThanOneInstanceAllowed()` governs the whole app rather than this API, and flipping it changes how double-clicking a project behaves. Worth knowing if you do weigh it: `anotherInstanceStarted()` is *already* overridden to open a `.mgd` passed to a second launch, and JUCE only calls that when instances are **not** allowed — so that handler is dead code today, and double-clicking a project currently opens a second instance instead of opening the file. That looks like the single-instance decision was made once and never wired up. Happy to open an issue for it.

lucaromagnoli: Opened as #2031 — the dead `anotherInstanceStarted()` handler, with the JUCE gate that never fires (`juce_ApplicationBase.cpp:290`) and the two directions it could be resolved in. Noted there that the per-instance records added here stay correct either way; they just stop being load-bearing if MAGDA goes single-instance.

Issue #2019, opened 2026-08-05
Engine core slice 10: crossfade changed ops across a plan swap


lucaromagnoli:
Slice 10 of #1889. Split out of #2014, which delivered the differ and the state carry.

Carrying state across a swap removes the clicks that come from starting again: a delay line that keeps its contents, a plugin that keeps its tail. It does not remove the ones that come from the signal itself changing. Inserting a device, reordering a chain, or repatching a send makes block N+1 a different signal from block N, and nothing about carrying state smooths that step.

The incumbent does not smooth it either. `NodePlayer::setNode` prepares the new graph against the old, lets every node adopt what it can by ID, and swaps pointers under a spin lock. So this is not parity work, and nothing is regressing while it is open: it is the epic's claim that only changed nodes crossfade, which is a step past what MAGDA ships on today.

## Why it did not fit in the differ slice

The obvious implementation does not work here. Rendering the old and new plans side by side for the length of a fade would process every shared plugin instance twice in one block, with the second call seeing the wrong input: instances are owned by the runtime store and deliberately shared across epochs, which is what keeps a plugin alive through a recompile in the first place. Two plans that share instances cannot both render.

## What it needs instead

A design where the fade is inside one plan rather than between two. Roughly: the differ reports which ops changed, and the compiler is asked for a plan that carries a transient second edge around them, with a crossfade op that reads both and is retired once the fade is done. Open questions, none of them small:

- What the transient edge does to the fingerprint, the value table and the buffer assignment, all of which are computed per plan and would now be computed for a plan that exists for twenty milliseconds.
- Whether the fade ends by compiling a third plan or by an op that retires itself, and what either does to the swap-per-edit cost when someone drags a device around.
- Where a fade is worth having at all. A device inserted mid-chain is audible; a send level changing is already a value and never touched the plan.
- Whether the same mechanism is what the launcher needs later, since #1894 has its own switching fades.

## Not blocking

Cutover does not wait for this. #2018 (the parallel executor) and the null-diff harness in #1896 both matter more, and a fade is only worth adding once there is something to measure it against.

PR #2027, opened 2026-08-06
Native engine: crossfade the edges an edit changed (#2019)

lucaromagnoli:
Carrying runtime state across a plan swap removes the clicks that come from starting again. It does not remove the ones that come from the signal itself changing: inserting a device makes block N+1 a different signal from block N, and nothing about carrying state smooths that step.

The obvious fix does not work here. Rendering the old and new plans side by side would process every shared plugin instance twice in one block, and moving the fade inside one plan does not help by itself, because a shadow subgraph naming a device the main path also names binds to the same instance. The constraint is not that two plans cannot both render. It is that any op carrying a store-owned instance may appear once.

So the fade goes on the changed edge rather than around the changed region. Insert a device D between A and B and the new plan already holds everything a fade needs: A runs once, D runs once, and a fade op reading both feeds B. Nothing is duplicated, because the old signal at that point is A's output, which is in the plan regardless.

- Crossfade in the op vocabulary, keyed to the op it feeds and the slot it fills, the same shape a Delay's identity has. The consumer's role goes in OpKey::index, because two ops at one model location differ only by their role and a fade keyed on the location alone would collide.
- insertCrossfades: a pass over a compiled plan rather than a mode of the compiler, so the compiler stays a pure function of the model and plans stay golden-testable. It fades an edge when the old signal is still in the plan, still upstream, and still computed from the same subtree. Removal, reordering and arity changes are counted rather than faded.
- An equal-gain ramp in the executor, carried across a swap like a delay line, so a re-prepare mid-fade resumes rather than restarting. Both executors render it through the same op body and stay bit-identical.
- A fade retires by being left out of the next plan. The pass asks the live executor which fades are still running: one that is gets re-emitted on the same key so it finishes, and one that is not disappears, which puts the plan back to what the compiler produced.
- Edits that move the latency along the edge leave the op in the plan but turn the fade off. A delay line built now starts flushed, so compensating the old side would fade in from silence, which is a dip where there was a step.


magda/engine/plan/PlanCrossfade.cpp
```diff
@@ -0,0 +1,325 @@
...
+    SlotDecision decide(OpId consumer, const PlanOp& before, std::size_t slot) const {
+        const auto& after = new_.ops[static_cast<std::size_t>(consumer)];
+        const auto& to = after.inputs[slot];
+        if (!isAudioPort(new_, to))
+            return {};
+
+        // Where the old edge was heading, which is what it carries by the time
+        // anyone hears this edit: a fade already on it is on its way to its new
+        // side, not sitting on its old one.
+        const auto heading = throughFades(old_, before.inputs[slot]);
```
lucaromagnoli, line 168: [P1] Preserve the current mixed signal across a mid-fade edit\n\nLooking through an unfinished fade to its destination side discards the signal actually heard at the swap boundary. For example, after 64 samples of a 221-sample fade from 1 to 0, the last output is about 0.71. If another device is inserted now, heading becomes the first fade new side (0), so the replacement ramp starts its first sample from 0 and creates a 0.71 step—the click this pass is meant to remove. The stillFading input identifies this exact case, but currently only keeps the old ramp when the compiled edge has already arrived at the same destination. When the destination changes again, please re-emit the unfinished fade as a producer and fade from its current output (a short fade chain), or otherwise initialize the replacement from the actual in-flight mixture.

magda/engine/plan/PlanCrossfade.cpp
```diff
@@ -0,0 +1,325 @@
...
+            const auto& fade = fades[nextFade];
+            const PortRef oldSide{moved[static_cast<std::size_t>(fade.oldSide.op)],
+                                  fade.oldSide.port};
+            const auto& newSide = op.inputs[static_cast<std::size_t>(fade.slot)];
+
+            PlanOp faded;
+            faded.kind = OpKind::Crossfade;
+            faded.key = op.key;
+            faded.key.role = OpRole::EdgeCrossfade;
+            faded.key.index = crossfadeIndex(op.key.role, fade.slot);
```
lucaromagnoli, line 278: [P1] Retain the consumer original index in the fade key\n\nThis overwrites the discriminator from op.key.index and encodes only the consumer role plus input slot. Repeated consumers at one model location therefore collide. A track can have two SendTap ops distinguished by their send indices; if both taps change source (for example both pre/post settings change in one model update), both inserted fades get the same location, EdgeCrossfade role, and crossfadeIndex(SendTap, 0). validatePlan then reports duplicate keys and the executor refuses the published plan. The same collision can happen when a second indexed edge changes while the first fade is still being re-emitted. Please include the original consumer index in the encoding/key identity as well as its role and input slot.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -430,6 +480,46 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+            continue;
+
+        // Both sides have to be the same moment of the timeline, and where the
+        // edit changed the latency along the edge they are not. Nothing here can
+        // fix that: a delay line built now starts flushed, so compensating the
+        // old side would fade in from silence, which is a dip where there was a
+        // step. The op stays and passes the new side through.
+        const auto sides = latency.portLatency[flatPort(op.inputs[0])] ==
+                           latency.portLatency[flatPort(op.inputs[1])];
+        if (!sides)
```
lucaromagnoli, line 502: [P1] Keep disabled crossfades out of latency propagation\n\nBy this point resolvePlanLatency has already treated Crossfade as a normal fan-in and assigned its output the maximum of both input latencies. On this branch the renderer instead passes inputs[1] through unchanged. When the old side has more latency than the new side, the audio therefore leaves at the new side lower latency while portLatency still advertises the old side higher latency. Any downstream fan-in then compensates its other branches against a phantom delay and produces misaligned audio (and outputLatency is overstated). The existing test covers only the opposite ordering, where new-side latency is already the maximum. Please make Crossfade latency follow its new side when the sides differ (or remove/mark the op before latency and buffer layout are resolved).

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -430,6 +480,46 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+        const auto& op = plan.ops[i];
+        if (op.kind != OpKind::Crossfade)
+            continue;
+
+        // Both sides have to be the same moment of the timeline, and where the
+        // edit changed the latency along the edge they are not. Nothing here can
+        // fix that: a delay line built now starts flushed, so compensating the
+        // old side would fade in from silence, which is a dip where there was a
+        // step. The op stays and passes the new side through.
+        const auto sides = latency.portLatency[flatPort(op.inputs[0])] ==
```
lucaromagnoli, line 500: This gate turns the ramp off correctly, but the latency walk that feeds it has already told downstream the wrong thing. resolvePlanLatency has no Crossfade case, so a fade goes through the generic branch and reports the max of both sides as its output latency. That is the wrong number for a disabled fade, and a disabled fade with the old side more latent is reachable: move a send tap from post a latent device to pre, or re-route a consumer from a latent chain's output to a direct source, and the pass emits a fade whose old side carries the chain's latency while the new side carries none. Prepare turns the ramp off on the inequality and the op passes the new side through, but this walk still reports max, so everything downstream compensates against latency the signal no longer has: the parallel paths at the next fan-in are misaligned by the difference, audibly, and it persists until the next structural edit retires the op. The fix is one special case beside the Delay one: a Crossfade's produced latency is arriving(inputs[1]), the new side. That is exact in all three states, since a disabled fade outputs the new side, a running fade requires the sides equal, and a spent fade is a pass-through of the new side. The executor's enable gate reads the input ports, not the fade's own, so it is unaffected. A test with the reverse of the existing latency-gate scenario, latent before rather than after, would have caught it: the current test only inserts latency, where max happens to be the right answer.

magda/engine/plan/PlanCrossfade.cpp
```diff
@@ -0,0 +1,325 @@
...
+        for (std::size_t i = 0; i < old_.ops.size(); ++i)
+            oldByKey_.emplace(old_.ops[i].key, static_cast<OpId>(i));
+        for (std::size_t i = 0; i < new_.ops.size(); ++i)
+            newByKey_.emplace(new_.ops[i].key, static_cast<OpId>(i));
+
+        // Carrying is local, and the old side of a fade has to be more than
+        // that: it has to be the signal that was actually there, which means the
+        // whole subtree behind it computing what it used to. One forward pass,
+        // because ops are in dependency order.
+        unchanged_.assign(new_.ops.size(), 0);
```
lucaromagnoli, line 91: A retired fade poisons this flag for everything downstream of it. When the publish pair is (faded plan, fresh compile), the consumer of a spent fade is not carried, since its input key changes from the fade back to the real producer, so unchanged_ goes zero there and propagates through every op below it, master mix included. The subtree is in fact computing exactly what it computed, because a spent fade is a pass-through of the very port the consumer now reads directly. The consequence is real if second-order: insert a device on a track, let the fade finish, then insert one on the master chain, and the master edit goes unfaded because the retirement in the same publish marked the master mix changed. The predicate that would fix it already exists in decide(): a slot that arrived through a fade that is not running is Unmoved. Running the same arrived-and-not-running test in a pre-pass and treating those consumers as carried when seeding unchanged_ would let simultaneous unrelated edits keep their fades. Low severity, worth doing while the pass is fresh; a test would be the master-after-track sequence above asserting inserted == 1 on the second edit.

lucaromagnoli (commented):
Reviewed the pass, the ramp, the key encoding and the executor integration in depth. The placement argument is the right one and the code honors it: the constraint really is op-level uniqueness of store-owned instances rather than plan-level exclusivity, and fading the changed edge inside the new plan satisfies it with zero duplication because the old signal is a port the plan computes anyway. The mechanics verify. Reorder detection falling out of dependency order as one index comparison is elegant and correct, and the reorder test pins all three edges with their three different answers. The in-place claim holds: every sample is read before the same index is written, so aliasing the destination with the new side is safe, and the tail copy guards the pointer. The ramp's position is counted off what remains rather than off the block, which is what makes a fade the same length at any block size, and the identity test drives it across the whole ramp on both executors. The carry protocol is the delay-line argument re-applied and correctly so, including hasConfiguration refusing a resumed ramp at a different rate. The retirement protocol, a fade retiring by being left out, is the cleanest part of the design: no timer, no message, the plan simply returns to what the compiler produced.

I also checked the second-edit-mid-fade semantics: fading from where the first fade was heading rather than from the audible mixture leaves a residual step bounded by the unfinished fraction of a five-millisecond ramp, and the alternative, chaining fades, grows without bound under drag-scrubbing. The collapse is the right trade and throughFades documents it honestly.

Two findings inline, one of them real misalignment. Two body-level nits: there is no test for the retirement path itself (publish, let the ramp spend, publish the same model again, and assert the plan equals a fresh compile by fingerprint with activeCrossfades zero); and withFades copies RenderPlan field by field, which silently drops whatever field the struct gains next. Copying the whole plan and rebuilding ops and outputOps would stay correct as the struct grows.

PR #2028, opened 2026-08-06
Native engine: crossfade the edges an edit moved

lucaromagnoli:
Slice 10 of #1889. Carrying state across a swap removes the clicks that come from starting again; it does nothing about the ones that come from the signal itself changing. Inserting a device makes block N+1 a different signal from block N, and this is where that step becomes a slope.

The fade lives inside one plan rather than between two: rendering the old plan beside the new one would process every shared plugin instance twice in a block, and instances are shared across epochs on purpose. So a pass over the compiler's output puts a Crossfade op on the audio edges the differ says moved, reading the edge as it was and the edge as it is, both of which the one plan already computes exactly once.

- OpKind::Crossfade and OpRole::EdgeCrossfade, keyed to the op a fade feeds and the slot it fills, with the consumer's role packed into the key index so two fades at one location cannot collide in the join.
- insertCrossfades(): fades an edge only when its old producer is still in the plan, earlier than the consumer, and computing what it used to. Removals, reorder cycles and anything touching a delay's edge step as they do today and are counted.
- An equal-gain linear ramp in the executor, shared with the epoch it came from when a fade is re-emitted, so a republish mid-fade resumes rather than restarts. The ramp is refused where the two sides do not arrive at the same latency.
- Retirement is a publish, never the audio thread: an arrived fade is re-emitted while it runs and dropped once it is spent, and the plan goes back to being byte for byte what the compiler produced.

The compiler stays a pure function of the model, published plans stay immutable, and both executors render the fade through one op body.


magda/engine/plan/PlanCrossfade.cpp
```diff
@@ -0,0 +1,332 @@
...
+    auto equivalent = true;
+
+    for (std::size_t slot = 0; slot < after.inputs.size(); ++slot) {
+        const auto beforeRef = before.inputs[slot];
+        const auto afterRef = after.inputs[slot];
+
+        // A fade already on this edge is not the old signal, it is the pair of
+        // signals it is between. What the edge was heading for is the old side
+        // of anything that replaces it, so fades collapse rather than stack.
+        const auto wasHeadingFor = throughFades(old_, beforeRef, 1);
```
lucaromagnoli: [P1] Preserve the signal actually audible when another edit lands mid-fade\n\nCollapsing an unfinished fade to its destination side introduces a discontinuity at the next swap. For a 221-sample fade from 1 to 0, after one 64-sample block the previous plan ends around 0.71; if another edit arrives then, wasHeadingFor is 0 and the replacement fade starts its first sample at exactly 0, producing the 0.71 step this feature is intended to remove. The same problem is larger if the second publish happens before the first fade renders at all. The topology test below verifies only that the destination becomes the next old side, not continuity across the rendered boundary. Please retain/re-emit the unfinished fade as the old-side producer and fade from its current output, or otherwise preserve its instantaneous mixture.

magda/engine/plan/PlanCrossfade.cpp
```diff
@@ -0,0 +1,332 @@
...
+        for (const auto& insertion : insertions_[i]) {
+            const PortRef oldSide{moved[static_cast<std::size_t>(insertion.oldSide.op)],
+                                  insertion.oldSide.port};
+            const auto newSide = op.inputs[insertion.slot];
+
+            PlanOp fade;
+            fade.kind = OpKind::Crossfade;
+            fade.key = op.key;
+            fade.key.role = OpRole::EdgeCrossfade;
+            fade.key.index = crossfadeIndex(now_.ops[i].key.role, static_cast<int>(insertion.slot));
```
lucaromagnoli: [P1] Include the consumer original index in crossfade identity\n\nThis starts from op.key but then overwrites both role and index, so consumers distinguished by their original index collapse to the same fade key. A track with two pre-fader SendTap ops has keys differing only by send index; switching both sends to post-fader in one model update inserts two valid fades, but both become the same track/location, EdgeCrossfade role, and crossfadeIndex(SendTap, 0). validatePlan then reports duplicate keys and preparation refuses the plan. It can also misassociate a re-emitted ramp with another indexed edge. Please encode the consumer original index alongside its role and input slot (or preserve it in another key field).

tests/test_plan_crossfade.cpp
```diff
@@ -0,0 +1,678 @@
...
+        REQUIRE(faded.inserted == 1);
+
+        PlanValues values;
+        for (const auto& message :
+             magda::engine::resolvePlanValues(faded.plan, {bench.track}, bench.master, values))
+            FAIL_CHECK(message);
+
+        std::unique_ptr<RenderThreadPool> pool;
+        if (numThreads > 0)
+            pool = std::make_unique<RenderThreadPool>(numThreads);
```
lucaromagnoli: This pool is constructed with the realtime default, so the identity test spawns actual realtime workers on machines that grant them, competing with whatever else is running for no benefit to the comparison; every other pool in the test suite passes realtime = false for exactly that reason (the pool's own constructor comment states it). One argument: RenderThreadPool(numThreads, false).

lucaromagnoli (commented):
Reviewed against PLAN-2019-crossfade.md line by line, then adversarially on its own terms. The implementation conforms to the spec completely: all five hard constraints hold, every component is where the plan put it, both named traps are closed from the first commit rather than patched after, and all twelve mandated tests exist with exact-value assertions (thirteen cases; the latency requirement correctly split into its two directions).

The two MUSTs deserve confirmation by name. A Crossfade's produced latency is arriving(inputs[1]), the side it is becoming, stated as a named case in resolvePlanLatency with the reasoning attached, and the reverse-direction test is exactly the adversarial shape the spec demanded: a hand-built plan with the old side 48 samples more latent, asserting the fade reports zero, both mix delays resolve to zero, and the ramp is refused. And the unchanged_ seed rule is implemented better than specified: instead of a separate pre-pass, visitConsumer returns the consumer's equivalence directly, so a spent fade's retirement leaves its consumer counting as unchanged in the same forward pass that decides everything else. One walk, no second traversal, and the retirement test pins the consequence with a fingerprint equality against a fresh compile.

Three places the implementation is stronger than the spec asked, all correct: the ramp reads both sides at an index before writing it, so the destination may alias either side, not just the new one; a running re-emitted fade marks its consumer as changed, so a simultaneous edit downstream cannot take a mid-mixture subtree as its old side, a conservative case the spec did not state; and a fade whose new side is a Delay output is refused alongside the consumer-is-delay case, strengthening the compensation invariant from both ends. The whole-struct copy in build() is there, the fast path returns the compiler's plan byte for byte, collapse-not-stack is tested with the fade op itself asserted absent from the second edit's old side, and the resume test checks the ramp continues from sample kBlockSize exactly.

One nit inline. Beyond it I have nothing: this is the design from the plan, implemented without drift, and the definition of done is now the suite plus make test-tsan TEST="[crossfade]" on the machine that runs CI.

lucaromagnoli (commented):
Follow-up on 0adb8f9: both P1 findings are fixed.\n\nThe continuity fix now preserves the entire running chain under its existing keys, emits it innermost-first, and places the new outer fade over the prior audible output. PlanDiff therefore adopts the existing ramps, and the first sample after the swap advances the inner ramp by exactly one sample while the new outer ramp begins at gain zero. The rendered boundary test pins that behavior rather than only checking topology.\n\nThe key encoding now includes the consumer role, its original key index, the input slot, and stack depth. canFade rejects values outside each field before encoding, so the packing is injective over every plan the pass accepts; the two-send regression covers the concrete collision.\n\nI also checked retained-chain ordering, oldSideFade indexing across multiple consumer slots, depth exhaustion, latency propagation through nested fades, and buffer aliasing. The generated chains remain dependency ordered, each retained fade matches its predecessor by unchanged key and sides, and each outer fade writes over only its own new side while reading the prior fade output.\n\nNo further findings in this update. Static review only, as requested.

lucaromagnoli (commented):
Second round, on the stacking commit. Both changes are correct, and the record should be straight about where they came from: neither is mine. My review endorsed the collapse as the right trade, and the send-tap collision is a hole in the spec I wrote and then did not catch reviewing the code that implemented it faithfully. The implementing session found both. Credit where due; both fixes are also implemented right, which I verified in depth since this is new surface, not a patch.

The stacking argument survives scrutiny. The collapse residual was not a rare corner: a second edit one millisecond into a five millisecond fade stepped by eighty percent of the difference, which is most of the click this slice exists to remove, and scripted edit sequences make back-to-back publishes realistic. The stack has perfect continuity instead, and the test pins it with the strongest assertion in the slice: the first sample after the swap equals sample kBlockSize of the ramp the previous epoch was rendering, exact float equality, explicitly not either destination. The growth worry that justified collapsing turns out to be self-limiting: a chain only deepens while edits land within one fade length, drag-scrubbing at UI rates never stacks because the previous fade has already spent, the whole chain retires together (the outermost spent implies every inner one is, since they started earlier at the same length, and I checked the rate-change edge: adoption is refused for every ramp at once, so they restart together and the invariant holds), and the depth cap degrades to exactly the old collapse semantics rather than to nothing. retain() keeping a chain's edges as they are rather than re-deciding them is the right reading, since anything that changed underneath has its own fade smoothing it at its own edge.

The widened key is sound: ten slot bits, twelve consumer-index bits, three depth bits, role above them with headroom to sixty-three before the sign bit is at risk, every field guarded in canFade so an out-of-range consumer index refuses rather than corrupting a neighbor field, and the two-send test pins the collision from the failing direction, since the old encoding would have tripped validatePlan's key-uniqueness inside the test's own FAIL_CHECK loop. The dump printing the decoded fields stays readable with depth in it.

The realtime nit is fixed. Nothing further from my side; I am updating the local plan document to match the landed semantics so the spec cannot re-derail a future session toward the collapse it used to mandate.

PR #2029, opened 2026-08-06
Stop moved clips leaving a ghost outline behind

### LINKED ISSUE #2026: [Bug]: GUI : Timeline : Moved Clips leaving ghost outline
### What happened?

Moving Clips around the timeline by dragging or the shortcut combo leaves a thin outline of previous positions. ( see video ).

### Steps to reproduce

https://github.com/user-attachments/assets/6249a2a1-f0e5-4181-a1fd-ee1898345f0e

### MAGDA version

0.18.0

### Operating system

Linux

lucaromagnoli:
The arrangement drew its outlines around the repaint damage instead of around the thing being outlined. Moving a clip invalidates exactly the rectangle it vacated, so the lane border came back as a complete box around those old bounds, and nothing ever repainted it away. Every position a dragged or nudged clip passed through kept one.

The damaged region decides whether to paint, never what to paint:

- the lane border and the group extent indicator now take the lane and the extent, not their intersection with the clip region
- the clip body, outline and header take the clip's own bounds, so a partial repaint no longer puts the rounded corners at the damage edge
- the broken-file placeholder is drawn once over the whole waveform area, because it hides its label below 54px and a narrow damage strip erased it

Linux showed it because X11 repaints the precise damage rects; macOS coalesces them into a union that usually still covers the lane, which hid the box under the moved clip. Latent everywhere.

test_repaint_invariance_juce renders each component whole and again through narrow partial repaints, and requires the two to be pixel-identical. All three cases fail without the fix.

Fixes #2026

### DIFF
--- a/tests/test_repaint_invariance_juce.cpp
+++ b/tests/test_repaint_invariance_juce.cpp
@@ -1,0 +1,145 @@
+#include <juce_gui_basics/juce_gui_basics.h>
+
+#include "JuceTestStateGuard.hpp"
+#include "magda/daw/core/ClipManager.hpp"
+#include "magda/daw/core/TrackManager.hpp"
+#include "magda/daw/ui/components/clips/ClipComponent.hpp"
+#include "magda/daw/ui/components/tracks/TrackContentPanel.hpp"
+
+/**
+ * Repaint invariance in the arrangement (#2026).
+ *
+ * A clip dragged or nudged across the timeline invalidates only the rectangle
+ * it left behind. Painting that reads its geometry from g.getClipBounds() draws
+ * its outlines around that damage instead of around the thing being drawn, so
+ * every position the clip passed through kept a thin box — the lane border, the
+ * clip outline, the rounded corners — and nothing ever repainted it away.
+ *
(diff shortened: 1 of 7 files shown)


tests/test_repaint_invariance_juce.cpp
```diff
@@ -0,0 +1,145 @@
...
+            if (a.getPixelAt(x, y) != b.getPixelAt(x, y))
+                ++differing;
+    return differing;
+}
+
+/// A clip whose source was never on disk — the broken-file placeholder is one
+/// of the things a partial repaint used to drop.
+ClipId createAudioClip(double startBeats, double lengthBeats) {
+    return ClipManager::getInstance().createAudioClipBeats(testTrackId, startBeats, lengthBeats,
+                                                           "/tmp/magda_repaint_test.wav",
```
lucaromagnoli: This assumes a fixed global path never exists. If /tmp/magda_repaint_test.wav is present, the test stops exercising the missing-file placeholder, and asynchronous thumbnail loading may make the render comparison flaky. Could we derive a unique nonexistent child path from JUCE’s temporary directory (and assert it does not exist) instead?

Issue #2034, opened 2026-08-06
Arranger clips slice 1: the clip snapshot


lucaromagnoli:
Slice 1 of #1890.

What a track plays, resolved by the model and published beside the plan. Clip positions are deliberately not in the plan (see RenderPlan.hpp): ops read them through a snapshot, so editing a clip never recompiles a plan.

## Contents

Per track, one entry per sounding clip:

- the audible span and the interior silenced ranges, from `computeAudibleSpans`,
- resolved fade in and fade out, from `ClipManager::effectiveFadesIn`, with their curves and behaviours,
- clip volume, gain and pan, and the launch fade,
- per event: the source and its file, the anchor and loop region in source samples, the interpretation (auto tempo, speed ratio, stretch mode, warp markers, pitch, reverse, active channels) and the event's own gain and fades.

Occlusion and fades are called once, here, so an op never sees an overlap: it plays a span with fades (#2003). An edit recompiles the snapshot the way a structural change recompiles the plan, so there is no sync protocol between two representations and no state that can go stale.

Takes and comping resolve here too. The fronted take is what the entry names, and an active comp names the comp render, which is what the incumbent plays; neither reaches the audio thread as a choice.

## Decisions

- Beats resolve to seconds here, through the tempo map, because a source is handed seconds (ClipPlacement.hpp) and there must not be a second tempo map on the audio thread. What does not resolve is anything whose ratio moves with the tempo, which is auto tempo and warp: those stay in beats and read the beat face of BlockInfo (slices 4 and 5).
- Keyed by TrackId, not by OpId, so it carries no plan fingerprint. Unlike PlanValues it is not resolved against a plan's structure, and a plan swap does not invalidate it.
- Published like TransportSnapshot: one farbot RealtimeObject, on the publishing thread.
- Audio entries carry interior silences, not just MIDI ones. Nothing on a lane is cut any more, so a clip dropped inside another leaves a hole in the middle of an audio clip and the voice has to play around it.

## Verification

A canonical dump plus golden tests, the same shape as PlanDump: a project fixture compiles to this text. The rules themselves are already pinned by tests/test_clip_occlusion.cpp and tests/test_clip_crossfades.cpp; what is new here is that the compile calls them and carries the result through.

lucaromagnoli:
Closed by #2041 (merged into dev/0.19.0). Slice 2 continues in #2035.

Issue #2003, opened 2026-08-03
[FR]: Clips : Allow overlapping clips


mikobuntu:
### What happened?

Clips cannot co-exist as a layer on top of each other without crossfading enabled and the moved clip's start and end position slice and remove the underlaying clip,

Even with crossfading enabled, once we completely cover a clip with the one we have dragged/moved  the stationary clip below becomes erased.

There is a lot going on here, so see my video for what I am trying to explain.

### Steps to reproduce

https://github.com/user-attachments/assets/9569dc9e-65c5-486e-b818-f56b210a9e53

### Expected behavior

clips should be layered and not erased.

### MAGDA version

0.18.0

### Operating system

Linux

### Crash log (if applicable)

```text
*
```

lucaromagnoli:
Yeah that's quite ugly. I'm not sure at which point it's been introduced

mikobuntu:
You may notice that cross-fading is enabled for some clips and not others, this is because I switched the mode in the settings during testing. It requires a restart to take effect and the clips remember their cross-fading state .

mikobuntu:
> Yeah that's quite ugly. I'm not sure at which point it's been introduced

It was at least a few versions ago, it will probably need some git dissecting to find the culprit. I could probably try and do that from the downloads ?

lucaromagnoli:
Yeah it's been added some time in April. In reality it's a little deeper than just a UI issue. The problem is that the data model doesn't account for two overlapping MIDI clips and as a side effect the clip is in effect an eraser and it creates those weird holes. I deliberately decided to go for this implementation, some time ago, but in hindsight its not acceptable.

mikobuntu:
ooops this has been here a long time, 1st test with v-0.13.0 still exhibits the behaviour ...rolling back further .... brb

/EDIT just noticed your reply that you have found the culprit, that is good news. yeah in theory it works for maybe 2 clips that you intentionally want this behavior , but it destroys moving things around. Bitwig has the exact same thing going on, which i reported a long time ago, but they haven't fixed it yet. Dragging to replace a section is workable in some cases, but when moving a clip by the shortcut acts as an eraser it becomes problematic.

I know you will figure this one out :)

lucaromagnoli:
Yeah it needs a different model. For this reason this is more of a new feature rather than a bug, so I will ship it next release cycle.

mikobuntu:
[...]

PR #2041, opened 2026-08-06
Native engine: what a track plays, resolved (#2034)

lucaromagnoli:
Clip positions are not in the render plan, so they reach the audio thread the way values and the transport do: as their own immutable snapshot, swapped in whole. Moving a clip therefore never compiles a plan.

Resolved is the point of it. Occlusion and crossfades are worked out once, at compile time, by the model functions that already own those rules, so a ClipAudio or ClipMidi op never sees an overlap: it plays a span with fades. An edit recompiles the snapshot the way a structural change recompiles a plan, which leaves no sync protocol between two representations of a lane and no state that can go stale (#2003).

Both faces of every instant are carried. Beats are the model's authority and what auto tempo and warp will keep working in; the seconds are derived here, once, through the tempo map, because a source is handed seconds and there must not be a second tempo map on the audio thread. The snapshot carries the fingerprint of the map they came from, since a tempo edit moves every one of them.

The two rule functions move to magda_types on the way: they are pure and beat domain, and the engine is meant to link the model leaf rather than the whole DAW. ClipManager keeps its names as aliases and forwarders, so every call site that asks the model reads as it always did.

### DIFF
--- a/magda/daw/CMakeLists.txt
+++ b/magda/daw/CMakeLists.txt
@@ -262,8 +262,18 @@ juce_add_binary_data(MagdaAssets
 add_library(magda_types STATIC
     core/ParameterUtils.cpp
     core/DeviceState.cpp
+    # Clip placement rules (#2003). Pure, beat domain, no manager: they answer
+    # against a lane of clips, which is what lets the arrangement ask about a
+    # drag it has not committed and the native engine's clip snapshot ask with
+    # no ClipManager at all (#2034). They live here rather than in magda_daw so
+    # the engine links a leaf instead of the whole DAW.
+    core/ClipOcclusion.cpp
+    core/ClipFades.cpp
     # value-type headers (listed to document the leaf's surface)
     core/TypeIds.hpp
+    core/ClipInfo.hpp
+    core/ClipOcclusion.hpp
+    core/ClipFades.hpp
     core/TechnicalText.hpp
     core/ChainNodePath.hpp
     core/ControlTarget.hpp
@@ -277,11 +287,15 @@ add_library(magda_types STATIC
     core/TempoUtils.hpp
 )
 target_compile_features(magda_types PUBLIC cxx_std_20)
+# juce_graphics as well as juce_core: ClipInfo names juce::Colour. Headers and
+# defines only, on the same terms as juce_core above.
 target_include_directories(magda_types
     PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}  # dependents resolve "core/<type>.hpp"
-    PRIVATE $<TARGET_PROPERTY:juce::juce_core,INTERFACE_INCLUDE_DIRECTORIES>)
+    PRIVATE $<TARGET_PROPERTY:juce::juce_core,INTERFACE_INCLUDE_DIRECTORIES>
+            $<TARGET_PROPERTY:juce::juce_graphics,INTERFACE_INCLUDE_DIRECTORIES>)
 target_compile_definitions(magda_types
-    PRIVATE $<TARGET_PROPERTY:juce::juce_core,INTERFACE_COMPILE_DEFINITIONS>)
+    PRIVATE $<TARGET_PROPERTY:juce::juce_core,INTERFACE_COMPILE_DEFINITIONS>
+            $<TARGET_PROPERTY:juce::juce_graphics,INTERFACE_COMPILE_DEFINITIONS>)
 
 # Core library — populated by per-subdir CMakeLists.txt below.
 add_library(magda_daw STATIC
--- a/magda/engine/clip/ClipSnapshotCompiler.cpp
+++ b/magda/engine/clip/ClipSnapshotCompiler.cpp
@@ -51,8 +51,14 @@ ClipSnapshot compileClipSnapshot(const std::vector<ClipLane>& lanes,
         TrackClipPlayback track;
         track.trackId = lane.trackId;
 
-        const auto audibleSpans = computeAudibleSpans(lane.clips);
-
+        // Rejected before anything is resolved, not while resolving. Occlusion
+        // and the crossfade queries take every clip in the lane they are given
+        // at face value: neither checks the view or the track, so a session
+        // clip or one belonging to somewhere else would cover its neighbours
+        // for as long as it sat in the lane, and diagnosing it afterwards would
+        // not give back what it had already silenced.
+        std::vector<ClipInfo> arrangement;
+        arrangement.reserve(lane.clips.size());
         for (const auto& clip : lane.clips) {
             const auto label = clipLabel(lane.trackId, clip.id);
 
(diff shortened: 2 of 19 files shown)


magda/engine/clip/ClipSnapshotCompiler.cpp
```diff
@@ -0,0 +1,226 @@
...
+            audio.clipId = clip.id;
+            audio.span = span;
+            audio.silenced = std::move(silenced);
+
+            // Straight from the model, at the tempo where the clip starts: the
+            // arrangement draws its curves from this same call, and a fade
+            // played differently from the one drawn is the bug the shared rule
+            // exists to prevent.
+            const auto fades =
+                effectiveFadesIn(lane.clips, clip.id, tempoMap.bpmAt(clip.placement.startBeat));
```
lucaromagnoli: **[P1] Resolve crossfade seconds from the overlap, not each clip start tempo.** `effectiveFadesIn` is called separately for each clip with the BPM at that clip placement start. When the clips begin in different tempo regions, the outgoing and incoming halves of the same crossfade receive different durations; a ramp across the overlap is also reduced to one instantaneous BPM. The snapshot contract says beat positions resolve through the tempo map, so derive crossfade seconds from `beatToTime(xf.endBeat) - beatToTime(xf.startBeat)` while preserving the existing fit-to-clip clamp. Please add a tempo-change or ramp case that asserts both halves agree.

magda/engine/clip/ClipSnapshotCompiler.cpp
```diff
@@ -0,0 +1,226 @@
...
+    std::unordered_map<SourceId, const ClipSourceInfo*> sourceById;
+    sourceById.reserve(sources.size());
+    for (const auto& source : sources)
+        sourceById[source.id] = &source;
+
+    for (const auto& lane : lanes) {
+        TrackClipPlayback track;
+        track.trackId = lane.trackId;
+
+        const auto audibleSpans = computeAudibleSpans(lane.clips);
```
lucaromagnoli: **[P2] Filter invalid lane members before computing occlusion.** `computeAudibleSpans` requires only arrangement clips from one track, but this passes the raw lane before the checks below reject session clips and clips whose `trackId` differs. An invalid upper clip can therefore shorten or fully suppress a valid clip even though that upper clip is later diagnosed and skipped. Build a validated arrangement-lane view first, while retaining the diagnostics, then use it for both occlusion and fade resolution.

magda/engine/clip/ClipSnapshotCompiler.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+                playback.autoPitch = event.autoPitch;
+                playback.autoPitchMode = event.autoPitchMode;
+                playback.pitchChange = event.pitchChange;
+                playback.transpose = event.transpose;
+
+                playback.reversed = event.reversed;
+                playback.leftChannelActive = event.leftChannelActive;
+                playback.rightChannelActive = event.rightChannelActive;
+                playback.gainDb = event.gainDB;
```
lucaromagnoli, line 226: **[P1] Preserve each audio event own fades.** This copies the per-event gain, but `AudioEventPlayback` has no fields for the event `fadeInSeconds`, `fadeOutSeconds`, curve types, or behaviours. Only the primary event values are promoted to `AudioClipPlayback`, so every non-primary event loses data. #2034 explicitly requires each event own gain and fades, and the model already supports N events. Please carry those fields on `AudioEventPlayback` and cover a multi-event clip with distinct fades.

magda/engine/clip/ClipSnapshotCompiler.cpp
lucaromagnoli: Under a multi-segment tempo map the two halves of one crossfade can come out different lengths: the left clip's fade-out is priced at bpmAt(left start), the right clip's fade-in at bpmAt(right start), and neither necessarily equals the overlap's exact seconds. Since `fades.xfIn`/`xfOut` already carry the overlap's beat endpoints, the crossfade seconds could be derived here by converting those endpoints through the map (same as spanFromBeats does for the clip spans), which would make both halves agree with each other and with the spans they sit inside.

Related doc nit: the header says this is the same question the arrangement asks when it draws, but ClipComponent::computeEffectiveFades passes parentPanel_->getTempo(), not bpmAt(clip start), so the two only agree on a flat map. Fine to leave until tempo curves actually drive playback, but worth deciding which side moves so drawn and played stay one rule.

magda/engine/clip/ClipSnapshotCompiler.cpp
```diff
@@ -0,0 +1,226 @@
...
+                                               std::to_string(primary.fadeOutType) +
+                                               " is not a curve, it fades linearly");
+            audio.fadeInBehaviour = primary.fadeInBehaviour;
+            audio.fadeOutBehaviour = primary.fadeOutBehaviour;
+
+            audio.gainDb = clip.volumeDB + clip.gainDB;
+            audio.pan = clip.pan;
+            audio.launchFadeSamples = clip.launchFadeSamples;
+
+            if (clip.events().empty()) {
```
lucaromagnoli: Small asymmetry: an audio clip with no model events is dropped from the snapshot entirely, while one whose only event names an unknown source stays in it with an empty events list (the diagnostics test pins the latter shape). Both play nothing but land differently in the dump. Also, this check sits after the fades and curve work above, which quietly reads audioEventRef's kNoEvent default; hoisting it above that block would skip the pointless work on the default event.

magda/engine/clip/ClipSnapshotCompiler.hpp
```diff
@@ -0,0 +1,71 @@
...
+ */
+struct ClipLane {
+    TrackId trackId = INVALID_TRACK_ID;
+    std::vector<ClipInfo> clips;
+};
+
+/**
+ * @brief Resolve @p lanes into a snapshot.
+ *
+ * Deterministic: the same model always compiles to the same snapshot, entry for
```
lucaromagnoli, line 53: The sorts make each track deterministic, but diagnostics keep the order lanes arrive in, so this claim quietly requires the caller to hand lanes over in a stable order. One sentence here pinning that as part of the input contract (or a stable sort of diagnostics) would make the golden guarantee airtight.

magda/daw/CMakeLists.txt
```diff
@@ -262,8 +262,18 @@ juce_add_binary_data(MagdaAssets
 add_library(magda_types STATIC
     core/ParameterUtils.cpp
     core/DeviceState.cpp
+    # Clip placement rules (#2003). Pure, beat domain, no manager: they answer
+    # against a lane of clips, which is what lets the arrangement ask about a
+    # drag it has not committed and the native engine's clip snapshot ask with
+    # no ClipManager at all (#2034). They live here rather than in magda_daw so
+    # the engine links a leaf instead of the whole DAW.
+    core/ClipOcclusion.cpp
```
lucaromagnoli, line 270: While this file changes targets: ClipOcclusion.hpp still says an audio clip cannot play around a hole, "so ClipManager splits audio clips before a hole can reach here". resolveOverlaps no longer splits (its comment now says the lane keeps whole clips in every case) and the snapshot this PR adds carries the holes for audio directly, so that sentence is stale on both counts. Worth updating in passing.

lucaromagnoli (commented):
Solid slice of the native-engine epic: the two rule functions move into magda_types verbatim with aliases keeping every ClipManager call site unchanged, the snapshot arrives resolved and immutable with both time domains carried, the feed reuses the exact farbot pattern EngineSession already uses for values and the transport, and the tests cover occlusion, fades, determinism, the feed and a golden dump. Nothing blocking; four small comments inline, mostly about tempo-curve edges and doc claims that will matter once multi-segment maps drive playback.

Issue #701, opened 2026-02-17
WebSocket API for external control and user scripting


lucaromagnoli:
## Summary

Add a WebSocket server to MAGDA that exposes TrackManager functionality to external systems. This enables user scripting, hardware control surfaces, web UIs, and AI agent integration — all language-agnostic.

## Use Cases

- **Elgato Stream Deck** — custom buttons for transport, bypass, presets, macros (Stream Deck SDK uses WebSocket)
- **User scripting** — Python/JS/any language scripts to automate workflows, batch operations, session templates
- **Web UI** — browser-based remote control panel
- **Touch interfaces** — phone/tablet control surfaces
- **AI agents** — LLMs controlling the DAW programmatically

### Protocol

JSON messages over WebSocket (default port TBD, e.g. 9800).

**Commands** (client → MAGDA):
\`\`\`json
{"method": "transport.play"}
{"method": "transport.stop"}
{"method": "device.setParam", "params": {"track": 1, "device": 2, "param": 0, "value": 0.5}}
{"method": "track.addDevice", "params": {"track": 1, "type": "reverb"}}
{"method": "track.list"}
\`\`\`

**Events** (MAGDA → client, push notifications):
\`\`\`json
{"event": "param.changed", "data": {"track": 1, "device": 2, "param": 0, "value": 0.75}}
{"event": "transport.started"}
{"event": "device.added", "data": {"track": 1, "device": 3, "type": "delay"}}
\`\`\`

**Queries** (client → MAGDA → client):
\`\`\`json
{"method": "track.getDevices", "params": {"track": 1}}
{"method": "device.getParams", "params": {"track": 1, "device": 2}}
\`\`\`

### Architecture

- TrackManager already centralizes all state and has `TrackManagerListener` for change notifications — maps directly to WebSocket events
- New `WebSocketServer` class that listens to TrackManager and routes incoming messages to TrackManager calls
- No changes to existing core code required
- Lightweight WebSocket library: `choc` (from Tracktion team) or `IXWebSocket` (header-only)

### Language Agnostic

Any language with a WebSocket client works: Python, JavaScript, Lua, Rust, Go, C#, Swift, or even command-line tools like `websocat`.

## Future Extensions

- Authentication for remote access
- Subscribe/unsubscribe to specific event types
- Bundled Python/JS client libraries for convenience
- Stream Deck plugin

lucaromagnoli:
## Sequencing: what's actually blocked, after auditing the tree

[...]

PR #2044, opened 2026-08-07
WebSocket API: subscriptions, revisioned events, and backpressure (#1857)

### LINKED ISSUE #1857: WebSocket API: subscriptions, revisioned events, and backpressure

lucaromagnoli:
Closes #1857. Part of #701, on top of #1855 and #1856.

## Summary

Remote clients can now be pushed changes instead of polling for them, without mirroring every internal model notification or flooding the JUCE message queue.

A new `SubscriptionHub` sits between `ChangeSource` (#1855), which says *that* a topic changed, and the transports, which have to say *what* changed to each client that asked. It is shared rather than private to the WebSocket adapter because MCP resource updates (#1858) need the identical projection and the identical resync rules, and the second implementation would drift from the first.

## What a client receives

A transport-neutral envelope of topic, type, revision, and payload:

```json
{ "topic": "clips", "type": "delta", "revision": 57, "payload": { "added": [], "updated": [], "removed": [] } }
```

- **`snapshot`** — complete state for the topic, in the same shape as the topic's read operation (`tracks` is `tracks.list`, and so on). Delivered in the reply to `subscribe`/`resync`, so a client is never subscribed without knowing what it is watching, and pushed as an event when a client has fallen behind.
- **`delta`** — `{added, updated, removed}` for `tracks`, `clips`, `automation`, and `session`. `project`, `transport`, `selection`, and `devices` send their whole state, because there is nothing useful to diff — a `DeviceId` is unique only within a track section, so an id-keyed diff of `devices` would be wrong rather than merely unhelpful.
- **`sample`** — `meters` and `playhead`. Opt-in, never delta-encoded, latest value wins.

A topic that was marked but projects identically sends nothing at all, and each topic is projected once per flush however many clients are watching.

## Meters and the playhead

Nothing in the model marks either — the playhead moves without notifying anyone and meters are a continuous signal — so they are sampled on their own 20 Hz timer. The live meter source drains a dedicated `MeteringBuffer` on `AudioBridge`, a third ring alongside the recording one: peeking the mixer's ring goes stale as soon as it fills with no UI meter draining it, which is exactly the headless case a remote subscriber is in. The audio thread is untouched — metering is already published lock-free and drained on the message thread.

## Backpressure

Dropped, not buffered. A refused event marks the client, so the next thing it takes is a `snapshot` rather than a delta it has no baseline for; persistent refusal disconnects it. Events share the reply queue for ordering but hold a separate budget, so a busy subscription cannot starve the answer a client is waiting on, and a burst of replies cannot silently cost a subscriber its stream.

## The contract

`subscriptions.subscribe`, `.unsubscribe`, `.list`, and `.resync` are declared in the shared registry and marked `transportScoped`: names and schemas in one place, execution in the adapter, and `system.describe` stays a complete catalogue. Dispatching one through the service fails with `invalid_request` rather than `unknown_operation` — the operation is real, and only the route is wrong. Also adds `automation.listLanes`, which the `automation` topic needed as a snapshot source, and `minItems` support in the schema validator.

## Two things the design had to get right

Projection reads live model state, so `handle()` validates on the calling thread and hops to the message thread for the rest, answering through a callback the way `dispatch` does.

Both timers are reconciled only on the message thread, and never from inside a timer callback: JUCE invokes a timer callback with its own lock released, so stopping one from elsewhere can free a timer whose callback is on the stack.

## Testing

- `test_remote_subscriptions.cpp` — load, ordering, resync, coalescing, and the slow-client policy against a mock, with `flush()` and `sampleNow()` driven by hand so nothing is timing-dependent.
- `test_remote_websocket_server.cpp` — subscribe, push, ordering, and the per-connection event budget end to end over a real socket.
- `test_remote_websocket_live_juce.cpp` — with the message-thread assertion enabled: the real 30 Hz flush timer, the sampling timer, and the hop itself.

Full Catch2 suite passes (2116 cases, 34 new); full JUCE suite passes; `make debug` builds clean. `clang-tidy` was not available locally, so `make lint-changed` did not run.

### DIFF
--- a/magda/daw/api/remote_subscriptions.cpp
+++ b/magda/daw/api/remote_subscriptions.cpp
@@ -19,12 +19,6 @@ std::size_t indexOf(Topic topic) {
     return static_cast<std::size_t>(topic);
 }
 
-/// Sampled rather than published: driven by their own timer, never by a model
-/// notification, and never delta-encoded.
-bool isSampled(Topic topic) {
-    return topic == Topic::Meters || topic == Topic::Playhead;
-}
-
 juce::var makeObject() {
     return juce::var(new juce::DynamicObject());
 }
@@ -442,7 +436,7 @@ void SubscriptionHub::applyTopology() {
             const auto topic = static_cast<Topic>(index);
             if (!anySubscriberLocked(topic))
                 continue;
-            if (isSampled(topic))
+            if (isContinuousTopic(topic))
                 wantSampler = true;
             else
                 wantChanges = true;
@@ -599,51 +593,73 @@ void SubscriptionHub::deliverLocked(Client& client, const SubscriptionEvent& eve
     ++client.consecutiveDrops;
 }
 
+bool SubscriptionHub::owesSnapshotLocked(Topic topic) const {
+    const auto index = indexOf(topic);
+    return std::any_of(clients_.begin(), clients_.end(), [index](const Client& client) {
+        return client.subscribed[index] && client.needsSnapshot[index];
+    });
+}
+
 void SubscriptionHub::publishTopicLocked(Topic topic, Revision revision) {
     const auto index = indexOf(topic);
+
+    // A client that refused an event is owed complete state, and it is owed it
+    // whether or not anything has changed since. Deciding that before the
+    // early-out below is what makes the promise good: otherwise the snapshot
+    // would arrive only on the next observable change, and a client that dropped
+    // one event from a project that then went quiet would stay stale until it
+    // thought to ask for a resync.
+    const bool owed = owesSnapshotLocked(topic);
+
     auto current = projectTopic(topic);
     auto& state = topics_[index];
+    const bool hadBaseline = state.hasBaseline;
 
     SubscriptionEvent delta{topic, SubscriptionEvent::Type::Delta, revision, {}};
-    bool haveDelta = false;
+    bool changed = !hadBaseline;
 
-    if (state.hasBaseline) {
+    if (hadBaseline) {
         if (!keyFieldsFor(topic).empty()) {
             const auto difference = diffElements(topic, state.baseline, current);
-            if (!difference.changed()) {
-                // The topic was marked but nothing observable changed — a
-                // parameter that came back to where it started, or a
-                // notification for state this projection does not expose. The
-                // coalescing that matters most is the event never sent.
-                state.baseline = current;
-                return;
-            }
-            delta.payload = difference.toJson();
+            changed = difference.changed();
+            if (changed)
+                delta.payload = difference.toJson();
         } else {
-            if (deepEquals(state.baseline, current)) {
-                state.baseline = current;
-                return;
-            }
-            delta.payload = current;
+            changed = !deepEquals(state.baseline, current);
+            if (changed)
+                delta.payload = current;
         }
-        haveDelta = true;
     }
 
     state.baseline = current;
     state.hasBaseline = true;
+
+    // The topic was marked but nothing observable changed — a parameter that
+    // came back to where it started, or a notification for state this projection
+    // does not expose — and nobody is behind. The coalescing that matters most
+    // is the event never sent.
+    if (!changed && !owed)
+        return;
 
     const SubscriptionEvent snapshot{topic, SubscriptionEvent::Type::Snapshot, revision, current};
 
     for (auto& client : clients_) {
         if (!client.subscribed[index])
             continue;
-        const bool complete = client.needsSnapshot[index] || !haveDelta;
-        deliverLocked(client, complete ? snapshot : delta);
+        if (client.needsSnapshot[index] || !hadBaseline) {
+            deliverLocked(client, snapshot);
+            continue;
+        }
+        // A client that is current and has nothing to be told: this pass exists
+        // only for the ones that are behind.
+        if (changed)
+            deliverLocked(client, delta);
     }
 }
 
 void SubscriptionHub::publish(const std::vector<ChangeSource::Change>& changes) {
     bool dropped = false;
+    std::vector<Topic> stillOwed;
     {
         const std::lock_guard<std::mutex> lock(mutex_);
         if (shutdown_)
@@ -652,7 +668,7 @@ void SubscriptionHub::publish(const std::vector<ChangeSource::Change>& changes) 
         for (const auto& change : changes) {
             // Nothing marks these, and if something did, a revisioned delta is
             // the wrong shape for a continuous signal.
-            if (isSampled(change.topic))
+            if (isContinuousTopic(change.topic))
                 continue;
             if (!anySubscriberLocked(change.topic))
                 continue;
@@ -660,6 +676,24 @@ void SubscriptionHub::publish(const std::vector<ChangeSource::Change>& changes) 
         }
 
         dropped = dropAbandonedLocked();
+
+        for (std::size_t index = 0; index < TOPIC_COUNT; ++index) {
+            const auto topic = static_cast<Topic>(index);
+            if (!isContinuousTopic(topic) && owesSnapshotLocked(topic))
+                stillOwed.push_back(topic);
+        }
+    }
+
+    // A client that refused again is owed a snapshot that nothing else will
+    // deliver: a quiet project produces no flush at all, so there would be no
+    // next attempt. Marking the topic dirty schedules one, which keeps the retry
+    // on the coalescing pump that already exists rather than giving it a timer
+    // of its own — and keeps the disconnect budget meaning what it says, since
+    // each retry is one more flush the client refused.
+    if (!stillOwed.empty()) {
+        const auto revision = service_.currentRevision();
+        for (const auto topic : stillOwed)
+            service_.changes().markChanged(topic, revision);
     }
 
     // Only when the client set actually changed. This runs inside the flush
@@ -690,18 +724,31 @@ bool SubscriptionHub::dropAbandonedLocked() {
     if (options_.droppedFlushesBeforeDisconnect <= 0)
         return false;
 
-    const auto abandoned =
-        std::remove_if(clients_.begin(), clients_.end(), [this](const Client& client) {
-            return client.consecutiveDrops >= options_.droppedFlushesBeforeDisconnect;
-        });
-    if (abandoned == clients_.end())
+    const auto abandoned = [this](const Client& client) {
+        return client.consecutiveDrops >= options_.droppedFlushesBeforeDisconnect;
+    };
+
+    // Collected before anything is removed, and deliberately not from the tail
+    // that `remove_if` leaves behind. That tail holds moved-from elements, not
+    // the ones that matched: with a stalled client followed by a healthy one,
+    // the healthy one is moved over the stalled one and what remains at the end
+    // is a hollowed-out copy whose `disconnect` is empty. The stalled client
+    // would then be dropped from the hub while keeping its socket, its thread,
+    // and its connection slot — silently, because the callback that was supposed
+    // to close it no longer exists.
+    std::vector<Disconnect> departing;
+    for (const auto& client : clients_)
+        if (abandoned(client) && client.disconnect != nullptr)
+            departing.push_back(client.disconnect);
+
+    const auto removed = std::remove_if(clients_.begin(), clients_.end(), abandoned);
+    if (removed == clients_.end())
         return false;
+    clients_.erase(removed, clients_.end());
 
-    for (auto it = abandoned; it != clients_.end(); ++it)
-        if (it->disconnect != nullptr)
-            it->disconnect("subscriber is not consuming events");
+    for (const auto& disconnect : departing)
+        disconnect("subscriber is not consuming events");
 
-    clients_.erase(abandoned, clients_.end());
     releaseIdleTopicsLocked();
     return true;
 }
@@ -713,7 +760,7 @@ bool SubscriptionHub::dropAbandonedLocked() {
 void SubscriptionHub::sendSnapshotsLocked(Client& client, const std::vector<Topic>& topics,
                                           Revision revision, juce::Array<juce::var>& into) {
     for (const auto topic : topics) {
-        if (isSampled(topic))
+        if (isContinuousTopic(topic))
             continue;
 
         const auto index = indexOf(topic);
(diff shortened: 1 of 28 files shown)


lucaromagnoli (commented):
I found several correctness and lifecycle issues in the subscription recovery path; details are inline.

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,900 @@
...
+            return;
+
+        for (const auto& change : changes) {
+            // Nothing marks these, and if something did, a revisioned delta is
+            // the wrong shape for a continuous signal.
+            if (isSampled(change.topic))
+                continue;
+            if (!anySubscriberLocked(change.topic))
+                continue;
+            publishTopicLocked(change.topic, change.revision);
```
lucaromagnoli, line 717: **[P1] Project replacement leaves non-`project` subscribers on the old project.** `RemoteApiService::projectReplaced()` discards pending changes and marks only `Topic::Project`, while this loop republishes only the topics present in `changes`. A client subscribed solely to tracks, clips, devices, session, selection, transport, or automation therefore receives nothing when a different project is opened and retains the previous project's state. Please invalidate/resnapshot every model topic on project replacement (or give the hub an explicit global-resync signal).

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,900 @@
...
+        case Topic::Clips:
+            return "clips.list";
+        case Topic::Devices:
+            return "devices.list";
+        case Topic::Selection:
+            return "selection.get";
+        case Topic::Transport:
+            return "transport.get";
+        case Topic::Session:
+            return "session.get";
```
lucaromagnoli, line 222: **[P2] The new session projection is not invalidated by session-clip mutations.** `session.get` derives its `slots` from clips, but `topicsFor("clips.*")` currently marks only `Topic::Clips`; `tracks.delete` also omits `Topic::Session`. Creating/deleting a clip with `view: "session"` can therefore leave a session-only subscriber stale indefinitely. Please include `Session` in the relevant dependent invalidations (and cover this with a cross-topic test).

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,900 @@
...
+        return;
+    }
+
+    // A refused sample is the policy working, not a client falling behind:
+    // meters and the playhead are latest-value-wins and the next one is along in
+    // a few tens of milliseconds.
+    if (event.type == SubscriptionEvent::Type::Sample)
+        return;
+
+    client.needsSnapshot[index] = true;
```
lucaromagnoli, line 609: **[P2] A refused event may never be replaced by the promised snapshot.** This records that the client needs a snapshot, but delivery is attempted only when another observable change for this topic is published. If the queue drains and the project becomes quiet after the dropped event, the client remains stale indefinitely unless it explicitly calls `resync`. The recovery path needs a queue-capacity/drained signal or another guaranteed retry mechanism.

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,900 @@
...
+            if (client.subscribed[indexOf(topic)])
+                deliverLocked(client, event);
+    }
+}
+
+bool SubscriptionHub::dropAbandonedLocked() {
+    if (options_.droppedFlushesBeforeDisconnect <= 0)
+        return false;
+
+    const auto abandoned =
```
lucaromagnoli: **[P1] Invoke disconnect callbacks before applying `remove_if`.** After `std::remove_if`, `[abandoned, end)` contains moved-from/unspecified elements, not reliably the clients that matched the predicate. With stalled client A followed by healthy client B, B is moved over A and A's `disconnect` callback can be destroyed before the loop below reaches it. A is erased from the hub but keeps its socket, thread, and connection slot. Collect/call the matching callbacks first, or erase clients one at a time.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -624,6 +706,24 @@ struct RemoteWebSocketServer::Impl {
...
+        if (subscriptions != nullptr) {
+            connection->subscriber = subscriptions->addClient(
+                [connection](const SubscriptionEvent& event) {
+                    return connection->enqueueEvent(notificationFor(event));
+                },
+                [connection](const juce::String&) {
+                    // Marking is all a foreign thread may do. The reader owns
+                    // the socket and notices within one read timeout, which is
+                    // the same shutdown path stop() uses.
+                    connection->markClosed();
```
lucaromagnoli, line 739: **[P1] A backpressure-disconnected client can keep executing RPC requests.** The hub's disconnect callback only calls `connection->markClosed()`, but this reader loop never checks that flag. If the peer keeps sending frames, `read()` continues succeeding and `handleMessage()` can dispatch mutations indefinitely; only its replies are discarded. Please expose/check the connection's closed state before processing another frame so marking it closed actually terminates the connection path.

magda/daw/api/remote_subscriptions.cpp
lucaromagnoli: Confirmed and fixed in cccd2c8c. `projectReplaced()` now invalidates every discrete topic rather than just `project`, so a client watching only `tracks` (or any other single topic) hears about the swap instead of continuing to show the outgoing project. The two continuous topics are excluded, since nothing marks a meter reading — it is sampled.

The resulting event is a delta computed against the old project's baseline, which is correct and idempotent: the old ids come through as `removed`, the new ones as `added`. Covered by "Opening a different project reaches subscribers of every topic", which subscribes only to `tracks`.

magda/daw/api/remote_subscriptions.cpp
lucaromagnoli: Confirmed and fixed in cccd2c8c, on both paths. `topicsFor` now returns `Session` for `clips.*` and adds it to `tracks.create`/`tracks.delete`; the model bridge marks it from `clipsChanged`, `clipPropertyChanged`, `clipPropertiesChanged`, and `tracksChanged`, so a UI-side edit invalidates it too.

Deliberately over-broad on the clip operations that cannot affect the grid — adding a note — because it costs nothing here: an unchanged projection is coalesced away before an event is built.

`noteModelChanged` gained an `initializer_list` overload so this stays one revision per edit rather than one per topic, which the existing re-entrancy guard already cared about. Cross-topic coverage in "A session subscriber hears about a clip that changes the grid" (hub) and "Clip and track notifications also invalidate the session grid" (bridge, against the real singletons).

magda/daw/api/remote_subscriptions.cpp
lucaromagnoli: Confirmed and fixed in cccd2c8c. A topic that still owes a snapshot is now re-marked on the change source at the end of `publish`, which schedules another flush. That keeps the retry on the coalescing pump that already exists instead of giving it a timer of its own, and it keeps `droppedFlushesBeforeDisconnect` meaning what it says, since each retry is one more flush the client refused.

`publishTopicLocked` also had to stop returning early when nothing changed: the debt is now checked before that early-out, or the snapshot would still only ride along with the next real change. Covered by "A refused event is followed by a snapshot even if nothing else changes", which drops one event and then flushes a completely quiet project — and asserts the retry stops once the debt is paid.

magda/daw/api/remote_subscriptions.cpp
lucaromagnoli: Confirmed and fixed in cccd2c8c. Callbacks are now collected before anything is removed. You are right about the mechanism and the consequence: the tail holds moved-from elements, so with a stalled client followed by a healthy one the callback read from the end is a hollowed-out copy of the healthy one, and the stalled client leaves the hub while keeping its socket, its thread, and its connection slot.

My existing test passed only because it had a single client, where `remove_if` moves nothing. Replaced with "The stalled client is the one disconnected, not whoever follows it", which uses the two-client ordering.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Confirmed and fixed in cccd2c8c. `Connection` gained `isClosed()`, and the reader loop checks it both in its condition and again after `read()` returns — the second check matters because the read is where that thread spends its time, so a connection retired while it was blocked would otherwise get one more request executed per frame the client kept sending.

The reader still owns the socket, so this stays a mark-and-notice rather than a cross-thread close, and it terminates within one read timeout as before. Covered by "A dropped subscriber stops being able to change the project", which drops a client for backpressure and then asserts `transport.play` never reaches the model.

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,900 @@
...
+// ---------------------------------------------------------------------------
+
+void SubscriptionHub::deliverLocked(Client& client, const SubscriptionEvent& event) {
+    const auto index = indexOf(event.topic);
+    const bool sent = client.sink != nullptr && client.sink(event);
+
+    if (sent) {
+        if (event.type == SubscriptionEvent::Type::Snapshot)
+            client.needsSnapshot[index] = false;
+        if (event.type != SubscriptionEvent::Type::Sample)
```
lucaromagnoli, line 598: **[P2] Track consecutive drops per flush, not per event.** This connection-wide counter is reset by any successful non-sample delivery. With multiple subscribed topics, an earlier topic can be accepted on every flush while a later topic is consistently refused (for example, an event budget of one); each success resets the counter before the refusal increments it back to one, so the later topic remains permanently stale and the disconnect threshold is never reached. Conversely, several refusals in one flush count as several "consecutive flushes." Please aggregate whether each client refused anything during `publish()` and update/reset the counter once per flush.

lucaromagnoli: Confirmed and fixed in db5ef345. You are right that the acceptance reset is the serious half: with two subscribed topics and an outgoing queue of one, the client sat at a count of one forever — connected, apparently healthy, and permanently stale on the topic it never managed to take, with the snapshot retry re-offering an event it refused again every flush.

Delivery now records only what happened to each client during the flush in progress, and `publish` folds that into `consecutiveDrops` once, after every topic has had its turn. A refusal outweighs an acceptance in the same flush, since keeping up on one topic while never taking another is not keeping up. A flush that delivered nothing to a client leaves its count alone — that is not evidence either way, and a subscriber to a quiet topic should not have its arrears forgiven by the silence.

Both directions are covered: "Keeping up on one topic is not keeping up" reproduces the livelock with a one-event queue drained between flushes, and "Several refusals in one flush are one flush" pins the over-counting. I checked the first against the old per-event counting to be sure it is a regression test rather than a passing one — it fails there with `disconnects == 0`.

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,992 @@
...
+        if (method == kSubscribe) {
+            for (const auto topic : requested)
+                entry->subscribed[indexOf(topic)] = true;
+
+            // A client that reconnects at the revision the project is still on
+            // has missed no committed change and may skip the snapshots. Any
+            // other revision — including one it never saw — is an explicit
+            // resync, because no per-revision history is kept to replay from.
+            const auto from = params["fromRevision"];
+            const bool current = !from.isVoid() && static_cast<juce::int64>(from) ==
```
lucaromagnoli: **[P2] `fromRevision` cannot prove that the client has the latest projected state.** `noteModelActivity()` deliberately publishes tracks/devices/session/automation changes without advancing the project revision, and the contract explicitly permits consecutive events with equal revisions. A client can disconnect at revision N, miss one of those events, then reconnect with `fromRevision: N`; this branch skips the snapshot and initializes the hub baseline from current server state, so the missed value is never replayed and the client remains stale until another change happens. Resumption needs a stream sequence/change epoch in addition to the commit revision, or it must snapshot whenever event continuity cannot be established.

lucaromagnoli: Confirmed and fixed in 4c2e90c7 — I took the second of your two options, and removed the parameter rather than patching it.

Your counter-example is a normal one, not a corner: `noteModelActivity` publishes for a clip's play state, a parameter following automation, a drag preview. A client that missed one of those has a cursor identical to a client that saw it, so the branch skipped the snapshots, seeded the baseline from current state, and left the client silently stale.

I went back and forth on the epoch. A per-topic sequence in every envelope, echoed back on subscribe, is the complete answer and would keep the optimisation — but it is a second cursor on the wire, permanently, to save a transfer on reconnect, which is rare by definition. Weighed against a failure mode that is silent staleness, it did not seem worth the protocol weight. If a client turns up that reconnects often enough to care, the sequence is the right thing to add and this is the place it goes.

The schema now rejects `fromRevision` rather than ignoring it, so a client that used to send it is told its guarantee is gone instead of quietly getting snapshots it thought it had skipped. `snapshot: false` stays as the opt-out — the client asserting it has state, rather than the server inferring it — with `resync` for when that assertion turns out to be wrong.

Three tests: "Subscribing snapshots even when the revision has not moved" reproduces exactly your scenario via `noteModelActivity` and asserts the reconnecting client receives the value it missed; plus one pinning the schema rejection and one for the `snapshot:false` path.

Issue #2039, opened 2026-08-06
Arranger clips slice 6: MIDI clip playback


lucaromagnoli:
Slice 6 of #1890, after slice 1. Independent of slices 3 to 5.

The source behind a track's `ClipMidi` op: what the arrangement's MIDI clips play. Audio and MIDI share the snapshot and the span rules, and share nothing below that.

- Notes with their velocities and lengths, CC and pitch bend with their curve types, tensions and bezier handles, and per-note MPE pitch expression.
- The clip's loop, which is in clip beats rather than source samples (`loopEnabled`, `loopStartBeats`, `loopLengthBeats`), plus `midiOffset` and `midiTrimOffset`.
- Groove templates and strength, which MAGDA takes from Tracktion today.
- Interior silences from the snapshot: a note starting inside a hole does not sound. MIDI plays around a hole rather than being cut, which is the rule the model already implements.
- Note lifetime across a loop wrap, a locate, a stop and a plan swap. Every note that sounded has to be ended, and this is the layer that owns it: a note left hanging by a seek is the failure the whole slice is judged on.
- A clip whose looping is turned off must not be shortened. Tracktion's `MidiClip::disableLooping` truncates the clip to one loop length, which is why flattening a looped MIDI clip plays only its first cycle today. The native engine does not inherit that.
- MIDI takes and comping resolve in the snapshot, so the source plays one assembled event list and never picks a take.

At most `kMaxMidiBytesPerPort` of encoded MIDI per block, which is what the executor reserved for the port.

Issue #2038, opened 2026-08-06
Arranger clips slice 5: warp markers, beat detection and loop info


lucaromagnoli:
Slice 5 of #1890, after slice 4.

Where a clip's own musical time comes from, and where it stops agreeing with the file's.

- **Warp markers.** The model's map is (sourceTime, warpTime) pairs in source seconds, piecewise linear, carrying on at slope 1 outside the marker range (`AudioEvent::warpedSourceSeconds`). Playback needs the inverse, warped time to source time, and needs it without scanning a marker vector on the audio thread: the snapshot compiles the inverted map, sorted and monotonic, and a voice steps through it. A warped clip is the stretch path with a ratio that changes at every marker, so this sits on slice 4 rather than beside it.
- **Auto detect beats** and beat sensitivity: transient detection over a source, which is where markers come from when the user has not placed them. Off the audio thread, cached per source, and an analysis MAGDA gets from Tracktion today, so it is a native implementation and not a binding.
- **Loop info**: the source BPM and beat count as a file reports them, seeded into an event's interpretation (`seedInterpretation`, `seedInterpretationFromSource`) and never overwriting what the user set. Tracktion reads these out of the file's own metadata today, so either the native reader gains that or the seeding loses its source.

Interpretation is the model's, not the engine's: what this slice owns is reproducing the mapping the model describes and supplying the analyses the model seeds itself from.

Issue #2037, opened 2026-08-06
Arranger clips slice 4: time stretch and pitch


lucaromagnoli:
Slice 4 of #1890, after slice 3.

Playing a clip at a speed and a pitch that are not the file's.

## Vendoring

Signalsmith and SoundTouch both live inside the Tracktion fork today (`third_party/tracktion_engine/modules/tracktion_engine/3rd_party`), and the engine target may not include TE: that boundary is checked at configure time. They move to `third_party/` with their licences. Signalsmith is MIT. SoundTouch is LGPL, so how it is linked is a decision to state here rather than inherit from the fork.

The persisted mode values stay pinned: `kDisabled`, `kSoundTouchNormal`, `kSoundTouchBetter`, `kSignalsmith` in TimeStretchModes.hpp are project-file integers and do not move. `AudioEvent::getEffectiveTimeStretchMode` is the rule for what actually runs, including the case where a mode left at Off still stretches because something else forced a stretcher on.

## Playback

- Speed ratio, which is a constant factor and resolves in the snapshot.
- Auto tempo, which is not. Its ratio is projectBpm over interpBpm and moves with the tempo curve, so it cannot be resolved to seconds ahead of the block. It reads the beat face of BlockInfo instead, which is why the clock publishes both faces of one instant: no second tempo map on the audio thread.
- Pitch: `pitchChange` and `transpose` through the stretcher, analog pitch by resampling instead (`isAnalogPitchActive`, in force only when nothing else has already forced a stretcher on).
- Speed ramp fades (`fadeInBehaviour` / `fadeOutBehaviour` set to speedRamp), which are a stretch feature rather than a gain one and are why fade behaviour is in the snapshot.

## The state question

A stretcher has internal state and latency, and both have to be answered rather than discovered:

- what a seek, a loop wrap and a plan swap do to it. One that keeps its state across a locate plays the wrong material; one that is rebuilt per block allocates on the audio thread.
- what its latency does to alignment against unstretched voices on the same track, and whether that is compensated here or reported to the plan's latency pass.

Issue #2036, opened 2026-08-06
Arranger clips slice 3: rate conversion, looping and reverse


lucaromagnoli:
Slice 3 of #1890, after slice 2.

What one voice reads, at unity speed. All of it is a mapping from the block's seconds onto the samples of a file, and none of it needs a stretcher.

- **Sample rate conversion**, for a file whose rate is not the device's. Left open on purpose by ClipPlacement.hpp: `sourceSampleAt` maps at the device's rate so the reader stays in step with the callback, which means a mismatched file plays at the wrong pitch until something resamples it. That something is here. Every file imported at a rate other than the device's is this.
- **Loop tiling** of an event's own source region (`loopEnabled` with `loopStartSamples` / `loopLengthSamples`): one event repeats to fill its length. A tile boundary is a position the reader was not pointed at, so it is a seek and has to be cued like a clip start.
- **Reverse** (`AudioEvent::reversed`). The stream reads forwards, so this is either a backwards prefetch or a chunk flipped on the way out, and picking between them is the decision this slice makes.
- **Active channels** (`leftChannelActive` / `rightChannelActive`) and the event's own gain trim, which sits under the clip's gain rather than replacing it.

PR #2045, opened 2026-08-07
Native engine: what a track's clips sum to (#2035)

### LINKED ISSUE #2035: Arranger clips slice 2: audio voices, spans and fades
Slice 2 of #1890, after slice 1.

The source behind a track's `ClipAudio` op: which clips sound in this block, and what they sum to. One prefetch stream (#2016) per sounding entry.

- A voice per sounding entry, taken from the snapshot. Overlapping clips sum, because the snapshot already decided which of them is audible where.
- A voice is cued before it starts rather than seeked when it starts. A read that does not continue the last one is a seek and costs a block of silence (#2016), and a clip entering on the beat is exactly that read; `PrefetchStream::seek` and `sourceSampleAt` are public so that whoever schedules material can avoid it. How far ahead to cue is this slice's number to pick.
- The audible span crops what a voice reads, the interior silences punch holes in it, and the resolved fades shape its edges. A crossfade between two clips is two voices playing the fades the snapshot handed them and nothing else.
- Clip volume, gain and pan, and the launch fade on the stopped to playing transition (`ClipInfo::launchFadeSamples`, 0 preserves the leading transient).
- How many voices a track may have at once is decided here rather than discovered. A track asking for more simultaneous clips than there are streams reports it the way an underrun is reported, because silence nobody counted is indistinguishable from a gap in the material.

Silence is what a voice with nothing to play renders, and the block is filled whatever happens: the executor hands the source an uncleared buffer.

Not here: rate conversion, looping and reverse (slice 3), stretch and pitch (slice 4).

lucaromagnoli:
Slice 2 of #1890, after #2034. The source behind a track's `ClipAudio` op, and the readers it plays through.

`ClipAudioSource` reads the snapshot slice 1 publishes, matches each sounding entry to a voice, and sums them. It decides nothing about overlaps: occlusion, crossfades and takes were resolved when the snapshot was compiled, so two clips over one stretch of lane simply add, and a crossfade is two voices each playing the fade it was handed (#2003).

A voice crops its read to the audible span, clears the interior silences out of what it read rather than skipping over them, shapes the span's edges with the resolved fades, and applies the clip's gain and pan with the incumbent's linear law. Reading across a hole is what keeps the far side of it from being a seek. A voice that begins mid-material has the step taken out of it by the launch ramp, which subtracts the starting offset rather than fading the audio, so a leading transient survives and a setting of zero preserves it exactly.

`ClipVoicePool` provisions the readers a second ahead of the transport, on a thread that may wait for a disk. A clip about to sound has its file opened and its stream pointed at the sample it starts on before any callback asks, which is the difference between a clip starting on the beat and a clip starting a block late (#2016). `PrefetchStream` gained `startAt` for that: a stream nobody has read from is moved rather than cued, so its first fill is not spent on the wrong part of the file and then thrown away.

Both ceilings are decided rather than discovered. Eight voices per track, because a crossfade is two, a clip dropped inside another is two, and a clip made of several events is one voice per event (#1901); past eight the track reports through `starvedVoices` the way an underrun is counted, since silence nobody counted is indistinguishable from a gap in the material.

`EngineSession` takes the pool, so publishing clips reaches the audio thread and the reader together, and the transport's position reaches the thread reading ahead of it. A snapshot arriving at one side and not the other would be a clip that is audible with nothing to read, which is the sync problem keeping one representation was supposed to have made impossible.

## Two calls worth review

The snapshot carries per-event fades and this applies only the clip's resolved pair. For the single event a clip has today the two are the same fade before and after resolution, so applying both would shape the edge twice; interior event fades belong with the clips that have interior events (#1901).

Event gain trim is not applied here. #2036 owns it, along with active channels.

## Verification

`tests/test_clip_voice.cpp` rolls a transport rather than skipping between blocks, because a read that does not continue the last one is a seek and would be testing a locate. It covers placement and trims, overlapping clips summing, a hole punched mid-block and the material continuing past it, every fade curve and where it is anchored, gain and pan, the launch ramp and its zero case, a starved clip, the voice ceiling, and a voice released in the block its clip ends in.

`tests/test_clip_voice_pool.cpp` covers the window, retirement and reopening, a file that will not open being reported once rather than retried, over-subscription, a sounding clip keeping its reader over one that has not started, and the end to end path from provisioning to a clip playing from its first sample with no underrun.

Confirmed to fail without their fix: disabling the hole clearing, the launch ramp and the voice-release sweep fails exactly those four tests and no others.

Not here: rate conversion, looping and reverse (#2036), stretch and pitch (#2037), warp (#2038), MIDI (#2039).

### DIFF
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -101,10 +101,12 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
                 });
 
             if (found == lastStream || soundingCount == kMaxVoicesPerTrack) {
-                // No reader standing by, or no room left for one. One number
-                // bounds both (kMaxVoicesPerTrack), so the second of these is
-                // only reachable through a table this source did not get from
-                // the pool; either way the clip does not sound and the count is
+                // No reader standing by, or no voice left to play it through.
+                // Both happen and they are different failures: a track may hold
+                // more readers than a callback has voices (kMaxReadersPerTrack
+                // is the larger, because a reader has to exist before its clip
+                // is due), so a block can be handed more entries than it can
+                // render. Either way the clip does not sound and the count is
                 // what says so.
                 starved_.fetch_add(1, std::memory_order_relaxed);
                 continue;
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -80,6 +80,14 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
     auto soundingCount = 0;
 
     for (const auto& clip : track->audio) {
+        // Sorted by where they start (ClipSnapshot.hpp), so once one begins at
+        // or after the end of this block, so does everything behind it. Without
+        // the break a track pays for its whole tail on every callback, all
+        // session, and the cost grows with the length of the arrangement rather
+        // than with what is playing.
+        if (clip.span.startSeconds >= block.endSeconds)
+            break;
+
         if (!reachesInto(clip.span, block))
             continue;
 
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -101,8 +101,11 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
                 });
 
             if (found == lastStream || soundingCount == kMaxVoicesPerTrack) {
-                // No reader standing by, or no room left for one. Either way
-                // this clip does not sound and the count is what says so.
+                // No reader standing by, or no room left for one. One number
+                // bounds both (kMaxVoicesPerTrack), so the second of these is
+                // only reachable through a table this source did not get from
+                // the pool; either way the clip does not sound and the count is
+                // what says so.
                 starved_.fetch_add(1, std::memory_order_relaxed);
                 continue;
             }
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -1,0 +1,139 @@
+#include "clip/ClipAudioSource.hpp"
+
+#include <algorithm>
+
+namespace magda::engine {
+
+namespace {
+
+/// Whether a span reaches into the block at all. Half-open at both ends, so a
+/// clip ending exactly on a block boundary contributes nothing to the block
+/// that starts there.
+bool reachesInto(const SnapshotSpan& span, const BlockInfo& block) {
+    return span.startSeconds < block.endSeconds && span.endSeconds > block.startSeconds;
+}
+
+}  // namespace
+
+ClipAudioSource::ClipAudioSource(TrackId trackId, ClipSnapshotFeed& clips, ClipStreamFeed& streams)
+    : trackId_(trackId), clips_(clips), streams_(streams) {}
+
+void ClipAudioSource::prepare(const RenderContext& context) {
+    scratch_.setSize(context.numChannels, context.maxBlockSize, false, true, false);
+    scratch_.clear();
+
+    for (auto& voice : voices_)
+        voice.prepare(context);
+}
+
+ClipVoice* ClipAudioSource::voiceFor(ClipId clipId, EventId eventId) {
+    for (auto& voice : voices_)
+        if (voice.playing(clipId, eventId))
+            return &voice;
+
+    for (auto& voice : voices_)
+        if (voice.clipId() == INVALID_CLIP_ID)
+            return &voice;
+
+    return nullptr;
+}
+
+void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float> out) {
+    out.clear();
+
+    const ClipStreamFeed::Reader streams(streams_);
+    const auto [firstStream, lastStream] =
+        streams ? streams->rangeFor(trackId_)
+                : std::pair<const ClipStreamTable::Entry*, const ClipStreamTable::Entry*>{nullptr,
+                                                                                          nullptr};
+
+    // Every stream this track has, sounding or not, once per block. The stream
+    // a cue is most useful to is one nobody is reading from: a clip that has
+    // not started would otherwise not hear about the position it was pointed at
+    // until the material was already due (#2016).
+    for (const auto* entry = firstStream; entry != lastStream; ++entry)
+        entry->stream->applyPendingCue();
+
+    const auto numSamples = std::min(block.numSamples, static_cast<int>(out.getNumSamples()));
+
+    const auto silence = [this] {
+        for (auto& voice : voices_)
+            voice.release();
+    };
+
+    if (!block.playing || numSamples <= 0)
+        return silence();
+
+    const ClipSnapshotFeed::Reader snapshot(clips_);
+    if (!snapshot)
+        return silence();
+
+    const auto* track = snapshot->find(trackId_);
+    if (track == nullptr)
+        return silence();
+
+    // What sounds, and through what. Gathered before anything is rendered so
+    // the voices that are not in it can be let go first: a voice still holding
+    // a clip that stopped last block would otherwise keep a slot a clip
+    // starting this one needs.
+    std::array<Sounding, kMaxVoicesPerTrack> sounding;
+    auto soundingCount = 0;
+
+    for (const auto& clip : track->audio) {
+        if (!reachesInto(clip.span, block))
+            continue;
+
+        for (const auto& event : clip.events) {
+            if (!reachesInto(event.span, block))
+                continue;
+
+            const auto* found =
+                std::find_if(firstStream, lastStream, [&](const ClipStreamTable::Entry& entry) {
+                    return entry.clipId == clip.clipId && entry.eventId == event.eventId;
+                });
+
+            if (found == lastStream || soundingCount == kMaxVoicesPerTrack) {
+                // No reader standing by, or no room left for one. Either way
+                // this clip does not sound and the count is what says so.
+                starved_.fetch_add(1, std::memory_order_relaxed);
+                continue;
+            }
+
+            sounding[static_cast<std::size_t>(soundingCount++)] =
+                Sounding{&clip, &event, found->stream.get()};
+        }
+    }
+
+    for (auto& voice : voices_) {
+        if (voice.clipId() == INVALID_CLIP_ID)
+            continue;
+
+        const auto stillSounding =
+            std::any_of(sounding.begin(), sounding.begin() + soundingCount, [&](const Sounding& s) {
+                return voice.playing(s.clip->clipId, s.event->eventId);
+            });
+
+        if (!stillSounding)
+            voice.release();
+    }
+
+    auto scratch =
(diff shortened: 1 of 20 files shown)


magda/engine/clip/ClipVoicePool.cpp
```diff
@@ -0,0 +1,219 @@
...
+                }
+
+                for (const auto& candidate : candidates) {
+                    const Key key{track.trackId, candidate.clipId, candidate.eventId};
+
+                    // Already provisioned: kept as it is, and never re-cued. A
+                    // stream that is playing cannot be pointed anywhere else
+                    // anyway, and one that is waiting is already pointed at the
+                    // sample the clip starts on.
+                    if (const auto found = streams_.find(key); found != streams_.end()) {
```
lucaromagnoli: **[P1] Reopen a stream when the event behind a stable ID changes.** This lookup keys reuse only by track/clip/event ID. If an edit changes the event’s file while preserving those IDs, the new snapshot is published but this keeps the reader for the old file, so playback can continue producing the wrong audio indefinitely. Anchor or trim changes likewise reuse prefetch data for the old placement and force an audio-thread seek/gap. Could the retained entry include the file and relevant placement identity (or a snapshot revision/fingerprint) and replace or re-cue the stream when they differ?

magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -0,0 +1,139 @@
...
+        for (const auto& event : clip.events) {
+            if (!reachesInto(event.span, block))
+                continue;
+
+            const auto* found =
+                std::find_if(firstStream, lastStream, [&](const ClipStreamTable::Entry& entry) {
+                    return entry.clipId == clip.clipId && entry.eventId == event.eventId;
+                });
+
+            if (found == lastStream || soundingCount == kMaxVoicesPerTrack) {
```
lucaromagnoli, line 103: **[P1] Cap actual overlap, not every entry intersecting the callback.** `soundingCount` includes every event touching this block, even when their sample ranges do not overlap. Nine short sequential clips inside one callback therefore drop the ninth although the maximum instantaneous voice count is one. The pool has the same issue when it truncates every candidate in the one-second window. This needs either within-block voice reuse/scheduling by sample range or a separate capacity for provisioned sequential entries.

magda/engine/clip/ClipVoice.cpp
```diff
@@ -0,0 +1,155 @@
...
+    // an underrun. All three start the material wherever it happens to be.
+    if (!sounded_ || !block.continuous)
+        applyStartDeClick(region, clip.launchFadeSamples);
+
+    out.getSubBlock(static_cast<std::size_t>(first), static_cast<std::size_t>(count)).add(region);
+
+    // Silence the reader could not fill is not sounding. Saying otherwise would
+    // let the block that resumes start mid-material with nothing to take the
+    // step out of it.
+    sounded_ = delivered > 0;
```
lucaromagnoli: **[P2] Treat a partial read as an underrun before the next block.** `PrefetchStream::read` may return fewer than `count` samples, leaving trailing silence, but any positive delivery marks the voice as continuous. If the reader catches up for the next block, playback resumes from silence without the launch ramp described above. Should this stay sounded only when `delivered == count`?

magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -0,0 +1,139 @@
...
+        return silence();
+
+    // What sounds, and through what. Gathered before anything is rendered so
+    // the voices that are not in it can be let go first: a voice still holding
+    // a clip that stopped last block would otherwise keep a slot a clip
+    // starting this one needs.
+    std::array<Sounding, kMaxVoicesPerTrack> sounding;
+    auto soundingCount = 0;
+
+    for (const auto& clip : track->audio) {
```
lucaromagnoli, line 82: Clips within a track are sorted by start (ClipSnapshot.hpp), so this walk can break once clip.span.startSeconds >= block.endSeconds: nothing later can reach into the block. As written, every clip on the track is visited every block on the audio thread, so a long arrangement pays for its whole tail all session. The clips before the cursor still have to be walked (an early start can still reach into the block), but the break bounds the far side for free.

magda/engine/clip/ClipVoicePool.cpp
```diff
@@ -0,0 +1,219 @@
...
+                    auto stream = open(*candidate.event, cueSeconds);
+                    if (stream == nullptr)
+                        ++unreadable;
+
+                    wanted.emplace(key, std::move(stream));
+                }
+            }
+        }
+
+        auto table = std::make_shared<ClipStreamTable>();
```
lucaromagnoli: A round that changes nothing still allocates a table here and publishes it, every 10 ms for the life of the session; the service() docstring says as much. wanted and streams_ compare by key and pointer, so an early `if (wanted == streams_)` bail after the walk would make the idle steady state allocation-free, and the retire loop and swap fall away with it. It also takes nonRealtimeReplace's wait-for-the-callback out of the idle path, which is currently paid at 100 Hz for nothing.

magda/engine/clip/ClipVoicePool.cpp
```diff
@@ -0,0 +1,219 @@
...
+};
+
+}  // namespace
+
+ClipVoicePool::ClipVoicePool(AudioFileReaderFactory& files, PrefetchThread& reader,
+                             const RenderContext& context, const PrefetchSettings& settings)
+    : files_(files), reader_(reader), context_(context), settings_(settings) {}
+
+ClipVoicePool::~ClipVoicePool() {
+    // Nothing may be reachable from the audio thread once this returns, and
```
lucaromagnoli, line 166: This is only safe once nothing can call service() any more: a round in flight would republish a full table right after this empty one, and the streams it names would then be pulled off the prefetch thread beneath it. That contract holds when ClipVoiceThread is destroyed first, but the pool does not own its thread and the production wiring is not in this slice, so nothing enforces or states the order. Worth a line here naming it, so whoever composes the two in the host slice cannot get it wrong quietly.

magda/engine/exec/EngineSession.hpp
```diff
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "clip/ClipSnapshotFeed.hpp"
+#include "clip/ClipVoicePool.hpp"
```
lucaromagnoli: Only a pointer is stored and every call on it lives in the .cpp, so `class ClipVoicePool;` would do here. The include pulls juce_core, std::map and PrefetchThread into everything that includes the session.

lucaromagnoli (commented):
Reviewed against the incumbent, since parity is the load-bearing claim here. What I checked:

- The four fade shapes are AudioFadeCurve's formulas exactly (Convex sin, Concave 1-cos, SCurve blend), and a fade out as the same curve at 1 - alpha is precisely what FadeInOutNode does through applyCrossfadeSection.
- The pan law is LiveClipLevel::getLeftAndRightGains verbatim: g - pan*g / g + pan*g.
- applyStartDeClick is a faithful port of applyAudioStartDeClick, including clamping the ramp to the buffer it was handed.
- The rounding pairing between BlockInfo::sampleForTime (lround) and sourceSampleAt (llround) is consistent, so rolling blocks stay contiguous and never cost a spurious seek, including a voice that begins mid-block off the sample grid.
- The publish-before-retire order in service() and the shared_ptr discipline mean a stream is only ever destroyed on the provisioning thread, after both the feed swap and the prefetch removal. Sound.

On the two calls flagged for review: I agree with both. The clip's resolved pair is what the lane decided the edges play, and applying the event's copy on top would shape every edge twice today; the event pair belongs with interior events and #1901. Event gain trim with active channels in #2036 is the right grouping.

One small question: fadeInBehaviour and fadeOutBehaviour ride along in the snapshot but a behaviour of 1 (speed ramp) plays as a gain fade here. Reasonable interim while stretch is #2037's business, but nothing says it is deliberate; worth a line if it is.

The rolling-transport rig is the right instrument for this layer, and the confirmed-to-fail-without-fix note is what makes the coverage claims credible. Four minor points inline, none blocking.

magda/engine/clip/ClipStreamFeed.hpp
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+namespace magda::engine {
+
+/**
+ * @brief The streams provisioned for every track, at one moment.
+ *
+ * Sorted by track, then clip, then event, so the audio thread can find a
+ * track's range without hashing and walk it without searching. A track's range
+ * is at most kMaxVoicesPerTrack long, which is what makes the walk free.
```
lucaromagnoli: Now that the reader budget is kMaxReadersPerTrack, a track's range can be sixteen entries rather than eight, so this names the wrong ceiling. The walk is still free at sixteen; only the constant in the sentence is stale.

lucaromagnoli (commented):
Re-reviewed the follow-up commit. Everything from the first round is resolved, and the new work holds up:

- The reader budget split from the voice ceiling is the right shape, and measuring overSubscribed as peak simultaneity through the sweep rather than window population fixes a misreport I had not flagged: chopped material would have looked like a track in trouble.
- The Reader identity (path, anchor) closes a real staleness hole under stable ids. I traced the branches: a swapped file falls through to open() and the retirement loop's pointer comparison retires the old stream; a trim on a not-started clip re-cues through seek(), which is the correct mechanism for a registered stream, and the target (span start) agrees with the original cue rule since not-sounding implies spanStart > windowStart. A trim on a sounding clip self-corrects on the next read at the cost of one counted block, as documented.
- sounded_ = delivered == count is right, and the half-fill test's continuation is the interesting part: the block after the gap reads at the advanced cursor without a seek, so only the ramp changes.
- The idle skip via map inequality is exact, and tablesPublished is a good tripwire for regressions there.

One stale line inline, nothing else.

magda/engine/clip/ClipAudioSource.cpp
lucaromagnoli: This remains blocking after 381a3d4. The pool now measures instantaneous overlap correctly, but `ClipAudioSource` still puts every entry touching the callback into the fixed eight-element `sounding` array. Nine back-to-back clips within one callback therefore have peak overlap one and `overSubscribed() == 0`, while the ninth is still dropped and increments `starvedVoices()`. The scale is also materially larger than the new comment suggests: at 44.1 kHz, nine clips inside the default 512-sample block means roughly one clip every 1.3 ms—not ten clips per millisecond—and larger device blocks make ordinary chopped material easier to hit. Please either reuse voices for non-overlapping sample regions within a block or define and report the ceiling consistently as entries per callback rather than simultaneous voices.

lucaromagnoli (commented):
Third pass, on the unified-ceiling commit. Nothing to flag.

I checked the sweep's block extension for exactness rather than taking conservative on faith, and it is exact: a clip holds a voice in every callback whose window intersects its span, so the set of clips live at sweep instant t, meaning from <= t < to + blockSeconds, is precisely the set a block ending at t touches, and the sweep's peak equals the worst clips-per-callback over every possible block placement. Since block phase is not locked to the timeline, that worst case is reachable, so reporting it is honest rather than pessimistic. The tie rule still lands right after the extension: a gap of exactly one block is a handover.

With one ceiling, the pool provisions at most sixteen entries per track, so the callback's voice-cap branch is back to unreachable from a pool-built table, and the rewritten comment states exactly that.

The two new tests bracket the boundary from both sides, and the short-slice one exercises the cost through the path it actually takes: the seventeenth slice starves for want of a reader, which is the same ceiling seen from the other half of the layer. The ClipStreamFeed comment naming the bound instead of a constant closes my last nit in a way that cannot go stale again.

This resolves the whole review. Good to merge from my side.

PR #2046, opened 2026-08-07
Native engine: rate conversion, looping and reverse (#2036)

### LINKED ISSUE #2036: Arranger clips slice 3: rate conversion, looping and reverse

lucaromagnoli:
Slice 3 of #1890, after #2045. What one voice reads, at unity speed.

Reverse, looping and rate conversion turn out to be one question asked three times: which of a file's samples answer a position. So none of them is processing, and each is a reader wrapped around the reader, built once when the pool opens a clip (`io/SourceReaders.hpp`). Everything above goes on seeing one forward file at the device's rate: `PrefetchStream` is untouched, and a voice still consumes one sample of its reading per output sample whether the clip is reversed, tiled, resampled or none of them.

## The decision this slice was to make

Reverse is neither a backwards prefetch nor a chunk flipped on the way out. A prefetch stream reads ahead and can only be pointed at one place, so a voice asking for descending blocks would leave the reader somewhere it was not pointed every block, which is a seek per block and silence for the whole take (#2016). The file is mirrored underneath instead: the stream reads forwards through a file whose samples are in the other order, and the flip is done by the thread that may wait for a disk rather than by the one that may not. The incumbent renders a reversed copy of the file to disk for this. Nothing is written here and nothing has to finish before the clip can play.

They compose in the incumbent's order, mirror then tile then convert. Tiling over the mirror is what makes a reversed loop the same region played backwards, which is what its reversed loop points say; converting last leaves everything below in the source's own samples, where the model holds them, and everything above in the device's, where a stretch whose ratio changes with the tempo will have to live (#2037).

## What each of them cost

**A loop wrap is no longer a seek.** Tiling happens below the stream, so the reader walks straight through a tile boundary. `PrefetchStream::seek` said a seamless return needed a second fill cursor and a second pool and belonged with whoever owned clip looping; it needed neither. A wrap no longer costs the block it happens in, and the underrun count says so.

**Rate conversion has no cursor and no carried phase.** An output sample's source position is its own position times the ratio and nothing else, so two reads that overlap agree to the last bit and a seek lands exactly where reading through would have. That is what lets a bounce that jumps into the middle of a clip render what the callback played on the way there; a stateful resampler would buy a fraction of a sample of accuracy at the price of a render that could disagree with what was heard. Cubic Lagrange, the family the incumbent's default resampling quality is from. Its sinc settings are a quality option and there is nothing in the model to select one with yet.

**Mirrored coordinates are derived once, for both sides** (`clip/EventPlacement.hpp`). The pool points the reader and the voice reads it a block later, and the two never meet: two derivations of the same mirror could disagree, and a clip playing the wrong part of its own file sounds like material rather than like a fault. The model keeps its own coordinates throughout, so an editor goes on showing the region the user chose. The incumbent writes the mirrored offset back over the clip when the flag is set; this converts on the way to the reader instead.

Also here: an event's active channels, and its own gain trim under the clip's rather than in place of it.

## Two calls worth review

**A channel that is off leaves the other heard on both sides**, rather than the clip going hard to one of them. Nothing played these before, so there was no incumbent behaviour to match. Where a clip sits in the image is what its pan decides, and silencing an output would spend that decision here.

**A reversed event whose source has no known length is dropped with a diagnostic.** A mirror needs an end to turn about. It sits beside the existing check for a source whose rate is unknown, and it is the same doctrine: a clip that will not sound says why.

`speedRatio` is still unity, including in the stretch a mirrored read turns about. It arrives with the stretcher (#2037), and the line says so.

## Verification

`tests/test_source_readers.cpp` covers each reader on its own: the mirror and its edges, that it turns about the length it was given rather than the file's, loop tiling through a wrap and from a phase in front of the region, a loop over a region the file does not reach playing the silence in it rather than reporting the end of a file, rate conversion up and down, that a resampled sample is the same sample however it was reached, that nothing below is ever asked for a position before the first sample, and that a reading asked for nothing is the file itself.

`tests/test_clip_voice.cpp` adds the same things heard through a rolling transport: a file at half the device's rate in step a hundred blocks later, a reversed clip counting down from the far end of its region and ending on the sample it was trimmed to, a looped clip carrying through the block its wrap falls in with no underrun, the channel cases, and an event trim under a clip gain.

`tests/test_clip_voice_pool.cpp` adds a reversed clip cued end to end, which is the test that would fail if the two derivations of the mirror ever parted, and a reader rebuilt when what its clip asks of the file changes.

Not here: stretch and pitch (#2037), warp (#2038), MIDI (#2039).


magda/engine/io/SourceReaders.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+    for (auto channel = 0; channel < destination.getNumChannels(); ++channel) {
+        auto* samples = destination.getWritePointer(channel) + destinationOffset;
+        std::reverse(samples, samples + available);
+    }
+
+    // Every one of these positions exists, whatever the file had at the far end
+    // of them. Reporting the file's count would say the material starts where
+    // the flip has just put the silence.
+    return found > 0 ? available : 0;
```
lucaromagnoli: A run that lies wholly in the mirrored leading silence comes back 0 from here, and that stalls the stream. The shape that reaches it is ordinary: a reversed event whose span outruns its source has a negative mirrored anchor, the pool cues the stream there, and the first fill's chunk lands entirely at or past the file's end underneath, so found is 0, this returns 0, and PrefetchStream::fill sets stalled_. Nothing clears it through the whole leading silence, because the callback's fully-negative reads take the early-out in PrefetchStream::read and never move the cursor; the first block that crosses zero mismatches nextSample_, seeks, and pays an underrun exactly where the material begins. The cue machinery is defeated for precisely this clip shape, and the read-ahead it did is thrown away.

JuceAudioFileReader already draws the line this needs: silence before the file's start is padded and reported as read, only a decode failure inside the advertised length reports 0. The mirrored counterpart of that silence is the file's past-the-end region, so a found of 0 where the underlying run lies wholly at or past the file's length should return available with the buffer this has already cleared, and 0 should be kept for a read that failed inside the material. A test with a reversed clip trimmed longer than its source, rolled through the stream, would have caught this: material's first block silent, one underrun counted.

magda/engine/io/SourceReaders.cpp
```diff
@@ -0,0 +1,253 @@
...
+                          std::floor(static_cast<double>(startSample + available - 1) * ratio_)) +
+                      2;
+    const auto count = static_cast<int>(last - first + 1);
+
+    const auto channels = destination.getNumChannels();
+    if (window_.getNumChannels() < channels || window_.getNumSamples() < count)
+        window_.setSize(std::max(channels, window_.getNumChannels()),
+                        std::max(count, window_.getNumSamples()), false, true, false);
+
+    if (readFrom(*file_, window_, 0, first, count) <= 0) {
```
lucaromagnoli, line 262: Same stall as the reversed reader, through the other layer: a rate-converted clip cued into structural leading silence (a negative anchor, or below a mirror) puts the whole window before the first sample, readFrom answers 0, this early-out returns 0, and the stream stalls until the zero crossing forces a seek. An unconverted forward clip in the same shape plays cleanly because JuceAudioFileReader pads before-start silence and reports it read; the layers should agree with the base reader on what structural silence is. Distinguishing silent-because-out-of-range (return available, buffer is already cleared) from silent-because-the-file-died (return 0) fixes both sites.

magda/engine/clip/EventPlacement.cpp
```diff
@@ -0,0 +1,101 @@
...
+        // rather than the event's, because a looped event reads the region
+        // round and round and its last sample is wherever the phase had got to.
+        //
+        // The incumbent works the same value out and writes it back over the
+        // clip's offset when the flag is set. Here the model keeps its own
+        // coordinates, which is what lets an editor go on showing the region
+        // the user chose, and the conversion happens on the way to the reader.
+        const auto last = event.anchorSamples + regionOf(event, sourceRate) - 1;
+
+        anchor = how.loopLengthSamples > 0
```
lucaromagnoli, line 86: This branch is the trickiest arithmetic in the slice and nothing executes it: the voice tests cover reverse and loop separately, the pool test covers reverse without loop, and the compose test builds its SourceRead by hand, so neither this anchor nor sourceReadFor's loop mirroring runs under any test. I worked it through for an anchor at the loop start and one mid-loop and both come out right, including the wrap landing back on the region's last sample. One rolling-transport test of a reversed looped clip would pin both derivations end to end, the way the reversed pool test already does for the unlooped mirror.

lucaromagnoli (commented):
The architecture is right and I verified the load-bearing math by hand: the mirrored anchor for a plain reversed event lands on the last sample of the region, the looped-reversed branch produces the correct phase in the mirrored loop for both an anchor at the loop start and one mid-loop, the tile mirror in sourceReadFor agrees with the compose test's hand-computed region, and the resampler's window indices stay in bounds at both edges (index - 1 >= 0 from the minus-one floor, index + 2 inside the count). The stateless resampler with the bit-exact overlap test is the right trade for bounce parity, and loop-below-the-stream genuinely kills the wrap seek, with underruns() == 0 proving it where it counts.

One real defect inline, in two places with one root cause: the wrapper readers report structural leading silence as zero, where JuceAudioFileReader pads it and reports it read. That stalls the prefetch stream mid-cue for a reversed clip whose span outruns its source, and for a rate-converted clip cued into leading silence, and the recovery path is a seek that costs a block exactly where the material begins.

On the two calls flagged for review: I agree with both. A disabled channel collapsing to the remaining one keeps the pan decision where it belongs, and matches what a mono source already does through the fan-out. The reversed-unknown-length diagnostic sits in the right place and reads like its neighbours.

One doc note that cannot take an inline comment because the file is untouched: PrefetchStream::seek's paragraph about a loop's return not being seamless, and a wrap costing the block it happens in, is now stale. Wraps happen below the stream and never reach seek, which the PR body itself points out; the comment should say so or go.

Test coverage is otherwise strong, and the reversed pool test guarding the two derivations of the mirror is exactly the test I would have asked for.

magda/engine/io/SourceReaders.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+    if (done < numSamples)
+        destination.clear(destinationOffset + done, numSamples - done);
+
+    // Silence inside the region is material: a loop over a region the file does
+    // not reach plays that silence every time round, and calling it the end of
+    // the file would leave the stream above asking for it again instead of
+    // moving on. A read with nothing anywhere in it is the other thing, and
+    // says so, because a reader that has stopped answering has to be visible.
+    return material ? numSamples : 0;
```
lucaromagnoli: A loop that partially extends beyond the file can stall permanently here. If its silent tail is longer than a prefetch chunk, a read wholly inside that tail leaves `material` false and returns `0`. `PrefetchStream` treats this as a permanent reader failure, so it never advances to the next wrap and the valid beginning of the loop never plays again. Could intentional out-of-file portions advance as silence, with a stream-level regression test where the silent tail exceeds the chunk size?

lucaromagnoli (commented):
Re-reviewed the structural-silence commit. Everything from the first round is resolved, and the fix is stronger than what I asked for.

insideMaterial draws the line in the one place the base reader draws it, and judging it against the file's own length rather than the mirror axis is the right call: a stale source table that believes the file longer puts the mirrored silence past the real material, which is structural, which walks. I traced the unreachable case too, from below zero cannot happen once available has been clamped, so the helper's two bounds are both doing work.

The PrefetchStream change is the part I had not asked for and it is necessary: dropping the fully-negative early-out is what makes the stream walk the leading silence instead of standing still beside it, so the read-ahead arrives at the material intact, which the new voice test pins with the underrun count. The knock-on I checked: a cue published mid-silence now waits, because the stream is legitimately sounding through its clip's own silence, which is consistent with how a sounding stream already treats a cue.

The loop-tail stall was a shape I had not called out, found and fixed here with the same doctrine, and the reversed-looped voice test covers exactly the two derivations I flagged as untested, with the expected-phase arithmetic agreeing with placementFor for both anchors, I re-derived both. The DeadReader sections keep the fault-visibility half of the contract honest on all three layers, and the rewritten seek() comment closes the doc note.

Nothing left from my side. Good to merge.

Issue #1892, opened 2026-07-27
Engine: rack graph (pins, summing, multi-out, nesting)


lucaromagnoli:
Part of #1882. Depends on the engine core and the parameter system.

Rack graph at parity with RackSyncManager's usage of TE racks.

- Pin-level connection graph with implicit summing at shared pins.
- Multiple output pairs, rack nesting (rack instances inside racks, recursion rejected), per-chain volume/pan/mute/solo.
- Per-chain MIDI bus semantics: instruments consume MIDI and become the audio source, non-instruments pass MIDI through.
- Sidechain sources for rack-internal devices.
- In the plan model these become compiler passes rather than a runtime object graph; the teardown-and-rebuild pattern in RackSyncManager disappears.

lucaromagnoli:
Scope note from breaking down #1896: this issue carries its own final parity slice, the way
#2040 was slice 7 of #1890.

Cases for what is ported here are written here, not in #1896. A parity case is worth what its
author's knowledge of the feature is worth, and that knowledge is at its peak on the day the
port lands. Written later by someone reading the code back, a case asserts the code rather than
the behaviour, which is the one thing a parity corpus must not do.

#1896 keeps the cross-cutting infrastructure the slice builds on: the whole-project case and the
tiered oracle (#2075), the goldens (#2076), the block-size gate (#2078), and the corpora that
span features (#2081, #2082).

lucaromagnoli:
Sliced, the way #1889, #1890 and #1891 were:

- #2135 aux outputs and multi-out tracks
- #2136 delta solo at device and rack scope
- #2137 nesting, recursion and op-key identity
- #2138 pins, channel counts and implicit summing
- #2139 parity cases for rack topologies

More of this issue already exists than its body implies. `Compiler::emitRack` compiles chains,
per-chain faders with volume, pan, mute and solo, the rack mix, the rack instance's own output
gains with their attenuate-only pan law, MIDI generation semantics, bypass, the empty-rack
passthrough, and nesting by recursion. Sidechain sources inside racks already contribute compile
ordering edges. What is left is what the compiler names in its own diagnostics, plus the channel
model, which is stereo everywhere and needs not to be.

lucaromagnoli:
All five slices are closed: aux outputs and multi-out tracks (#2135), delta solo (#2136), nesting and op-key identity (#2137), pins and implicit summing (#2138), parity cases (#2139).

PR #2048, opened 2026-08-07
Docs: a map of the native engine, and a shape for docs/ (#1882)

lucaromagnoli:
A map of the native engine (#1882), for reading before the code.

The engine is nine slices of #1889 and three of #1890 deep, spread over six directories, and the only way in today is to open a header and hope it is one of the ones that explains itself. Most of them do. What none of them can say is how the pieces stand to each other, because each is written from inside its own problem.

## What is in it

- **Where it stands**, issue by issue, including the slices of #1889 and #1890 individually.
- **The one idea**: four immutable values compiled off the audio thread and swapped in whole, separate because they change at wildly different rates. With the rule of thumb worth remembering, that moving a clip or a fader never recompiles a plan.
- **What a plan is**, with the compiler's own golden dump of a one-track session, and why one device is three ops.
- **The life of an edit**: compile, validate, diff, bind, prepare, publish, retire, and what each of those is protecting.
- **Who runs what**: the five threads, and the one rule they exist to keep.
- **How a clip plays**: snapshot to pool to streams to voices, the two ceilings and why they are different numbers, and the reading chain.
- **What is not built yet**, so nobody goes looking for racks or automation.
- Where the code lives, and how to build and test it.

Six mermaid diagrams, because the shapes here are graphs and threads, and a paragraph describing a graph is a paragraph nobody finishes. GitHub renders them inline.

## What it does not do

It does not restate the headers. They stay the documentation: every file in `magda/engine/` opens with a block explaining why it is shaped the way it is, and this says so at the top. When the two disagree the headers are right and this is stale. Each section points at the files that own the detail.

## Checked rather than remembered

- The golden dump is byte-identical to `test_render_plan_compiler.cpp` (my first pass had a space wrong).
- The ceilings, the cue-ahead window and the op vocabulary come from the constants and the enum on this branch.
- The test tags are the ones that exist here. My first pass listed `[source]`, which on dev belongs to unrelated model tests and only becomes an engine tag when #2046 lands.
- All six diagrams were rendered through mermaid-cli rather than eyeballed.
- The reading-chain section is flagged as slice 3 in review, because those files are on #2046 and not on dev yet.

---

# And the directory the map lives in

Same branch, second commit, because the two are one piece of work: a map is no use in a directory nobody can navigate.

`docs/` was thirty-odd files in one flat list, in three naming conventions, where a January bug report sat beside the testing guide with nothing to say which was still true. Now six directories with the rule for each written down in a new `docs/README.md` index.

| Directory | What belongs in it |
| --- | --- |
| `architecture/` | how a subsystem works and why it is shaped that way |
| `specs/` | a feature or a piece of work described before it exists |
| `development/` | how to work on the codebase |
| `ci/` | the repository itself: builds, runners, protections, templates |
| `product/` | what MAGDA is for, and how it is shown |
| `archive/` | finished. Kept because it explains how something got this way |

`issues/` and `smoke-tests/` are untouched working notes.

**Archive earns the rest.** Seven documents were true when written and describe work long since done: two workflow setup summaries, a February bug analysis, the local LLM POC, the Faust free-tier POC notes, an earlier manifesto draft, and the January test plan the testing guide replaced. Filed rather than deleted, so nobody has to open one to find out it is history.

**Two names that lied.** `AUTOMATION_ARCHITECTURE.md` is an analysis of Tracktion parameter writes and is now called that. `FAUST_POOL_REFACTOR.md` describes the parameter pool that shipped, not a pending refactor. Everything else is kebab-case, which the newer half already used.

**References followed the files**: both READMEs, four Faust headers citing the pool spec, `.gitleaksignore` (which pins a false positive to a path *and* a line), and three cross-links inside docs. Every link in the tree resolves, checked with a script rather than by eye, index included.

## Two notes

**`README_CN.md` shows as fully rewritten, and that is line endings.** It was still CRLF, so the repository's own `mixed-line-ending` hook normalised it the moment a link edit staged it. The alternative was leaving the Chinese README pointing at moved files.

**The reorganisation commit is marked `[skip ci]`.** What is not documentation in it is four comment lines and one allowlist path, both checked by hand: the header diffs are single-line comment replacements, and line 187 of the moved settings guide is still the AWS example key gitleaks was told to ignore there. Merging into dev runs the pipeline over all of it.


docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+    participant P as publishing thread
+    participant S as RuntimeStateStore
+    participant A as audio thread
+
+    U->>P: the model changed
+    P->>P: compile a new plan
+    P->>P: validate it
+    P->>P: diff it against the live one
+    P->>S: bind ops to instances
+    S-->>P: plugins and readers, kept by model id
```
lucaromagnoli: `RuntimeStateStore` does not bind or own file readers. It realizes devices, track-level clip sources, hardware inputs, and meter taps; the readers and streams are opened and owned independently by `ClipVoicePool` and never enter `PlanBindings` or the published plan epoch. Could this label the objects the store actually returns so the lifecycle here agrees with section 6?

docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+```
+
+What travels on which channel, and what it costs:
+
+| Channel | Carries | Changes when | Cost of a change |
+| --- | --- | --- | --- |
+| `RenderPlan` | signal topology only | a device moves, a track is added, a chain is bypassed | a compile, at human speed |
+| `PlanValues` | gains, faders, pans, sends, mutes | a mixer move | resolve a flat table, no compile |
+| `ClipSnapshot` | what every track plays, resolved | a clip moves, is trimmed, is faded | recompile the snapshot, no plan touched |
+| `TransportSnapshot` | where the timeline is | every block | published by the clock |
```
lucaromagnoli: `TransportSnapshot` is not published every block and does not carry the live playhead position. `EngineSession::publishTransport` replaces it only for tempo, loop, click, or transport-request changes; `TransportClock::advance` owns and advances the cursor on the audio thread, exposing that position separately. Could the table and diagram describe the snapshot as clock input rather than something published by the clock each block?

docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+---
+
+## 7. What is not built yet
+
+Worth knowing before reading the code and wondering where something is:
+
+- **Parameters and automation** ([#1891](https://github.com/Conceptual-Machines/magda-core/issues/1891)).
+  `PlanValues` carries mixer values; automation curves, modifiers and macros do not exist in the
+  engine yet.
+- **Racks** ([#1892](https://github.com/Conceptual-Machines/magda-core/issues/1892)). The op
```
lucaromagnoli: Basic rack compilation already exists: `PlanCompiler::emitRack` emits rack-chain faders, rack mixes, rack output faders, MIDI merges, and nested racks, with compiler and executor tests covering them. What remains for #1892 is the full pin graph, auxiliary/multi-output routing, and related parity. Could this distinguish that missing work from the rack support already emitted?

docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+[  7] Gain        det   T1:trackMute                   in=6:0              out=audio       deps=1
+[  8] MixAudio    det   T-2:trackAudioInput            in=7:0              out=audio       deps=1
+[  9] Fader       det   T-2:trackFader                 in=8:0,-            out=audio       deps=1
+[ 10] Meter       det   T-2:trackMeter                 in=9:0              out=audio       deps=1
+[ 11] Gain        det   T-2:trackMute                  in=10:0             out=audio       deps=1
+[ 12] Output      det   T-2:hardwareOutput             in=11:0             out=-           deps=1
+ready=0
+```
+
+`T-2` is the master. `det` is the op's liveness domain, carried from day one and unused until
```
lucaromagnoli: Liveness is already read today: `PlanCrossfade` uses it to prevent a live producer from being introduced ahead of a deterministic consumer, and `validatePlan` checks its provenance. The anticipative executor is still its larger future purpose, but calling it unused is no longer accurate.

docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+
+The pool runs a second ahead of the transport (`kCueAheadSeconds`), opening files and pointing
+readers at the sample their clip starts on. That is the difference between a clip starting on
+the beat and a clip starting a block late: a read that does not continue the last one is a seek,
+and a seek costs a block of silence.
+
+Two ceilings, and they answer different questions. `kMaxVoicesPerTrack` is 16, and it is how
+many clips a track can sound in one callback. `kMaxReadersPerTrack` is twice that, because a
+reader has to exist before its clip is due. Past either, the track says so through
+`starvedVoices` and `overSubscribed` rather than quietly dropping the extras, because silence
```
lucaromagnoli: The reader ceiling is not reported through `overSubscribed`. That counter is specifically peak concurrent clips beyond `kMaxVoicesPerTrack`; a crowded one-second reader window is normally harmless, and only candidates dropped inside `kReadAheadBridgeSeconds` increment `unbridged`. `starvedVoices` then counts clips actually missed by the callback. Could this name `unbridged` and keep the two failure modes separate?

docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+    H --> I["8 MixAudio<br/>master input"]
+    I --> J["9 Fader<br/>master"]
+    J --> K["10 Meter"]
+    K --> L["11 Gain<br/>master mute"]
+    L --> M["12 Output"]
+```
+
+Two things that look like clutter and are not:
+
+**One device is three ops.** Its processing, its gain trim and its meter tap are separate, so
```
lucaromagnoli: This is the default shape for an ordinary processing device, but it is not universal: analysis devices compile to a bare process op, and disabling device meters removes the meter op. Could this qualify the three-op statement so readers do not treat it as an invariant?

docs/architecture/native-engine.md
```diff
@@ -0,0 +1,404 @@
...
+handed, and nothing downstream pairs them up.
+
+The pool runs a second ahead of the transport (`kCueAheadSeconds`), opening files and pointing
+readers at the sample their clip starts on. That is the difference between a clip starting on
+the beat and a clip starting a block late: a read that does not continue the last one is a seek,
+and a seek costs a block of silence.
+
+Two ceilings, and they answer different questions. `kMaxVoicesPerTrack` is 16, and it is how
+many clips a track can sound in one callback. `kMaxReadersPerTrack` is twice that, because a
+reader has to exist before its clip is due. Past either, the track says so through
```
lucaromagnoli: This names the wrong counter for one of the two ceilings. Past the voice ceiling the report is overSubscribed and the audible cost is starvedVoices, as written; but past the reader budget the pool reports through unbridged(), the gauge for dropped window clips that cannot be provisioned before they are due, and overSubscribed stays at zero for that case by design, since a lane of sequential slices fills the reader budget many times over without two of them ever sounding together. Naming unbridged here also earns the section its third sentence: the two ceilings really do answer different questions, and they answer through different counters.

lucaromagnoli (commented):
Reviewed as a fact-check, since a map that is wrong is worse than no map. Everything checkable I checked against the branch, and it nearly all holds:

- The golden dump is byte-identical to test_render_plan_compiler.cpp, ops, spacing, ready line and all.
- The op vocabulary matches OpKind exactly, in enum order.
- The ceilings are right: kMaxVoicesPerTrack 16, kMaxReadersPerTrack 2x, kCueAheadSeconds one second, 10 ms rounds.
- Every referenced file and function exists where the doc points, and the bare core/ paths are exactly how engine sources spell those includes, so a reader grepping will land.
- All fifteen test tags exist in the test tree; make debug and make test are real targets; the engine option defaults ON; the Tracktion configure-time guard is where the doc says; magda_engine is linked by magda_tests and nothing else.
- The differ sentence matches the implementation: key join, then kind, outputs and input count, then per-slot port and producer key.
- The issue table sampled correct on both sides: 1888, 1901, 2010, 2019 closed; 1891, 2036 open; and saying slices rather than done for the open 1889 epic is the honest phrasing.

One sentence names the wrong counter, inline. The reading-chain section being flagged as on the slice-3 branch is the right way to handle a doc racing its code, and the staleness rule at the top, headers win, is the best line in the file.

Issue #1854, opened 2026-07-25
Remote API: define transport-neutral operations and DTO schemas


lucaromagnoli:
Part of #701.

## Summary

Define the canonical MAGDA Remote API contract shared by WebSocket, MCP, and future transports. The contract sits above `MagdaApi`; transport adapters must not serialize core model objects or call managers directly.

> **Status:** the bulk of this landed in 0.17.0 (709111df). See the [audit comment](https://github.com/Conceptual-Machines/magda-core/issues/1854#issuecomment-5159295864) for detail. Remaining work is the unchecked boxes below — this issue is **not** a blocker for #1855, #1856, or #1858.

## Scope

- [x] Define stable operation names and API versioning
- [x] Add JSON-safe DTOs for project, tracks, clips, devices, selection, transport, session, racks, and automation
- [x] Add shared input/output JSON Schemas and validation
- [x] Define consistent error codes and envelopes
- [x] Implement one discoverable operation registry
- [ ] Extract reusable serialization from CLI and Lua bindings where appropriate
- [x] Add contract tests using `MockMagdaApi`
- [ ] Document fields intentionally excluded from the remote API
- [ ] Add the operations that depend on #1859 — device add/remove, parameter read/write, automation editing, MIDI note editing

Do not expose complete `TrackInfo`, `ClipInfo`, or `DeviceInfo` objects: they include UI state, transient AI state, plugin internals, caches, and file paths.

## Acceptance

- WebSocket and MCP consume one registry without duplicating schemas or handlers.
- DTOs round-trip through JSON in tests.
- DTOs that name a model object can address exactly one — see the `DeviceId`-per-section correction in #1991.
- Invalid IDs, enums, ranges, NaN/Infinity, and unknown fields return structured validation errors.
- No response exposes raw pointers, filesystem paths, plugin state blobs, or transient UI/AI state.

## Out of scope

Network listeners, message-thread dispatch, subscriptions, and transport-specific framing.

lucaromagnoli:
## Audit: most of this has already shipped

Checking the tree before starting work here — the bulk of this issue landed in 0.17.0 (709111df) and is live in `magda/daw/api/remote_api.{hpp,cpp}` + `remote_api_projection.cpp` (~1800 lines). Recording what's actually done so this stops reading as unstarted work blocking #1855, #1856, and #1858, which are in fact unblocked.

**Landed:**

[...]

PR #2049, opened 2026-08-07
MCP: expose the MAGDA Remote API as tools and resources (#1858)

### LINKED ISSUE #1858: MCP: expose the MAGDA Remote API as tools and resources

lucaromagnoli:
Closes #1858.

Adds an MCP server over the transport-neutral contract from #1854/#1855, so an external AI host can inspect and control the running session using the same operations the WebSocket already uses — same validation, revisions, undo grouping, error taxonomy. No MCP handler reaches a manager.

## Shape

Two units, mirroring the WebSocket adapter's split:

- **`McpEndpoint`** (`remote_mcp.*`) — the protocol, with no sockets and no `httplib.h`. Registry→tools projection, the resource table, error mapping, version negotiation. A later stdio proxy reuses it unchanged.
- **`RemoteMcpServer`** (`remote_mcp_server.*`) — Streamable HTTP on its own loopback port.

Also `remote_http_auth`, extracted so both transports enforce one bearer-token and `Origin` policy, and `devices.catalog` added to the shared registry to back `magda://devices/catalog`.

## MCP shipped a new revision mid-flight

**2026-07-28** landed while this was being built. It removes protocol-level sessions, the `initialize` handshake, the standalone GET stream, and `resources/subscribe`; adds mandatory `server/discover` and mirrored `MCP-Protocol-Version` / `Mcp-Method` / `Mcp-Name` headers.

Hosts have not all moved, so the endpoint is **dual-era** and decides per request:

| Condition | Era |
|---|---|
| `params._meta` names a modern version | Modern, stateless |
| Method is `initialize` | Legacy; session minted |
| Live `Mcp-Session-Id` | Legacy, at that session's version |
| None of the above | `400` + `-32602` — the signal a dual-era client uses to fall back |

Supported: `2026-07-28`, `2025-11-25`, `2025-06-18`. Nothing is pinned — the acceptance criterion about `2024-11-05` is covered by a test.

## `magda-mcp`, the stdio bridge

MAGDA's port is ephemeral and its **token is regenerated every launch**, so a static `url` + `Authorization` pair in a host's config breaks the first time the DAW restarts. The bridge resolves both from the discovery record per request, so the host config is one command and never changes:

```json
{ "mcpServers": { "magda": { "command": "/Applications/MAGDA.app/Contents/MacOS/magda-mcp" } } }
```

It re-resolves on a dead port or rotated token, and re-handshakes when a restart invalidates a legacy session — the host sent `initialize` once at startup and cannot know. It links `juce_core` + cpp-httplib only (~1.6 MB), deliberately not `magda_daw`. It also reaches hosts that speak only stdio, which #1858 listed as a follow-on.

## Decisions worth reviewing

- **Array results are wrapped** as `{"items": [...]}` on the tool surface. MCP types `outputSchema` *and* `structuredContent` as objects and clients enforce both, so `tracks.list` and the three other list reads cannot be expressed directly. Schema and value are wrapped together. Resources are untouched and still read as plain arrays.
- **Idempotency is opt-in**, via a client-supplied key in `_meta`. MCP has no session and every client behind the one bearer token looks identical, so deriving a key from the JSON-RPC id would let two clients that both counted from 1 receive each other's cached writes.
- **`meters` and `playhead` are not resources.** An MCP update carries only a URI, which is useless for a meter, and 20 Hz of them is abusive. They stay on the WebSocket, which can carry a payload.
- **A second listener**, not routes on the WebSocket's: an SSE stream holds a connection thread for its lifetime and needs a separate budget, and a bug in one transport shouldn't take the other's listener down.

## UI and packaging

AI Settings gains a **Remote** tab — a toggle that opens/closes the listener immediately rather than at next launch, live status, and the host config composed from this install's own paths.

The bridge is staged beside the executable on all three platforms, signed in its own right on macOS and Windows (a host launches it directly, so Gatekeeper and SmartScreen judge it alone), and published as a standalone Linux asset because an AppImage cannot expose a stable path to anything inside it.

## Testing

Full suite green: **2226 cases, 557,830 assertions**. New coverage:

- `test_remote_mcp.cpp` — protocol layer, no sockets
- `test_remote_mcp_server.cpp` — real loopback HTTP, both eras, SSE, limits, shutdown
- `test_remote_mcp_bridge.cpp` — forks the real binary over real pipes, including MAGDA restarting underneath it

Mechanized acceptance criteria: `tools/call`, `resources/read`, and direct dispatch return byte-identical structured data; one mutation is one undo compound.

Driven end to end from Claude Code against a running MAGDA — 38 tools, resources, reads and writes, undo, session recovery across a restart.

## Notes for the reviewer

Four defects only surfaced against a real host, three from the spec's prose disagreeing with what clients enforce (`outputSchema` and `structuredContent` are both documented as permitting arrays; neither does). Tests now pin each.

The `~13` pre-existing mojibake sites elsewhere in the UI — `juce::String(const char*)` decodes as ASCII, so bare non-ASCII literals render as Latin-1 — are **not** in this PR. Same one-line `fromUTF8` fix each; happy to do them separately.


lucaromagnoli (commented):
I found six lifecycle, packaging, and stdio bridge issues that can leave the MCP integration listening but unusable, produce stale host configuration, or leave requests unanswered.

magda/daw/ui/dialogs/AISettingsDialog.cpp
```diff
@@ -2451,6 +2460,260 @@ class AISettingsDialog::ModelDownloadsPage : public juce::Component {
...
+        auto& config = magda::Config::getInstance();
+        config.setRemoteApiEnabled(enabled);
+        config.save();
+
+        // Absent in the headless CLI and in tests. The config is still written,
+        // so the next launch honours it — there is simply no live listener here
+        // to reconcile.
+        if (auto* host = magda::remote::activeHost()) {
+            if (enabled)
+                host->start();
```
lucaromagnoli, line 2597: **[P1] Recreate the service before re-enabling.** `RemoteApiHost::stop()` permanently calls `shutdown()` on both `RemoteApiService` and `SubscriptionHub`, but this path later calls `start()` on the same host. The listeners bind again around those shut-down objects, so operations return `Cancelled` and subscriptions cannot register. Please either make this a listener-only pause or recreate the service/hub before restarting.

magda/daw/ui/dialogs/AISettingsDialog.cpp
```diff
@@ -2451,6 +2460,260 @@ class AISettingsDialog::ModelDownloadsPage : public juce::Component {
...
+#else
+        // Resolve the real binary rather than argv[0]: inside an AppImage the
+        // type-2 runtime leaves argv[0] pointing at the outer .AppImage, so the
+        // sibling lookup would land in the user's download folder.
+        char buffer[PATH_MAX];
+        if (const auto length = ::readlink("/proc/self/exe", buffer, sizeof(buffer) - 1);
+            length > 0) {
+            buffer[length] = '\0';
+            const juce::File selfExe(juce::String::fromUTF8(buffer));
+            if (const auto sibling = selfExe.getParentDirectory().getChildFile("magda-mcp");
```
lucaromagnoli, line 2655: **[P1] Don’t publish an AppImage mount path.** In the shipped AppImage this sibling exists under `/tmp/.mount_…/usr/bin`, so the generated MCP configuration captures a transient mount point that changes on every launch. Once the host or bridge restarts, its configured command no longer exists. Linux needs to point at the separately installed bridge asset (or prompt the user to install/select a stable path).

magda/mcp_bridge/main.cpp
```diff
@@ -0,0 +1,621 @@
...
+        // to be answered with SSE, so the non-streaming path below still handles
+        // an SSE body it did not expect — it just gets to do it after the fact.
+        if (method == "subscriptions/listen") {
+            // A stream never completes on its own, so no read timeout may apply
+            // to it. Cancellation and shutdown both stop the client instead,
+            // which is prompt regardless.
+            client->set_read_timeout(0, 0);
+            registerStream(id, client);
+
+            auto result = client->Post(endpoint.path, headers, body, "application/json",
```
lucaromagnoli: **[P1] Forward non-SSE responses from streaming requests.** A content receiver receives the body for every HTTP status. If `subscriptions/listen` is rejected with a `400` or `503` JSON-RPC body, this callback feeds it to `consumeSse()`, `result->body` stays empty, and the `status != 401` return below reports the request as answered without writing anything to stdout. The host then waits forever for this request ID. Please inspect the response status/content type before selecting the SSE parser, or buffer and emit non-SSE bodies.

magda/mcp_bridge/main.cpp
```diff
@@ -0,0 +1,621 @@
...
+
+    static bool isEventStream(const httplib::Response& response) {
+        return response.get_header_value("Content-Type").rfind("text/event-stream", 0) == 0;
+    }
+
+    /// Turn an SSE byte stream back into stdio messages: one `data:` payload per
+    /// line, comments discarded.
+    void consumeSse(const std::string& chunk) {
+        std::lock_guard<std::mutex> lock(sseMutex_);
+        sseBuffer_ += chunk;
```
lucaromagnoli: **[P2] Keep SSE parser state per stream.** Every concurrent `subscriptions/listen` request calls this with chunks from a different HTTP response, but they all append to the same `sseBuffer_`. If one callback receives a partial frame and another stream runs before its remainder, the two responses are concatenated and the acknowledgment/notification is corrupted or discarded. The buffer needs to belong to the individual request/stream.

magda/mcp_bridge/main.cpp
```diff
@@ -0,0 +1,621 @@
...
+        // forwards. On stdio it is how a host closes a `subscriptions/listen`
+        // stream; on HTTP the equivalent is closing the response, which is
+        // exactly what stopping that request's client does.
+        if (method == "notifications/cancelled") {
+            cancel(message["params"]["requestId"]);
+            return;
+        }
+
+        std::lock_guard<std::mutex> lock(threadsMutex_);
+        threads_.emplace_back([this, message] { forward(message); });
```
lucaromagnoli: **[P2] Reap completed request threads.** Every inbound message appends a new joinable thread to `threads_`, and completed threads are not joined or removed until stdin closes. A long-running MCP host therefore accumulates thread handles/native resources without bound. Please use a bounded worker pool or continuously reap completed workers.

magda/daw/api/remote_mcp_server.cpp
```diff
@@ -0,0 +1,1226 @@
...
+                juce::String::fromUTF8(request.body.data(), static_cast<int>(request.body.size())),
+                parsed)
+                .failed() ||
+            parsed.getDynamicObject() == nullptr) {
+            writeJson(response, httplib::StatusCode::BadRequest_400,
+                      jsonRpcError({}, McpError{MCP_PARSE_ERROR, "Malformed JSON"}));
+            return;
+        }
+
+        const auto id = parsed["id"];
```
lucaromagnoli, line 948: **[P2] Validate the JSON-RPC envelope before routing.** This extracts the ID and method without checking that `jsonrpc` is the string `"2.0"`. Requests with a missing or incorrect version can therefore execute tools and mutations and receive a 2.0 response. MCP is JSON-RPC 2.0; malformed envelopes should be rejected as invalid requests before era resolution or dispatch.

lucaromagnoli:
All six addressed in 835920f1. Every one was real — thanks, the two P1 lifecycle issues in particular were paths my own tests never walked.

**[P1] Recreate the service before re-enabling.** Confirmed: `shutdown_` is a one-way `exchange(true)` and `retireState()` never restores, so the toggle produced listeners bound around a dead dispatcher — answering `Cancelled` while looking perfectly healthy. Took the listener-only route: `stop()` is now the destruction path alone, and a new `stopListening()` closes the sockets and withdraws the token while leaving the service, bridge, and hub intact. New test toggles off and on and then dispatches an operation through the restarted host.

**[P1] Don't publish an AppImage mount path.** Right, and embarrassing — the doc I wrote in this same PR already says Linux ships the bridge as a separate asset, and then the page emitted the in-AppImage path anyway. It now detects the AppImage runtime (`APPIMAGE`/`APPDIR`), prints a placeholder with instructions, and disables Copy rather than handing over a command that stops existing when MAGDA quits.

**[P1] Forward non-SSE responses from streaming requests.** Confirmed exactly as described. Now checks the content type after the response completes, and emits the raw body when it wasn't an event stream — the accumulated bytes come from the new per-request parser below, since `Result::body` is empty once a content receiver has run. Falls back to an explicit JSON-RPC error if the body was empty too, so the id is always answered.

**[P2] Keep SSE parser state per stream.** Extracted an `SseParser` owned by the request rather than the bridge. It also holds the raw bytes, which is what makes the fix above possible.

**[P2] Reap completed request threads.** Each worker now carries a completion flag and finished ones are joined and dropped before each new dispatch.

**[P2] Validate the JSON-RPC envelope before routing.** Added, matching what the WebSocket transport has always done. New test asserts a mutating call with no `jsonrpc` member is refused *and* that the project is untouched — i.e. rejected before dispatch, not after.

Full suite green: 2228 cases, 557,835 assertions.

magda/daw/api/remote_api_host.cpp
```diff
@@ -217,25 +262,45 @@ bool RemoteApiHost::start() {
...
+void RemoteApiHost::stopListening() {
+    // Before the hub, like the WebSocket server: an open notification stream is
+    // a registered hub client, and the hub must not be shut down underneath one.
+    if (mcpServer_ != nullptr) {
+        mcpServer_->stop();
+        mcpServer_.reset();
+    }
     if (server_ != nullptr) {
         server_->stop();
         server_.reset();
```
lucaromagnoli, line 282: **[P1] Avoid reusing cached WebSocket request identities after restart**

`stopListening()` now preserves `RemoteApiService` and its idempotency cache, but recreating `RemoteWebSocketServer` resets `nextConnectionId` to 1. The first connection after re-enabling therefore becomes `ws:1` again, and a request that reuses (for example) JSON-RPC id `1` gets the same cache key as the old connection. A write can then return the stale cached response without executing. Please include a listener-generation/nonce in the WebSocket client identity, keep connection IDs unique across restarts, or otherwise invalidate the transport-scoped cache when listeners are replaced.

magda/mcp_bridge/main.cpp
```diff
@@ -0,0 +1,697 @@
...
+ * request, which is an ordinary JSON-RPC error rather than an event stream —
+ * and by then `Result::body` is empty, so this is the only copy left.
+ */
+class SseParser {
+  public:
+    using Emit = std::function<void(const std::string&)>;
+
+    void consume(const std::string& chunk, const Emit& emit) {
+        std::lock_guard<std::mutex> lock(mutex_);
+        raw_ += chunk;
```
lucaromagnoli: **[P1] Do not retain the full body of a long-lived SSE stream**

Every received chunk is appended to `raw_`, but a successful SSE notification stream can remain open indefinitely. Although parsed frames are removed from `pending_`, `raw_` never shrinks, so normal subscriptions consume memory for their entire lifetime and can eventually exhaust the bridge process. Please retain the raw body only while the response might be a non-SSE error (or cap/clear it once the response is known to be SSE); a valid stream should keep only its incomplete frame.

magda/daw/api/remote_mcp_server.cpp
```diff
@@ -0,0 +1,1238 @@
...
+
+        // MCP is JSON-RPC 2.0, and the envelope is checked before anything is
+        // routed. Without this a request carrying no `jsonrpc` at all — or a
+        // different version — still reaches a tool and mutates the project,
+        // then receives a 2.0 reply it never asked for. The WebSocket transport
+        // has always rejected these; this one had not.
+        if (parsed["jsonrpc"].toString() != "2.0") {
```
lucaromagnoli: **[P2] Require `jsonrpc` to be a string**

Comparing `toString()` alone also accepts a numeric JSON value such as `"jsonrpc": 2.0`, because JUCE serializes that number as `"2.0"`. JSON-RPC requires this member to be the string `"2.0"`. Please check `isString()` before comparing the value and add a numeric-version regression test.

lucaromagnoli (commented):
Three remaining correctness and resource-lifetime issues:

lucaromagnoli:
All three fixed in fbd22817. Two were consequences of my own earlier fixes, which is worth saying plainly.

**[P1] Cached request identities after restart.** Correct, and I introduced it: before `stopListening()` existed, `stop()` shut the service down and the cache died with it. Keeping the service alive kept keys whose scope had expired — a WebSocket key is `ws:<connection id>`, and a fresh server counts from 1 again. Added `RemoteApiService::clearIdempotencyCache()` and call it from `stopListening()`, the only place that replaces a listener without replacing the service. The new test drives a keyed write, confirms the replay behaviour that makes the collision dangerous, restarts, and asserts the same key now executes.

**[P1] Retaining a long-lived stream's body.** Also mine — the raw buffer only existed because of the previous round's "forward non-SSE responses" fix, and I gave it no upper bound. The first parsed frame proves the response is an event stream, so the raw copy is released there rather than merely capped; before that it is capped, since an error body is small and anything larger is not one.

**[P2] `jsonrpc` must be a string.** Right — JUCE renders the number `2.0` as the text `"2.0"`, so the rendering comparison accepted it. Now checks `isString()` first, with a numeric-version regression test. **Also fixed in the WebSocket transport**, which had the identical check; not reported, but the same mistake.

One thing worth flagging about how I tested the second fix. My first attempt asserted on the bridge process's resident set size after driving traffic through a live stream — and it **passed with the fix reverted**. Forty flushes of keep-alives don't move RSS past page granularity and allocator reuse, so it proved nothing. I've extracted `SseParser` into its own header and test the retained bytes directly instead; that version does fail without the fix (two assertions), which I verified by reverting it.

Full suite: 2252 cases, 558,920 assertions.

lucaromagnoli:
Both fixed in 96363b61. You were right on both, and the first one shows my previous fix was aimed at the wrong thing entirely.

**[P1] Cache clearing.** Clearing was wrong twice over, as you say — racy against a write already queued on the message thread, and it discarded MCP keys that are *supposed* to survive a restart, which is precisely what makes a retry safe after the endpoint went away mid-request. So I took the generation route: a WebSocket client id is now `ws:<listener generation>:<connection id>`, unique by construction. No clearing, no race, MCP keys untouched. `clearIdempotencyCache()` is deleted rather than left unused — it was the wrong mechanism, and leaving it would invite the next person to call it.

Two tests now: a real socket round-trip across a restart reusing JSON-RPC id 1 on connection id 1 (fails without the generation — tempo stays at the cached value), and an MCP-style UUID key that still dedupes *across* a restart, which the clearing approach would have broken.

**[P2] `pending_` unbounded.** Correct, and the sharper point is that my accessor measured the buffer I had fixed rather than the memory the object was using — which is why the junk test passed while holding four megabytes. `retainedBytes()` now reports both buffers, and `pending_` is capped at 1 MiB with a resync: past the cap no frame can complete, so the partial is dropped and framing skips to the next separator, discarding the truncated tail rather than emitting it. Added a test for that resync path.

Both verified by reverting each fix and confirming the tests fail — three assertions across the two.

The restart test also stopped hand-building a `RequestContext`. The identity under test is the one the transport constructs, and a hand-written one just asserts whatever shape the test picked — which is exactly what changed here.

Full suite: 2254 cases, 558,935 assertions.

lucaromagnoli:
Fixed in 1dd771bc. Correct — clearing the whole buffer threw away separators that arrived in the same chunk, and the flag then ate the next good frame too. Two messages lost for one bad one, silently, on the stream a host relies on for state changes.

Resynchronisation now discards *through* the first buffered separator and keeps everything after it, which the loop parses normally. The flag is only set when no separator has arrived yet — the case it was actually meant for, where the eventual terminating segment is the tail of something already truncated and is rightly dropped.

New test reproduces your case directly: an oversized frame whose terminator and a following valid frame share the chunk that crosses the limit, plus another frame in the next chunk. Reverting to the wholesale clear makes it fail with zero frames emitted where two were expected, exactly as you described.

Full suite: 2255 cases, 558,945 assertions.

Worth noting this is the third round where the finding was in a fix from the round before rather than in the MCP work itself — the protocol layer has been stable since the first review, and every regression since has been in the buffering and lifetime plumbing around it. Appreciate you running these down.

lucaromagnoli:
Fixed in 98712750, and the ordering was the actual defect — thanks for pinning it that precisely.

Completeness is a property of a frame, not of what arrived behind it. Frames terminated by a separator are now parsed first and judged on their own size, so an oversized one is dropped individually and costs nothing either side of it; only the incomplete remainder is capped afterwards. That also deletes the find-the-separator special case my previous fix needed, since the loop has consumed every separator by the time the cap runs.

Two tests, both reproducing your numbers when the ordering is put back: a good frame ahead of an oversized partial (emitted 0, buffer still over the cap), and an oversized frame between two good ones.

Full suite: 2257 cases, 558,939 assertions.

For what it's worth, the last four findings have all been in this one function, each in the fix for the one before — non-SSE bodies needed a raw buffer, the raw cap left `pending_` unbounded, bounding it broke resync, and the resync fix put the cap on the wrong side of the parse. The current shape is simpler than any of the intermediate ones, which is usually the sign it is the right one; I should have arrived at "parse what is complete, then bound what is not" several rounds ago instead of patching each symptom where it surfaced.

PR #2050, opened 2026-08-07
Docs: give the docs directory a shape

lucaromagnoli:
Thirty-odd files in one flat directory, in three naming conventions, where a January bug report sat beside the testing guide with nothing to say which of them was still true. Nobody reads a directory like that. They grep it, and grep does not tell you whether what it found is current.

## Six directories, and a rule for which is which

Written down in a new `docs/README.md`, because a structure nobody can state the rule for decays back into a flat list within a month.

| Directory | What belongs in it |
| --- | --- |
| `architecture/` | how a subsystem works and why it is shaped that way |
| `specs/` | a feature or a piece of work described before it exists |
| `development/` | how to work on the codebase |
| `ci/` | the repository itself: builds, runners, protections, templates |
| `product/` | what MAGDA is for, and how it is shown |
| `archive/` | finished. Kept because it explains how something got this way |

`issues/` and `smoke-tests/` are working notes on particular issues and manual passes, and are untouched.

## Archive is the part that earns the rest

Seven documents were true when they were written and describe work long since done: two workflow setup summaries, a February bug analysis, the local LLM proof of concept, the Faust free-tier POC notes, an earlier manifesto draft, and the January test plan that the testing guide replaced.

Filed rather than deleted. Each explains how something got the way it is, and the point of the directory is that nobody has to open one to find out it is history.

## Two names that lied

- `AUTOMATION_ARCHITECTURE.md` is an analysis of Tracktion parameter writes and had nothing to do with automation. It is now `architecture/tracktion-parameter-writes.md`.
- `FAUST_POOL_REFACTOR.md` describes the parameter pool that shipped, not a refactor still pending. It is now `architecture/faust-param-pool.md`.

Everything else is kebab-case, which is what the newer half of the tree already used.

## References followed the files

- both READMEs, which linked two of the moved CI docs
- four Faust headers that cite the pool spec in their comments
- `.gitleaksignore`, which pins a false positive to a path **and a line number**
- three cross-links inside docs

Every link in the tree resolves. Checked with a script rather than by eye, including the new index.

## Two things worth knowing before merging

**`README_CN.md` comes out fully rewritten, and that is line endings rather than content.** It was still CRLF, so the repository's own `mixed-line-ending` hook normalised the whole file the moment a link edit staged it. The alternative was leaving the Chinese README pointing at two files that have moved.

**Merge #2048 first if you can.** The index links `architecture/native-engine.md`, which arrives with that PR. Merged the other way round, the index carries one dead link until #2048 lands. Nothing else in either PR overlaps.


lucaromagnoli:
Folded into #2048 so the docs work is one branch and one review. That PR now carries the map and the reorganisation together, which also removes the merge-order question: the index links native-engine.md and both live on the same branch.

Issue #2040, opened 2026-08-06
Arranger clips slice 7: null-diff corpus for clip playback


lucaromagnoli:
Slice 7 of #1890. Needs the harness from #1896.

Parity is claimed by rendering, not by reading the code: a corpus of projects rendered through both engines to a near null difference.

Coverage: plain placement and trims, fades and every curve, clip loops and tiling, files at a rate other than the device's, reverse, speed ratio, auto tempo across a tempo change, both stretch engines at both qualities, warped clips, comped and multi-take clips, and dense MIDI with CC, pitch bend and MPE.

Overlaps stay out of it on purpose (#2003). Tracktion has no correct incumbent behaviour to diff against, and giving it one is the double work that decision avoided. They are covered instead by the model suites, which pin every rule, and by reference-executor assertions: a fully covered region renders silent, a crossfade sums to unity, a play-through overlap sums both sources.

Issue #1891, opened 2026-07-27
Engine: parameters, modifiers, macros, automation playback


lucaromagnoli:
Part of #1882. Depends on the engine core.

The parameter and modulation system, designed around the lessons encoded in the TE fork patches.

- Two write lanes (host values, modulation) resolved once per block into a flat value buffer; devices read resolved values only. The host-write-dropped-under-modifier bug class becomes structurally impossible.
- Modifier engines: LFO (custom waveforms, gating, retrigger, skip-resync), ADSR, envelope follower with external input, random. Per-link depth and bipolarity stay on the MAGDA side as they do today.
- Macros at track, rack, and device scope.
- Automation playback as a bake target for the MAGDA-owned curve model.
- The dB-to-fader-position curve must match TE's bit-exactly; it is baked into persisted automation value domains.

lucaromagnoli:
Please make the parameter precedence and within-block representation explicit. “Host values + modulation” leaves automation ambiguous, while “resolved once per block” can conflict with sample-accurate ramps.

A useful contract would distinguish:
- stored/base value, including UI or control-surface writes;
- automation events/segments;
- modulation contribution;
- final resolved value stream consumed by the device.

The issue should define:
- precedence when a host write occurs under active automation;
- write/touch/latch behaviour, even if only read mode ships initially;
- whether modulation is additive, multiplicative, or parameter-defined;
- clamping/normalisation point;
- sample-offset ordering for simultaneous MIDI and parameter events;
- representation of ramps inside a block.

“Resolved once per block” can still work if the flat buffer contains constant/ramp segments or per-sample values where required. Devices should never need to rediscover automation/modulation precedence themselves.

lucaromagnoli:
Scope note from breaking down #1896: this issue carries its own final parity slice, the way
#2040 was slice 7 of #1890.

Cases for what is ported here are written here, not in #1896. A parity case is worth what its
author's knowledge of the feature is worth, and that knowledge is at its peak on the day the
port lands. Written later by someone reading the code back, a case asserts the code rather than
the behaviour, which is the one thing a parity corpus must not do.

#1896 keeps the cross-cutting infrastructure the slice builds on: the whole-project case and the
tiered oracle (#2075), the goldens (#2076), the block-size gate (#2078), and the corpora that
span features (#2081, #2082).

lucaromagnoli:
Broken into eight slices. The order below is the dependency order: everything after the first
writes into the lane the first defines.

[...]

PR #2052, opened 2026-08-07
Native engine: time stretch and pitch (#2037)

### LINKED ISSUE #2037: Arranger clips slice 4: time stretch and pitch

lucaromagnoli:
Slice 4 of #1890, closes #2037.

Playing a clip at a speed and a pitch that are not its file's: speed ratios, auto tempo, pitch, analog pitch and speed ramp fades.

## One function

All of it is where in the reading a moment of the timeline sits, so all of it is `readingPositionAt` in `clip/EventPlacement.hpp` answering differently.

| What the clip asks for | What the position does |
| --- | --- |
| its file's own speed | one reading sample per output sample, as in slice 3 |
| a speed ratio | advances by the ratio, resolved in the snapshot |
| auto tempo | advances by the beats that have passed, times a beat of the file |
| analog pitch | advances by the ratio the model already folded the pitch into, no stretcher |
| a speed ramp fade | the moment itself is warped near the clip's edge, and the rate with it |
| warp (#2038) | another answer from the same place |

Auto tempo is the one to read twice. Its ratio is the project's tempo over the file's own and moves with the tempo curve, so it cannot be resolved to seconds ahead of a block. What saves it is that the integral of that ratio is beats: the material an instant has consumed is how many beats have passed since the event began, times what a beat of the file is worth. That is why the clock publishes both faces of one instant, and it is what keeps a second tempo map off the audio thread.

A block then reads exactly `round(P(end)) - round(P(start))` samples and hands them to a stretcher to come back as the block's own length. The ratio a block runs at is whatever its own two ends say, so a tempo curve and a speed ramp cost nothing extra, and nothing accumulates: both ends are rounded rather than counted forward, so one block's reading ends where the next one's begins.

## Where a stretcher lives, and what that answers

One per provisioned event, built and configured by `ClipVoicePool` on the thread that opens the file, carried to the callback in the same table as the stream. Not in the reading chain, which is random access and pure; not in a voice, which is claimed and released per block and would be allocating one on the audio thread.

- **A plan swap** does nothing to it: it never enters a plan epoch.
- **A loop wrap** does nothing to it: tiling happens below the stream, so a wrap is a discontinuity in the material rather than a change of position.
- **A locate** resets and re-primes something that already exists.
- **An underrun** deliberately does not re-prime. Priming reads behind the position the block wants, so a stream that came back short and was primed again would be sent backwards every block, and every one of those is a seek that guarantees the next block is short too.

**Latency is answered here rather than reported upwards.** Every engine wants material from before the first sample to be heard and says how much; the pool cues the stream that far back, so a voice's first read is one contiguous read that begins with the priming samples. A `ClipAudio` op still reports no latency and stretched voices stay aligned with unstretched ones on the same track.

## Vendoring

`third_party/signalsmith-stretch` (MIT, header only) and `third_party/soundtouch` (LGPL-2.1), with their licences and a `SOURCES.md` each, out of the Tracktion fork that the engine may not include.

SoundTouch is its own unmodified static target rather than files folded into `magda_engine`, which is what keeps the relink option open. MAGDA ships GPL-3.0 and LGPL-2.1 is compatible with it, so nothing more is owed today. It is carried at all because `kSoundTouchNormal` and `kSoundTouchBetter` are pinned project-file integers and sessions saved with them have to play through the algorithm they were made with.

One thing the plan did not anticipate: the fork compiles its own copy of those sources into its unity build, so `soundtouch_ac_test` was defined twice and the final link refused it. Ours is renamed through a compile definition rather than by touching the vendored source, and that goes away with the fork.

## One deliberate divergence

The fork's Signalsmith wrapper primes with the material *at* the start rather than before it, so it begins every stretched clip roughly `inputLatency + rate * outputLatency` late, about 120 ms at the default preset. This primes from the material before the start, which is what the library's `outputSeek` is for. The null-diff corpus (#2040) will show that as a fixed offset on stretched clips.

## Also here

- Auto pitch plays its own transpose alone and the snapshot says so in a diagnostic: the other half of it is an offset from the project's pitch sequence, and there is no pitch track in the engine yet (#1891).
- The clip snapshot carries its two fade lengths on the beat face as well, because a speed ramp on an auto tempo clip has to be measured in the domain its position is derived from.
- The cubic curve the rate converter uses is now shared, so a file at another rate and a clip playing fast are not two different sounds.

## Testing

`make debug` clean, and the full Catch2 suite green (2198 passed, 13 skipped for missing fixtures).

29 new cases, tagged `[stretch]`. The position map is tested as arithmetic, exactly. The resampling path is tested sample for sample through a counting reader, which works because cubic interpolation of a straight line is exact. The stretch engines are tested the way a listener would judge them: the block is full, the level survived, and a stretched clip at double speed keeps its 1 kHz where a resampled one goes to 2 kHz.

Everything rolls, as the other clip rigs do, and rolls in from before the clip starts, which is where the pool would have cued it.

### DIFF
--- a/magda/engine/clip/ClipStretcher.cpp
+++ b/magda/engine/clip/ClipStretcher.cpp
@@ -1,0 +1,477 @@
+#include "clip/ClipStretcher.hpp"
+
+#include <SoundTouch.h>
+#include <signalsmith-stretch.h>
+
+#include <algorithm>
+#include <cmath>
+#include <vector>
+
+#include "core/TimeStretchModes.hpp"
+#include "io/SourceReaders.hpp"
+
+namespace magda::engine {
+
+namespace {
+
+/// Channel pointers, gathered without allocating. Every implementation needs
+/// them, both libraries want an array of them, and a block cannot be handed over
+/// as one.
+class ChannelPointers {
+  public:
+    explicit ChannelPointers(int numChannels)
+        : reading_(static_cast<std::size_t>(numChannels)),
+          writing_(static_cast<std::size_t>(numChannels)) {}
+
+    const float* const* gather(juce::dsp::AudioBlock<const float> block, int offset = 0) {
+        for (std::size_t channel = 0; channel < reading_.size(); ++channel)
+            reading_[channel] = channel < block.getNumChannels()
+                                    ? block.getChannelPointer(channel) + offset
+                                    : nullptr;
+        return reading_.data();
+    }
+
+    float* const* gather(juce::dsp::AudioBlock<float> block) {
+        for (std::size_t channel = 0; channel < writing_.size(); ++channel)
+            writing_[channel] =
+                channel < block.getNumChannels() ? block.getChannelPointer(channel) : nullptr;
+        return writing_.data();
+    }
+
+  private:
+    std::vector<const float*> reading_;
+    std::vector<float*> writing_;
+};
+
+int samplesOf(juce::dsp::AudioBlock<const float> block) {
+    return static_cast<int>(block.getNumSamples());
+}
+
+/// A pre-roll buffer big enough for the rate an event usually runs at, with room
+/// for it to wander. Auto tempo's rate moves with the tempo curve, so the rate at
+/// the moment of a locate is not the rate this was sized for; a quarter again
+/// covers an ordinary tempo ramp, and beyond that priming is short rather than
+/// wrong.
+constexpr double kPreRollHeadroom = 1.25;
+
+/**
+ * @brief Signalsmith Stretch, the default engine.
+ *
+ * A phase vocoder, so the ratio is whatever a call is handed: input samples in,
+ * output samples out, and the two lengths are the rate. That is what makes a
+ * tempo curve free here, where an engine with a set tempo has to be told.
+ *
+ * Configured the way the incumbent configures it, down to the formant base and
+ * the pitch-compensated neutral formant shift, because a project that sounded a
+ * particular way through the fork has to go on sounding that way.
+ *
+ * Priming is `outputSeek`, which is the library's own answer to starting in the
+ * middle of a file: hand it the material leading up to the first sample wanted
+ * and it pre-computes the output that would have led there, so the next call
+ * begins aligned rather than fading in over a window. The incumbent primes with
+ * the material *after* the start instead, and so begins every stretched clip
+ * about a preset window late; that is the one place this deliberately differs.
+ */
+class SignalsmithClipStretcher final : public ClipStretcher {
+  public:
+    explicit SignalsmithClipStretcher(const StretchSetup& setup)
+        : channels_(setup.numChannels), pointers_(setup.numChannels) {
+        // splitComputation: the FFT work is spread across calls rather than
+        // landing in whichever block crosses a window boundary. This runs on the
+        // audio thread, so the flat cost is the one that matters.
+        stretch_.presetDefault(channels_, static_cast<float>(setup.sampleRate), true);
+        stretch_.setFormantBase(static_cast<float>(200.0 / setup.sampleRate));
+        stretch_.setTransposeSemitones(setup.semitones);
+        stretch_.setFormantFactor(1.0f, true);
+
+        allocatePreRoll(channels_, static_cast<int>(std::ceil(preRollSamples(setup.nominalRate) *
+                                                              kPreRollHeadroom)));
+    }
+
+    int preRollSamples(double rate) const override {
+        return static_cast<int>(std::ceil(stretch_.outputSeekLength(static_cast<float>(rate))));
+    }
+
+    void reset() override {
+        stretch_.reset();
+    }
+
+    void prime(PrefetchStream& stream, std::int64_t until, int samples, double) override {
+        const auto before = readPreRoll(stream, until, samples);
+        const auto count = samplesOf(before);
+
+        if (count <= 0) {
+            stretch_.reset();
+            return;
+        }
+
+        // outputSeek resets on its way in, and infers the rate from how much it
+        // was given, which is why preRollSamples above is the length to give it.
+        stretch_.outputSeek(pointers_.gather(before), count);
+    }
+
+    void process(juce::dsp::AudioBlock<const float> input, double, double,
+                 juce::dsp::AudioBlock<float> output) override {
+        const auto in = samplesOf(input);
+        const auto out = static_cast<int>(output.getNumSamples());
+        if (out <= 0)
+            return;
+
+        if (in <= 0) {
+            output.clear();
+            return;
+        }
+
+        const auto* const* sources = pointers_.gather(input);
+        auto* const* destinations = pointers_.gather(output);
+        stretch_.process(sources, in, destinations, out);
+    }
+
+  private:
+    int channels_ = 2;
+    ChannelPointers pointers_;
+    signalsmith::stretch::SignalsmithStretch<float> stretch_;
+};
+
+/**
+ * @brief SoundTouch, for the two pinned modes that name it.
+ *
+ * Carried because kSoundTouchNormal and kSoundTouchBetter are project-file
+ * integers: sessions saved with either have to go on playing through the
+ * algorithm they were made with, whatever else is available now.
+ *
+ * A pipe rather than a function, which is the awkward part. It takes input and
+ * gives back whatever it has finished, so a block cannot demand an exact length
+ * the way the phase vocoder can. Two things follow. Its tempo is set per block
+ * from the two lengths, so a moving rate still works. And what it holds back is
+ * absorbed by priming with more material than alignment alone needs: enough that
+ * once the pre-roll's own output has been discarded there is a block's worth
+ * already waiting, so the first block a listener hears is full rather than the
+ * fourth.
+ */
+class SoundTouchClipStretcher final : public ClipStretcher {
+  public:
+    explicit SoundTouchClipStretcher(const StretchSetup& setup)
+        : channels_(setup.numChannels), maxBlockSamples_(setup.maxBlockSamples) {
+        touch_.setChannels(static_cast<unsigned int>(channels_));
+        touch_.setSampleRate(static_cast<unsigned int>(setup.sampleRate));
+
+        if (setup.mode == time_stretch_mode::kSoundTouchBetter) {
+            touch_.setSetting(SETTING_USE_AA_FILTER, 1);
+            touch_.setSetting(SETTING_AA_FILTER_LENGTH, 64);
+            touch_.setSetting(SETTING_USE_QUICKSEEK, 0);
+            touch_.setSetting(SETTING_SEQUENCE_MS, 60);
+            touch_.setSetting(SETTING_SEEKWINDOW_MS, 25);
+        }
+
+        touch_.setPitchSemiTones(setup.semitones);
+        touch_.setTempo(std::clamp(setup.nominalRate, kMinStretchRate, kMaxStretchRate));
+
+        // Sized for the worst block this can be asked for: the largest output
+        // block, times the fastest the reading may be consumed for it. Grown
+        // here, on the thread that made this, and never again.
+        const auto frames =
+            static_cast<int>(std::ceil(setup.maxBlockSamples * kMaxStretchRate)) + 1;
+        interleaved_.resize(static_cast<std::size_t>(frames * channels_));
+        deinterleaved_.resize(static_cast<std::size_t>(setup.maxBlockSamples * channels_));
+
+        allocatePreRoll(channels_, static_cast<int>(std::ceil(preRollSamples(setup.nominalRate) *
+                                                              kPreRollHeadroom)));
+    }
+
+    int preRollSamples(double rate) const override {
+        // What the pipe holds before it will answer at all, plus a block's worth
+        // of surplus so that the block after priming is already there. Both are
+        // input samples, so the surplus is counted at the rate it will be
+        // consumed at.
+        const auto latency = touch_.getSetting(SETTING_INITIAL_LATENCY);
+        const auto batch = touch_.getSetting(SETTING_NOMINAL_OUTPUT_SEQUENCE);
+        const auto cushion = std::max(maxBlockSamples_, std::max(batch, 0));
+
+        return latency + static_cast<int>(std::ceil(cushion * rate));
+    }
+
+    void reset() override {
+        touch_.clear();
+        discard_ = 0;
+    }
+
+    void prime(PrefetchStream& stream, std::int64_t until, int samples, double rate) override {
+        touch_.clear();
+        discard_ = 0;
+        touch_.setTempo(std::clamp(rate, kMinStretchRate, kMaxStretchRate));
+
+        const auto before = readPreRoll(stream, until, samples);
+        const auto count = samplesOf(before);
+        if (count <= 0)
+            return;
+
+        // In pieces, because the interleaving buffer is sized for a block rather
+        // than for a whole pre-roll, and because putSamples does not care how
+        // many calls the material arrives in.
+        const auto stride =
+            static_cast<int>(interleaved_.size() / static_cast<std::size_t>(channels_));
+
+        for (auto done = 0; done < count;) {
+            const auto run = std::min(stride, count - done);
+            write(before, done, run);
+            done += run;
+        }
+
+        // Everything the pre-roll will eventually come back out as. Discarding
+        // it is what puts the output at the first sample meant to be heard;
+        // draining happens as it becomes available, over however many blocks
+        // that takes, because the pipe cannot be made to answer sooner.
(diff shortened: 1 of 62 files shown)


magda/engine/clip/ClipVoicePool.cpp
```diff
@@ -309,25 +363,40 @@ void ClipVoicePool::service() {
...
-                        // is playing cannot be pointed elsewhere and the next
-                        // read corrects it anyway, at the cost of a seek.
-                        if (reuse.anchorSamples != placement.sourceOffsetSamples) {
+                        // A stretch setting changed, which the file did not.
+                        // Rebuilding the whole reader for it would close a file
+                        // and pay a seek to reopen it at the same place, so only
+                        // the stretcher is replaced; it is state and a warm-up,
+                        // both of which a rate or pitch change invalidates
+                        // anyway.
+                        if (reuse.setup != setup) {
```
lucaromagnoli, line 379: The replacement stretcher is never primed during a live rate, pitch, or mode edit. `ClipVoice::primed_` remains true because the clip/event IDs have not changed, so the newly constructed stateful engine reaches `process()` without a reset or pre-roll. SoundTouch can consequently output silence/latency, while the other engines can start discontinuously. Could the table expose a stretcher generation (or the voice retain the pointer) and clear `primed_` whenever the instance changes?

magda/engine/clip/ClipStretcher.cpp
```diff
@@ -0,0 +1,477 @@
...
+            //
+            // A speed ramp is the exception to the rate of one. Its edges are
+            // not at one however the clip is set, and a ramp is a tape effect in
+            // any case, so this is what plays it.
+            if (!setup.speedRamp && std::abs(setup.nominalRate - 1.0) < 1.0e-9)
+                return nullptr;
+
+            return std::make_unique<ResamplingClipStretcher>(setup);
+
+        default:
```
lucaromagnoli, line 654: Returning `nullptr` here does not actually produce unity playback when the event has a non-unity speed. `readingPositionAt()` still advances by the requested ratio, while the null-stretcher path reads `count` contiguous samples. At 2x, for example, each block plays internally at unity and the next block begins 2x farther into the file, skipping material at every boundary. Please either make the position map fall back to unity when the selected engine is unavailable or provide an explicit supported fallback.

magda/engine/clip/EventPlacement.cpp
```diff
@@ -98,4 +102,105 @@ ClipPlacement placementFor(const AudioEventPlayback& event, double deviceSampleR
...
+                             const RenderContext& context) {
+    StretchSetup setup;
+    setup.mode = event.timeStretchMode;
+
+    // What the incumbent reads, and the two are not added: a clip following the
+    // pitch track transposes by its own offset, and one that is not plays the
+    // pitch change its editor set. Auto pitch's other half, the offset from the
+    // project's pitch sequence, needs a pitch track the engine does not have
+    // yet (#1891); until it does, such a clip plays its transpose alone.
+    setup.semitones = event.autoPitch ? static_cast<float>(event.transpose) : event.pitchChange;
```
lucaromagnoli, line 194: Auto-pitch can calculate a non-zero transpose here but still create no stretcher. `AudioEvent::getEffectiveTimeStretchMode()` upgrades Off only for auto-tempo, warp, speed ratio, or `pitchChange`; it does not consider `autoPitch`/`transpose`. Thus auto-pitch=true, transpose!=0, pitchChange=0, and unity speed leaves `setup.mode` disabled, `makeStretcher()` returns null, and this transpose is silently ignored. The current test masks this by retaining `pitchChange = 3`. Please force an available stretch mode whenever the selected semitone adjustment is non-zero (and add the auto-pitch-only playback case).

lucaromagnoli (commented):
Found three playback correctness issues in the new stretch paths.

magda/engine/clip/ClipStretcher.cpp
```diff
@@ -0,0 +1,477 @@
...
+                 juce::dsp::AudioBlock<float> output) override {
+        const auto in = samplesOf(input);
+        const auto out = static_cast<int>(output.getNumSamples());
+        if (out <= 0)
+            return;
+
+        if (in > 0) {
+            touch_.setTempo(
+                std::clamp(static_cast<double>(in) / out, kMinStretchRate, kMaxStretchRate));
+            write(input, 0, in);
```
lucaromagnoli: This write is unchunked, and `in` can exceed what interleaved_ holds. The buffer is sized ceil(maxBlock * kMaxStretchRate) + 1 frames, but the voice clamps `wanted` to stretchScratchSamples - maxBlock, which is int(maxBlock * kMaxStretchRate) + 4: three frames more. write() then stores in * channels floats in one pass, which is a heap write past the vector's end.

The only route to it is an auto tempo step past kMaxStretchRate, since the constant-ratio branch clamps and a ramp only slows below its base, so it takes a project-to-file tempo ratio above ten (an interpBpm of 15 under a 160 bpm project). Rare, but user-reachable, and the failure is memory corruption rather than bad audio.

prime() already chunks by stride for exactly this reason; process() should reuse that loop, or the voice's clamp ceiling should be int(maxBlock * kMaxStretchRate) + 1 so every buffer sized against the rate ceiling actually covers what a block can carry. Doing both costs nothing.

magda/engine/clip/EventPlacement.cpp
```diff
@@ -98,4 +102,105 @@ ClipPlacement placementFor(const AudioEventPlayback& event, double deviceSampleR
...
+    // Seconds of the file's own material, which is what the reading is counted
+    // in: it was converted to the device's rate below the stream, so a second of
+    // it is deviceSampleRate samples whatever the file was recorded at.
+    double elapsed = 0.0;
+
+    if (event.autoTempo && event.interpBpm > 0.0) {
+        const auto at = ramped(clip, beat, clip.span.startBeat, clip.span.endBeat, clip.fadeInBeats,
+                               clip.fadeOutBeats);
+
+        elapsed = (at - event.span.startBeat) * 60.0 / event.interpBpm;
```
lucaromagnoli, line 167: The constant-ratio branch below clamps to what a stretcher will run at, and this branch does not: a degenerate interpBpm makes the effective rate here arbitrarily large. The voice's scratch clamp keeps it memory-safe in principle, but the block then truncates its reading, the next block's positions jump past what was read, and every block after that is a seek: silence with underruns for as long as the clip is in reach. It is also the one route to the overflow flagged in ClipStretcher.cpp.

Clamping elapsed here per block would break the rounded-ends continuity, so the clamp belongs on the consumption side: bounding wanted in ClipVoice to int(maxBlock * kMaxStretchRate) + 1 keeps the position map pure and makes the degenerate case merely sound wrong, the way the constant branch already merely sounds wrong.

magda/engine/clip/ClipVoice.cpp
```diff
@@ -89,17 +94,79 @@ bool ClipVoice::render(const AudioClipPlayback& clip, const AudioEventPlayback&
...
+    // continues it rather than seeking again.
+    //
+    // Not after an underrun, deliberately, and this is the difference between a
+    // reader catching up and one that never does. Priming reads behind the
+    // position the block wants, so a stream that came back short and was primed
+    // again for it would be sent backwards every block, and every one of those
+    // is a seek that guarantees the next block is short too. A stretcher handed
+    // silence and then material has a discontinuity in what it is playing, which
+    // is what a loop wrap is as well, and neither needs anything said about it.
+    if (stretcher != nullptr && (!primed_ || !block.continuous)) {
```
lucaromagnoli: primed_ survives a stretcher the pool has replaced. An edit to pitch, rate or mode on a sounding clip swaps in a fresh stretcher without touching the stream, the next table hands this voice the new instance, and this condition sees primed_ true and a continuous block, so the new stretcher runs cold: Signalsmith unprimed is roughly its full latency late, 120 ms at the default preset, and it stays late until the next locate or restart because nothing ever primes it. Dragging a pitch control during looped playback is exactly this gesture.

The voice cannot see the swap today because it keeps no memory of which stretcher it primed. One pointer member, primed-against, compared each block and re-priming on change, turns the edit into the one counted seek it should cost. The rebuild-on-setup-change path in the pool is right; this is the audio-thread half it still needs.

magda/engine/exec/RenderContext.hpp
```diff
@@ -130,6 +130,28 @@ struct BlockInfo {
      * edges of a region, and a region that runs to the end of the block ends at
      * the sample after the last one, the way every half-open range does.
      */
+    /**
```
lucaromagnoli, line 144: The insertion point orphans sampleForTime's docblock: its comment now sits directly above this one, so the tools and the reader both hand it to beatAtTime and sampleForTime is left bare. Moving beatAtTime above that docblock or below sampleForTime restores both pairings.

lucaromagnoli (commented):
Reviewed the first-party surface; the vendored sources I took as given, and their wiring (own static SoundTouch target, SYSTEM includes, the ac_test rename through a compile definition, licences and SOURCES.md) is clean.

What I verified rather than trusted:

- The cue, the prime and the first read are exactly contiguous: cueFor is position plus readAhead minus preRoll, the prime reads [readFrom - preRoll, readFrom), and the block's read continues at readFrom, so the pool's startAt lands on the first sample the prime will ask for. No seek anywhere in the start path.
- The rounded-ends invariant holds under auto tempo: beatAtTime is exact at a block's own ends, block N's endBeat is block N+1's startBeat, so llround(closes) of one block is llround(opens) of the next to the bit. Nothing accumulates.
- fadeRampPosition is the fork's rescale integrals term for term, and both engine configurations match the fork's wrappers exactly (formant base 200/rate, neutral formant shift, and the four Better-mode SoundTouch settings).
- The analog pitch claim is true in the model: ClipManager folds 2^(pitch/12) into speedRatio when isAnalogPitchActive, and the effective mode stays kDisabled, so the resampling path is what plays it.
- The scratch aliasing is safe: the render region and the reading never overlap, and the reading offset is fixed by maxBlockSize rather than by the block.
- cueSamples as an identity rather than a position is a good catch; storing the transport-dependent cue would have republished the table every round for any clip the cursor was inside.

Three findings inline: one reachable buffer overflow (small, degenerate input, but heap), one alignment gap on mid-play stretch edits, and the unclamped auto-tempo branch that triggers the first. The doc-placement nit from RenderContext.hpp rides along.

The test suite earns its claims: the position map as exact arithmetic, resampling sample for sample through the counting reader, and the engines judged by what a listener would check. The locate test and the two SoundTouch modes are exactly the right pins.

magda/engine/clip/ClipStretcher.cpp
```diff
@@ -0,0 +1,488 @@
...
+            // A mode this build has no engine for, which is every value
+            // Tracktion's enum holds that MAGDA never wrote. The incumbent
+            // answers this with its default engine (TimeStretcher::
+            // checkModeIsAvailable) and so does this: the clip is asking to be
+            // stretched, and the position map has already decided how much
+            // material a block will consume. Returning nothing here would leave
+            // a block reading that material at unity and the next one starting
+            // where the ratio says, which is a skip at every block boundary
+            // rather than a fallback.
+            return std::make_unique<SignalsmithClipStretcher>(setup);
```
lucaromagnoli, line 664: This branch, and the two engine branches above it, build an engine on the strength of the mode field alone, and the incumbent does not: usesTimeStretchedProxy (tracktion_AudioClipBase.cpp:1901) engages a stretcher only when something is actually asked, auto tempo, auto pitch, a pitch change or a ratio off unity, and the mode alone never does. A clip whose mode was set once and asks for nothing, unity ratio, no pitch, no ramp, plays clean through the fork and through a phase vocoder at one-to-one here, which is not transparent. That case predates this commit for the two named engines and this fallback widens it to every unknown mode; the null-diff corpus (#2040) will flag every project that ever touched the dropdown without stretching.

The asks-nothing test the kDisabled branch already runs, unity nominal rate, no semitones, no ramp, belongs in front of the engine branches too, with one carve-out: auto tempo keeps its engine even at a unity average, both because its per-block rate varies around that average and because the incumbent engages on getAutoTempo() alone. As it stands today an auto tempo clip always arrives here with an upgraded mode, so the carve-out is one autoTempo check on the setup, which stretchSetupFor already has in hand.

magda/engine/clip/ClipStretcher.cpp
```diff
@@ -0,0 +1,488 @@
...
+        }
+
+        touch_.setPitchSemiTones(setup.semitones);
+        touch_.setTempo(std::clamp(setup.nominalRate, kMinStretchRate, kMaxStretchRate));
+
+        // Sized for the worst block this can be asked for: the largest output
+        // block, times the fastest the reading may be consumed for it. Grown
+        // here, on the thread that made this, and never again.
+        const auto frames =
+            static_cast<int>(std::ceil(setup.maxBlockSamples * kMaxStretchRate)) + 1;
```
lucaromagnoli: maxReadingSamples' comment now says these numbers are not allowed to be separately derived, and this line still derives it separately: ceil(maxBlock * kMaxStretchRate) + 1, which is numerically maxReadingSamples today and harmless under writeAll either way, but the sentence is only true once this calls the function.

lucaromagnoli (commented):
Re-reviewed the response commit. All four findings are resolved the right way round: the overflow is closed from both ends (writeAll chunks the process path and maxReadingSamples is the one bound every buffer and the voice's clamp now share), primed-against-identity is exactly the fix with a direct voice-level test that covers first prime, no re-prime while continuous, re-prime on swap and re-prime on jump, the auto tempo ceiling is enforced at consumption with the position map left pure and a SoundTouch rig test at fifteen times the ceiling proving it, and sampleForTime has its docblock back.

The two additions beyond the review are both justified, and I checked the reasoning on the first: a null stretcher under a non-unity ratio really would skip at every block boundary, since the voice's unstretched path reads one sample per sample while the positions advance at the ratio. Falling back to the default engine is also what the incumbent does with a mode it cannot run.

One parity gap opens alongside it, inline at the fallback, and one one-line consistency nit beside it. Neither blocks the merge if the parity case is accepted as a known #2040 delta, but the fix is small and the corpus will otherwise flag every project that ever touched the mode dropdown.

lucaromagnoli (commented):
Third pass, on the mode-is-a-preference commit. Both remaining items are resolved and nothing new opens.

The asks-nothing gate sits in front of every engine branch with exactly the carve-out it needed, and followsTempo is derived from the same condition the position map uses for the auto tempo path, so the two cannot disagree about which clips follow the tempo. I traced the edges of the refactored kDisabled branch: analog pitch at zero pitch now correctly gets nothing through the top gate, analog pitch with a folded ratio passes the gate on its nonzero semitones and still reaches the resampler, which ignores them, a speed ramp at unity keeps its engine, and kDisabled with followsTempo is unreachable because analog pitch requires auto tempo off. The unity threshold asymmetry between the gate (1e-9) and the model's effective-mode upgrade (1e-3) leaves a microscopic band where a raw ratio like 1.0005 resamples here and stretches in the fork; that band is a model threshold question older than this PR and inaudible, so I am noting it for #2040's records rather than asking for anything.

The two new tests pin both sides of the gate, mode-set-but-idle across all three engines and auto tempo at a unity average, and the SoundTouch buffer now derives from maxReadingSamples, making the not-separately-derived sentence true.

This resolves the whole review. Good to merge from my side.

PR #2056, opened 2026-08-08
Remote API: permissions, audit, and cross-transport conformance (#1860)

### LINKED ISSUE #1860: Remote API: permissions, settings, audit log, and conformance tests

lucaromagnoli:
Closes #1860.

Remote clients could do anything the token holder could. This adds the permission model that makes the settings UI worth opening.

## Scopes

`read`, `edit`, `transport`, `session`, `hardware-midi` — declared once per operation in a contiguous policy table in `remote_api.cpp`, enforced in `RemoteApiService` so both transports get one answer rather than two.

The split is by what a user would regret, not by which manager the code calls. Driving the transport is separable from editing because a remote that only starts and stops playback is a thing people want and must not also be able to delete a track. Session launch is neither.

Two orderings are load-bearing. Permission is checked **before input validation**, so a client that may not call something cannot use the refusal as an oracle for a schema it is not allowed to reach. And **before the idempotency cache**, so a revoked grant cannot be handed a replayed success — the one path where revocation would otherwise not take effect.

A write with no policy entry keeps the `read` default. The registry constructor turns that into a startup failure, and a test re-checks it in release builds where `jassert` is a no-op.

## Identity and grants

A client names itself — `?client=name` on the WebSocket upgrade, `clientInfo.name` for MCP. Normalised, so `Cursor` / `cursor ` / `cursor` are one entry. Read-only on first sight, and registered the moment it asks for anything so it appears in the settings list straight away.

The grant is looked up **per request**, never cached on the connection. That is the entire mechanism behind "revoking takes effect without restart": there is no copy to go stale, so a change applies to the next request on the socket the client is already holding.

Grants persist under `remoteApi.clients` in config. No credential is in there.

## Also in scope

- **Audit log** — bounded, in memory, 512 entries: time, client, connection, transport, operation, request id, outcome, short reason. No payloads, and failures recorded by error *code* rather than message, because a validation message quotes the offending value.
- **Redaction** — registered secrets plus a `Bearer` shape rule, applied once inside `record` rather than at each call site. The startup log now names the discovery file instead of its path.
- **Token rotation** — implemented as a listener restart, which is what guarantees every existing connection is gone: they were admitted by a token that no longer exists.
- **Settings UI** — rotation on the Remote tab; a new **Clients** tab with a scope checkbox per client, connected state and transport, disconnect, forget, and the recent-activity tail, so a refusal and the checkbox that fixes it are on one screen.

## What this is not

A sandbox, and the docs lead with that. The token is the security boundary, it lives in a file any process running as the user can read, and the client name is self-declared — a token holder can claim any name and get whatever it was granted. Per-client secrets would be published in the same readable file and buy nothing.

What the model does buy is the thing that actually goes wrong: an assistant you gave read access does not silently delete a track, and when something is refused you can see it and grant it in one place. `docs/architecture/remote-api-permissions.md` states the threat model in full.

## `hardware-midi` has no operations yet

The registry exposes no hardware MIDI surface. The scope is declared now because grants are persisted: a word invented later reads as "not granted" on every existing client — the correct answer, but only if the word already exists when those grants are written.

## Bugs found and fixed along the way

- Inserting fields into the middle of `McpEndpoint::Call` silently re-mapped positional aggregate init in `test_remote_mcp.cpp`, so `idempotencyScope` became the client name. The helpers now assign by name.
- The audit wrapper originally captured `this`; a completion running after `shutdown()` would have dereferenced a destroyed dispatcher. It now carries the log pointer by value.
- `persistGrants` was reachable from a transport thread when a new client registered itself, and `Config::save()` reads the whole unlocked singleton. It now hops to the message thread, capturing a snapshot rather than `this`.

## Tests

- `test_remote_permissions.cpp` — vocabulary, registry, enforcement, audit, redaction.
- `test_remote_permission_conformance.cpp` — one vector table driven through **both** transports over real sockets, asserting they agree; plus revocation without reconnect, disconnect, anonymity, fail-closed with no registry, and a concurrent-load case that flips the grant underneath four clients.
- Host-level coverage for rotation, grant persistence across a restart, and a brand-new client being read-only end to end.

A transport configured without a registry refuses **everything**, reads included — so a future transport that forgets to consult it fails loudly on its first test run rather than handing out the project silently.

## Verification

Full Catch2 suite green (2294 passed, 13 skipped, 0 failed). Both live JUCE suites green. Debug and release builds clean. Load case run three times for flakiness.

Two caveats:

- **Limits** (payload, rate, concurrency, queue) were already implemented by #1856/#1858. This documents and tabulates them; it does not add new ones.
- **Packaging** verified on macOS only. The change adds no new artifacts — only sources compiled into `magda_daw` — so there is nothing to stage differently, but Linux and Windows will only be confirmed by CI. The Clients tab compiles in debug and release but has not been seen rendered in a running app.

### DIFF
--- a/magda/daw/api/remote_api_host.cpp
+++ b/magda/daw/api/remote_api_host.cpp
@@ -1,9 +1,13 @@
 #include "remote_api_host.hpp"
 
+#include <juce_events/juce_events.h>
+
 #include <random>
 
 #include "AppPaths.hpp"
 #include "Config.hpp"
+#include "remote_audit.hpp"
+#include "remote_clients.hpp"
 #include "remote_mcp_server.hpp"
 #include "remote_model_bridge.hpp"
 #include "remote_service.hpp"
@@ -173,21 +177,68 @@ RemoteApiHost* activeHost() {
 }
 
 RemoteApiHost::RemoteApiHost(MagdaApi& api, AudioEngine* engine)
-    : service_(std::make_unique<RemoteApiService>(api)),
+    : audit_(std::make_unique<RemoteAuditLog>()),
+      clients_(std::make_unique<RemoteClientRegistry>()),
+      service_(std::make_unique<RemoteApiService>(api)),
       bridge_(std::make_unique<ModelChangeBridge>(*service_)),
       subscriptions_(std::make_unique<SubscriptionHub>(api, *service_)) {
     if (engine != nullptr)
         subscriptions_->setMeterSource(makeLiveMeterSource(*engine));
+
+    service_->setAuditLog(audit_.get());
+
+    // Grants are restored before the change handler is installed, so loading
+    // them does not immediately write them back out again.
+    clients_->loadGrantsFromJson(Config::getInstance().getRemoteApiClients());
+    clients_->setChangeHandler([this] { persistGrants(); });
+
     activeHostInstance = this;
 }
 
 RemoteApiHost::~RemoteApiHost() {
     stop();
+
+    // Before the members go: the handler captures `this` and the registry is
+    // free to notify from a transport thread, so leaving it installed while the
+    // object unwinds is a call into a half-destroyed host. `stop()` has already
+    // joined the transports, so nothing can be mid-notification here.
+    if (clients_ != nullptr)
+        clients_->setChangeHandler(nullptr);
+    if (service_ != nullptr)
+        service_->setAuditLog(nullptr);
+
     // Only if we are still the registered one. Tests construct hosts in
     // sequence, and an earlier one destructing after a later one was registered
     // must not clear the later one's registration.
     if (activeHostInstance == this)
         activeHostInstance = nullptr;
+}
+
+void RemoteApiHost::persistGrants() {
+    // Snapshotted here, applied possibly later, and deliberately capturing no
+    // `this`: the host may be destroyed between the two, and the queued write
+    // needs nothing from it.
+    auto grants = clients_->grantsToJson();
+    auto apply = [grants = std::move(grants)] {
+        auto& config = Config::getInstance();
+        config.setRemoteApiClients(grants);
+        config.save();
+    };
+
+    // `Config` is a singleton with no internal locking, and `save()` reads every
+    // member of it while the settings pages write them. This is reached from a
+    // transport thread whenever a client MAGDA has not seen registers itself, so
+    // writing from here would race the UI over the whole config file — for a
+    // change that is not urgent by any measure.
+    //
+    // Inline when there is no message loop to post to, which is the headless
+    // case: a test or a console host, where nothing else is writing config.
+    if (juce::MessageManager::existsAndIsCurrentThread() ||
+        juce::MessageManager::getInstanceWithoutCreating() == nullptr) {
+        apply();
+        return;
+    }
+    juce::MessageManager::callAsync(std::move(apply));
 }
 
 bool RemoteApiHost::start() {
@@ -217,15 +268,23 @@ bool RemoteApiHost::start() {
         return false;
     }
 
+    // The token is a credential that will appear in headers this process logs
+    // around. Registering it means any line that quotes one is masked before it
+    // reaches the audit log or the app log.
+    registerRemoteSecret(token_);
+
     RemoteWebSocketServer::Options options;
     options.bearerToken = token_;
     options.port = config.getRemoteApiPort();
+    options.clients = clients_.get();
+    options.audit = audit_.get();
     for (const auto& origin : config.getRemoteApiAllowedOrigins())
         options.allowedOrigins.push_back(juce::String::fromUTF8(origin.c_str()));
 
     server_ = std::make_unique<RemoteWebSocketServer>(*service_, options, subscriptions_.get());
     if (!server_->start()) {
         server_.reset();
+        forgetRemoteSecret(token_);
         token_ = {};
         return false;
     }
@@ -239,6 +298,8 @@ bool RemoteApiHost::start() {
     mcpOptions.bearerToken = token_;
     mcpOptions.port = config.getRemoteApiMcpPort();
     mcpOptions.allowedOrigins = options.allowedOrigins;
+    mcpOptions.clients = clients_.get();
+    mcpOptions.audit = audit_.get();
 
     mcpServer_ = std::make_unique<RemoteMcpServer>(*service_, mcpOptions, subscriptions_.get());
     if (!mcpServer_->start()) {
@@ -249,13 +310,16 @@ bool RemoteApiHost::start() {
     // A running listener whose token nobody can read is useless and still a
     // listener, so a failure to publish takes the servers down with it.
     if (!writeTokenFile(tokenFile(), token_, server_->boundPort(), mcpPort())) {
-        DBG("RemoteApiHost: could not write " + tokenFile().getFullPathName());
+        // The name, not the path. This line goes to the app log, which users
+        // paste into bug reports, and the directory is their home directory.
+        DBG("RemoteApiHost: could not write " + redactedFileName(tokenFile()));
         if (mcpServer_ != nullptr) {
             mcpServer_->stop();
             mcpServer_.reset();
         }
         server_->stop();
         server_.reset();
+        forgetRemoteSecret(token_);
         token_ = {};
         return false;
     }
@@ -285,7 +349,32 @@ void RemoteApiHost::stopListening() {
     // would advertise a port that is no longer listening and a credential that
     // no longer works.
     tokenFile().deleteFile();
+    forgetRemoteSecret(token_);
     token_ = {};
+
+    // Connections were tracked by the servers that have just gone. Anything
+    // still listed refers to a socket that no longer exists — a settings table
+    // showing phantom clients, with a disconnect button that would answer
+    // false. Grants are untouched: those are the user's decisions, not
+    // per-run state.
+    if (clients_ != nullptr) {
+        for (const auto& client : clients_->connections())
+            clients_->noteDisconnected(client.connectionId);
+    }
+}
+
+bool RemoteApiHost::rotateToken() {
+    if (!isRunning())
+        return false;
+
+    // A restart rather than a swap of the credential in place. Both servers
+    // read `bearerToken` from an immutable Options copy taken at construction,
+    // and making it mutable would mean a live comparison against a value
+    // changing under the comparing thread — for the one operation where being
+    // half-applied is worst. Rebuilding is what guarantees every existing
+    // connection is gone: they were admitted by a token that no longer exists.
+    stopListening();
+    return start();
 }
 
 void RemoteApiHost::stop() {
@@ -328,6 +417,14 @@ RemoteApiService& RemoteApiHost::service() {
 
 SubscriptionHub& RemoteApiHost::subscriptions() {
     return *subscriptions_;
+}
+
+RemoteClientRegistry& RemoteApiHost::clients() {
+    return *clients_;
+}
+
+RemoteAuditLog& RemoteApiHost::audit() {
+    return *audit_;
 }
 
 }  // namespace remote
--- a/magda/daw/api/remote_api_host.cpp
+++ b/magda/daw/api/remote_api_host.cpp
@@ -215,13 +215,38 @@ RemoteApiHost::~RemoteApiHost() {
 }
 
 void RemoteApiHost::persistGrants() {
-    // Snapshotted here, applied possibly later, and deliberately capturing no
-    // `this`: the host may be destroyed between the two, and the queued write
-    // needs nothing from it.
+    // Read before taking the writer's lock: `grantsToJson` takes the registry's,
+    // and holding two is worth avoiding even where the order happens to be safe.
     auto grants = clients_->grantsToJson();
-    auto apply = [grants = std::move(grants)] {
+
+    auto writer = grantWriter_;
+    {
+        const std::lock_guard<std::mutex> lock(writer->mutex);
+        writer->latest = std::move(grants);
+        // Already a write on its way. It reads `latest` when it runs, so it will
+        // carry this change too — and posting a second task would be how the
+        // older of two snapshots ends up applied last.
+        //
+        // That reordering is the reason this is a slot rather than a captured
+        // value: a transport thread can snapshot, the user can then revoke a
+        // scope in the settings dialog and have it saved inline, and the
+        // transport's older snapshot would land afterwards and put the revoked
+        // grant back — silently, and permanently, because config is what the
+        // next launch reads.
+        if (writer->posted)
+            return;
+        writer->posted = true;
+    }
+
+    auto apply = [writer] {
+        juce::var latest;
+        {
+            const std::lock_guard<std::mutex> lock(writer->mutex);
+            latest = writer->latest;
+            writer->posted = false;
+        }
         auto& config = Config::getInstance();
-        config.setRemoteApiClients(grants);
+        config.setRemoteApiClients(latest);
         config.save();
     };
 
@@ -229,7 +254,8 @@ void RemoteApiHost::persistGrants() {
     // member of it while the settings pages write them. This is reached from a
     // transport thread whenever a client MAGDA has not seen registers itself, so
     // writing from here would race the UI over the whole config file — for a
-    // change that is not urgent by any measure.
+    // change that is not urgent by any measure. Hopping also gives the
+    // coalescing above one thread to serialise on.
     //
     // Inline when there is no message loop to post to, which is the headless
     // case: a test or a console host, where nothing else is writing config.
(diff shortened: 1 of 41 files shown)


magda/daw/api/remote_websocket_server.cpp
```diff
@@ -618,15 +718,21 @@ struct RemoteWebSocketServer::Impl {
                 return;
             }
             subscriptions->handle(connection->subscriber, method, params,
-                                  [connection, id](Response response) {
+                                  [this, connection, id, method, idKey](Response response) {
```
lucaromagnoli: [P1] This completion can run asynchronously on the message thread, but it captures the server's raw `this`. If listener shutdown or token rotation destroys `RemoteWebSocketServer` while the transport thread queues the subscription job, the completion later calls `audit()` through freed storage. Please capture the audit log and immutable request data independently, or guard the server lifetime.

magda/daw/api/remote_api_host.cpp
```diff
@@ -173,23 +177,70 @@ RemoteApiHost* activeHost() {
...
     // must not clear the later one's registration.
     if (activeHostInstance == this)
         activeHostInstance = nullptr;
 }
 
+void RemoteApiHost::persistGrants() {
+    // Snapshotted here, applied possibly later, and deliberately capturing no
+    // `this`: the host may be destroyed between the two, and the queued write
+    // needs nothing from it.
+    auto grants = clients_->grantsToJson();
```
lucaromagnoli: [P1] This snapshots grants before posting the save to the message thread. A worker can take an old snapshot, the UI can revoke a scope and save the new state inline, and then the worker can enqueue its older snapshot, restoring the revoked grant in config and after restart. Please serialize or coalesce these writes, or version snapshots and discard stale applications.

magda/daw/ui/dialogs/AISettingsDialog.cpp
```diff
@@ -2741,6 +2796,327 @@ class AISettingsDialog::RemoteApiPage : public juce::Component, private juce::Ti
...
+            emptyLabel_.setText("The remote API is not available in this session.",
+                                juce::dontSendNotification);
+            emptyLabel_.setVisible(true);
+            return;
+        }
+
+        const auto grants = host->clients().grants();
+        const auto connections = host->clients().connections();
+
+        juce::StringArray names;
```
lucaromagnoli, line 3033: [P2] `forget()` intentionally leaves live connections open, but this list is built only from grants. After Forget, the row therefore disappears while the client remains connected, leaving no way to disconnect it from this UI until another request recreates its grant. Please build rows from the union of grants and connections, or disconnect as part of forgetting.

magda/daw/api/remote_mcp.cpp
```diff
@@ -820,6 +823,9 @@ void McpEndpoint::readResource(const Call& call, Completion onComplete) {
 std::optional<RequestContext> McpEndpoint::requestContext(const Call& call, McpError& error) const {
     RequestContext context;
     context.clientId = call.clientId;
+    context.clientName = call.clientName;
+    context.transport = TRANSPORT_MCP;
+    context.scopes = call.scopes;
```
lucaromagnoli, line 859: [P2] Propagating scopes here only covers calls that reach `requestContext()` and the dispatcher. Direct MCP methods handled earlier (`tools/list`, `resources/list`, and `resources/templates/list`) return before this and never check `call.scopes`. Consequently `Options::clients == nullptr` still permits those methods, contrary to the documented fail-closed invariant. Please require `Scope::Read` on those direct read surfaces and add MCP coverage to the no-registry conformance test.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Fixed in f37808b8 — right, and it is the same bug I had already fixed one floor up in `RemoteApiService::dispatch` and then missed here.

The completion now captures the audit log by value and goes through a free `recordAudit(log, connection, …)` rather than `Impl::audit`, so it reaches back into the server for nothing. The connection is already a `shared_ptr`, so the client fields stay valid regardless.

Worth naming the trigger explicitly, since it is not hypothetical: token rotation calls `stopListening()` then `start()` while connections are live, so a hub completion outliving the server that queued it is a routine path rather than a shutdown-only edge.

magda/daw/api/remote_api_host.cpp
lucaromagnoli: Fixed in f37808b8. The interleaving you describe restores a revoked grant in the file the next launch reads — silent and permanent, which for this feature is the worst available failure.

Replaced the captured snapshot with a shared slot: whoever finds `posted` false owns the hop, everyone after simply replaces `latest`, and the pending task re-reads `latest` when it runs on the message thread. So the newest state always wins, a burst of changes costs one write, and the ordering problem cannot arise because there is only ever one write in flight. The slot is held by `shared_ptr` and the task captures nothing else, so it is safe for it to outlive the host; if `callAsync` fails on a quitting loop the flag is cleared rather than left stuck.

magda/daw/ui/dialogs/AISettingsDialog.cpp
lucaromagnoli: Fixed in f37808b8 — rows are now built from the union of grants and connections.

I kept `forget()` non-disconnecting (it is the weaker of the two actions, and pairing it with a disconnect would leave no way to say "revert this client to read-only" without also cutting it off). The update loop had the same defect and is fixed with it: it iterated grants, so a row without one would never have refreshed. It now iterates the rows and falls back to `defaultClientScopes()` for a client with no grant — which is exactly what that client will get on its next request, so the row shows the truth rather than a blank.

magda/daw/api/remote_mcp.cpp
lucaromagnoli: Fixed in f37808b8. This one was a documentation-versus-code mismatch, which is the worst kind: both `remote-api-permissions.md` and the no-registry conformance test asserted an invariant the code did not honour.

The three catalogue methods now require `Scope::Read`. `server/discover` and `initialize` deliberately still do not, and the reason is written down at the call site: they are how a client learns the endpoint exists and negotiates a protocol version, they carry no catalogue and no project data, and a client that could not complete a handshake could not be told why it was refused.

The no-registry conformance test now covers MCP as well as WebSocket, and there is a companion asserting the catalogue is still readable by an ordinary client, so the gate cannot be tightened into a regression. I checked the new test fails without the gate rather than trusting that it passes with it — 1 failed at `test_remote_permission_conformance.cpp:498` with the check stubbed out.

magda/daw/api/remote_api_host.cpp
lucaromagnoli: [P1] This still permits the same stale ordering because the snapshot is taken before `writer->mutex`. An older callback can snapshot A and pause; a newer callback snapshots and stores B; then the older callback resumes, replaces `latest` with A, sees `posted`, and returns. The pending task consequently persists A. Please serialize the snapshot itself under the writer lock, or attach a registry generation and reject older snapshots.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: [P1] The token-rotation path is fixed, but the replacement still captures a raw `RemoteAuditLog*`. `SubscriptionHub::shutdown()` clears its hub gate, yet a queued job subsequently invokes this completion with `Cancelled`. If the host has finished destructing before that message runs, `audit_` has also been destroyed and `recordAudit(log, ...)` dereferences freed memory. Member declaration order only protects destruction within the host; it cannot protect work that runs after the entire destructor. Please give queued completions shared audit ownership or an audit lifetime gate.

tests/test_remote_permission_conformance.cpp
```diff
@@ -0,0 +1,711 @@
...
+    // afterwards cannot have been.
+    const auto total = kClients * kRequestsEach;
+    const auto waitForProgress = [&](int target) {
+        const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(30);
+        while (completed.load() < target && std::chrono::steady_clock::now() < deadline)
+            std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        return completed.load() >= target;
+    };
+
+    REQUIRE(waitForProgress(total / 4));
```
lucaromagnoli: [P2] This is still scheduler-dependent: `completed >= total / 4` does not mean the main thread resumes at that boundary. It may wake after the workers have advanced through most or all requests, leaving too few denials for the lower bound and recreating the CI flake this change is meant to remove. Also, if either `REQUIRE(waitForProgress(...))` fails, stack unwinding destroys joinable worker threads before the join loop and calls `std::terminate`. Please coordinate explicit phases with barriers/latches and make joining RAII-safe.

lucaromagnoli: Both correct, fixed in 2c1ce726. You are right that "provably" was the wrong word — `completed >= total / 4` says a quarter of the work has finished, not that the main thread resumes at that boundary, so it can wake with the workers nearly done and have too few requests left to deny. Same flake, longer fuse.

Replaced the polling with explicit rounds and a barrier between them. The grant now only moves while every worker is parked, so a request belongs to the round it was issued in by construction: round 1 is all-allowed, round 2 all-denied, and the bounds are arithmetic rather than a race the test usually wins. Round 3 keeps the actual concurrency — the grant moving continuously under the per-request lookup — and deliberately contributes to neither bound, so its only way to fail is by producing an answer that is neither.

The join is now RAII: the threads live in a pool whose destructor aborts the barrier first and then joins, since workers parked at a barrier that will never open would otherwise deadlock the join they are being joined by. I verified that rather than assuming it — forcing a failure mid-test gives exit 1 with a legible Catch2 report instead of 134.

While in there I fixed a third defect neither of those covers: `runOverWebSocket` asserts with `REQUIRE`, and four workers were calling it concurrently. Catch2 assertion macros are not thread-safe, so that was undefined behaviour dressed up as a passing test. The workers now use a macro-free helper that reports through its return value, and every assertion happens on the main thread after the join.

10 consecutive runs clean locally.

lucaromagnoli: [P2] The barriers fix the phase-boundary and unwind problems, but round 3 can still pass without exercising concurrent access. `openRound(2)` wakes the workers before this thread starts toggling; if the main thread is descheduled there, the workers can finish all 40 racing-round requests, making the loop condition false before a single grant write occurs. Please start a dedicated flipper before releasing this round, or gate worker requests until flipping has demonstrably begun.

magda/daw/api/remote_api_host.cpp
lucaromagnoli: Right again — I fixed the symptom and left the cause. Taking the snapshot before the lock leaves the identical reordering one level down, exactly as you traced it: read A, get preempted, another thread reads and stores B, then A resumes, overwrites `latest`, sees `posted` and returns.

Fixed in a1c1628a by taking the snapshot *under* the writer lock, which is what makes `latest` actually the latest. That means holding the registry lock inside the writer lock; it is safe in this order and only this order, because the registry always calls its change handler after releasing its own lock, so the pair is never taken the other way round. I preferred that to a generation counter since it removes the stale snapshot rather than detecting it.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Correct, and the distinction you draw is the one I got wrong: declaration order sequences destruction *within* the host, which says nothing about work that runs after the whole destructor. The `Cancelled` path you describe is the clearest case — the job still completes, and completing is what writes the audit line.

Fixed in a1c1628a by giving the log shared ownership end to end: `RemoteApiHost` holds a `shared_ptr`, `RemoteApiService::setAuditLog` takes one, both transports' `Options::audit` is one, and every queued completion captures a share. Whoever still holds it keeps the log alive for exactly as long as something might write to it, which removes the lifetime question rather than reasoning about it.

The client registry is deliberately still a raw pointer: it is only ever touched synchronously on a connection thread inside a live request, never from a queued completion, so it has the same lifetime as the `RemoteApiService&` the transport already holds by reference.

tests/test_remote_permission_conformance.cpp
lucaromagnoli: Correct — releasing the workers and then starting to toggle on this thread is not the same as toggling while they run, and if the handover deschedules the main thread the round completes having exercised nothing. The test would have gone green while proving the opposite of what it claims.

Fixed in a1c1628a. The flipper is now its own thread, started *before* the round is released, and the workers are not released until it has demonstrably flipped at least once. The round then ends on the workers parking at the next barrier rather than on a completion counter this thread was racing, so the flipper is still running for the whole round by construction — and the test asserts the flip count advanced across it, so a flipper that died early fails rather than passes quietly. It joins through a guard, like the workers.

8 consecutive runs clean.

lucaromagnoli: [P2] The dedicated thread helps, but this assertion can still be satisfied before the racing round opens. The flipper may increment after `flipsBefore` is loaded but before `openRound(2)` releases the workers; it can then be descheduled for the whole round, and `flips > flipsBefore` still passes without any grant write occurring while requests were allowed to run. Please count flips only inside an explicit round-active interval (set before release and cleared after `allParked`), or make the flipper participate in the same start barrier.

magda/daw/api/remote_api_host.cpp
```diff
@@ -173,23 +177,112 @@ RemoteApiHost* activeHost() {
...
+    // member of it while the settings pages write them. This is reached from a
+    // transport thread whenever a client MAGDA has not seen registers itself, so
+    // writing from here would race the UI over the whole config file — for a
+    // change that is not urgent by any measure. Hopping also gives the
+    // coalescing above one thread to serialise on.
+    //
+    // Inline when there is no message loop to post to, which is the headless
+    // case: a test or a console host, where nothing else is writing config.
+    if (juce::MessageManager::existsAndIsCurrentThread() ||
+        juce::MessageManager::getInstanceWithoutCreating() == nullptr) {
```
lucaromagnoli, line 287: [P2] The no-message-loop path can still run `Config` writes concurrently. `apply()` clears `writer->posted` under the mutex and then releases it before `setRemoteApiClients()` and `save()`. A second transport thread can enter after that clear, see `posted == false`, and execute another inline `apply()` while the first is still writing. `RemoteApiHost` starts multi-threaded WebSocket/MCP transports in this headless/test case, while `Config` and its `juce::var` have no internal locking, so two newly seen clients can race both the in-memory assignment and file save. Please serialize the entire inline apply, or protect Config writes with a separate mutex.

tests/test_remote_permission_conformance.cpp
lucaromagnoli: Correct — sampling on this thread leaves a gap at each end, and a flip landing in either one satisfies "the count moved" while nothing raced anything.

Fixed in d0d30f3f by moving the observation into the workers rather than tightening the window: each reads the flip counter either side of *its own* racing requests, and a worker that saw it move was demonstrably issuing requests while the grant was being rewritten. That is the property stated directly, so there is no interval for the main thread to race — which felt like the better answer than a third attempt at getting a window right.

I verified it has teeth by stopping the flipper entirely: the test now fails on `observedConcurrent > 0`.

Worth recording that the first attempt at that verification failed for the wrong reason. The readiness wait was 30s, and a worker sitting in it is not in `read()` — so the server's ping timeout dropped all four connections and the run died with forty transport errors instead of one legible assertion. The wait is now 500ms: free when the flipper is already spinning, and short enough that the assertion is what reports the problem rather than the transport.

magda/daw/api/remote_api_host.cpp
lucaromagnoli: Correct, and it is the hop's own purpose defeated on the path that cannot hop. Clearing `posted` under the mutex and then writing unlocked lets a second transport thread see `posted == false` and enter while the first is still inside `save()` — and `RemoteApiHost` really does start both transports with real threads in the headless case, so two clients connecting at once hit it.

Fixed in d0d30f3f: the whole apply is serialized on its own mutex. Kept separate from the one guarding `latest` deliberately, so a thread recording a newer grant is never blocked behind a file write — it just leaves `latest` for whoever is inside the write to pick up. Taken before `mutex` and never after, and `persistGrants` only ever takes `mutex`, so there is no path that acquires them in the other order.

Issue #2057, opened 2026-08-08
TrackManager::getRackByPath answers a chain path with the chain's parent rack


lucaromagnoli:
Found while adding the facade's path-addressed rack and chain surface (#1993). Not introduced by it — this is long-standing behaviour that the new tests had to be written around.

## What it does

`getRackByPath` walks the path accumulating `currentRack`, and returns it at the end without checking what the final step actually was (`TrackManager.cpp:2425-2481`):

```cpp
RackInfo* TrackManager::getRackByPath(const ChainNodePath& rackPath) {
    RackInfo* currentRack = nullptr;
    ChainInfo* currentChain = nullptr;
    for (const auto& step : rackPath.steps) {
        // Rack steps assign currentRack; Chain steps assign currentChain
        // and leave currentRack where it was.
    }
    return currentRack;   // <- whatever rack was walked through last
}
```

So asking for a rack with a **chain** path succeeds and hands back that chain's parent:

```cpp
const auto rackPath  = ChainNodePath::rack(track, rack);
const auto chainPath = rackPath.withChain(chain);

getRackByPath(chainPath) == getRackByPath(rackPath);   // true, today
```

The same holds for a device path: `Rack > Chain > Device` returns the rack two levels up.

## Why it is worth an issue

It is a lookup that does not fail when it should, which is the failure mode that costs the most to find. A caller that passes the wrong path gets a plausible neighbour rather than `nullptr`, so the mistake surfaces later as a write landing on the wrong node rather than immediately as a null check.

It also caught us out concretely. `MockTrackApi` was written to the stricter rule — return nothing unless the last step names a rack — and agreed with itself and with the mock-driven tests. The live suite against the real `TrackManager` is what surfaced the disagreement, and the mock had to be made *less* correct to match. Both suites now pin the current behaviour with a comment saying it is surprising, in:

- `tests/test_track_api_nested_racks.cpp` — "An unresolvable path yields nothing rather than the wrong node"
- `tests/test_track_api_nested_racks_live_juce.cpp` — "An unresolvable path changes nothing"

## Why it was not fixed there

Out of scope for a facade change, and not obviously safe. `getRackByPath` is called from the UI, the chain editor, sidechain wiring, and preset application, and some of those may be passing a deeper path deliberately and relying on getting the enclosing rack. Tightening it is a behaviour change that needs its own audit.

## Scope

- [ ] Audit `getRackByPath` callers for ones that pass a non-rack path and depend on the leniency
[...]

PR #2058, opened 2026-08-08
TrackApi: address racks and chains by path (#1993)

### LINKED ISSUE #1993: MagdaApi: TrackApi rack and chain operations cannot address nested racks
Found while building the device catalogue for #1859 (#1992).

## Problem

Every rack and chain operation on `TrackApi` is addressed by a `(trackId, rackId, chainId)` triple:

```cpp
ChainId addChainToRack(TrackId, RackId, const juce::String& name);
void removeChainFromRack(TrackId, RackId, ChainId);
void setChainVolume(TrackId, RackId, ChainId, float);
void setChainOutput(TrackId, RackId, ChainId, int);
void setRackBypassed(TrackId, RackId, bool);
const ChainInfo* getChain(TrackId, RackId, ChainId) const;
// ...and the rest of the rack/chain surface
DeviceId addDeviceToChain(TrackId, RackId, ChainId, const DeviceInfo&);
```

A triple describes exactly one level of nesting. MAGDA's rack model nests arbitrarily — `Track > Rack > Chain > Rack > Chain > Device` — so nothing inside a nested rack is reachable through the facade at all. There is no triple that names the inner chain.

This is the same limitation the remote API had before #1991, where `AutomationTargetDto` and `DeviceDto` carried a device id and an immediate parent instead of a route.

## It is only the facade

`TrackManager` already has a path-based equivalent for every one of these, and they are the ones the UI uses:

| Triple-based on `TrackApi` | Path-based on `TrackManager` |
|---|---|
| `addChainToRack(t, r, name)` | `addChainToRack(rackPath, name)` |
| `removeChainFromRack(t, r, c)` | `removeChainByPath(chainPath)` |
| `getChain(t, r, c)` | `getChainByPath(chainPath)` |
| `setChainMuted/Bypassed/Solo/Volume/Pan(t, r, c, …)` | same names taking `chainPath` |
| `setRackBypassed(t, r, …)` / `setRackVolume` | `setRackBypassedByPath(rackPath, …)` / `setRackVolume(rackPath, …)` |
| `getRack(t, r)` | `getRackByPath(rackPath)` |
| `addDeviceToChain(t, r, c, device)` | `addDeviceToChainByPath(chainPath, device)` |

So this is a facade gap, not a model gap. Widening it is mostly surfacing what exists.

## Scope

- [ ] Add `ChainNodePath` overloads for the rack and chain surface on `TrackApi`, forwarding to the existing `TrackManager` path methods
- [ ] Decide the fate of the triple-based methods — keep as thin shims over the path form, or migrate call sites and remove
- [ ] Extend `MockMagdaApi` to model nesting deeper than one level, so the gap cannot silently return
- [ ] Facade tests covering a device and a chain inside a nested rack

## Why it matters here

#1859 requires "Support top-level and rack-chain placement" and #701 requires remote adapters to reach racks and chains through the facade rather than around it. With the triple-based surface, a remote caller can address a device in a nested rack — `DevicePathDto` handles that since #1991 — but cannot address the *chain* it sits in, so operations like "create a chain inside this rack" stop working at depth two.

## Out of scope

Device lifecycle and parameter writes, which are #1859.

lucaromagnoli:
Closes #1993.

## The gap

Every rack and chain operation on the facade took a `(trackId, rackId, chainId)` triple, which names exactly one level of nesting. The model nests arbitrarily — `Track > Rack > Chain > Rack > Chain > Device` — so nothing inside a nested rack was reachable through `TrackApi` at all: there is no triple that names the inner chain.

A caller could already address a *device* at depth, since `DevicePathDto` has carried a full route since #1991 — but not the chain containing it. So "add a chain to this rack" stopped working at depth two, and the remote API's `racks.create` still describes itself as creating a *top-level* rack.

## The change

`ChainNodePath` overloads for the whole rack and chain surface, forwarding to the `TrackManager` methods the UI has always used. Mostly surfacing what already existed.

The names deliberately mirror TrackManager's own (`getRackByPath`, `removeChainByPath`, `addDeviceToChainByPath`, and plain overloads where TrackManager overloads). This facade is a view onto it, and inventing a third vocabulary for the same calls would be one more mapping to hold in your head.

**Two of them had nothing to forward to.** The issue says TrackManager "already has a path-based equivalent for every one of these" — it did not. `setChainOutput` and `setChainName` were triple-only, so the facade could otherwise reach a nested chain's mute and volume but not its name or output, which is not a line anyone would have drawn on purpose. Both now have path forms on TrackManager, with the triple versions delegating so there is one body rather than two that can drift.

**The triples are kept as shims** at depth one rather than removed. The agent DSL is their only substantial consumer and its grammar is inherently one-rack-at-a-time; migrating it means changing that grammar, which is its own piece of work. `(t, r, c)` is `ChainNodePath::chain(t, r, c)` and the tests pin that equivalence.

## The mock was part of the problem

`MockTrackApi` stubbed every rack and chain method to `INVALID_*_ID` or `nullptr`. That agreed perfectly with a facade that could not reach depth — it had nothing to disagree with — so the gap was invisible from the tests. It now holds the same nested tree `TrackManager` holds and resolves paths the same way.

## Tests, and what the second suite caught

Two halves, and the second earned its place immediately:

- `test_track_api_nested_racks.cpp` — the facade's shape through the mock.
- `test_track_api_nested_racks_live_juce.cpp` — the same shapes through `TrackApiLive` against the real `TrackManager`.

The live suite exists because a mock I wrote passing tests I wrote proves the mock, not the forwarding. **It failed on first run.** `TrackManager::getRackByPath` returns the deepest rack it walked through regardless of the final step, so a *chain* path answers with that chain's parent rack rather than nothing — and I had written the mock to the tidier rule. The mock-only test had encoded my opinion rather than the model's behaviour.

The mock now matches the model, and both suites pin the real semantics with a comment saying it is surprising. `TrackManager` itself is unchanged: that leniency is long-standing, `getRackByPath` is called from the UI, the chain editor, sidechain wiring, and preset application, and some of those may pass a deeper path deliberately. Tracked as #2057.

## Verification

Full Catch2 suite green (2326 passed, 13 skipped). Both new suites green. `Remote Service Live` still green. Release build clean.

## Follow-ups, not in scope

- **#2057** — whether `getRackByPath` should reject a non-rack path, or be renamed to say what it does. Also covers the asymmetry with `getChainByPath`, which is strict.
- **The remote API is now narrower than the facade.** `racks.*` still takes ids and still says "top-level"; widening it to a `DevicePathDto`-style route is the natural next step and what #701 wanted. Separate PR.

### DIFF
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -2635,12 +2635,16 @@ const ChainInfo* TrackManager::getChain(TrackId trackId, RackId rackId, ChainId 
     return nullptr;
 }
 
+void TrackManager::setChainOutput(const ChainNodePath& chainPath, int outputIndex) {
+    if (auto* chain = getChainByPath(chainPath)) {
+        chain->outputIndex = outputIndex;
+        notifyTrackDevicesChanged(chainPath.trackId);
+    }
+}
+
 void TrackManager::setChainOutput(TrackId trackId, RackId rackId, ChainId chainId,
                                   int outputIndex) {
-    if (auto* chain = getChain(trackId, rackId, chainId)) {
-        chain->outputIndex = outputIndex;
-        notifyTrackDevicesChanged(trackId);
-    }
+    setChainOutput(ChainNodePath::chain(trackId, rackId, chainId), outputIndex);
 }
 
 void TrackManager::setChainMuted(TrackId trackId, RackId rackId, ChainId chainId, bool muted) {
@@ -2679,15 +2683,19 @@ void TrackManager::setChainPan(TrackId trackId, RackId rackId, ChainId chainId, 
     }
 }
 
-void TrackManager::setChainName(TrackId trackId, RackId rackId, ChainId chainId,
-                                const juce::String& name) {
-    if (auto* chain = getChain(trackId, rackId, chainId)) {
+void TrackManager::setChainName(const ChainNodePath& chainPath, const juce::String& name) {
+    if (auto* chain = getChainByPath(chainPath)) {
         auto trimmed = name.trim();
         if (trimmed.isEmpty() || trimmed == chain->name)
             return;
         chain->name = trimmed;
-        notifyTrackPropertyChanged(trackId);
+        notifyTrackPropertyChanged(chainPath.trackId);
     }
+}
+
+void TrackManager::setChainName(TrackId trackId, RackId rackId, ChainId chainId,
+                                const juce::String& name) {
+    setChainName(ChainNodePath::chain(trackId, rackId, chainId), name);
 }
 
 void TrackManager::setRackVolume(TrackId trackId, RackId rackId, float volume) {
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -2431,42 +2431,55 @@ RackInfo* TrackManager::getRackByPath(const ChainNodePath& rackPath) {
     RackInfo* currentRack = nullptr;
     ChainInfo* currentChain = nullptr;
 
+    // A step that resolves to nothing ends the walk. This used to leave the
+    // previous `currentRack` in place and carry on, which made a broken path
+    // resolve to whatever it had passed through — so `outer > missingChain >
+    // missingRack` answered with `outer`. Worse, a missed Chain step left
+    // `currentChain` null, so the *next* Rack step searched the track's
+    // top-level list again and `outer > missingChain > someOtherTopLevelRack`
+    // answered with that unrelated rack.
+    //
+    // Every path-based rack mutator resolves through here, so that was a write
+    // landing silently on a node the caller never named. Failing closed is the
+    // only answer that cannot do that.
+    //
+    // Deliberately unchanged: a path that resolves *completely* but ends on a
+    // Chain or Device step still returns the enclosing rack rather than
+    // nothing. That leniency is long-standing, callers may rely on it, and it
+    // is a different question — #2057.
     for (const auto& step : rackPath.steps) {
         switch (step.type) {
             case ChainStepType::Rack: {
-                if (currentChain == nullptr) {
-                    // Top-level rack in track's chainElements
-                    for (auto& element : track->chain.fxChainElements) {
-                        if (magda::isRack(element)) {
-                            if (magda::getRack(element).id == step.id) {
-                                currentRack = &magda::getRack(element);
-                                break;
-                            }
-                        }
-                    }
-                } else {
-                    // Nested rack within a chain
-                    for (auto& element : currentChain->elements) {
-                        if (magda::isRack(element)) {
-                            if (magda::getRack(element).id == step.id) {
-                                currentRack = &magda::getRack(element);
-                                currentChain = nullptr;  // Reset chain context
-                                break;
-                            }
-                        }
+                // Top-level racks live in the track's own FX list; a nested one
+                // lives in the chain the previous step reached.
+                auto& elements =
+                    currentChain != nullptr ? currentChain->elements : track->chain.fxChainElements;
+                RackInfo* found = nullptr;
+                for (auto& element : elements) {
+                    if (magda::isRack(element) && magda::getRack(element).id == step.id) {
+                        found = &magda::getRack(element);
+                        break;
                     }
                 }
+                if (found == nullptr)
+                    return nullptr;
+                currentRack = found;
+                currentChain = nullptr;  // Reset chain context
                 break;
             }
             case ChainStepType::Chain: {
-                if (currentRack != nullptr) {
-                    for (auto& chain : currentRack->chains) {
-                        if (chain.id == step.id) {
-                            currentChain = &chain;
-                            break;
-                        }
+                if (currentRack == nullptr)
+                    return nullptr;
+                ChainInfo* found = nullptr;
+                for (auto& chain : currentRack->chains) {
+                    if (chain.id == step.id) {
+                        found = &chain;
+                        break;
                     }
                 }
+                if (found == nullptr)
+                    return nullptr;
+                currentChain = found;
                 break;
             }
             case ChainStepType::Device:
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -2443,13 +2443,40 @@ RackInfo* TrackManager::getRackByPath(const ChainNodePath& rackPath) {
     // landing silently on a node the caller never named. Failing closed is the
     // only answer that cannot do that.
     //
+    // The structure has to hold as well as the ids. Racks contain chains and
+    // chains contain racks, so a route alternates `Rack > Chain > Rack > Chain`;
+    // a step that cannot follow the one before it describes a route the model
+    // has no way to express. Checking only that each id resolves is not enough,
+    // because two consecutive steps of the same kind then move sideways through
+    // the tree using ids that all genuinely exist:
+    //
+    //   rack(A).withRack(B)            — B is a *sibling* top-level rack, and
+    //                                    with `currentChain` null the second
+    //                                    Rack step searched the track list
+    //                                    again and found it.
+    //   rack(R).withChain(c1).withChain(c2)
+    //                                  — c2 is a sibling chain of c1 in the
+    //                                    same rack, reached by a route that
+    //                                    never went through anything.
+    //
+    // Both resolved, and every path-based rack mutator would then write to a
+    // node on the other side of the tree from the one the caller named.
+    //
     // Deliberately unchanged: a path that resolves *completely* but ends on a
     // Chain or Device step still returns the enclosing rack rather than
-    // nothing. That leniency is long-standing, callers may rely on it, and it
-    // is a different question — #2057.
-    for (const auto& step : rackPath.steps) {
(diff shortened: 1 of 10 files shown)


tests/MockMagdaApi.hpp
```diff
@@ -406,32 +406,275 @@ class MockTrackApi : public TrackApi {
...
+                auto& elements = chain != nullptr ? chain->elements : track->chain.fxChainElements;
+                rack = nullptr;
+                for (auto& element : elements) {
+                    if (magda::isRack(element) && magda::getRack(element).id == step.id) {
+                        rack = &magda::getRack(element);
+                        break;
+                    }
+                }
+                chain = nullptr;
+                if (rack == nullptr)
```
lucaromagnoli: [P1] Do not let broken paths fall back to an enclosing rack

This does not actually match `TrackManager::getRackByPath` for a failed intermediate step. The real resolver never clears `currentRack`/`currentChain` when an ID lookup misses, so a path such as `outerRack -> missingChain -> missingRack` returns `outerRack`; this mock returns `nullptr` at the first miss. Through the newly exposed live methods, `setRackVolume`, `setRackBypassedByPath`, or `addChainToRack` can therefore mutate the outer rack while these mock tests report that the invalid path changes nothing. This is separate from #2057’s successfully resolved chain-path leniency. Please make the real traversal fail closed on a missing step and add this deep invalid-path case to the live suite.

lucaromagnoli: Correct, and this one is a genuine bug rather than a documented quirk. Fixed in e0b11f15.

Tracing it through, it is worse than the enclosing-rack fallback you describe. A missed Chain step left `currentChain` null, so the *next* Rack step searched the track's top-level list again — meaning `outer > missingChain > someOtherTopLevelRack` resolved to that unrelated rack, on the other side of the tree entirely. Not just a parent, a stranger.

You are also right that this is mine to fix rather than to note. `setRackVolume`, `setRackBypassedByPath` and `addChainToRack` by path are new facade surface in this branch, so the branch grew the blast radius of a resolver that silently writes to a node the caller never named.

A step that resolves to nothing now ends the walk. I left the #2057 leniency alone deliberately: a path that resolves *completely* but ends on a Chain or Device step still returns the enclosing rack. That is a type mismatch on a successful resolution, which is a different question from a broken one, and callers may lean on it.

On the mock: it was right and the model was wrong. It always failed closed, which is exactly why the mock-driven tests could report "an invalid path changes nothing" while the real resolver was mutating the outer rack — the agreement between them was the lie, not the mock. Now that both fail closed the agreement means something.

Both suites gain the deep invalid-path case, including your `outer -> missingChain -> missingRack` shape and the nastier one where the trailing step names a rack that really exists elsewhere, plus assertions that `setRackVolume` / `setRackBypassedByPath` / `addChainToRack` through a broken path are no-ops.

Verified by reverting the resolver rather than trusting the new tests pass with it: four failures, including `Expected value within 0.001 of: 0, Actual value: -24` — the write landing on the outer rack.

Full suite (2326), all three live JUCE suites, and the release build green.

magda/daw/core/TrackManager.cpp
```diff
@@ -2431,42 +2431,55 @@ RackInfo* TrackManager::getRackByPath(const ChainNodePath& rackPath) {
...
-                            if (magda::getRack(element).id == step.id) {
-                                currentRack = &magda::getRack(element);
-                                currentChain = nullptr;  // Reset chain context
-                                break;
-                            }
-                        }
+                // Top-level racks live in the track's own FX list; a nested one
+                // lives in the chain the previous step reached.
+                auto& elements =
+                    currentChain != nullptr ? currentChain->elements : track->chain.fxChainElements;
```
lucaromagnoli, line 2491: [P1] Reject structurally impossible path steps too

The new miss checks still allow a broken path to cross the tree when each ID happens to exist. After resolving a Rack, `currentChain` is null, so a consecutive Rack step searches the track-level list again: `rackAPath.withRack(rackB)` resolves unrelated top-level `rackB`, and the newly exposed rack mutators will write to it. Consecutive Chain steps have the same problem because `currentRack` remains set, allowing a route to jump between sibling chains. These are invalid intermediate routes, not #2057’s deliberately retained terminal-type leniency, and the mock currently duplicates them. Please enforce the expected Rack/Chain alternation and add live cases with real sibling IDs.

lucaromagnoli: Correct on both shapes, and the ids-all-exist framing is the part I had missed — my previous fix only rejected a step that failed to *resolve*, not one that could not legally follow the step before it. Fixed in cb74d8cd.

The traversal now enforces the `Rack > Chain > Rack > Chain` alternation, treats a Device as a leaf that nothing may follow, and allows a Segment only as the leading step.

Worth flagging that the first attempt at this was still wrong, and your "the mock currently duplicates them" note is what made me check it rather than assume. `getChainByPath` needed the guard *separately*: it drops the final step and resolves the prefix, and for `rack > chain1 > chain2` that prefix is a perfectly legal route — so the search then found `chain2` sitting beside `chain1` in the same rack. Adding alternation to the rack walk alone left the sibling-chain case resolving exactly as before, and both suites failed on it. Both the model and the mock now require the step before a Chain to be a Rack.

Tests use real sibling ids throughout, so only the shape is wrong: `rack(A).withRack(B)` with B a real top-level rack, `rack(R).withChain(c1).withChain(c2)` with both chains real and in the same rack, a real device with a step after it, and a trailing Segment — plus assertions that `setRackVolume` / `setRackBypassedByPath` / `addChainToRack` through such a route are no-ops rather than writes to the sibling.

Still deliberately unchanged: a route that resolves *completely* but ends on a Chain or Device step returns the enclosing rack. That is a terminal-type leniency on a path that exists, which is #2057 rather than this.

Full Catch2 suite (2327 passed), the whole `magda_juce_tests` binary as CI runs it, and the release build green.

magda/daw/core/TrackManager.cpp
```diff
@@ -2431,49 +2431,98 @@ RackInfo* TrackManager::getRackByPath(const ChainNodePath& rackPath) {
...
                 }
+                if (found == nullptr)
+                    return nullptr;
+                currentChain = found;
                 break;
             }
             case ChainStepType::Device:
-                // Devices don't contain racks, skip
+                // A leaf. Devices contain nothing, so nothing may follow one.
+                if (!isLast)
```
lucaromagnoli: [P1] Resolve terminal steps instead of only checking their position

A Device leaf is never resolved here, so `rackAPath.withDevice(9999)` returns `rackA` even though a Device cannot directly follow a Rack and id 9999 need not exist; `rackA > validChain > missingDevice` does the same. Rack mutators can therefore still write through a route that does not exist. A leading Segment has a parallel hole: `Segment(PostFx) > Rack(realTopLevelRack)` is accepted, the segment is ignored, and the Rack lookup searches the main FX list. Neither route is #2057’s terminal-type leniency because neither resolves completely. Please require a Device to follow the current Chain and exist there, and either honor the selected Segment container or reject segmented paths from the rack resolver; mirror both cases in the mock/live tests.

magda/daw/core/TrackManager.cpp
lucaromagnoli: Follow-up to the above, found by auditing what the new rules would reject rather than by a further review round — but it is the same bug and worth flagging here rather than leaving in a commit message.

`TrackManagerDevices.cpp` carried a **second copy** of `getChainByPath`, and `getDeviceInChainByPath` used that one. So the structural guards landed in `getChainByPath` and not in the copy: `rack > chain1 > chain2` was refused by the model's own lookup while the device resolver still walked it into the sibling `chain2` and returned whatever device sat there.

That is the shape that matters most of the three. Rack and chain paths are built by typed helpers; **device** paths are the ones a remote client supplies verbatim, since `toChainNodePath` assembles them from whatever steps arrive on the wire. So this was the one route where a malformed path came from outside the process — and post-#1860 a client holding `edit` could reach a device in a part of the tree it never named.

Fixed in 44b8e35c by deleting the duplicate and delegating. Two copies of a resolver is how the rules diverged in the first place; teaching the second copy the same lesson would have left the next one to diverge again.

The live suite gains a device case — a real device in a real sibling chain, reachable by its own route and by neither the doubled-Chain nor the doubled-Rack one. Verified by restoring the duplicate: assertion 3 fails.

Full Catch2 (2327), the whole `magda_juce_tests` binary, and the release build green.

magda/daw/core/TrackManager.cpp
lucaromagnoli: Fixed in 2c2aef68.

`getRackByPath` now validates that a terminal Device follows the current Chain and names a real device before returning the enclosing rack, so the documented #2057 behavior remains for valid nested-device routes without accepting missing or misplaced leaves. Explicit Segment paths are rejected because post-fx and mixer-analysis sections are flat and cannot contain racks; contradictory track-level/top-level-device representations fail closed as well.

`MockTrackApi` mirrors the same rules. The mock and live suites now cover a valid nested Device, missing and misplaced Device leaves, `Segment(PostFx) > realRack`, and no-op writes through the invalid routes.

Local verification: 87 focused Catch2 assertions, 55 live JUCE assertions, and 214 existing TrackManager path assertions all pass. The prior Linux JUCE failure was an unrelated connection reset while downloading sccache; the new CI run is pending.

PR #2060, opened 2026-08-08
Native engine: warp markers, beat detection and loop info (#2038)

### LINKED ISSUE #2038: Arranger clips slice 5: warp markers, beat detection and loop info

lucaromagnoli:
Slice 5 of #1890, after slice 4. Closes #2038.

Where a clip's own musical time comes from, and where it stops agreeing with the file's.

## Warp adds nothing below the position map

Slice 4 turned playback into one function. The voice already asks `readingPositionAt` at both ends of a block and hands the difference to the stretcher, so a ratio that changes at every marker costs nothing that a moving auto-tempo ratio did not. `ClipVoice`, the stretchers, the pool, the stream and the reading chain are untouched by this slice.

Signalsmith has no warp-marker API and does not need one: `process(inputs, inputSamples, outputs, outputSamples)` takes the ratio per call. Tracktion does not push markers into a stretcher either; `AudioSegmentList` chops the clip into constant-ratio segments.

`clip/WarpMap` compiles the model's `(sourceTime, warpTime)` pairs into the inverse playback needs: sorted, strictly increasing on both sides, coincident markers collapsed, and anything that cannot be part of a monotonic map dropped at compile time with a counted diagnostic rather than divided by on the audio thread. The snapshot carries that instead of the raw marker vector. A binary search rather than a cursor, because a locate breaks sequential stepping and a reversed read walks the map the other way.

## Reverse and warp compose, which the incumbent cannot

The fork bakes warp into a rendered proxy file and can only bake one thing per clip: `WaveAudioClip::createRenderJob` checks `getIsReversed()` first and returns the reverse job before it ever reaches the warp one, so a reversed warped clip there plays reversed and unwarped and its markers are silently lost. That is a constraint of the proxy architecture, not a semantic, and computing both live has no such constraint.

The map is **not** mirrored at compile time. A reversed event walks it backwards from the far end of what it reads and mirrors the answer, exactly as an unwarped reversed event's anchor is mirrored. Mirroring the map instead would need the length of the region the event reads, which is itself an answer from the map, and the two would define each other.

Removing that circularity fixes something older with it: `regionOf` was the event's span at unity, so slice 4 already placed a reversed **stretched** clip's anchor with the wrong region length. It is now the exact material the event consumes, warp and rate included.

## A warped loop folds in warp time

The one case where the reading chain cannot do its own tiling. Folding below the stream works because the reading advances linearly, and under warp it does not: a position that had already been through the map would fold in the wrong domain, the map would go on extending at slope 1 past the loop's end, and every pass after the first would play straight.

So a warped loop folds above the map and `sourceReadFor` leaves the tiling below switched off. The reading saws back at each wrap instead of climbing, which the stream reads as a seek: one per pass, bounded. This is reachable rather than theoretical, since a session clip dropped from the media DB with saved markers both loops and warps.

Stretcher sizing reads the map's steepest segment rather than its average, because a warped event has no single rate.

## The two analyses

**`analysis/TransientDetector`** is the fork's detector reproduced coefficient for coefficient: a normalise pass, two one-pole followers, a differentiator, a third follower, a threshold at `-10 - sensitivity * 30` dB, a retrigger lockout and a spacing rule. Time domain, not spectral. Reproduced rather than improved, because a detector that found different transients would move every auto-detected marker in every project that already has one.

**`io/SourceLoopInfo`** is not an analysis at all. The fork reads a source's tempo and beat count out of `juce::AudioFormatReader::metadataValues` -- the acid chunk that JUCE's `WavAudioFormat` already parses -- so this is a parse over a metadata map, testable with no fixture and no format registered. Every field is optional because the model's seeding rule (never overwrite what the user set) depends on telling a silence from a value.

Neither is wired to the model yet. `WarpMarkerManager` and the clip model still get both from Tracktion; swapping them over belongs with switching the engine on rather than with implementing it.

## Tests

`tests/test_clip_warp.cpp`, `tests/test_transient_detector.cpp`, `tests/test_source_loop_info.cpp`. All arithmetic, against numbers worked out by hand rather than against another run of the same code, and against computed readers rather than fixtures.

Full suite: 2305 cases, 560k assertions, 13 pre-existing Spleeter model-file skips.

### DIFF
--- a/magda/engine/analysis/TransientDetector.cpp
+++ b/magda/engine/analysis/TransientDetector.cpp
@@ -144,7 +144,15 @@ std::vector<double> detectTransients(AudioFileReader& reader,
     // from. Placing the marker where the sound starts rather than where the
     // detector noticed is the difference between a warped beat landing on the
     // grid and landing just behind it.
-    const auto rewind = static_cast<int>(sampleRate * 0.0005);
+    //
+    // Rounded up, which is not a preference. The incumbent truncates after the
+    // subtraction (`int (i - sampleRate * 0.0005)`), and for a whole-numbered i
+    // that is the same as subtracting the rounded-up rewind: at 44100 its
+    // effective rewind is 23 samples, not the 22 that truncating the rewind on
+    // its own would give. Reproducing this detector exactly is the whole
+    // argument for having written it this way, and one sample is a marker that
+    // moved.
+    const auto rewind = static_cast<int>(std::ceil(sampleRate * 0.0005));
 
     int countdown = 0;
 
--- a/magda/engine/analysis/TransientDetector.cpp
+++ b/magda/engine/analysis/TransientDetector.cpp
@@ -1,0 +1,184 @@
+#include "analysis/TransientDetector.hpp"
+
+#include <juce_audio_basics/juce_audio_basics.h>
+
+#include <algorithm>
+#include <cmath>
+#include <cstdint>
+
+namespace magda::engine {
+
+namespace {
+
+/// The incumbent reads in chunks of this, and where a trigger's rewind lands is
+/// measured from the start of one. Matching it is what keeps a detected marker
+/// in the same place rather than half a millisecond from it.
+constexpr int kBlockSamples = 32768;
+
+/// A one-pole follower with separate attack and release, where 1 is instant and
+/// 0 never moves. The incumbent's, coefficient for coefficient.
+class EnvelopeFollower {
+  public:
+    void setCoefficients(float attack, float release) noexcept {
+        attack_ = attack;
+        release_ = release;
+    }
+
+    void process(float* samples, int numSamples) noexcept {
+        for (int i = 0; i < numSamples; ++i) {
+            const auto in = std::abs(samples[i]);
+
+            if (envelope_ < in)
+                envelope_ += attack_ * (in - envelope_);
+            else if (envelope_ > in)
+                envelope_ -= release_ * (envelope_ - in);
+
+            samples[i] = envelope_;
+        }
+    }
+
+  private:
+    float envelope_ = 0.0f;
+    float attack_ = 1.0f;
+    float release_ = 1.0f;
+};
+
+/// Sample-to-sample difference. What turns a smoothed envelope into something
+/// that peaks where the envelope was climbing fastest, which is the onset.
+class Differentiator {
+  public:
+    void process(float* samples, int numSamples) noexcept {
+        for (int i = 0; i < numSamples; ++i) {
+            const auto current = samples[i];
+            samples[i] = current - last_;
+            last_ = current;
+        }
+    }
+
+  private:
+    float last_ = 0.0f;
+};
+
+/// The file's own peak, which the threshold is relative to. A quiet recording
+/// and a loud one have the same transients, and a fixed threshold would find
+/// them only in the second.
+float peakOf(AudioFileReader& reader, juce::AudioBuffer<float>& block, std::int64_t totalSamples) {
+    float peak = 0.0f;
+
+    for (std::int64_t at = 0; at < totalSamples; at += kBlockSamples) {
+        const auto count =
+            static_cast<int>(std::min<std::int64_t>(kBlockSamples, totalSamples - at));
+
+        if (reader.read(block, 0, at, count) <= 0)
+            continue;
+
+        const auto range =
+            juce::FloatVectorOperations::findMinAndMax(block.getReadPointer(0), count);
+        peak = std::max(peak, std::max(std::abs(range.getStart()), std::abs(range.getEnd())));
+    }
+
+    return peak;
+}
+
+/// Thin @p transients until none are closer together than @p spacing.
+///
+/// Backwards, keeping the later of any pair, and repeated because one pass can
+/// leave a new pair adjacent where it removed what sat between them. The
+/// incumbent gives up after ten passes and so does this: what it is thinning is
+/// a detector's own retriggers, and a file that still has them after ten rounds
+/// has something the spacing rule was never going to fix.
+void thin(std::vector<double>& transients, double spacing) {
+    if (transients.size() < 2)
+        return;
+
+    for (int pass = 0; pass < 10; ++pass) {
+        const auto before = transients.size();
+        auto last = transients.back();
+
+        for (int i = static_cast<int>(transients.size()) - 2; i >= 0; --i) {
+            const auto at = transients[static_cast<std::size_t>(i)];
+
+            if (last - at < spacing)
+                transients.erase(transients.begin() + i);
+            else
+                last = at;
+        }
+
+        if (transients.size() == before)
+            return;
+    }
+}
+
+}  // namespace
+
+std::vector<double> detectTransients(AudioFileReader& reader,
+                                     const TransientDetectionSettings& settings) {
+    std::vector<double> transients;
+
+    const auto sampleRate = reader.sampleRate();
+    const auto totalSamples = reader.lengthInSamples();
+
+    if (!(sampleRate > 0.0) || totalSamples <= 0)
+        return transients;
+
+    // One channel, which is what the reader hands the file's first through and
+    // what the incumbent detects on. A transient is in every channel of a
+    // recording that has one.
+    juce::AudioBuffer<float> block(1, kBlockSamples);
+
+    const auto peak = peakOf(reader, block, totalSamples);
+    const auto scale = peak > 0.0f ? 1.0f / peak : 1.0f;
+
+    EnvelopeFollower followers[3];
+    for (auto& follower : followers)
+        follower.setCoefficients(1.0f, 0.002f);
+
+    Differentiator differentiator;
+
+    const auto threshold = juce::Decibels::decibelsToGain(
+        -10.0f - std::clamp(settings.sensitivity, 0.0f, 1.0f) * 30.0f);
+    const auto lockout = static_cast<int>(sampleRate * settings.retriggerSeconds);
+
+    // Half a millisecond, because what crosses the threshold is the differential
+    // of a smoothed envelope and that peaks a little after the attack it came
+    // from. Placing the marker where the sound starts rather than where the
+    // detector noticed is the difference between a warped beat landing on the
+    // grid and landing just behind it.
+    const auto rewind = static_cast<int>(sampleRate * 0.0005);
+
+    int countdown = 0;
+
+    for (std::int64_t at = 0; at < totalSamples; at += kBlockSamples) {
+        const auto count =
+            static_cast<int>(std::min<std::int64_t>(kBlockSamples, totalSamples - at));
+
+        reader.read(block, 0, at, count);
+
+        auto* samples = block.getWritePointer(0);
+        juce::FloatVectorOperations::multiply(samples, scale, count);
+
+        followers[0].process(samples, count);
+        followers[1].process(samples, count);
+        differentiator.process(samples, count);
+        followers[2].process(samples, count);
+
+        for (int i = 0; i < count; ++i) {
+            if (countdown > 0)
+                --countdown;
+
+            if (samples[i] > threshold) {
+                if (countdown == 0)
+                    transients.push_back(static_cast<double>(at + std::max(0, i - rewind)) /
+                                         sampleRate);
+
+                countdown = lockout;
+            }
+        }
+    }
+
+    thin(transients, settings.minimumSpacingSeconds);
+
+    return transients;
+}
+
+}  // namespace magda::engine
--- a/magda/engine/clip/ClipSnapshot.hpp
+++ b/magda/engine/clip/ClipSnapshot.hpp
@@ -4,6 +4,7 @@
 #include <string>
 #include <vector>
 
+#include "clip/WarpMap.hpp"
 #include "core/ClipInfo.hpp"
 #include "core/TypeIds.hpp"
 
@@ -118,8 +119,12 @@ struct AudioEventPlayback {
     /// has already forced a stretcher on (AudioEvent::isAnalogPitchActive).
     bool analogPitch = false;
 
-    bool warpEnabled = false;
-    std::vector<WarpMarker> warpMarkers;
+    /// Compiled and inverted, never the model's marker list (WarpMap.hpp).
+    /// Empty is identity, which is both "warp is off" and "warp is on and there
+    /// are no markers": the two describe the same playback and nothing
+    /// downstream has to tell them apart. Already mirrored where the event is
+    /// reversed, so the audio thread does one lookup whichever way it plays.
+    WarpMap warp;
 
     bool autoPitch = false;
     int autoPitchMode = 0;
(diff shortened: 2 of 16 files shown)


magda/engine/clip/ClipSnapshot.hpp
```diff
@@ -118,8 +119,12 @@ struct AudioEventPlayback {
     /// has already forced a stretcher on (AudioEvent::isAnalogPitchActive).
     bool analogPitch = false;
 
-    bool warpEnabled = false;
-    std::vector<WarpMarker> warpMarkers;
+    /// Compiled and inverted, never the model's marker list (WarpMap.hpp).
+    /// Empty is identity, which is both "warp is off" and "warp is on and there
+    /// are no markers": the two describe the same playback and nothing
+    /// downstream has to tell them apart. Already mirrored where the event is
```
lucaromagnoli: This contradicts the implementation and the PR description: the map is not mirrored at compile time. compileWarpMap never sees the reversed flag; reverse is resolved per lookup in EventPlacement's warpedReadingSample, which walks the map backwards from the far end and mirrors the answer. Reads like a leftover from an earlier draft where mirroring was the plan; worth rewording before someone builds on it.

lucaromagnoli (commented):
Slice 5 read against the incumbent it reproduces. The map compile, the placement arithmetic, the reversed walk and the warp-time loop fold all check out against the hand-worked values in the tests, and the detector matches TE's coefficient for coefficient with one exception noted inline (the rewind truncation, which moves every detected marker by one sample at 44.1k-family rates). The loop-info parse's swapped time-signature fields turn out to be faithful to tracktion_LoopInfo.cpp rather than a typo, which deserves saying in the code. Two comments in the snapshot layer still describe an earlier design where the map was mirrored at compile time. Nothing blocking.

magda/engine/clip/ClipSnapshotCompiler.cpp
```diff
@@ -230,8 +231,32 @@ ClipSnapshot compileClipSnapshot(const std::vector<ClipLane>& lanes,
                 playback.speedRatio = event.speedRatio;
                 playback.timeStretchMode = event.getEffectiveTimeStretchMode();
                 playback.analogPitch = event.isAnalogPitchActive();
-                playback.warpEnabled = event.warpEnabled;
-                playback.warpMarkers = event.warpMarkers;
+                // Inverted, mirrored where the event is reversed, and known to
```
lucaromagnoli: Same stale claim as on ClipSnapshot.hpp's warp field: nothing here mirrors the map for a reversed event. compileWarpMap only sorts, validates and drops; reverse stays a read-time coordinate change in EventPlacement.

magda/engine/analysis/TransientDetector.cpp
```diff
@@ -0,0 +1,184 @@
...
+    const auto threshold = juce::Decibels::decibelsToGain(
+        -10.0f - std::clamp(settings.sensitivity, 0.0f, 1.0f) * 30.0f);
+    const auto lockout = static_cast<int>(sampleRate * settings.retriggerSeconds);
+
+    // Half a millisecond, because what crosses the threshold is the differential
+    // of a smoothed envelope and that peaks a little after the attack it came
+    // from. Placing the marker where the sound starts rather than where the
+    // detector noticed is the difference between a warped beat landing on the
+    // grid and landing just behind it.
+    const auto rewind = static_cast<int>(sampleRate * 0.0005);
```
lucaromagnoli: One sample off the incumbent at 44.1k-family rates. TE truncates after the subtraction (`int (i - sampleRate * 0.0005)` in tracktion_WarpTimeManager.cpp), so at 44100 the effective rewind is 23 samples; truncating the rewind first gives 22. Since the argument for reproducing this detector verbatim is that an auto-detected marker must not move, this moves every marker on a 44.1k file by one sample.

`static_cast<int>(std::ceil(sampleRate * 0.0005))` matches TE at every rate (for integer i and r > 0, `trunc(i - r) == i - ceil(r)` whenever the result is non-negative, integer r included), and kRewindSamples in test_transient_detector.cpp becomes 23.

magda/engine/io/SourceLoopInfo.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (!info.bpm)
+        info.bpm = doubleOf(metadata, "tempo");
+
+    if (!info.numBeats)
+        info.numBeats = doubleOf(metadata, "beat count");
+
+    if (const auto timeSignature = valueOf(metadata, "time signature")) {
+        if (!info.denominator)
+            info.denominator =
+                timeSignature->upToFirstOccurrenceOf("/", false, false).getIntValue();
```
lucaromagnoli: Deliberately bug-compatible, I assume: tracktion_LoopInfo.cpp does the same swap (denominator from before the slash, numerator after), and the test's 6/8 section codifies it. Without a comment saying so, both read as typos and will eventually get "fixed" into disagreeing with the incumbent. Worth a line here citing the TE parse.

Two other divergences from the same TE path worth confirming as intentional: TE also derives a root note from the "key signature" key here, which this drops; and TE's beat-count branch computes bpm as `beats / duration / 60` (a genuine TE bug, seeding 0.03 bpm where this seeds 120), which this quietly corrects.

magda/engine/clip/WarpMap.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    auto& points = result.map.points;
+    points.reserve(sorted.size());
+
+    for (const auto& point : sorted) {
+        // Strictly increasing on both sides. Equal on either is a segment with
+        // no span, which is an infinite rate one way and a zero-divide the
+        // other; decreasing is a map with no inverse. Both are the user having
+        // dragged a marker past its neighbour, and both cost that marker.
+        if (!points.empty() && (point.sourceSeconds <= points.back().sourceSeconds ||
```
lucaromagnoli: Minor, and only about which marker survives: this greedy pass keeps the earlier of a crossing pair, so a marker whose warp time was dragged up past several neighbours survives while each neighbour it cleared is dropped, and the diagnostic counts N markers rather than 1. The header's "a user who dragged one marker past its neighbour should lose that marker" describes the drag-down case only. Fine as a defensive compile, but the doc and the drop rule currently disagree about who loses.

tests/test_clip_warp.cpp
```diff
@@ -0,0 +1,334 @@
...
+ */
+std::vector<magda::WarpMarker> markers() {
+    return {{0.0, 0.0}, {2.0, 1.0}, {3.0, 3.0}};
+}
+
+/**
+ * @brief A clip whose interpretation makes a warp second a timeline second.
+ *
+ * The event is on the beat face, as every warped event is, so its elapsed is
+ * beats over its own bpm. Interpreting the file at the timeline's own tempo
```
lucaromagnoli, line 79: The rig makes a warp second equal a timeline second (interpBpm equal to the timeline's tempo), which keeps every expectation readable but also makes the beat-face conversion in readingPositionAt a multiply by one. One case with interpBpm differing from the timeline tempo (a 60 bpm interpretation under this 120 bpm timeline, say) would catch a units slip between beats, warp seconds and source seconds that this rig cannot. Reverse and loop together is the other untested composition; each is covered alone.

magda/engine/io/SourceLoopInfo.cpp
```diff
@@ -0,0 +1,90 @@
...
+SourceLoopInfo loopInfoFrom(const juce::StringPairArray& metadata, double sampleRate,
+                            std::int64_t lengthInSamples) {
+    SourceLoopInfo info;
+
+    // ---- The acid chunk, which is what loop libraries write ----------------
+
+    info.numBeats = doubleOf(metadata, juce::WavAudioFormat::acidBeats);
+    info.numerator = intOf(metadata, juce::WavAudioFormat::acidNumerator);
+    info.denominator = intOf(metadata, juce::WavAudioFormat::acidDenominator);
+    info.bpm = doubleOf(metadata, juce::WavAudioFormat::acidTempo);
```
lucaromagnoli: **[P2] Zero-valued ACID fields block the intended inference for real reader metadata.** JUCE emits both acid tempo and acid beats keys for every ACID chunk, including values of 0. That engages these optionals, so the fallback below only tests presence and will not derive BPM from positive beats when tempo is zero, or beats from positive tempo when beats is zero. The current tests omit the opposite key, unlike the map produced by JUCE. Please treat non-positive values as unspecified for inference and add a test with both keys present.

magda/engine/clip/WarpMap.cpp
lucaromagnoli: **[P2] This greedy filter can discard an otherwise valid suffix.** For source-sorted warp values [0, 100, 1, 2, 3], it keeps 100 and drops the next three markers, although dropping only 100 yields a valid monotonic map. That contradicts the stated behavior that an offending marker costs only itself and can radically change the remaining warp. Please select a maximal monotonic subset with deterministic tie-breaking, or otherwise resolve isolated conflicts without always privileging the previous point.

magda/engine/clip/ClipSnapshotCompiler.cpp
```diff
@@ -230,8 +231,32 @@ ClipSnapshot compileClipSnapshot(const std::vector<ClipLane>& lanes,
...
+                // be invertible before anything plays it (WarpMap.hpp). A
+                // marker that cannot be part of a monotonic map is dropped
+                // here; saying how many is the difference between a clip whose
+                // timing is not what its editor shows and one that says so.
+                if (event.warpEnabled) {
+                    auto warp = compileWarpMap(event.warpMarkers);
+                    playback.warp = std::move(warp.map);
```
lucaromagnoli, line 249: **[P1] Preserve warpEnabled separately from the compiled map.** An empty WarpMap currently represents both warp disabled and warp enabled with no valid markers, but the model treats the latter as identity mapping that still follows interpBpm and forces a stretcher. Downstream usesBeatFace and followsTempo check map emptiness, so a warp-enabled clip with zero markers, or with every marker dropped, plays using the ordinary seconds-based rate. Please carry the enabled flag in AudioEventPlayback and use it for musical-time and stretcher decisions while leaving the empty map as the identity lookup.

magda/engine/io/SourceLoopInfo.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    if (!info.bpm)
+        info.bpm = doubleOf(metadata, "tempo");
+
+    if (!info.numBeats)
+        info.numBeats = doubleOf(metadata, "beat count");
+
+    if (const auto timeSignature = valueOf(metadata, "time signature")) {
+        if (!info.denominator)
+            info.denominator =
```
lucaromagnoli, line 93: **[P2] Slash-form time signatures are parsed backwards.** JUCE writes a time signature such as 6/8 as numerator/denominator, but this assigns the text before the slash to denominator and the text after it to numerator, producing 8/6. The new test currently codifies the reversed result. Please assign the left side to numerator and the right side to denominator.

magda/engine/clip/WarpMap.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    // By source time, because that is the side the model's own map is a
+    // function of. Storage order is not guaranteed either way
+    // (AudioEvent::warpedSourceSeconds says so), so this is a sort rather than
+    // a check.
+    std::sort(sorted.begin(), sorted.end(), [](const WarpMap::Point& a, const WarpMap::Point& b) {
+        return a.sourceSeconds < b.sourceSeconds;
+    });
+
+    // Coincident on the source side first. Two markers on one instant of the
```
lucaromagnoli: "The earlier one is kept so that which survives does not depend on the order they were stored in" is not quite delivered: std::sort is unstable and the comparator orders on sourceSeconds alone, so for two markers sharing a source time, which lands first after the sort is unspecified and can depend on the input order. It also matters to the LIS which one survives: with (0,0), (1,50), (1,2), (2,3), keeping (1,2) preserves three points while keeping (1,50) preserves two. Tie-breaking the comparator on warpSeconds fixes both at once - the sort becomes fully determined, and the collapse then keeps the smallest-warp duplicate, which never shortens the subsequence (any chain through the larger duplicate admits the smaller one in its place).

magda/engine/io/SourceLoopInfo.cpp
```diff
@@ -0,0 +1,115 @@
...
+    info.denominator = positiveIntOf(metadata, juce::WavAudioFormat::acidDenominator);
+    info.bpm = positiveDoubleOf(metadata, juce::WavAudioFormat::acidTempo);
+
+    if (const auto oneShot = valueOf(metadata, juce::WavAudioFormat::acidOneShot))
+        info.oneShot = *oneShot == "1";
+
+    // The chunk carries a root note and a flag saying whether it means
+    // anything. A note that has not been marked as set is not a note, which is
+    // the chunk's own rule rather than a caution taken here.
+    if (metadata[juce::WavAudioFormat::acidRootSet] == "1")
```
lucaromagnoli, line 72: positiveIntOf here quietly adds a filter the chunk's own rule does not have: the incumbent's acid path accepts the root note whenever acidRootSet is 1, zero included, while this now reads note 0 (C-1) as absent - and the comment above still claims the chunk's rule is the only thing deciding. Root note 0 in real material is vanishingly rare, so either way is defensible, but the comment and the behaviour should agree: keep the raw intOf under the flag, or say the zero filter is deliberate the way the MidiUnityNote comment below does.

lucaromagnoli (commented):
Second pass over the review-fixes commit. The rewind parity, the rewritten snapshot comments, the warpEnabled flag and its wiring, the LIS compile and the new tests all check out; I verified the LIS reconstruction (chain length provably equals tails.size(), strict on the warp side via lower_bound), the ceil identity against TE at 44.1k, 48k, 88.2k and 22.05k, and the new test expectations by hand. The voice pool composes with a marker edit correctly too: reader identity is cue plus setup, and per-block positions come fresh from the swapped snapshot. Two small residues inline.

PR #2061, opened 2026-08-08
OSC control surfaces: input path (#1757)

lucaromagnoli:
First slice of #1757. MAGDA listens on UDP and a TouchOSC-style template
drives it, with no mapping step. One-way: feedback, OSC bindings/learn, the
settings UI and the shipped template are later slices.

`juce_osc` was already in the JUCE submodule and just not in the link line,
so this adds no third-party dependency.

## The namespace

Parsed strictly. A wildcard where a track number belongs, a leading zero, an
out-of-range index, or a trailing component are all rejected rather than
coerced — OSC address *patterns* are a real part of the protocol, and quietly
reading one as an index would mean a surface driving every strip at once when
its author meant one.

Track numbers are mixer positions, not `TrackId`s. IDs survive a reload but
are sparse — delete tracks 2 and 3 and the rest are 1, 4, 5 — and a template's
eight faders are not. A position with no track behind it is ignored, so a
sixteen-strip template on an eight-track project drives eight faders rather
than failing.

Momentary buttons and toggles are distinguished, because surfaces send both
down one address: a `Trigger` ignores a zero argument (the release half of a
press), a `Toggle` treats it as a state and flips when sent no argument at
all. Without that, releasing Play would stop the transport.

## Why it coalesces rather than queues

A tablet mixer sends continuously while a finger is down — 100 Hz per control
is the normal case, and eight strips moving at once is a chord change. Posting
each message would put the UI thread on the far end of a network firehose, and
posting them in order would leave the user's last fader position waiting behind
several hundred stale ones.

So the receive thread publishes values instead of posting messages. Every
address owns a preallocated slot, a store is one relaxed write plus a dirty
bit, and the message thread drains the whole table at once. At most one drain
is ever outstanding whatever the rate; the receive thread allocates nothing and
takes no lock, so a blocked message thread makes MAGDA coarser, never the
network thread slower. The tradeoff is that values are not applied in arrival
order *across* addresses, which is why the namespace carries levels and states
rather than anything sequenced.

## Two paths into the model, deliberately

Levels go through `ControllerParamWriter` — the writer the MIDI surfaces
already use — so an OSC fader, a MIDI fader and an automation curve at the same
position produce the same gain, and its host-write path keeps working when an
LFO or macro is already driving the target. Everything else goes through
`MagdaApi`, which routes states the way the on-screen controls do; play in
particular dispatches through the same playhead-aware path as the Play button.

Notably it does **not** route through `RemoteApiService`. That layer guarantees
one undo step per mutation, which is right for scripted edits and wrong here —
a fader sweep would become a hundred undo entries.

## Security posture

OSC is unauthenticated UDP; the protocol has no notion of a credential. So the
bind address is the whole of the access control and is a setting rather than a
constant, and the service binds its own `DatagramSocket` instead of calling
`OSCReceiver::connect`, which binds every interface unconditionally. Off by
default, and a failed bind fails closed rather than falling back to every
interface. Only control values and transport are exposed — no file or project
operations.

## Incidental fix

`SendLevel` shared a bare `break` with `Tempo` in
`DefaultControllerParamWriter::write`, so a resolved send target went through
the whole binding pipeline and did nothing. Same gap
`test_controller_track_level_write_juce.cpp` was written about when the
track-level kinds had it. OSC is the first caller that needs it, so it is
wired, with a test in that file. `Tempo` keeps the fall-through — OSC sets
tempo through `ProjectApi`, where the project's limits live.

## Testing

- `magda_tests`: the address grammar, the coalescer, and the sink against
  `MockMagdaApi`.
- `magda_juce_tests`: socket lifecycle and real datagrams through a live
  receive thread, plus the send-level writer against the shared engine.

Two bugs were found and are now guarded, each verified by reintroducing it and
watching the test fail:

- the drain released its scheduled flag *after* walking the table, stranding
  any value published mid-drain — for a fader just released, the wrong value
  stays on screen until it is touched again;
- the rebind guard compared the configured port against the *bound* one, so an
  ephemeral port rebound on every unrelated config save.

Bundles are unpacked, which is not optional: JUCE hands realtime listeners the
bundle rather than its messages, so implementing only the message callback goes
silent against any surface that bundles — and many do by default.

**Not verified:** no run against a real surface. OSC is off by default and
there is no settings UI until a later slice, so that needs a hand-edited config
file.

### DIFF
--- a/magda/daw/api/osc_command_sink_live.cpp
+++ b/magda/daw/api/osc_command_sink_live.cpp
@@ -2,6 +2,7 @@
 
 #include "../audio/controllers/ControllerParamWriter.hpp"
 #include "../core/ControlTarget.hpp"
+#include "../core/MixerStripOrder.hpp"
 #include "../core/TrackInfo.hpp"
 #include "focused_api.hpp"
 #include "magda_api.hpp"
@@ -26,23 +27,16 @@ OscCommandSinkLive::~OscCommandSinkLive() = default;
 // ============================================================================
 
 TrackId OscCommandSinkLive::trackAtPosition(int position) const {
-    if (position < 1)
-        return INVALID_TRACK_ID;
-
-    // Mixer visibility rather than the raw track list: the strips the user can
-    // see are the ones a mixer template is counting, and a hidden track would
-    // otherwise consume a fader that appears to do nothing. This is the same
-    // filter `TrackManager::getVisibleTracks` applies, walked through the
-    // facade so the sink has no second route into the model — and so it stops
-    // at the strip it wants rather than building the whole list to index once.
-    int seen = 0;
-    for (const auto& track : api_.tracks().getTracks()) {
-        if (!track.isVisibleIn(ViewMode::Mix))
-            continue;
-        if (++seen == position)
-            return track.id;
-    }
-    return INVALID_TRACK_ID;
+    // The mixer's own rule for what a strip is, not a second approximation of
+    // it: a hidden track, an aux return, or the child of a collapsed group all
+    // take no position, and getting any of them wrong lands a fader on the
+    // strip beside the one the user is looking at.
+    //
+    // Always ViewMode::Mix, even when the user is looking at another view. A
+    // surface's fader 3 must not become a different track because someone
+    // switched to the arrangement — the numbering a template was built against
+    // is the mixer's.
+    return mixerStripAtPosition(api_.tracks().getTracks(), ViewMode::Mix, position);
 }
 
 int OscCommandSinkLive::sendBusForPosition(TrackId trackId, int position) const {
--- a/magda/daw/api/osc_command_sink_live.cpp
+++ b/magda/daw/api/osc_command_sink_live.cpp
@@ -1,0 +1,148 @@
+#include "osc_command_sink_live.hpp"
+
+#include "../audio/controllers/ControllerParamWriter.hpp"
+#include "../core/ControlTarget.hpp"
+#include "../core/TrackInfo.hpp"
+#include "focused_api.hpp"
+#include "magda_api.hpp"
+#include "project_api.hpp"
+#include "track_api.hpp"
+#include "transport_api.hpp"
+
+namespace magda {
+
+using osc::OscCommand;
+using osc::OscCommandKind;
+
+OscCommandSinkLive::OscCommandSinkLive(MagdaApi& api, std::unique_ptr<ControllerParamWriter> writer)
+    : api_(api), writer_(std::move(writer)) {
+    jassert(writer_ != nullptr);
+}
+
+OscCommandSinkLive::~OscCommandSinkLive() = default;
+
+// ============================================================================
+// Addressing
+// ============================================================================
+
+TrackId OscCommandSinkLive::trackAtPosition(int position) const {
+    if (position < 1)
+        return INVALID_TRACK_ID;
+
+    // Mixer visibility rather than the raw track list: the strips the user can
+    // see are the ones a mixer template is counting, and a hidden track would
+    // otherwise consume a fader that appears to do nothing. This is the same
+    // filter `TrackManager::getVisibleTracks` applies, walked through the
+    // facade so the sink has no second route into the model — and so it stops
+    // at the strip it wants rather than building the whole list to index once.
+    int seen = 0;
+    for (const auto& track : api_.tracks().getTracks()) {
+        if (!track.isVisibleIn(ViewMode::Mix))
+            continue;
+        if (++seen == position)
+            return track.id;
+    }
+    return INVALID_TRACK_ID;
+}
+
+int OscCommandSinkLive::sendBusForPosition(TrackId trackId, int position) const {
+    const auto* track = api_.tracks().getTrack(trackId);
+    if (track == nullptr || position < 1 || position > static_cast<int>(track->sends.size()))
+        return -1;
+    return track->sends[static_cast<size_t>(position - 1)].busIndex;
+}
+
+bool OscCommandSinkLive::resolveToggle(float value, bool current) {
+    return value == osc::kOscToggleRequest ? !current : value != 0.0f;
+}
+
+void OscCommandSinkLive::writeLevel(const ControlTarget& target, float value) {
+    // The writer takes a resolved target; these are already concrete, so there
+    // is nothing for the alias/resolver machinery to do first.
+    ResolveResult resolved;
+    resolved.target = target;
+    resolved.resolved = true;
+    writer_->write(resolved, value);
+}
+
+// ============================================================================
+// Applying
+// ============================================================================
+
+void OscCommandSinkLive::apply(const OscCommand& command, float value) {
+    switch (command.kind) {
+        case OscCommandKind::TransportPlay:
+            api_.transport().play();
+            return;
+        case OscCommandKind::TransportStop:
+            api_.transport().stop();
+            return;
+        case OscCommandKind::TransportRecord:
+            api_.transport().setRecording(resolveToggle(value, api_.transport().isRecording()));
+            return;
+        case OscCommandKind::TransportLoop:
+            api_.transport().setLoopEnabled(resolveToggle(value, api_.transport().isLoopEnabled()));
+            return;
+        case OscCommandKind::TransportTempo:
+            // The facade clamps against the project's real tempo range; a
+            // surface sending 0 or 5000 is not this layer's to reinterpret.
+            api_.project().setTempo(value);
+            return;
+        case OscCommandKind::TransportPosition:
+            api_.transport().setPositionBeats(value);
+            return;
+
+        case OscCommandKind::MasterVolume:
+            writeLevel(ControlTarget::trackVolume(MASTER_TRACK_ID), value);
+            return;
+        case OscCommandKind::MasterPan:
+            writeLevel(ControlTarget::trackPan(MASTER_TRACK_ID), value);
+            return;
+
+        case OscCommandKind::FocusedMacro:
+            // OSC numbers macros from 1, the macro array from 0. With nothing
+            // focused this is a no-op inside the facade, which is the right
+            // answer for a knob the user has not pointed at anything yet.
+            api_.focused().setMacroValue(command.index - 1, value);
+            return;
+
+        case OscCommandKind::TrackVolume:
+        case OscCommandKind::TrackPan:
+        case OscCommandKind::TrackMute:
+        case OscCommandKind::TrackSolo:
+        case OscCommandKind::TrackSend:
+            break;  // handled below, once the position has a track behind it
+    }
+
+    const TrackId trackId = trackAtPosition(command.index);
+    if (trackId == INVALID_TRACK_ID)
+        return;  // a template with more strips than the project has tracks
+
+    switch (command.kind) {
+        case OscCommandKind::TrackVolume:
+            writeLevel(ControlTarget::trackVolume(trackId), value);
+            return;
+        case OscCommandKind::TrackPan:
+            writeLevel(ControlTarget::trackPan(trackId), value);
+            return;
+        case OscCommandKind::TrackMute:
+            if (const auto* track = api_.tracks().getTrack(trackId))
+                api_.tracks().setTrackMuted(trackId, resolveToggle(value, track->muted));
+            return;
+        case OscCommandKind::TrackSolo:
+            if (const auto* track = api_.tracks().getTrack(trackId))
+                api_.tracks().setTrackSoloed(trackId, resolveToggle(value, track->soloed));
+            return;
+        case OscCommandKind::TrackSend: {
+            const int bus = sendBusForPosition(trackId, command.subIndex);
+            if (bus >= 0)
+                writeLevel(ControlTarget::sendLevel(trackId, bus), value);
+            return;
+        }
+        default:
+            jassertfalse;  // the first switch returned for every other kind
+            return;
+    }
+}
+
+}  // namespace magda
--- a/magda/daw/audio/osc/OscAddress.cpp
+++ b/magda/daw/audio/osc/OscAddress.cpp
@@ -4,6 +4,61 @@ namespace magda::osc {
 
 namespace {
 
+/// The deepest address in the namespace is `/magda/track/{n}/send/{m}`.
+constexpr int kMaxSegments = 5;
+
+/**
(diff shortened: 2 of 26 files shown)


magda/daw/audio/osc/OscAddress.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// ============================================================================
+// Parsing
+// ============================================================================
+
+std::optional<OscCommand> parseOscAddress(juce::StringRef address) {
+    // Split rather than match incrementally: the grammar is three to five
+    // components deep and every branch needs the component count anyway.
+    auto parts = juce::StringArray::fromTokens(juce::String(address), "/", "");
```
lucaromagnoli: **[P2] The receive path allocates despite its stated no-allocation contract.** Every message builds a StringArray and component strings through fromTokens; scheduling a drain also allocates an AsyncCallInvoker through MessageManager::callAsync. This undermines the design claim that increased surface traffic never slows the receive thread. Please parse directly without token allocations and use a preallocated async-notification mechanism, or narrow the documented guarantee.

magda/daw/audio/osc/OscRouter.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+void OscRouter::submit(const OscCommand& command, float value) {
+    const int slot = oscSlotIndex(command);
+    jassert(slot >= 0 && slot < kOscSlotCount);
+
+    // Value first, then the bit that advertises it: a drain that observes the
+    // bit is guaranteed to observe at least this value, and a store it misses
+    // leaves the bit set for the next one.
+    values_[static_cast<size_t>(slot)].store(value, std::memory_order_relaxed);
```
lucaromagnoli, line 150: **[P2] Edge-triggered commands cannot use latest-value coalescing.** Bare toggles are transitions rather than values: two pending bare mute messages should cancel, but this slot collapses them into one flip; mute=1 followed by a bare flip is reduced to only the flip and can produce the opposite final state. Separate play and stop slots also drain in enum order rather than arrival order. Please preserve toggle parity or operation composition and represent play/stop through one transport-state slot, or queue ordered commands separately while retaining coalescing for continuous values.

magda/daw/audio/osc/OscRouter.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        case OscArgKind::Toggle:
+            if (!argument.has_value())
+                return kOscToggleRequest;
+            return *argument != 0.0f ? 1.0f : 0.0f;
+
+        case OscArgKind::Normalized:
+            if (!argument.has_value())
+                return std::nullopt;
+            return juce::jlimit(0.0f, 1.0f, *argument);
```
lucaromagnoli: **[P1] Reject non-finite OSC arguments before they reach the model.** A valid OSC float can contain NaN or infinity. juce::jlimit does not sanitize NaN, so normalized controls can write NaN into volume, pan, sends, or macros; transport position also forwards NaN into Tracktion tempo conversion. Since this is unauthenticated UDP, please reject any numeric argument for which std::isfinite is false before applying the argument convention.

lucaromagnoli (commented):
Read the whole slice against the model code it lands on. The coalescer's memory ordering is correct (release on the dirty bit pairs with the drain's acquire exchange, the value store is ordered before the bit by the release, and clearing the scheduled flag before the walk closes the stranding window the PR describes), the slot mapping is verified exhaustively by its own test, the send-level write matches AutomationPlaybackEngine's SendLevel path exactly (normalizedToReal to dB, pow(10, dB/20), setSendLevel - I checked both sides), kMaxSendNumber and kMaxMacroNumber match the constants they claim to, and the socket lifecycle and fail-closed bind are right. Three findings inline; the first is the one I would fix before merging, since it breaks the feature's own core promise in a common project shape.

Minor notes not worth threads: a double flip of one toggle inside a single drain window coalesces to one flip, which is inherent to latest-value-wins and fine for fingers, worth a line in the class doc at most; setSink and setDrainScheduler are unsynchronized against a live receive thread but both carry the constraint in their docs; handleMessage returns false both for "not ours" and for "ours, deliberately ignored" (a trigger's release), which the bindings slice will want to distinguish so a release does not fall through to a binding on the same address.

magda/daw/api/osc_command_sink_live.cpp
```diff
@@ -0,0 +1,148 @@
...
+        return INVALID_TRACK_ID;
+
+    // Mixer visibility rather than the raw track list: the strips the user can
+    // see are the ones a mixer template is counting, and a hidden track would
+    // otherwise consume a fader that appears to do nothing. This is the same
+    // filter `TrackManager::getVisibleTracks` applies, walked through the
+    // facade so the sink has no second route into the model — and so it stops
+    // at the strip it wants rather than building the whole list to index once.
+    int seen = 0;
+    for (const auto& track : api_.tracks().getTracks()) {
```
lucaromagnoli: This is not the filter the mixer builds strips from, and the difference is reachable in ordinary projects. MixerView::rebuildChannelStrips additionally skips TrackType::Aux (aux strips render in their own section) and skips children of collapsed groups. So with one aux track mid-list, OSC position N is not the Nth strip on screen: the aux consumes a position and every fader after it drives the strip one to the left of the one the user is looking at. A collapsed group is worse, since its children take positions the mixer is not showing at all, which is exactly the "fader that appears to do nothing" this comment says the filter avoids.

The underlying problem is that this is a second, approximate statement of what a mixer strip is. I would extract the strip order into one shared helper (TrackManager, next to getVisibleTracks) and have both the view and this sink call it, so the two cannot drift when the mixer's rules change again.

magda/daw/audio/osc/OscRouter.hpp
```diff
@@ -0,0 +1,181 @@
...
+ * address in the fixed namespace owns a slot (`oscSlotIndex`), a store into
+ * that slot overwrites whatever had not been applied yet, and the message
+ * thread drains the whole table at once. What lands is the most recent value
+ * per address, at whatever rate the message thread can absorb, and the cost of
+ * a faster surface is finer resolution rather than a longer queue.
+ *
+ * The consequence worth knowing is that values are not applied in arrival
+ * order across different addresses — a drain walks slots. Within one address
+ * order is preserved trivially, because only the latest value exists. This is
+ * the standard control-surface tradeoff and it is why the namespace carries
```
lucaromagnoli: There is one place where this tradeoff bites harder than the paragraph admits: play and stop are two addresses controlling one state, and a drain applies them in slot order, which puts play (slot 0) before stop (slot 1) unconditionally. Stop therefore wins any drain that holds both, whatever order they arrived in - and a bundle of [stop, position, play], which is a bog-standard OSC cue from show-control software, lands stopped. Fingers cannot hit the window; a bundle hits it every time, and bundles are delivered atomically into one drain by design.

The fix that stays inside the coalescing model is to give the pair one slot: play publishes 1, stop publishes 0, and latest-value-wins then resolves them by arrival exactly as it does for a fader. If that is deferred to a later slice, I would at least say so here, because "levels and states rather than anything sequenced" reads as if no pair of addresses shares a state.

magda/daw/audio/osc/OscRouter.cpp
```diff
@@ -0,0 +1,165 @@
...
+    switch (argKindFor(command.kind)) {
+        case OscArgKind::Trigger:
+            // A momentary button sends 1 on press and 0 on release. Acting on
+            // the release would make holding Play stop the transport when the
+            // finger came off.
+            if (argument.has_value() && *argument == 0.0f)
+                return std::nullopt;
+            return 1.0f;
+
+        case OscArgKind::Toggle:
```
lucaromagnoli, line 81: A message whose first argument exists but is not numeric lands here as "no argument" and flips the toggle: /magda/track/1/mute with a string argument toggles the mute. For Normalized the same situation is correctly ignored (the test covers the string-volume case), but for Toggle the absence of a numeric argument is load-bearing - it is the flip request - so an argument we cannot read is not the same thing as no argument at all. Distinguishing message.isEmpty() from "has arguments, none numeric" and ignoring the latter would keep an unknown dialect from changing state. Same question applies to Trigger, where a string argument currently fires as if the message were bare, though acting on press is at least the harmless reading there.

lucaromagnoli:
Thanks — all six were real. Verified each against the code before changing anything, and each fix is negatively verified (reintroduced the bug, watched the specific test fail).

**[P1] Non-finite arguments.** Confirmed: `jlimit` compares, and every comparison against NaN is false, so it returns NaN untouched. Arguments are now three-valued — absent, unusable, present — which also fixes your separate note about a non-numeric argument reading as "no argument": for a `Toggle` those are opposites, since absence *is* the flip request, so a string argument no longer flips a mute. Covered by `Non-finite arguments never reach the model` and `An argument we cannot read is not the same as none`.

**[P2] Edge-triggered commands.** You were right that this isn't fixable within latest-value-wins, so triggers and toggles now go through a bounded ring in arrival order while continuous values keep their slots. That fixes toggle parity and the play/stop ordering with one mechanism, so I didn't need the single-transport-state slot — arrival order resolves them directly. The drain applies values before edges, so your `[stop, position, play]` cue locates first and then rolls, rather than rolling from the old position and jumping.

Sizing the ring turned up something: at 256 the test that sends every address at once dropped two. The discrete address space is 128 tracks × (mute+solo) + 4 transport = 260, so a surface's connect-time state dump is exactly the worst realistic case. It's 512 now with a `static_assert` tying it to `kMaxTrackNumber`, and the round-trip test asserts nothing was dropped.

**[P2] Allocation on the receive path.** Correct, and the doc overclaimed. Parsing is now a scan over the address bytes with no allocation. The `callAsync` allocation I kept and narrowed the claim around instead: since at most one drain is outstanding, it's bounded by the drain rate rather than the message rate — which is what the guarantee was actually about. `getAddressPattern().toString()` is a refcount bump, not an allocation.

**Mixer strip order.** Confirmed by reading `rebuildChannelStrips` — aux and collapsed-group children both, both reachable in ordinary projects. Took the shared-helper suggestion: `core/MixerStripOrder.hpp` now owns the rule and both the view and the sink call it. Tests cover an aux mid-list and a collapsed group in both states. One decision worth flagging: the sink asks for `Mix` order even when another view is on screen, so a template's fader 3 doesn't become a different track when the user switches to the arrangement.

**Your minor notes.** The double-flip-in-one-window case is now genuinely fixed rather than documented, since edges are ordered. `setSink`/`setDrainScheduler` keep their documented constraints. `handleMessage` returning false for both "not ours" and "ours, deliberately ignored" I've left as-is — agreed the bindings slice needs to distinguish them so a trigger's release doesn't fall through to a binding on the same address, and it's better decided there than guessed at now.

Full suite green on both targets, app builds.

lucaromagnoli (commented):
Second pass over the two fix commits. All three findings are resolved the right way: MixerStripOrder is one rule used by both the view and the sink, with the aux and collapsed-group cases tested on the sink side and the expand case covered too; the ordered ring is a correct SPSC queue (I traced the pairing - entry copy is ordered before the read-index release, the producer's acquire of the read index keeps it from clobbering an uncopied slot, and the unsigned indices are wrap-safe with the lag bounded by the capacity check); and the three-state argument reader closes both the string-flips-a-toggle case and the NaN/infinity hole I had not flagged, which was worth closing on an unauthenticated port. The allocation-free splitter checks out on the edge cases (empty address, bare slash, trailing slash, over-deep addresses, multi-byte UTF-8 in an index position, signed-char bytes). Values-before-edges is the right drain order for the cue shape and is now documented.

One new defect from the composition, inline below. Also one doc nit not worth a thread: the ring's single-producer claim rests on handleMessage only ever being called from one receive thread at a time, which is true today (one OSCReceiver, and a rebind tears the old thread down before the new one starts) but is an invariant worth a sentence on handleMessage itself, since a second caller silently breaks pushOrdered.

magda/daw/audio/osc/OscRouter.cpp
```diff
@@ -0,0 +1,260 @@
...
+        orderedRead_.store(read + 1, std::memory_order_release);
+        sink_->apply(entry.command, entry.value);
+    }
+
+    // Hit the cap with work still queued. Whoever pushed the remainder may have
+    // found the drain already scheduled, so ask for the next one here rather
+    // than leaving those commands until the surface is touched again.
+    if (orderedRead_.load(std::memory_order_relaxed) !=
+        orderedWrite_.load(std::memory_order_acquire))
+        scheduleDrain();
```
lucaromagnoli: This re-schedule defeats its own cap when it matters. In the app the scheduler is the default postDrain, and postDrain runs drainPending inline when already on the message thread - which this call always is. So hitting the cap with work still queued recurses: drainPending -> drainOrdered -> scheduleDrain -> postDrain -> drainPending, one stack level per 512 applied commands, with no return to the event loop in between. A loopback flooder that keeps the ring non-empty holds the message thread in this recursion indefinitely and grows the stack while doing it - exactly the hold-the-thread scenario the per-drain bound was written to prevent, and the bound only works when the next drain arrives via callAsync.

Two ways out: have this tail path post the follow-up drain with callAsync directly (skipping the on-message-thread shortcut, so the remainder waits its turn behind other events), or drop the inline shortcut from postDrain and let scheduleDrain always post when a message manager exists. The first is smaller and keeps the headless inline path for tests.

lucaromagnoli:
Confirmed and fixed. You're right that `scheduleDrain` from the tail runs the default scheduler, and that scheduler applies inline when already on the message thread — which at that point it always is. The follow-up now posts directly and never goes through the scheduler; with no message loop it does nothing instead, since a headless host drives its own drains and recursing there would be the same bug.

Worth recording how narrow the window is, because it explains why nothing caught it. For the tail CAS to succeed, no producer can have called `scheduleDrain` since `drainPending` cleared the flag — and work can only remain if a producer pushed. Single-threaded those are contradictory, so the flag always wins the race. It's reachable only because `pushOrdered` publishes `orderedWrite_` *before* `submit` calls `scheduleDrain`: a drain reading the index inside that gap sees work pending with the flag still clear. So it needs two threads, and I couldn't make it deterministic. Agreed it should be fixed anyway — stack safety resting on that ordering happening to hold is the kind of thing that stops being true after an unrelated edit.

The cap it protects *is* deterministic, so that got the regression test instead: a sink that re-sends on every apply. I verified it by removing the cap — the test doesn't fail, it hangs, which is the honest shape of the failure (killed it at 12s).

Took the doc note too: the single-producer invariant is now on `handleMessage` rather than only on the ring, since that's where a second caller would come from.

Full suite green on both targets, app builds.

Issue #1193, opened 2026-05-05
Incorrect note display in the piano roll.


nowgad:
### What happened?

Ghost notes from the previous clip remain. These notes can be edited within the bounds of the second clip—you can add new ones, delete them, and change their length. When selecting notes from the first clip in this area and adding a new note, the focus shifts to the first clip.
UPD. During this clip collision, when playing back, the DAW crashes or freezes completely.

### Steps to reproduce

Precondition: Place two MIDI clips sequentially on the same track.

1 Two MIDI files on the same track.

2 Trim one of the clips by a few bars.

3 Move another MIDI file into those bars, with the piano roll open.

4 Bug. The notes from the trimmed section of the first clip remain visible in the piano roll.

https://github.com/user-attachments/assets/7f514a6c-b6c4-46d1-b56d-f6ca6a70e68c

### Expected behavior

There should be no extra information in this area.

This is an interesting bug :) – maybe it could be used as a ghost note/clip feature for other MIDI clips.

### MAGDA version

0.7.1

### Operating system

Windows

### Crash log (if applicable)

```text

```

lucaromagnoli:
Hey man good to see you again!

It's actually even worse. Basically you can keep adding notes past the end of the clip and this will bleed into the next one. Similarly when you trim stale notes are left. Fix on the way

nowgad:
> Hey man good to see you again!

Same here! I'm not going anywhere, man :)

The bug hasn't gone away.

1 Now, if you overlap one clip with another, the DAW freezes completely.
2 The ability to combine and overlap different MIDI clips is gone. Now, when you do this, the clip that is "below" gets trimmed.
Magda 0.7.2

lucaromagnoli:
Thanks for the follow-up!

**On the freeze:** I haven't been able to reproduce this on Windows or macOS. Could you give me exact steps to reproduce it? Does it freeze on overlap or only on playback? Is there anything else going on at the time that might have caused it (plugins loaded, recording armed, etc.)?

**On the overlap/trimming:** I'm a bit confused — I thought overlapping clips on the same track wasn't desirable (like most DAWs). Could you give an example of when you'd want that?

nowgad:
https://github.com/user-attachments/assets/3f2ba0c9-d5f1-4dfe-846e-1c17dbdd9182 
- Condition: two clips are "touching".
[MagdaSynth.zip](https://github.com/user-attachments/files/27459484/MagdaSynth.zip)

[...]

PR #2064, opened 2026-08-10
Native engine: MIDI clip playback (#2039)

### LINKED ISSUE #2039: Arranger clips slice 6: MIDI clip playback

lucaromagnoli:
Slice 6 of #1890. Closes #2039.

The source behind a track's `ClipMidi` op. Audio and MIDI share the snapshot, the span and the interior silences, and share nothing below that: no file, no reader, no stretcher, no pool. What there is instead is a question audio never has to answer, which is what a note that has already sounded is owed.

## Compiled, not carried

Slice 1 put the model's own lists in the snapshot. `MidiClipPlayback` now holds one sorted array of short messages in content beats instead, the same move `AudioEventPlayback` makes with `WarpMap`. The curve densification, the MPE channel assignment, the same-pitch overlap rule and the two offsets resolve once, off the audio thread, and the callback does a binary search and a walk.

The fork arrives at the same place by a longer road (`MidiList::createDefaultPlaybackMidiSequence`). The difference is that its sequence lives inside `te::MidiClip`, so editing a curve rebuilds it, `TreeWatcher` sees the tree change and playback restarts under a rolling transport, which is the click `ClipSynchronizer` warns about in three places.

## A loop is a coordinate change, not a copy

The fork unrolls one copy of the sequence per repetition, rebuilt whenever a note moves. A block here is a beat range, and folding it through the loop gives a handful of sub-ranges over the one list. The per-pass clipping rule is kept exactly, because it is what puts every note's off in the same pass as its on.

Nothing reads a loop as a length either, so `MidiClip::disableLooping` truncating a clip to one loop length has no counterpart here: the span is the length. There is no second visible-range crop, because the span already is the crop.

## Note lifetime

One invariant, held by `ActiveNoteList` rather than by care: **a note-off is never emitted for a note the source did not start, and never withheld from one it did.** Five things end a note and only a pass or a span ending falls out of the material; the rest are a locate or wrap, a stop, and a snapshot swap reconciled against what the new snapshot says should be sounding here.

The port budget cannot break it. Drops fall on note-ons and controllers, and a note-on that is dropped is never entered in the active list, so it never comes to owe an off. An off that still will not fit is carried to sample zero of the next block. Late is a longer tail; missing is a note that rings until the session ends.

## Groove runs per pass

Ported natively, and the one thing here evaluated on the audio thread rather than compiled. Groove is anchored to the project grid, so a loop length that is not a whole multiple of the template's period grooves each pass differently, which is what the fork delivers by re-timing its sequence to `clipRange.start + loopIndex * loopLength` once per pass and grooving it there (`LoopedMidiEventGenerator::setLoopIndex`). Baking before the fold would diverge on every odd-length loop.

The block's event search widens by the compiled table's own worst-case displacement, which is exact rather than sampled. Two guards the fork lacks: grooved events emit in grooved order, and a grooved note-off never precedes its own note-on.

`compileClipSnapshot` takes a `GrooveTemplateSet` and gets an empty one everywhere so far. Wiring the app's templates to it belongs with switching the engine on, like the transient detector and the loop-info parse.

## Two deliberate divergences

Both recorded in `docs/architecture/native-engine.md` beside the Signalsmith priming one.

**Controller density.** Messages go out on every change of the quantised value and no closer together than about a millisecond, not on the sync layer's 1/16-beat grid. That grid is anchored to the wrong axis, so it is both too dense and too sparse: a ramp of one unit over eight bars is 512 near-identical messages on it and 2 here, while a pitch-bend dive over 100 ms gets 3 grid points at 120 BPM and about 100 here. It also moves with tempo, running at 8 Hz at 30 BPM and 128 Hz at 480 for the same drawn curve, when smoothness is a wall-clock property. The floor bounds the cost where the grid was aimed, about ten messages per block per controller, and bounds it in wall-clock rather than in beats, so #1193 stays shut.

Because this alters the controller stream feeding an arbitrary synth, #2040 needs to change shape rather than change tolerance: MIDI event streams compared as their own artifact, where the difference is exact and event-addressable, and curve-driven audio out of the near-null audio assertion rather than loosening it until it passes.

**`midiOffset` on an unlooped clip.** The fork's arranger path drops it (`setOffset(0)` in the else branch of `syncMidiClipToEngine`) while its session path applies it. Applying it either way means a project that leant on the current behaviour plays differently under the engine.

## Tests

`tests/test_midi_clip_playback.cpp` and `tests/test_groove_template.cpp`.

The spine is `Recorder::hanging()`, which models what a receiver actually holds, one bit per (channel, note). Every test that moves the transport asserts it empty after a stop: wraps, a locate forwards, a locate backwards, a stop mid-note, a swap that deletes the clip, a swap that moves the note, and a span that ends mid-note.

Densification is asserted as two properties rather than a message count (every message changes the value; none closer than the floor), then the curve shape against the editor's own formulas, then the two cases the grid got wrong in opposite directions. Groove has the odd-length-loop case that decided where it runs. The budget has both halves: an overflow reports rather than allocating, and offs owed by a full block arrive in the next one.

Full suite green: 2461 passed, 13 skipped.


magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+                const auto isOff = event.isNoteOff();
+                const auto isOn = event.isNoteOn();
+                const auto phaseOf = isOn ? 2 : (isOff ? 1 : 0);
+                if (phaseOf != phase)
+                    continue;
+
+                auto timelineBeat = pass.timelineOfContentZero + event.beat;
+                if (event.isNoteEdge())
+                    timelineBeat = clip.groove.groovyBeat(timelineBeat);
```
lucaromagnoli, line 290: A grooved note-on can escape its own pass here, and the pass-end endClip below then fires before it in sample order. The walk's filter is the block range (`from`/`to`), not the pass's stretch of it, so a note-on near the loop's end whose grooved timeline lands past the wrap is emitted after the wrap sample - and `endsPass` endClip has already ended it at the wrap, which in MidiBuffer sample order puts the off *before* the on. The receiver is left with a sounding note the active list no longer knows about, which breaks the comment at line 261: something the pass started has outlived it.

It is masked in practice - a looped clip re-strikes the same pitch next pass and that pass's off ends the dangling one, so it degrades to an audible overhang into the next pass rather than an eternal hang, and the span-end case is saved by `to` clamping to the span - which is also why no Recorder test catches it: the bitset resolves. But it is reachable with ordinary swing on an odd-length loop (displacement 0.165 beats at 0.66 on a 2-per-beat grid, note within that of the window end).

The symmetric fix to the off-clamped-to-its-on rule: clamp a grooved note-on's timeline into its pass, `min(grooved, passEndTimeline - oneSample)` when the pass ends inside the block, so "every note that sounds inside a pass is ended inside the same pass" holds under groove too. Plus the test: odd-length loop, swung note near the window end, asserted on message order rather than the bitset.

magda/engine/clip/MidiEventList.hpp
```diff
@@ -0,0 +1,237 @@
...
+    /// path applies it; that is a gap in the sync layer rather than a semantic,
+    /// and it is recorded as a divergence.
+    double offsetBeats = 0.0;
+
+    bool loopEnabled = false;
+    double loopStartBeats = 0.0;
+    double loopLengthBeats = 0.0;
+
+    /// The content beat sounding at timeline beat @p beat.
+    double contentAt(double beat) const;
```
lucaromagnoli: contentAt has no caller - foldBlock derives the same mapping inline and nothing uses this. That makes it dead code and, worse, a second statement of the fold arithmetic that can drift from the one that plays (the exact two-derivations hazard the warp slice's placementFor comment warns about). Either delete it, or keep it as the single source and have foldBlock's pass construction go through it - and if it stays, a test asserting contentAt agrees with the passes foldBlock emits would make the pair self-checking.

lucaromagnoli (commented):
Read the whole slice, traced the fold arithmetic both directions (contentAt and foldBlock agree, timeline reconstruction round-trips, the half-open nextafter at the block end keeps a wrap-aligned block from opening the pass it never reached), and verified the budget invariant end to end: the fits() reservation caps concurrently active notes at 1365, which is exactly what one port's 4096 bytes can flush, so the owed-offs drain can never overflow and the break-then-clear at the top of render can never lose an off. That arithmetic being tight rather than approximately safe is the best thing in the PR. The swap reconcile, the hole rules, the three-phase walk (MidiBuffer's sample-plus-insertion ordering doing the compile's tie-break for free), the MPE assignment and the groove port against the fork's formula all check out, and track.midi is sorted where the early break assumes it.

Two findings inline. One test gap worth closing in the same breath as the first finding: the PR body's own test plan promised "under MPE the expression is reconstructed on the note's own channel" on a locate, and no test asserts it - controllerStateAt should deliver it structurally, but the chase section stops at plain CC.

One observation not worth a thread: the double-strike guard in startNote compares two timeline doubles that arrive by different arithmetic routes (the outer chase passes `from`, the pass chase passes contentZero + windowStart), so at an exact wrap the dedup rests on those cancelling to the same bits. If they ever differ by an ulp the cost is a duplicate note-on at one sample, which a receiver reads as one note, so it is cosmetic - but a comment saying the equality is load-bearing would stop someone "simplifying" one side later.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+        const auto searchTo = std::min(pass.windowEnd, pass.contentEnd + widen);
+
+        const auto first = list.lowerBound(searchFrom);
+        const auto last = list.lowerBound(searchTo);
+
+        // Three ordered walks rather than a scratch sort. juce::MidiBuffer
+        // inserts by sample and keeps insertion order for ties, so walking
+        // controllers, then note-offs, then note-ons gives the compile's own
+        // ordering at every sample for free, groove or no groove.
+        for (auto phase = 0; phase < 3; ++phase) {
```
lucaromagnoli: **P1 — Process note edges in timestamp order.** This outer phase loop updates note state in message-type order, so when a complete note falls inside one callback its off is visited before its on. The off sees no active note and is skipped; the later on is then left active until a boundary or stop. `MidiBuffer` sorting fixes output order but cannot repair `active_`. Please walk note edges chronologically, using the rank only to break equal-timestamp ties, and add a regression test for a note whose on/off are in the same block.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+                }
+
+                emit(out, block.sampleForBeat(timelineBeat), event);
+            }
+        }
+
+        // The pass ran out inside this block, so nothing it started may outlive
+        // it. This is what the fold owes the invariant: every note that sounded
+        // in a pass is ended inside the same pass.
+        if (pass.endsPass)
```
lucaromagnoli, line 350: **P1 — Keep grooved note-ons inside the current fold pass.** The earlier range check constrains `timelineBeat` only to the whole render range, not to this pass. A note near `windowEnd` can be grooved past the loop boundary, then this cleanup inserts its off at the earlier boundary and clears the internal state. The receiver sees off-then-on and can hang until a later pass. Reject or clamp shifted starts at the pass boundary and cover a positively shifted note at the end of a loop.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+        }
+        emit(out, sample, clip.events.events[static_cast<std::size_t>(index)]);
+    }
+
+    // Then the notes this instant is inside. Gathered into a second pass over
+    // the scratch because startNote uses it.
+    scratch_.clear();
+    clip.events.notesSoundingAt(contentBeat, scratch_);
+
+    const auto sounding = scratch_;
```
lucaromagnoli: **P2 — Avoid allocating in the audio callback.** Copying `scratch_` constructs and allocates a new vector during every non-empty chase. The same pattern occurs for `stale` in snapshot reconciliation. Both can iterate the already-reserved scratch storage directly because neither `startNote` nor `endNote` mutates it. This currently breaks the class real-time allocation guarantee.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+            dropped_.fetch_add(1, std::memory_order_relaxed);
+            continue;
+        }
+        emit(out, sample, clip.events.events[static_cast<std::size_t>(index)]);
+    }
+
+    // Then the notes this instant is inside. Gathered into a second pass over
+    // the scratch because startNote uses it.
+    scratch_.clear();
+    clip.events.notesSoundingAt(contentBeat, scratch_);
```
lucaromagnoli: **P2 — Chase using groove-adjusted note boundaries.** `notesSoundingAt(contentBeat)` evaluates raw event beats, while normal playback grooves the on and off independently. A locate between a raw onset and its delayed onset starts the note too early; an advanced onset produces the inverse miss. The sounding-state query needs to evaluate the grooved edges for this fold pass.

magda/engine/clip/MidiClipCompiler.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+        list.longestNoteBeats = std::max(list.longestNoteBeats, note.lengthBeats);
+
+        // ---- Per-note pitch expression -------------------------------------
+        //
+        // Densified like any other curve rather than on the sync layer's
+        // 1/16-beat grid: that grid is not TE's rule, TE emits one raw wheel per
+        // expression point with no interpolation at all. Expression is pitch,
+        // which is where a coarse grid is heard most directly.
+        if (list.mpe && note.hasPitchExpression()) {
```
lucaromagnoli, line 322: **P2 — Reset pitch bend when an MPE member channel is reused.** Once any note has expression, all notes use member channels, but a non-expressive note emits no pitch-wheel value here. After the allocator eventually reuses a channel whose previous expressive note ended away from center, that note inherits the stale bend. Emit a center value before non-expressive note-ons on reused channels, or otherwise restore member-channel state.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+            const auto from = std::max(block.startBeat, clip.span.startBeat);
+            if (foldBlock(clip.fold, from, block.endBeat, clip.span.endBeat, passes) == 0)
+                continue;
+
+            scratch_.clear();
+            clip.events.notesSoundingAt(passes[0].contentStart, scratch_);
+
+            for (const auto index : scratch_) {
+                const auto& event = clip.events.events[static_cast<std::size_t>(index)];
+                expected.set(event.channel(), static_cast<int>(event.data1));
```
lucaromagnoli: **P1 — Preserve clip ownership during snapshot reconciliation.** This mask records only `(channel, note)`, while `ActiveNoteList` also uses `clipId` to decide which later off is valid. If active clip A is removed and the new snapshot exposes clip B sounding the same channel/note, this bit keeps A registered as owner. B is not chased on a continuous block, and its eventual off is rejected because the owner is still A, so the note can hang until stop. The expected state must include ownership, ending/restarting when the expected clip differs.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    // Only for a note this source started, and never suppressed
+                    // by a hole. Both edges are grooved independently, so an off
+                    // can land before its own on: it is held back to it rather
+                    // than sent early.
+                    if (!active_.active(channel, note) ||
+                        active_.owner(channel, note) != clip.clipId)
+                        continue;
+
+                    const auto onBeat = active_.startBeat(channel, note);
+                    endNote(out, block.sampleForBeat(std::max(timelineBeat, onBeat)), channel,
```
lucaromagnoli: **P1 — Clamp grooved note-offs to the fold-pass end.** A positive groove can move this off beyond `pass.timelineOfContentZero + pass.windowEnd`. Processing clears the active entry and inserts the off after the wrap; the subsequent pass-end cleanup then sees nothing to end at the boundary. The receiver hears the note ring into the next pass, where the delayed off may also cut a newly started note. Apply the pass boundary to offs as well as ons and add a regression with an off shifted across an in-block wrap.

magda/engine/clip/MidiEventList.cpp
```diff
@@ -0,0 +1,126 @@
...
+        return;
+
+    // No note-on further back than the longest note can still be sounding, so
+    // the walk is bounded by the material rather than by the clip's length.
+    auto index = lowerBound(beat - longestNoteBeats);
+    const auto end = lowerBound(beat);
+
+    for (; index < end; ++index) {
+        const auto& event = events[index];
+        if (!event.isNoteOn() || event.endsAt < 0)
```
lucaromagnoli: **P2 — Chase retriggered notes before the retrigger occurs.** `endsAt < 0` also represents the first of two overlapping same-pitch notes, whose own off was intentionally dropped. During ordinary playback that first note is still sounding from its onset until the later note-on replaces it. A locate inside that interval currently skips it and produces silence. The compiled representation needs an effective chase end—such as the retrigger beat—even when no MIDI note-off is emitted.

PR #2070, opened 2026-08-11
Native engine: the null-diff corpus (#2040)

### LINKED ISSUE #2040: Arranger clips slice 7: null-diff corpus for clip playback

lucaromagnoli:
Closes #2040.

Parity claimed by rendering rather than by reading the code: twenty-one projects built as model values, each rendered through Tracktion and through the native engine, compared as audio or as a MIDI stream.

## How it is arranged

**The comparator is model-only and has its own tests against known-bad pairs.** It is the one piece nothing else checks, and a comparator that passes everything reports parity it never measured. It is fed a one-sample placement error, a missing note, a value two units out, a stream a beat late, a controller that stops halfway, a burst in the wrong place, and a transposed tone, and it has to say so for each.

**Material is chosen per case rather than tolerance per case.** Impulses and steps where the engines must agree sample for sample; band-limited tones wherever an interpolator or a stretcher stands between them, because broadband material through either produces a residual tens of decibels above anything a placement bug makes.

**Tempo maps and warp maps are compared as functions**, to the microsecond, so a mapping disagreement fails once with a number instead of twenty-one times as a waveform.

**Two ways the incumbent leg could lie are handled rather than left to chance.** It renders float, checked on read-back, because sixteen bits would put quantisation noise above the floor. And it waits for proxies through the render manager, failing a case as unmeasurable rather than reporting a race as a parity failure.

## Six engine bugs the corpus found, all fixed here

- `renderOffline` never told the voice pool where it was, so an **offline bounce of arrangement audio opened no readers and rendered silence**. It now services and fills the pool in step, and `PrefetchThread` gained a manual mode so an offline render's reading is deterministic rather than racing a thread.
- The launch ramp was **clamped to whatever fitted in the first block**, so it decayed over `min(256, blockSize)` samples and never continued.
- The launch ramp also ran on a voice **starting at the beginning of its own material**, where nothing precedes it: a clip whose first sample is a transient lost the transient and gained 256 samples of tail.
- A rate that varies within a block was resolved from that block's own two ends, and a stretcher framed whatever sizes it was handed, so both made output **a function of how the callback was cut up** — which `RenderContext` forbids and which means a bounce and playback at different block sizes disagree. Clips that consume their reading at a rate are now fed on a grid anchored to where their event begins. The block-size-dependent list went from five cases to one.
- An **MPE note opened with no timbre and no pressure**, though the specification asks for both and a member channel is reused round-robin, so a synth inherited another note's expression.
- Fixing that exposed `emit` building **every message as three bytes**, which made channel pressure malformed.

## What it asserts

Twelve cases hold, including `fades.curves` at **-141 dBFS** (the fade-curve port is exact), placement, trims, loop tiling, reverse, analog pitch and comping at a perfect null, and all five MIDI cases — `midi.fold` matching 24 grooved notes across an odd loop length, which pins slice 6's trickiest ported semantic against the thing itself.

Two divergences are pinned as declared constants with their mechanisms rather than absorbed into tolerances: the fork drops `midiOffset` on an unlooped arranger clip, and it ends every note 0.0001 s early to force event ordering.

Eight cases are **under calibration**: measured and printed on every run, not yet held to a bound, because a bound nobody can explain is how a real difference ends up inside an expected one. The list is a set asserted in both directions, so a case cannot join or leave it quietly. `rate.48k` is the instructive one — it sits 0.852 of a sample out, which looked exactly like a difference between two interpolation kernels, so the machinery to pin and align a fixed sub-sample offset went in; aligning by it made the case *worse*, which rules that mechanism out, and the case now says so where the next person will read it.

## Testing

- 406 tests pass in `magda_tests`, including 35 for the comparator and the corpus's own declarations.
- `Null Diff Maps` and `Null Diff Corpus` pass in `magda_juce_tests`, registered as two ctest entries.

### DIFF
--- a/magda/engine/clip/ClipVoice.cpp
+++ b/magda/engine/clip/ClipVoice.cpp
@@ -12,6 +12,11 @@ namespace magda::engine {
 void ClipVoice::prepare(const RenderContext& context) {
     sampleRate_ = context.sampleRate;
     maxBlockSamples_ = context.maxBlockSize;
+
+    // One cell of output, off the callback. What a cell produces beyond the
+    // block that asked for it waits here.
+    held_.setSize(std::max(1, context.numChannels), kCellSamples, false, true, false);
+
     release();
 }
 
@@ -20,6 +25,126 @@ void ClipVoice::release() {
     eventId_ = INVALID_EVENT_ID;
     sounded_ = false;
     primed_ = nullptr;
+    deClick_.reset();
+    pending_ = false;
+    pendingCount_ = 0;
+    pendingRead_ = 0;
+    skip_ = 0;
+}
+
+bool ClipVoice::renderThroughCells(const AudioClipPlayback& clip, const AudioEventPlayback& event,
+                                   const BlockInfo& block, PrefetchStream& stream,
+                                   ClipStretcher& stretcher, int preRoll,
+                                   juce::dsp::AudioBlock<float> scratch,
+                                   juce::dsp::AudioBlock<float> region, double windowStart,
+                                   int count) {
+    // The grid, in samples of the timeline, anchored to where the event begins
+    // rather than to where this block does. That anchor is the whole point: two
+    // renders that cut the timeline into different blocks still divide it into
+    // the same cells, so the stretcher is handed the same input in the same
+    // order both times and gives back the same samples.
+    const auto eventStartSample =
+        static_cast<std::int64_t>(std::llround(event.span.startSeconds * sampleRate_));
+    const auto windowStartSample =
+        static_cast<std::int64_t>(std::llround(windowStart * sampleRate_));
+
+    // Beginning rather than carrying on: the first block of a voice, the first
+    // after the timeline jumped, and one whose stretcher the pool has replaced.
+    // Not after an underrun, deliberately, and that is the difference between a
+    // reader catching up and one that never does: priming reads behind the
+    // position wanted, so a stream that came back short and was primed again
+    // would be sent backwards every block.
+    auto needsPrime = primed_ != &stretcher || !block.continuous || !pending_;
+
+    if (needsPrime) {
+        // Back to the cell boundary at or before where playback resumes, and
+        // the head of that cell is produced and dropped. Resuming mid-cell
+        // would anchor the grid to wherever the transport happened to land,
+        // which is the block dependency again wearing a locate's clothes.
+        const auto offset = windowStartSample - eventStartSample;
+        const auto index = static_cast<std::int64_t>(
+            std::floor(static_cast<double>(offset) / static_cast<double>(kCellSamples)));
+
+        nextCell_ = eventStartSample + index * kCellSamples;
+        pendingCount_ = 0;
+        pendingRead_ = 0;
+        skip_ = static_cast<int>(windowStartSample - nextCell_);
+
+        stretcher.reset();
+        primed_ = &stretcher;
+        pending_ = true;
+    }
+
+    auto produced = 0;
+    auto full = true;
+
+    while (produced < count) {
+        if (pendingRead_ < pendingCount_) {
+            const auto take = std::min(pendingCount_ - pendingRead_, count - produced);
+            for (std::size_t channel = 0; channel < region.getNumChannels(); ++channel) {
+                const auto* source = held_.getReadPointer(static_cast<int>(
+                    std::min(channel, static_cast<std::size_t>(held_.getNumChannels() - 1))));
+                auto* destination = region.getChannelPointer(channel);
+                std::copy(source + pendingRead_, source + pendingRead_ + take,
+                          destination + produced);
+            }
+
+            pendingRead_ += take;
+            produced += take;
+            continue;
+        }
+
+        // One whole cell, at its own two ends. The ratio a cell runs at is what
+        // its own boundaries say, exactly as a block's used to be, and the
+        // boundaries no longer move.
+        const auto cellStartSeconds = static_cast<double>(nextCell_) / sampleRate_;
+        const auto cellEndSeconds = static_cast<double>(nextCell_ + kCellSamples) / sampleRate_;
+
+        const auto opens = readingPositionAt(clip, event, cellStartSeconds,
+                                             block.beatAtTime(cellStartSeconds), sampleRate_);
+        const auto closes = readingPositionAt(clip, event, cellEndSeconds,
+                                              block.beatAtTime(cellEndSeconds), sampleRate_);
+        const auto step = (closes - opens) / kCellSamples;
+
+        const auto ahead = stretcher.readAheadSamples();
+        const auto readFrom = static_cast<std::int64_t>(std::llround(opens)) + ahead;
+        const auto readTo = static_cast<std::int64_t>(std::llround(closes)) + ahead;
+        const auto wanted = static_cast<int>(
+            std::clamp<std::int64_t>(readTo - readFrom, 0, maxReadingSamples(maxBlockSamples_)));
+
+        if (needsPrime) {
+            // At the cell's own boundary rather than at the block's, so what
+            // the stretcher is primed with is the material leading up to a
+            // fixed instant. Read out of the same stream in the same pass, so
+            // the cell's own read continues it rather than seeking again.
+            stretcher.prime(stream, readFrom, preRoll, step);
+            needsPrime = false;
+        }
+
+        auto reading = scratch.getSubBlock(static_cast<std::size_t>(maxBlockSamples_),
+                                           static_cast<std::size_t>(wanted));
+        const auto delivered = stream.read(readFrom, reading, wanted);
+
+        juce::dsp::AudioBlock<float> cell(held_);
+        auto cellOut = cell.getSubBlock(0, static_cast<std::size_t>(kCellSamples));
+        stretcher.process(reading, opens - static_cast<double>(readFrom), step, cellOut);
+
+        nextCell_ += kCellSamples;
+        pendingCount_ = kCellSamples;
+        pendingRead_ = std::min(skip_, kCellSamples);
+        skip_ = std::max(0, skip_ - kCellSamples);
+
+        // A cell the reader could not fill is a cell that ends in silence, and
+        // the voice is no longer carrying on from anywhere. Said rather than
+        // acted on: the block is still filled to its end, exactly as the plain
+        // path fills it, because stopping here would leave the rest of the
+        // region holding whatever the scratch held last and would put the voice
+        // a block behind for the rest of the take.
+        if (delivered < wanted)
+            full = false;
+    }
+
+    return full;
 }
 
 void ClipVoice::applyFade(juce::dsp::AudioBlock<float> region, int regionFirstSample,
@@ -67,6 +192,7 @@ bool ClipVoice::render(const AudioClipPlayback& clip, const AudioEventPlayback& 
         eventId_ = event.eventId;
         sounded_ = false;
         primed_ = nullptr;
+        deClick_.reset();
     }
 
     const auto nothing = [this] {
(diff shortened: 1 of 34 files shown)


tests/NullDiffCompare.hpp
```diff
@@ -0,0 +1,436 @@
...
+    /// is how that would have gone unnoticed.
+    int nativeUncompared = 0;
+    int incumbentUncompared = 0;
+
+    /// Every disagreement, addressed: which note, which controller, which
+    /// instant. A parity failure that says "the streams differ" is a day of
+    /// somebody's life.
+    std::vector<std::string> problems;
+
+    bool passed() const {
```
lucaromagnoli, line 361: This pass condition excludes `nativeUncompared`, `incumbentUncompared`, and the problems produced for them. A different number of channel-pressure events therefore still passes; with equal counts, wrong channels, values, or timestamps also pass. Since the MPE cases intentionally emit channel pressure, please compare those events by channel/value/time (or fail on every unmatched message) and include that result in `passed()`.

tests/test_null_diff_juce.cpp
```diff
@@ -0,0 +1,543 @@
...
+        options.sampleRate = value.sampleRate;
+        options.measureShift = value.verdict == Verdict::Shift;
+        options.maxShiftSamples = value.maxShiftSamples;
+
+        report.hasAudio = true;
+        report.audio = compareAudio(aligned, incumbent.audio, options);
+
+        switch (value.verdict) {
+            case Verdict::Null:
+                report.passed = report.audio.withinFloor() && report.audio.refusal.empty();
```
lucaromagnoli: `compareAudio` records `lengthDifference` but compares only the overlapping prefix, and `withinFloor()` ignores the length. Consequently, a shorter incumbent render with a bit-identical prefix is certified as a Null pass even though the comparator contract explicitly calls a short render a bug. Please require `lengthDifference == 0` for Null verdicts and add a truncated-buffer regression test.

magda/engine/clip/ClipVoice.cpp
```diff
@@ -20,6 +25,126 @@ void ClipVoice::release() {
...
+            needsPrime = false;
+        }
+
+        auto reading = scratch.getSubBlock(static_cast<std::size_t>(maxBlockSamples_),
+                                           static_cast<std::size_t>(wanted));
+        const auto delivered = stream.read(readFrom, reading, wanted);
+
+        juce::dsp::AudioBlock<float> cell(held_);
+        auto cellOut = cell.getSubBlock(0, static_cast<std::size_t>(kCellSamples));
+        stretcher.process(reading, opens - static_cast<double>(readFrom), step, cellOut);
```
lucaromagnoli, line 130: The output passed here is always a 128-sample cell, but SoundTouch is still provisioned with `setup.maxBlockSamples` and caps `receiveSamples` to `maxBlockSamples_`. With a 64-sample device block, it therefore writes only the first 64 samples of `held_` and zero-fills the other 64, producing alternating audio and silence across callbacks. The input clamp above can also truncate fast cells because it is derived from the block rather than the cell. Please provision the scratch and stretcher capacities for at least `max(maxBlockSize, kCellSamples)` (or use dedicated cell capacity), and cover a stretched 64-sample-block render.

magda/engine/clip/ClipVoice.cpp
```diff
@@ -20,6 +25,126 @@ void ClipVoice::release() {
...
+                                             block.beatAtTime(cellStartSeconds), sampleRate_);
+        const auto closes = readingPositionAt(clip, event, cellEndSeconds,
+                                              block.beatAtTime(cellEndSeconds), sampleRate_);
+        const auto step = (closes - opens) / kCellSamples;
+
+        const auto ahead = stretcher.readAheadSamples();
+        const auto readFrom = static_cast<std::int64_t>(std::llround(opens)) + ahead;
+        const auto readTo = static_cast<std::int64_t>(std::llround(closes)) + ahead;
+        const auto wanted = static_cast<int>(
+            std::clamp<std::int64_t>(readTo - readFrom, 0, maxReadingSamples(maxBlockSamples_)));
```
lucaromagnoli, line 113: The cell's read budget is still the block's. A cell always consumes up to kCellSamples * kMaxStretchRate + 1 = 1281 samples of reading, but this clamp allows maxReadingSamples(maxBlockSamples_), which at a 64-sample block size is 641. At block sizes below kCellSamples with a rate past maxBlockSamples/kCellSamples times the ceiling, every cell's read is truncated - systematically, not as an underrun, so `full` goes false but the audio is also wrong, and block-size invariance breaks for exactly the small block sizes a low-latency interface uses. The scratch provisioned by scratchSamples(maxBlockSamples) has the same assumption baked in.

Budget and provision against max(maxBlockSamples_, kCellSamples): the clamp here, the scratch sizing in prepare, and held_ are all sized from the same answer, so one helper (cell-aware maxReadingSamples) keeps them agreeing.

magda/engine/clip/ClipVoice.cpp
```diff
@@ -20,6 +25,126 @@ void ClipVoice::release() {
...
+
+        // One whole cell, at its own two ends. The ratio a cell runs at is what
+        // its own boundaries say, exactly as a block's used to be, and the
+        // boundaries no longer move.
+        const auto cellStartSeconds = static_cast<double>(nextCell_) / sampleRate_;
+        const auto cellEndSeconds = static_cast<double>(nextCell_ + kCellSamples) / sampleRate_;
+
+        const auto opens = readingPositionAt(clip, event, cellStartSeconds,
+                                             block.beatAtTime(cellStartSeconds), sampleRate_);
+        const auto closes = readingPositionAt(clip, event, cellEndSeconds,
```
lucaromagnoli, line 105: BlockInfo::beatAtTime is a lerp over the block's own endpoints, extrapolating linearly outside them - and a cell regularly outlives its block. Within one tempo region the extrapolation is exact, so the grid is sound there. But a cell straddling a step tempo change gets its far end's beat computed at the producing block's slope, not the map's: the blocks themselves are cut at tempo boundaries, so the producing block sits wholly in the old region and evaluates the cell's post-change stretch at the pre-change tempo. That is deterministic and block-size independent, so the invariance test cannot see it, but it is wrong against the exact map - the cell's ratio is briefly off across every tempo step, which is a residual localized at changes on exactly the auto-tempo material tempo.auto plays. It may be a contributor to why that case is still under calibration.

The fix that stays inside the grid's design: truncate a cell at a tempo boundary the way blocks already are. Tempo changes are timeline-anchored, so a boundary-split grid is still a function of position and nothing else, and every cell then lives inside one region where the lerp is exact.

tests/test_null_diff_native_leg.cpp
```diff
@@ -0,0 +1,145 @@
...
+    std::set<std::string> failed;
+
+    for (const auto& value : buildCorpus(scratch())) {
+        if (value.capturesMidi())
+            continue;
+
+        INFO(value.name);
+
+        auto small = value;
+        small.blockSize = 128;
```
lucaromagnoli: 128 and 1024 are both multiples of kCellSamples, and every stretched clip in the corpus starts at beat 0, so in both runs the event-anchored cell grid coincides exactly with the block grid: pendingRead_/held_ carry-over, the skip_ head-drop, and mid-cell resume never execute under this test. The invariance claim is strongest precisely where the new machinery does real work, and right now that part runs in production only. One stretched case whose clip starts off the 128-sample grid (any non-integer beat does it), plus one misaligned size here - 96 or 480 - would put the holdover paths under the same assertion as the aligned ones.

lucaromagnoli (commented):
Read all of it: the comparator (shift search, envelope, spectra, the controller step-function comparison and its known-bad tests), both legs, the six engine fixes, and the runner. The structure is right where it matters: the comparator is the one unchecked piece and it has 24 adversarial tests; the calibration set is asserted in both directions so nothing joins or leaves it quietly; the incumbent leg's two ways of lying (bit depth, proxy races) are closed and checked rather than assumed; and the rate.48k note that records a ruled-out mechanism with the evidence that killed it is exactly how that file should read. The MPE fix is verified ordered-before-the-note-on on the member channel, the two-byte message fix charges the budget conservatively so the reservation invariant still holds, and the owed-offs arithmetic survives the emit change.

Three findings inline, all on the cell grid, which is the one piece of new engine machinery. Two body-level points:

**speed.ratio deserves louder billing than a calibration-list comment.** "The two disagree about what a speed ratio means with stretching switched off" is a content-level disagreement on a shipped, stretcher-free path - plain varispeed - and of the eight parked cases it is the only one that cannot be a timing subtlety. It should be an issue with a name before this merges, because inside the calibration list it reads like one more sub-sample mystery when it is a semantic one.

**The PR says Closes #2040 while eight of twenty-one cases are measured but not bounded.** The bidirectional set stops them regressing silently inside the code, but closing the issue makes the debt invisible outside it. Either keep #2040 open for calibration or open a successor issue that names the eight cases and what each is waiting on; the in-code comments are good enough that the issue can mostly quote them.

magda/engine/clip/ClipVoice.cpp
lucaromagnoli: Re-review at `202204e2`: this production issue remains. The new 96-vs-480 regression test now exposes it, but the change only moves `tempo.auto` into `knownDependent` and documents that the cell ratio is wrong across every tempo step. That converts the failure into an expected one without fixing playback. Please split the stretch cell at the tempo boundary (or evaluate both endpoints from the actual tempo map) before merging.

tests/test_null_diff_juce.cpp
```diff
@@ -0,0 +1,543 @@
...
+        // difference, and the material these cases play has one to correlate.
+        if (report.shiftCorrelationEnvelope < 0.98) {
+            logMessage("  " + juce::String(value.name) + ": the envelopes do not correlate (" +
+                       juce::String(report.shiftCorrelationEnvelope, 3) + ") at any offset");
+            return false;
+        }
+
+        const auto shift = static_cast<int>(std::llround(report.measuredShift));
+
+        report.envelope = compareEnvelopes(native.audio, incumbent.audio, shift, value.sampleRate);
```
lucaromagnoli, line 391: The new `nulled()` guard closes the short-render hole for Null cases, but this Shift path still ignores both `report.audio.lengthDifference` and `report.audio.refusal`. `compareEnvelopes` and `compareSpectra` each trim to the overlapping prefix, so a correctly aligned stretched render that stops early can still satisfy every check here. Please reject refused or unequal-length audio before applying the verdict-specific stretched metrics, and add the same truncated-buffer regression for a Shift verdict.

tests/NullDiffCompare.cpp
```diff
@@ -0,0 +1,1076 @@
...
+    const auto floorLinear = fromDb(options.floorDb);
+    auto peak = 0.0;
+    auto sumOfSquares = 0.0;
+
+    for (auto channel = 0; channel < native.getNumChannels(); ++channel) {
+        const auto* a = native.getReadPointer(channel);
+        const auto* b = incumbent.getReadPointer(channel);
+
+        for (auto index = begin; index < end; ++index) {
+            const auto residual =
```
lucaromagnoli: A non-finite sample can still make this comparator certify a perfect null. If either input is NaN, `residual` and `magnitude` are NaN; both `magnitude > peak` checks are false, so `peak` stays zero, `peakDb` becomes `-inf`, and `nulled()` passes. `rmsDb` becomes NaN but no verdict consults it. This also lets the ReportOnly case pass despite its contract saying the measurement must be finite. Please reject non-finite source/residual samples (as a refusal or explicit failure) and add an adversarial NaN-buffer test.

Issue #2077, opened 2026-08-11
Validation slice 3: differ property tests over random edit sequences


lucaromagnoli:
Slice 3 of #1896. Unblocked, and the highest-risk untested surface in the engine core.

Per the property-test paragraph on #1896: generate random structural edit sequences and assert
what the differ decided, rather than asserting the handful of edits I thought to write down.

The differ (#2014) and the crossfade of changed ops across a swap (#2019) are tested today by
cases I authored, which means they are tested against my reading of the compatibility matrix.
That is the same weakness #2040 was built to fix on the playback side, and it applies here with
more force: the differ's inputs are pairs of plans, the space of pairs is enormous, and the
failures that matter are the ones nobody thought to write.

Properties to assert over a generated sequence of edits:

- **State compatibility decisions.** An op judged compatible keeps its runtime state; an op
  judged incompatible is rebuilt. A voice that carried its position across an edit it should
  not have is a click at best and a wrong reading at worst.
- **Correct retirement.** Every retired op is retired exactly once, after the epoch that can
  still read it, and nothing retires while the audio thread can reach it.
- **PDC alignment.** Latency is consistent across the swap, and a change in one branch does not
  silently move another.
- **Silence and discontinuity bounds outside the changed region.** This is the one that catches
  the most: an edit to track 3 must be inaudible on track 1, sample for sample, and an edit to
  track 3 itself must not exceed the declared discontinuity bound. Rendering the same sequence
  with and without the unrelated edits and requiring a null on the untouched tracks is the
  assertion.

Failures shrink to a minimal edit sequence and print it as a reproducible case, because a
property test that reports "failed after 4000 random edits" has told you nothing you can act
on.

Seeds are recorded and a failing seed is checked in as a regression case.

Issue #1893, opened 2026-07-27
Engine: external plugin hosting and hardware inserts


lucaromagnoli:
Part of #1882. Depends on the engine core and the parameter system.

External plugin hosting on juce::AudioPluginInstance, plus hardware inserts.

- Async instantiation, latency reporting into PDC, wet/dry wrapper parameters.
- State restore sequencing at parity with today's three-step contract: saved parameter baseline, then native chunk as authoritative overlay, then parameter cache refresh. The new engine should make this a first-class API instead of requiring internal headers.
- Hardware insert as a graph-level send/return op pair (no special casing by chain position), plus insert render capture for bounce.
- The device op contract is location-transparent from day one: in-process and out-of-process are two implementations of the same op, so sandboxing later is an add-on, not a rework.

lucaromagnoli:
Scope note from breaking down #1896: this issue carries its own final parity slice, the way
#2040 was slice 7 of #1890.

Cases for what is ported here are written here, not in #1896. A parity case is worth what its
author's knowledge of the feature is worth, and that knowledge is at its peak on the day the
port lands. Written later by someone reading the code back, a case asserts the code rather than
the behaviour, which is the one thing a parity corpus must not do.

#1896 keeps the cross-cutting infrastructure the slice builds on: the whole-project case and the
tiered oracle (#2075), the goldens (#2076), the block-size gate (#2078), and the corpora that
span features (#2081, #2082).

PR #2083, opened 2026-08-12
Native engine: whole-project cases and the tiered oracle (#2075)

### LINKED ISSUE #2075: Validation slice 1: whole-project cases and the tiered comparison oracle
Slice 1 of #1896. Unblocked, and everything else here depends on it.

#2040 built the rig: two legs handed the same model values, a comparator with its own
known-bad tests, MIDI capture at the instrument's input, tempo and warp maps compared as
functions, and a canonical report printed on every run rather than only on failure. It works,
and it is shaped for one thing. A case is clip model values, the rig is fixed, and the verdict
vocabulary is null, shift, midi and report only.

This slice widens the case without widening the tolerances.

**A case becomes a project.** Tracks, devices, routing, sends and automation, not clips on a
bare track. The builders behind `NullDiffCorpus` grow accordingly, and the existing twenty-one
cases keep passing unchanged, because a refactor that quietly relaxes the clip corpus is a
regression wearing a green tick.

**A case records its environment**, per the tiered-oracle comment on #1896: sample rate, block
size, channel layout, random seeds, plugin versions, and its expected determinism class. Today
those are implicit in the runner. Left implicit, a null failure says more about the fixture
environment than about the engine, and the first time that happens someone spends a day inside
the engine looking for a bug that is in the harness.

**The verdict vocabulary becomes the tiers.** Five, each with what justifies it:

- Bit exact or a tight epsilon, for deterministic internal DSP and simple routing.
- Latency aligned numerical comparison, for PDC, sends, bypass and graph transitions.
- Time and frequency domain bounds, for stretch, warp and resampling. The stretch rows already
  work this way after #2040's calibration run: a pinned shift, envelope timing within a sample,
  and a magnitude bound with its window and hop stated.
- State, latency, event order, tail and no-discontinuity invariants, for nondeterministic
  external plugins, where nulling is not a question that has an answer.
- Scripted interactive checks, for launcher, monitoring, hardware input and recording, where an
  offline render is not representative of what the code does.

A case declares its tier and the mechanism that puts it there. A tier is a description of what
can be true about a comparison, never a way to make a failing case pass, and the rule #2040 set
still holds: change the shape of the comparison, never the size of the allowance.

The last two tiers have no cases yet, because the engine has nothing to put in them. They are
defined here so that #1893 through #1895 have somewhere to land, rather than inventing a sixth
shape each.

lucaromagnoli:
Closes #2075.

Three commits, and the first is not part of the slice. It is here rather than in
its own PR because a push runs CI, and two PRs would run it twice for one
afternoon's work.

## One clang-format, pinned

Three authorities disagreed about what formatted meant:

| | Version | Scope |
| --- | --- | --- |
| the commit hook | pinned 17.0.6 | every tracked file |
| `make format` | whatever is on PATH | `magda tests` |
| CI | Ubuntu's own apt package | `magda/daw magda/agents tests` |

So `make format` rewrote ten files, the hook put three of them back on the way
in, and CI had an opinion about neither. `magda/engine` and `magda/music` were
never format-checked at all.

All three now go through the pin that was already in `.pre-commit-config.yaml`
and that nothing was using. The ten files this normalises were committed
unformatted under any of the three versions, which is what happens when the
check that would have caught them was looking somewhere else. `make format`
twice in a row is now a no-op.

## The slice

A case was a track with clips on it, judged by one of four verdicts, rendered at
unity through a default master. The mixer was never touched, which means the
fader law, the pan law, mute inheritance and solo through destination routing
have never been compared against the incumbent at all.

**The four verdicts conflated two independent questions.** `Midi` meant "no audio
assertion", so a project with an instrument track and audio tracks had to pick
one. They separate: the audio tier is a determinism class, declared because of
what stands between the two engines rather than as a preference, and whether the
MIDI streams are compared is its own flag.

| Tier | What it asserts | Cases |
| --- | --- | --- |
| `None` | nothing; the MIDI comparison carries it | the four `midi.*` |
| `Exact` | residual under the floor, nothing allowed for | twenty |
| `Aligned` | one declared offset, undone, then `Exact` | none yet |
| `Spectral` | pinned shift, envelope timing, magnitude bound | the four stretched |
| `Invariants` | finite, equal length, bounded step, decayed tail | none yet |
| `Measured` | measured and printed, asserted only to be finite | `stretch.broadband` |

`Invariants` is implemented and tested against known-bad pairs with no corpus
case using it. An external plugin owes nobody a sample, so #1893 needs this
instead of a residual, and the same checks are what #2077 asserts outside a
changed graph region. Neither it nor `Aligned` can be declared loosely: a case
without a declared bound, or without a declared offset, is refused rather than
passed.

`Scripted`, the fifth class on #1896, is deliberately not a tier. A launcher or a
monitoring case is not an offline render of a range, so it needs a different
runner, and a value in this enum would be a promise the runner does not keep.

**Six mixer cases**, the first in the corpus with more than one track: summing,
the fader across its range including silence, the pan law at its ends, mute,
solo, and the master's own fader and pan. All null.

That falsifies the one prediction worth making here. I expected `mix.pan` to fail
if the incumbent's `VolumeAndPanPlugin` were on a constant-power law where MAGDA
is linear, which would have meant a pad sitting 3 dB down through a rack. It is
linear on both sides.

The incumbent leg drives them through the same four calls
`AudioBridge::trackPropertyChanged` makes, in the same order, because the sync
layer is part of what is being validated.

**The environment is recorded on the case** and printed wherever it deviates from
the corpus default, so a number measured at another rate can never be read as a
number measured here.

## What is left out, and why

**Sends.** The compiler emits `SendTap` pre and post fader and the value layer
resolves the levels, so the native leg could render one today. The incumbent's
sends live on `te::AuxSendPlugin` instances that `PluginManagerSync` creates from
a `PluginManager` and the device layer behind it, and writing those plugins
straight into the leg would be the second sync this corpus refuses to have. They
belong with the rest of the routing graph, in #1892.

**Real devices.** Nothing yet runs one under both engines, so no case carries one.

The eight cases already under calibration from #2040 still fail exactly as they
did, asserted in both directions as before.

## Verified

`make format` is idempotent, `make debug` is clean, 2522 Catch2 cases pass, and
the corpus runs 27 of 27 through both engines.


.github/workflows/ci.yml
```diff
@@ -297,31 +297,21 @@ jobs:
...
-        sudo apt-get update -o Acquire::Retries=3
-        sudo apt-get install -y clang-format
+    # Formatting is checked with the clang-format version pinned in
+    # .pre-commit-config.yaml, which is the same one the commit hook and
+    # 'make format' now use. Installing the runner's own clang-format made this
+    # job a third opinion: a different version over a narrower set of
+    # directories than either of the other two, so it passed files the hook
+    # would rewrite and never looked at magda/engine or magda/music at all.
+    - name: Install pre-commit
+      run: pipx install pre-commit
```
lucaromagnoli, line 307: The "Cache apt packages" step above this is now caching archives nothing installs. Worth dropping it, or repointing it at ~/.cache/pre-commit keyed on hashFiles('.pre-commit-config.yaml') so the clang-format hook environment isn't rebuilt from scratch on every run.

Makefile
```diff
@@ -362,13 +362,20 @@ rebuild: clean debug
...
+# happens to be on PATH means every developer formats to their own version, the
+# commit hook puts the file back on its way in, and the pair churn forever
+# against each other. One pinned binary, one set of files, used by this target,
+# by the hook and by CI.
 format:
 	@echo "🎨 Formatting code..."
-	@if command -v clang-format >/dev/null 2>&1; then \
-		find magda tests \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) | xargs clang-format -i; \
+	@if command -v pre-commit >/dev/null 2>&1; then \
+		pre-commit run clang-format --all-files || true; \
```
lucaromagnoli: One behavioural change worth knowing about: --all-files means all *tracked* files, so a brand-new file that hasn't been `git add`ed is skipped here, where the old `find` formatted it. The commit hook still catches it once staged, so this is fine as is; noting it because "make format did nothing to my new file" will look like a bug the first time it happens.

docs/architecture/native-engine.md
```diff
@@ -518,7 +531,98 @@ there while its session path applies it, which is a gap in the sync layer rather
...
+| Tier | What it asserts | Where it applies |
+| --- | --- | --- |
+| `None` | nothing; the MIDI comparison carries the case | the four `midi.*` |
+| `Exact` | residual under the floor, nothing allowed for | deterministic DSP, routing, the mixer |
+| `Aligned` | one declared offset, undone, then `Exact` | anything whose whole effect is a delay |
+| `Spectral` | pinned shift, envelope timing, magnitude bound | a phase vocoder in the path |
+| `Invariants` | finite, equal length, bounded step, decayed tail | a plugin that owes nobody a sample |
+| `Measured` | measured and printed, asserted only to be finite | `stretch.broadband` |
+
+`Aligned` and `Invariants` have no corpus case yet and are implemented and tested anyway.
```
lucaromagnoli: Half of this claim is ahead of the code. Invariants has the six known-bad-pair tests, but nothing anywhere constructs an AudioTier::Aligned case: its refusal path and its judged path exist only in the runner's switch and have never executed. Either soften this to Invariants, or give Aligned the same known-bad-pair treatment; as written, the day someone declares the first Aligned case is also the first time that logic runs.

lucaromagnoli (commented):
The tier split is clean, and I checked the leg's sync claims against the code rather than taking the comments' word: trackPropertyChanged makes those four calls in that order, AudioBridgeMixer is the pass-through it says it is, and the master resolution matches AudioBridgeMixer::setMasterVolume exactly. Five small things inline: two on the format toolchain, a doc overclaim about Aligned, a message wording, and a report gap on the mixed audio+MIDI cases this redesign exists to enable.

tests/NullDiffCompare.cpp
```diff
@@ -703,6 +703,132 @@ AudioComparison compareAudio(const juce::AudioBuffer<float>& native,
...
+    result.worstStepAt =
+        result.worstStepNative >= result.worstStepIncumbent ? nativeAt : incumbentAt;
+
+    result.continuous = result.worstStepNative <= options.maxStepPerSample &&
+                        result.worstStepIncumbent <= options.maxStepPerSample;
+    if (!result.continuous)
+        result.problems.push_back(
+            "a step of " +
+            formatDb(toDb(std::max(result.worstStepNative, result.worstStepIncumbent))) +
+            " dBFS at sample " + std::to_string(result.worstStepAt) + " passes the bound of " +
```
lucaromagnoli: This prints when the step failed the bound, but "passes the bound" reads as satisfying it. "exceeds the bound" says what happened.

tests/NullDiffCompare.cpp
```diff
@@ -1067,22 +1197,41 @@ std::string formatDb(double db) {
...
+                line, sizeof(line), "step=%-9s tail=%-9s length=%s",
+                formatDb(toDb(std::max(invariants.worstStepNative, invariants.worstStepIncumbent)))
+                    .c_str(),
+                formatDb(std::max(invariants.tailPeakNativeDb, invariants.tailPeakIncumbentDb))
+                    .c_str(),
+                invariants.lengthsMatch ? "ok" : "differs");
+            out << line;
+            if (!invariants.refusal.empty())
+                out << " refused: " << invariants.refusal;
         } else if (report.hasMidi) {
```
lucaromagnoli: This chain prints exactly one comparison per case, which was right while the verdict was exclusive, but the point of this slice is that audio and MIDI are now independent assertions. The first case that asserts both will print notes=/cc= and drop the residual, and an Invariants case with MIDI prints only step/tail. `passed` still ANDs both, so judgment is unaffected, but the number that moved goes missing from the report on exactly the mixed cases the redesign enables. Worth making the MIDI segment additive before the first instrument-plus-audio project lands.

tests/NullDiffCorpus.cpp
```diff
@@ -531,6 +573,144 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
...
+
+    {
+        // The fader across its range, which is a slider position rather than a
+        // gain: unity, six down, and the top of the range, which is six up
+        // rather than twice. Silence is in here too, and it is the interesting
+        // one: the position floors at -100 dB, so whether that is zero or merely
+        // very quiet is a thing the two engines can disagree about at a level
+        // the floor would catch.
+        auto value = newMixCase(
+            "mix.volume", "the fader law: unity, attenuation, the top of the range, and silence",
```
lucaromagnoli: **P1 — Do not certify cases that violate the Tracktion graph invariant.** On the refreshed head, running `Null Diff Corpus` still emits `JUCE Assertion failure in tracktion_NodePlayerUtilities.h:123` for both `mix.volume` and `mix.pan`; the duplicate ID is reported for two `ArrangerLauncherSwitchingNode`s. Because this is a non-fatal `jassert`, the suite still reports all 27 cases passed. Please fix the duplicate node IDs or restructure these cases, and make this class of assertion fail the test rather than certifying an invalid graph.

tests/NullDiffCase.hpp
```diff
@@ -57,7 +57,18 @@ struct Case {
...
+    /// declared because of what the signal goes through, and the mechanism
+    /// below says what that is.
+    AudioTier tier = AudioTier::Exact;
+
+    /// Whether the captured MIDI streams are compared.
+    ///
+    /// Independent of the tier, because they are independent questions. A
+    /// project with an instrument track and audio tracks asserts both, and the
+    /// four MIDI-only cases assert this one with the audio tier at None.
+    bool compareMidiStreams = false;
```
lucaromagnoli, line 71: **P2 — The incumbent leg does not yet support the mixed projects this flag enables.** `renderIncumbent` always inserts one `MidiCapturePlugin` on `value.tracks.front()`, while the native leg aggregates every device capture. If the instrument track is not first, or if a project has multiple instrument tracks, the incumbent stream is empty or incomplete. Please attach captures to the tracks/devices that actually consume MIDI and aggregate them symmetrically.

tests/NullDiffCorpus.cpp
```diff
@@ -531,6 +573,144 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
...
+            {mixTrack(1, "Unity"), mixTrack(2, "Down"), mixTrack(3, "Up"), mixTrack(4, "Silent")});
+
+        const std::array<float, 4> volumes{1.0f, 0.5f, 2.0f, 0.0f};
+        for (auto index = 0; index < 4; ++index) {
+            value.tracks[static_cast<std::size_t>(index)].volume =
+                volumes[static_cast<std::size_t>(index)];
+
+            const auto source = writeSource(scratchDirectory, "vol" + juce::String(index), steps());
```
lucaromagnoli: **P2 — Give each fader/pan track distinguishable material.** Every `mix.volume` track receives the same step waveform, and `mix.pan` repeats the same construction below. Swapping fader or pan values between track IDs therefore leaves the summed render unchanged and passes even though the value-to-track mapping is wrong. The ambiguity also produces `shift=-32` for both successful cases in the local report. Please use distinct patterns per track, as `mix.summing` already does.

tests/NullDiffCorpus.cpp
```diff
@@ -531,6 +573,144 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
...
+        }
+
+        corpus.push_back(std::move(value));
+    }
+
+    {
+        // The master's own fader and pan, applied once to the sum. Two tracks
+        // rather than one, so that a master gain applied per track instead of
+        // once to the mix would be a different number rather than the same one.
+        auto value = newMixCase("mix.master", "master volume and pan, applied once",
```
lucaromagnoli: **P2 — This render cannot prove that master processing is applied once.** Gain and linear pan distribute over addition: `M(a + b) == M(a) + M(b)`. Two tracks therefore produce the same samples whether the master matrix is applied after the sum or separately to every input, contrary to the comment's stated coverage. This needs a structural assertion or a non-linear/order-sensitive element if placement is part of the claim.

Makefile
```diff
@@ -362,13 +362,26 @@ rebuild: clean debug
...
+# file that has not been git added yet is skipped here where the old find would
+# have formatted it. The commit hook catches it the moment it is staged, so
+# nothing reaches a commit unformatted; "make format did nothing to my new file"
+# is this, and not a bug.
 format:
 	@echo "🎨 Formatting code..."
-	@if command -v clang-format >/dev/null 2>&1; then \
-		find magda tests \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) | xargs clang-format -i; \
+	@if command -v pre-commit >/dev/null 2>&1; then \
+		pre-commit run clang-format --all-files || true; \
```
lucaromagnoli: **P2 — Do not swallow every formatter failure.** Exit 1 is expected when pre-commit rewrites files, but `|| true` also hides hook-environment failures, download errors, crashes, and invalid configuration while printing success. One safe pattern is to run the hook once to apply changes and then require a second invocation to succeed.

lucaromagnoli (commented):
Five remaining findings after refreshing against `fdabdbb9`. The audio+MIDI reporting issue from the earlier review is fixed and omitted here. I rebuilt both null-diff targets and reran them; Catch2 passed 970 assertions, while the JUCE corpus still logged the duplicate-node assertions described inline.

tests/test_null_diff_corpus.cpp
lucaromagnoli, line 1: **P2 — Validate the actual MIDI-consuming track**

This still assumes that a MIDI-comparison case has its device on `tracks.front()`. The new incumbent capture logic intentionally supports an audio track followed by an instrument track, but that valid mixed case would fail this corpus-shape test before it can run. Please check whether any declared track satisfies `chainConsumesMidi()` instead of inspecting the first track. (This refers to the check at lines 164–165.)

tests/test_null_diff_juce.cpp
```diff
@@ -49,6 +49,59 @@ juce::File scratchDirectory() {
...
+
+  private:
+    void logMessage(const juce::String& message) override {
+        // The exact prefix JUCE writes, not a substring match. This logger is
+        // the current one while the corpus runs, so the runner's own reporting
+        // passes through it too: a substring match reads the line that says a
+        // case asserted as another assertion, and every case after the first
+        // inherits the one before it.
+        if (message.startsWith("JUCE Assertion failure"))
+            assertions_.push_back(message);
```
lucaromagnoli: **P2 — Synchronize assertion collection**

`Logger` is application-wide, so `logMessage()` may run on an engine/background thread while the test thread calls `take()`. Appending to `assertions_` concurrently with moving/clearing it is a data race and can corrupt or lose exactly the assertion this watcher is meant to surface. Please protect the buffer in both methods with a mutex.

tests/test_null_diff_juce.cpp
```diff
@@ -49,6 +50,69 @@ juce::File scratchDirectory() {
...
+ * which is the right answer for a build where the claim was never made.
+ */
+class AssertionWatch final : public juce::Logger {
+  public:
+    AssertionWatch() : previous_(juce::Logger::getCurrentLogger()) {
+        juce::Logger::setCurrentLogger(this);
+    }
+
+    ~AssertionWatch() override {
+        juce::Logger::setCurrentLogger(previous_);
```
lucaromagnoli: **P2 — Keep the installed logger alive until callbacks are quiescent**

The mutex protects `assertions_`, but this destructor still replaces and destroys a process-wide logger while engine threads may call `Logger::writeToLog()`. JUCE stores `currentLogger` as an unsynchronized raw pointer, so a thread can read `this`, lose the race to this teardown, and then call `logMessage()` on the destroyed watcher. Please keep the watcher alive through engine/test-process shutdown, or otherwise guarantee all logger callbacks have stopped before restoring the previous logger.

tests/test_null_diff_juce.cpp
```diff
@@ -49,6 +50,82 @@ juce::File scratchDirectory() {
...
+    std::vector<juce::String> take() {
+        const std::lock_guard<std::mutex> lock(mutex_);
+        auto taken = std::move(assertions_);
+        assertions_.clear();
+        return taken;
+    }
+
+  private:
+    AssertionWatch() {
+        juce::Logger::setCurrentLogger(this);
```
lucaromagnoli: **P2 — Install the logger before any worker threads start**

Keeping the watcher alive fixes teardown, but `instance()` is first called from `NullDiffCorpusTests::runTest()`. In a full `magda_juce_tests` run, earlier suites may already have initialized the shared engine and its long-lived threads. This call then writes JUCE’s unsynchronized `currentLogger` pointer while one of those threads may read it in `writeToLog()`, which is still a data race. Please install the process-lifetime watcher from `main()` before `ScopedJuceInitialiser_GUI` or any engine/thread initialization.

tests/AssertionWatch.hpp
```diff
@@ -0,0 +1,88 @@
...
+    // Never runs. Private so that nobody can stack-allocate one and reintroduce
+    // the teardown this type exists to avoid.
+    ~AssertionWatch() override = default;
+
+    void logMessage(const juce::String& message) override {
+        // The exact prefix JUCE writes, not a substring match. This is the
+        // current logger for the whole process, so a test reporting that a case
+        // asserted passes through here too, and a substring match would read
+        // that line as another assertion.
+        if (message.startsWith("JUCE Assertion failure")) {
```
lucaromagnoli, line 73: **P2 — Exercise the assertion safety net in a test**

No test currently emits a captured assertion: every corpus case is assertion-free now, so this prefix/filter, the process-wide installation, `take()`, and the branch that turns a render into `unmeasurable` can all regress while CI remains green. Since this mechanism exists specifically to prevent an invalid graph from being certified, please add a focused test that writes a synthetic `JUCE Assertion failure...` message and proves it is captured (ideally also that the corpus verdict is failed/unmeasurable).

PLAN-2075-project-cases.md
```diff
@@ -0,0 +1,134 @@
...
+audio tracks cannot say what it wants: it has to pick one. So the two questions separate.
+
+**What is asserted about the audio** becomes a tier, which is the determinism class from the
+issue rather than a per-case preference:
+
+| Tier | What it asserts | Cases today |
+| --- | --- | --- |
+| `None` | nothing; the case is carried by MIDI | the four `midi.*` |
+| `Exact` | residual under the floor, nothing allowed for | fourteen |
+| `Aligned` | one pinned offset, measured at the start, then `Exact` | `midi.offset`, and the three still calibrating |
```
lucaromagnoli: **P3 — Keep the tier table aligned with the corpus**

There are five `midi.*` cases, not four, and `midi.offset` remains `AudioTier::None` with a declared MIDI shift; it is not an `Aligned` audio case. The table currently says both the opposite tier and the wrong count, which makes the new vocabulary harder to use as its design reference. Please move `midi.offset` back to the `None` row and update the count here and in the architecture table.

lucaromagnoli (commented):
Two follow-up findings after the latest fix.

docs/architecture/native-engine.md
```diff
@@ -518,7 +531,104 @@ there while its session path applies it, which is a gap in the sync layer rather
...
+allowed anything.
+
+The issue's fifth class, scripted interactive checks, is deliberately not a tier. A launcher or
+a monitoring case is not an offline render of a range, so it needs a different runner rather
+than a different verdict.
+
+**The mixer** is where the corpus first has more than one track. `resolvePlanValues` implements
+the fader law, the linear pan law, mute inheritance and solo through destination routing, and
+until [#2075](https://github.com/Conceptual-Machines/magda-core/issues/2075) none of it had been
+compared against the incumbent at all. Six cases now do: summing, the fader across its range
```
lucaromagnoli: **P3 — Update the corpus and mixer case counts**

This section still describes the pre-slice corpus as current: line 539 says the corpus has 21 projects, and this paragraph says six mixer cases. `sharedCorpus()` now returns 28 projects, including seven `mix.*` cases because the fader-bottom check was split into `mix.volume.silent`. Please update both counts so the architecture document matches the canonical report (`cases=28`) and the builders.

tests/NullDiffCorpus.cpp
```diff
@@ -540,7 +774,8 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
             clip.midiNotes.push_back(note(60 + index, static_cast<double>(index),
                                           index % 2 == 0 ? 0.75 : 0.5, 40 + index * 10));
         value.clips.push_back(std::move(clip));
-        value.verdict = Verdict::Midi;
+        value.tier = AudioTier::None;
```
lucaromagnoli, line 905: **P2 — Exercise a real mixed audio+MIDI project**

Every case that enables `compareMidiStreams` also sets `AudioTier::None`, and every audio case leaves the MIDI flag off. As a result, the corpus still never runs both assertions for one project, and all five MIDI cases keep the instrument as the only/first track. The new non-first/multiple-track capture placement in both render legs, the aggregation, and the final `audio && midi` verdict can therefore regress while the corpus and the synthetic report test remain green. Please add at least one real case with an audio track first and an instrument track second (ideally also two MIDI-consuming tracks) and a non-`None` audio tier so this redesign is exercised end to end.

tests/test_null_diff_juce.cpp
```diff
@@ -274,19 +275,43 @@ class NullDiffCorpusTests : public juce::UnitTest {
...
+        for (const auto& value : corpus) {
+            auto report = run(value);
+
+            // An engine that objects to its own graph has not produced a render
+            // worth comparing, whatever the residual says. Reported as
+            // unmeasurable rather than as a residual, for the same reason a
+            // proxy that never arrived is: the number would be about something
+            // other than parity, and it would look like a parity failure.
+            if (auto fired = watch.take(); !fired.empty()) {
+                report.passed = false;
```
lucaromagnoli, line 370: **P1 — Make engine assertions fail even calibration cases**

Setting `report.passed = false` is not enough for the eight entries in `underCalibration`: the final loop explicitly expects each of those names to be in `failing`, so an assertion in `tempo.auto`, `warp.audio`, or any stretched calibration case merely replaces its expected comparison failure with an `unmeasurable` report and the JUCE suite still exits successfully. That lets the exact invalid-graph condition this watcher was added to prevent remain green for a third of the corpus. Track assertion occurrence separately and call `expect(false, ...)` (or otherwise fail unconditionally) before applying the calibration expectations; please add a test covering this interaction too.

tests/test_null_diff_juce.cpp
```diff
@@ -341,13 +421,37 @@ class NullDiffCorpusTests : public juce::UnitTest {
             if (!value.passed)
                 failing.insert(value.name);
 
-        for (const auto& value : reports)
-            if (underCalibration.count(value.name) == 0)
-                expect(value.passed, juce::String(value.name) + " did not hold");
+        const auto complaints = judgeSuite(asserted, failing, underCalibration);
```
lucaromagnoli: **P1 — Do not let calibration forgive unmeasurable cases**

Assertions are separated now, but every other refusal is still collapsed into `failing` before `judgeSuite`: native/incumbent render failures, proxy timeouts, plan diagnostics, fixed-point output, comparator refusals, and length mismatches all set `unmeasurable` and leave `passed == false`. If any of those happens to `tempo.auto` or another calibration entry, its name appears in `failing`, satisfies the expected-failure membership, and the suite stays green despite producing no comparison number at all. The calibration list is meant to forgive an explained comparison result without a bound, not an inability to measure. Please track `report.unmeasurable` separately and fail it unconditionally alongside assertions, with a rule test for an unmeasurable calibration case.

tests/NullDiffCorpus.cpp
```diff
@@ -531,6 +589,244 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
...
+        // have regressed with every case still green.
+        //
+        // An audio track first, then two instrument tracks. The order is the
+        // point of the first: a leg that put its capture on tracks.front() would
+        // hand back an empty stream against a full one. The second instrument is
+        // the point of the pair: one capture read where two exist compares half
+        // a project against all of it.
+        //
+        // The two instruments play different pitches at different instants, so a
+        // stream attributed to the wrong track, or one of the two dropped, is a
```
lucaromagnoli: **P2 — Preserve the capture target when comparing multiple instruments**

The claim here is not currently true. `MidiEvent` has no track/device identity, and both render legs concatenate every capture into one flat `MidiStream` before `compareMidi()`. If Synth A receives Synth B's four notes and Synth B receives Synth A's four notes, the aggregate still contains the same eight timestamp/status/data tuples and `project.mixed` passes; distinct pitches only detect a missing stream, not attribution to the wrong track. Since this case is meant to validate per-track capture placement, please compare streams keyed by track/device (or tag captured events with that identity) before aggregating for reporting.

tests/NullDiffNativeLeg.cpp
```diff
@@ -228,6 +234,7 @@ NativeRender renderNative(const Case& value) {
                 auto device = std::make_unique<MidiCapture>(value.sampleRate);
```
lucaromagnoli: **P2 — Capture once per track, not once per Device op**

The incumbent leg inserts one capture at the head of each MIDI-consuming track, but the native leg creates a capture for every `OpKind::Device` and later concatenates all of those captures into one `midiByTrack[trackId]`. On a track with an instrument followed by another device that receives chain MIDI, both Device ops can see the same incoming clip stream (`emitDevice` preserves `signal.midi` unless a device replaces it), so the native result contains each note twice while the incumbent result records it once. More generally, these are different comparison points: one per track versus every device stage. The new per-track map therefore still rejects a valid multi-device project. Please bind an explicit capture once at the equivalent track/chain-head point, or retain device/stage identity and compare like-for-like; a two-device MIDI case would pin this down.

tests/NullDiffNativeLeg.cpp
```diff
@@ -228,6 +237,24 @@ NativeRender renderNative(const Case& value) {
...
+                // head of the track's plugin list, so that is what it records:
+                // the MIDI arriving at the chain, once.
+                //
+                // Reading every device instead would compare a different thing
+                // and get the wrong answer for it. A chain whose MIDI survives
+                // its first device hands the same notes to the second, and
+                // emitDevice preserves signal.midi unless a device replaces it,
+                // so a two-instrument track would report each note twice
+                // against the incumbent's once.
+                if (!recordedDevice.contains(op.key.trackId)) {
```
lucaromagnoli: **P2 — Do not choose an audio-only Device op as the MIDI tap**

`recordedDevice` is populated for the first Device op on every track, regardless of whether that op has a connected MIDI input. A valid chain such as audio effect → instrument therefore selects the audio-only effect: its capture sees an empty `midiIn`, while the later instrument capture that actually receives the clip is discarded. The incumbent capture at the plugin-list head still records the notes, so the legs disagree solely because of the harness. There is a second manifestation in mixed projects: an audio-only track with any device gets an empty `midiByTrack` entry here, while the incumbent only creates entries for `chainConsumesMidi` tracks, and the runner reports that track as captured by the engine only. Please restrict the designated tap to a capture-eligible track and an op with the equivalent incoming MIDI (for example, validate `op.inputs[1]`/the chain-consumer predicate), and cover an audio effect before an instrument.

tests/NullDiffNativeLeg.cpp
```diff
@@ -225,10 +248,51 @@ NativeRender renderNative(const Case& value) {
...
+                // nothing, so it passes signal rather than recording or
+                // clearing it.
+                //
+                // That one test is also what keeps a track the incumbent never
+                // captures from getting an entry here. A track whose chain
+                // consumes no MIDI compiles no ClipMidi op, so there is no chain
+                // MIDI for any of its devices to be wired to, and an audio track
+                // carrying an effect is skipped without having to ask the model
+                // a second question about it.
+                const auto readsChainMidi = op.inputs.size() > 1 && op.inputs[1].valid();
```
lucaromagnoli: **P2 — A routed MIDI source can have a connected slot without consuming MIDI**

The premise above is not always true. `emitTrack()` also compiles `ClipMidi` when `midiSourceTracks_` contains the track (for example, another track has an active MIDI-input route from it), even if `chainConsumesMidi(track)` is false. Since every Device op currently receives chain MIDI, an audio-effect-only source track then satisfies `readsChainMidi` and gets a native `midiByTrack` entry. The incumbent leg still skips that same track at `if (!chainConsumesMidi(track)) continue`, so a project with a MIDI clip/effect on the routed source and an instrument on the destination reports the source as captured by the engine only. Please gate taps by the same capture-eligible track set as the incumbent, in addition to requiring a connected slot, and add a routed-source case.

tests/NullDiffNativeLeg.cpp
```diff
@@ -225,10 +260,50 @@ NativeRender renderNative(const Case& value) {
...
+                //
+                // Everything else on the track stands where the incumbent has
+                // nothing, so it passes signal rather than recording or
+                // clearing it.
+                //
+                // Both tests are needed and neither implies the other. A wired
+                // slot without eligibility is a routed MIDI source; eligibility
+                // without a wired slot is a track that consumes MIDI and has no
+                // clip to feed it.
+                const auto readsChainMidi = op.inputs.size() > 1 && op.inputs[1].valid();
```
lucaromagnoli: **P2 — Preserve empty capture streams for eligible tracks**

The comment names an eligible track with no wired slot, but the implementation still makes the two legs disagree on exactly that case. The incumbent creates one `MidiCapturePlugin` for every `chainConsumesMidi` track and later executes `result.midiByTrack[trackId]` even when it captured zero events, so an empty stream is present. Here `isTap` is false when there is no MIDI source and no entry is ever added. The runner treats a missing map entry as a one-leg-only capture, not as an empty stream, so a mixed project containing an otherwise valid but empty instrument track fails solely because of the harness. Please initialize `midiByTrack` for every eligible track (or normalize empty streams symmetrically), and cover an eligible track with no clip/input beside a populated one.

.github/workflows/ci.yml
```diff
@@ -791,9 +843,41 @@ jobs:
...
+          # PowerShell, and what reaches here is only an exit code.
+          $bootstrapped = $false
+          foreach ($attempt in 1..3) {
+            try {
+              & "$vcpkgRoot\bootstrap-vcpkg.bat" -disableMetrics
+            } catch {
+              Write-Host "vcpkg bootstrap threw: $_"
+            }
+
+            if (Test-Path "$vcpkgRoot\vcpkg.exe") { $bootstrapped = $true; break }
```
lucaromagnoli: **P2 — Do not treat a failed bootstrap's output path as success**

`Test-Path` is weaker than the validation the bootstrap just performed. At this pinned commit, `scripts/bootstrap.ps1` passes `$vcpkgRoot\vcpkg.exe` directly to `tls12-download.exe`, then checks the downloader exit code and finally runs `vcpkg.exe version`. A failed/interrupted transfer can therefore leave the destination path present, and any failure after creating it still returns nonzero through the batch file. This loop ignores that exit code, immediately declares success, skips the remaining attempts, and then copies the unvalidated file into the tool cache. Please require `$LASTEXITCODE -eq 0` and/or run `& $exe version --disable-metrics` successfully before accepting/caching it; remove the failed destination before retrying so the next attempt cannot inherit it.

.github/workflows/ci.yml
```diff
@@ -791,9 +843,70 @@ jobs:
...
+          Copy-Item "$toolCache\vcpkg.exe" "$vcpkgRoot\vcpkg.exe" -Force
+          if (Test-VcpkgTool "$vcpkgRoot\vcpkg.exe") {
+            Write-Host "vcpkg.exe restored from the tool cache"
+          } else {
+            Write-Host "the cached vcpkg.exe does not run, bootstrapping instead"
+            Remove-Item "$vcpkgRoot\vcpkg.exe" -Force -ErrorAction SilentlyContinue
+          }
+        }
+
         if (-not (Test-Path "$vcpkgRoot\vcpkg.exe")) {
```
lucaromagnoli, line 889: **P2 — Validate an existing tool before skipping bootstrap**

This guard still treats the persistent `$vcpkgRoot\vcpkg.exe` as valid solely because it exists. On the self-hosted runner that root survives across runs, and a third failed bootstrap attempt leaves whatever partial destination it produced in place before this step throws. The next run therefore skips both the validated cache restore and the entire retry loop, copies the bad file into `$toolCache`, and only fails later at `vcpkg install`. The same applies to any corrupt tool already present in `C:\\vcpkg-ci`. Please run `Test-VcpkgTool` on the existing root executable first and remove it when validation fails, so this guard is reached only for a working tool.

PR #2088, opened 2026-08-13
Plan goldens in CI (#2076)

### LINKED ISSUE #2076: Validation slice 2: plan goldens in CI
Slice 2 of #1896. Unblocked.

`dumpPlan` and `dumpClipSnapshot` are canonical text already, and #2010 built them for exactly
this. What exists today is a handful of raw-string comparisons inline in
`tests/test_render_plan_compiler.cpp`, which pin two plans and no more.

This slice turns that into a fixture corpus: model fixtures in, checked-in snapshots beside
them, a regeneration path for when a change to the plan is intended, and a CI entry that fails
on an unintended one.

The value is in the failure mode it catches, which the null-diff corpus cannot. A render says
the output is wrong. A plan golden says which of the compiler's decisions changed, before any
audio exists, and a diff of two canonical dumps is readable in a way a waveform is not.

**The dump has to exclude anything unstable**, per the goldens comment on #1896: addresses,
allocation order, iteration order over unordered containers, and anything else that is an
implementation detail rather than a decision. If a golden churns for reasons nobody caused,
it gets regenerated on sight and stops being a test. `dumpPlan` is believed to be stable
already, since #2010 asserts two compiles of the same model print identically. Believed is not
the same as pinned, so this slice pins it: the same fixture compiled twice in one process, and
compiled in two processes, prints the same text.

Fixtures cover what the compiler actually decides: ordering, buffer assignment, latency
compensation, the differ's op identities, and the crossfade ops from #2019.

lucaromagnoli:
Closes #2076. Slice 2 of #1896.

Eleven fixtures in `tests/goldens/plan`, one file each: the compiled plan, the layout it prepares to, and for the edit fixtures the edited plan, the diff, and the crossfaded result.

Regenerate with `MAGDA_UPDATE_PLAN_GOLDENS=1 ./magda_tests "[goldens]"`. A normal run cannot rewrite them, which is the point - the diff is the reviewable artefact, and a golden that gets regenerated without being read has stopped being a test.

CI already runs `magda_tests` bare, so the goldens fail the existing job on an unintended change. No new entry needed.

## Two dumps that did not exist

`dumpPlan` was already there. Buffer assignment and latency could not be pinned without these:

- `dumpPlanLayout` - delay hold, port slots, in-place writes and elided delays, against the device latencies they were resolved with.
- `dumpPlanDiff` - what carried and what retired, by **op key** rather than op index. Inserting a device moves every index behind it, so an index-keyed golden would churn on an edit that carried correctly.

## Stability

Pinned, not assumed. Every fixture compiles twice in-process and the two dumps are compared; the files are the cross-process half, since each was written by a different run than the one asserting it.

Two more tests keep the corpus honest: fixture names must be unique, and every file on disk must belong to a fixture - a renamed fixture cannot leave a stale golden behind that still looks like coverage.

## What the fixtures cover

`bare-track`, `device-chain`, `instrument-track`, `fan-in`, `cascaded-latency`, `rack-chains`, `sidechain`, `group-routing`, and three edits: `edit-insert-device`, `edit-remove-device`, `edit-reorder-devices` (the last one is the crossfade case from #2019 - reordering moves an edge without adding or removing anything, so every fade in the golden is one the pass chose).

## Note on the trailing-whitespace handling

Both new dumps trim trailing spaces and the writer names its line endings explicitly. Without that, the repo's whitespace hook and JUCE's CRLF default each rewrite every golden on first commit, and every run afterwards compares against a file it did not write. Both were hit and fixed rather than reasoned about.

## Verified

Full Catch2 suite: 2547 cases, 571846 assertions. A deliberately corrupted golden fails with the offending line named.

### DIFF
--- a/magda/engine/exec/PlanLayoutDump.cpp
+++ b/magda/engine/exec/PlanLayoutDump.cpp
@@ -2,21 +2,14 @@
 
 #include <sstream>
 
+#include "plan/DumpFormat.hpp"
+
 namespace magda::engine {
 namespace {
 
-std::string padded(std::string text, std::size_t width) {
-    if (text.size() < width)
-        text.append(width - text.size(), ' ');
-    return text;
-}
-
-std::string rightAligned(int value, std::size_t width) {
-    auto text = std::to_string(value);
-    if (text.size() < width)
-        text.insert(text.begin(), static_cast<long>(width - text.size()), ' ');
-    return text;
-}
+using dump_format::padded;
+using dump_format::rightAligned;
+using dump_format::writeLine;
 
 /// An op's output ports, as the arena and slot each renders into.
 std::string describePorts(const RenderPlan& plan, const BufferLayout& layout,
@@ -42,7 +35,7 @@ std::string describePorts(const RenderPlan& plan, const BufferLayout& layout,
 ///
 /// An op with no output ports has no latency to report rather than a latency
 /// of zero, and printing zero for it would read as "nothing reaches this",
-/// which of the hardware output is the opposite of true.
+/// which for the hardware output is the opposite of true.
 std::string describeLatency(const RenderPlan& plan, const PlanLatency& latency,
                             const std::vector<int>& portOffsets, std::size_t op) {
     if (plan.ops[op].outputs.empty())
@@ -52,20 +45,15 @@ std::string describeLatency(const RenderPlan& plan, const PlanLatency& latency,
     return port < latency.portLatency.size() ? std::to_string(latency.portLatency[port]) : "-";
 }
 
-/// Trailing spaces would not survive the repo's whitespace hook, which would
-/// rewrite every golden the first time one was committed.
-void writeTrimmed(std::ostringstream& out, std::string line) {
-    while (!line.empty() && line.back() == ' ')
-        line.pop_back();
-    out << line << "\n";
-}
-
 }  // namespace
 
 std::string dumpPlanLayout(const RenderPlan& plan, const std::vector<int>& deviceLatency) {
-    const auto portOffsets = portOffsetsOf(plan);
-    const auto latency = resolvePlanLatency(plan, portOffsets, deviceLatency);
-    const auto layout = assignBuffers(plan, portOffsets, latency.delaySamples);
+    // The same call PlanExecutor::prepare makes, which is what makes the
+    // golden pin the path the engine runs rather than a copy of it.
+    const auto prepared = resolveLayout(plan, deviceLatency);
+    const auto& portOffsets = prepared.portOffsets;
+    const auto& latency = prepared.latency;
+    const auto& layout = prepared.buffers;
 
     std::ostringstream out;
     out << "magda-plan-layout v1\n";
@@ -90,7 +78,7 @@ std::string dumpPlanLayout(const RenderPlan& plan, const std::vector<int>& devic
         if (i < layout.elided.size() && layout.elided[i] != 0)
             line << " elided";
 
-        writeTrimmed(out, line.str());
+        writeLine(out, line.str());
     }
 
     return out.str();
--- a/magda/engine/exec/PlanLayoutDump.cpp
+++ b/magda/engine/exec/PlanLayoutDump.cpp
@@ -1,0 +1,99 @@
+#include "exec/PlanLayoutDump.hpp"
+
+#include <sstream>
+
+namespace magda::engine {
+namespace {
+
+std::string padded(std::string text, std::size_t width) {
+    if (text.size() < width)
+        text.append(width - text.size(), ' ');
+    return text;
+}
+
+std::string rightAligned(int value, std::size_t width) {
+    auto text = std::to_string(value);
+    if (text.size() < width)
+        text.insert(text.begin(), static_cast<long>(width - text.size()), ' ');
+    return text;
+}
+
+/// An op's output ports, as the arena and slot each renders into.
+std::string describePorts(const RenderPlan& plan, const BufferLayout& layout,
+                          const std::vector<int>& portOffsets, std::size_t op) {
+    const auto& outputs = plan.ops[op].outputs;
+    if (outputs.empty())
+        return "-";
+
+    std::string text;
+    for (std::size_t i = 0; i < outputs.size(); ++i) {
+        if (i > 0)
+            text += ",";
+
+        const auto port = static_cast<std::size_t>(portOffsets[op]) + i;
+        text += outputs[i] == SignalKind::Midi ? "m" : "a";
+        text += std::to_string(layout.portSlots[port]);
+    }
+    return text;
+}
+
+/// The latency reaching an op's first output port, which is the number that
+/// says where in the graph a delay had to go.
+///
+/// An op with no output ports has no latency to report rather than a latency
+/// of zero, and printing zero for it would read as "nothing reaches this",
+/// which of the hardware output is the opposite of true.
+std::string describeLatency(const RenderPlan& plan, const PlanLatency& latency,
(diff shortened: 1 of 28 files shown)


tests/test_plan_goldens.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+        INFO("fixture: " << fixture.name);
+        INFO("golden: " << file.getFullPathName());
+
+        if (!file.existsAsFile()) {
+            FAIL_CHECK("no golden: regenerate with MAGDA_UPDATE_PLAN_GOLDENS=1");
+            continue;
+        }
+
+        const auto expected = file.loadFileAsString().toStdString();
```
lucaromagnoli, line 197: The hosted windows-latest fallback from pick-windows-runner checks out with Git for Windows' core.autocrlf=true, so the goldens arrive as CRLF there, getline keeps the \r on every expected line, and all eleven fixtures fail on their first line. Whether this run trips it depends only on which runner got picked. The writer half is already pinned; the reader half needs a .gitattributes line, `tests/goldens/plan/*.txt -text` (or `text eol=lf`), so git never translates them on any checkout.

magda/engine/exec/PlanLayoutDump.cpp
```diff
@@ -0,0 +1,99 @@
...
+        line.pop_back();
+    out << line << "\n";
+}
+
+}  // namespace
+
+std::string dumpPlanLayout(const RenderPlan& plan, const std::vector<int>& deviceLatency) {
+    const auto portOffsets = portOffsetsOf(plan);
+    const auto latency = resolvePlanLatency(plan, portOffsets, deviceLatency);
+    const auto layout = assignBuffers(plan, portOffsets, latency.delaySamples);
```
lucaromagnoli: These three calls mirror PlanExecutor::prepare exactly (PlanExecutor.cpp:356-360), which is what makes the golden pin the real path. But the mirror holds by convention: a step added to prepare between these calls would leave the golden pinning a path the engine no longer runs, and nothing would fail. Extracting the triple into one PlanLayout function both call, say resolveLayout(plan, deviceLatency) returning latency plus layout, makes the drift impossible rather than unlikely.

magda/engine/plan/PlanDiffDump.cpp
```diff
@@ -0,0 +1,56 @@
+#include "plan/PlanDiffDump.hpp"
+
+#include <sstream>
+
+namespace magda::engine {
+namespace {
+
+std::string padded(std::string text, std::size_t width) {
```
lucaromagnoli: Third copy of padded/rightAligned (PlanDump.cpp and PlanLayoutDump.cpp have the others), and this file inlines the trailing-space trim that PlanLayoutDump has as writeTrimmed. One small internal header for the dump formatting helpers would keep the three dumps in lockstep, which matters more than usual here since column drift rewrites every golden.

magda/engine/exec/PlanLayoutDump.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return text;
+}
+
+/// The latency reaching an op's first output port, which is the number that
+/// says where in the graph a delay had to go.
+///
+/// An op with no output ports has no latency to report rather than a latency
+/// of zero, and printing zero for it would read as "nothing reaches this",
+/// which of the hardware output is the opposite of true.
```
lucaromagnoli: "which of the hardware output" reads garbled; "which for the hardware output", I think.

lucaromagnoli (commented):
Read the whole diff and checked it against the tree. dumpPlanLayout's derivation matches PlanExecutor::prepare's exact sequence, the plan pipeline uses no unordered containers so the cross-platform determinism the files rely on holds, the orphan/uniqueness/double-compile guards close the usual golden loopholes, and Catch2 is the right home for model-level fixtures. One real gap, inline at the reader: the goldens are not eol-pinned in git, and the hosted Windows fallback runner checks out with autocrlf=true, which CRLFs every golden on checkout. Two structure notes and a typo besides.

lucaromagnoli:
**[P1] Pin the plan goldens to LF**

The new snapshot files have no `eol` attribute. With `core.autocrlf=true`, Git checks them out as CRLF; `loadFileAsString()` preserves the `\r`, while `renderGolden()` produces LF-only text, so `firstDifference()` reports every line as different. I reproduced the checkout conversion locally with `git -c core.autocrlf=true checkout-index`. Please add `tests/goldens/plan/*.txt text eol=lf` to `.gitattributes` (or normalize line endings before comparison).

tests/CMakeLists.txt
```diff
@@ -70,6 +70,8 @@ set(TEST_SOURCES
     test_momentary_param_button.cpp
     test_delta_solo.cpp
     test_hash_combine.cpp
+    PlanGoldenFixtures.cpp
```
lucaromagnoli: **[P2] Gate the golden sources behind `MAGDA_BUILD_NATIVE_ENGINE`**

These sources are added unconditionally, but they include native-engine headers and call symbols from `magda_engine`; that target is only created and linked when `MAGDA_BUILD_NATIVE_ENGINE` is enabled. Configuring the supported `MAGDA_BUILD_NATIVE_ENGINE=OFF` variant will therefore fail to compile or link `magda_tests`. Please move both entries into the existing native-engine conditional block below.

tests/test_plan_goldens.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+juce::File goldenFile(const std::string& name) {
+    return juce::File(kGoldenDir).getChildFile(juce::String(name) + ".txt");
+}
+
+/// Device latency per op, which is what the layout pass takes. The fixture
+/// names devices; only Device ops carry one, and everything else is zero.
+std::vector<int> deviceLatencyPerOp(const RenderPlan& plan,
+                                    const std::vector<std::pair<DeviceId, int>>& byDevice) {
+    std::map<DeviceId, int> lookup;
```
lucaromagnoli: **[P2] Key fixture latency by the full device identity**

This map drops the track/rack/chain/section location and keeps only `DeviceId`. Device IDs are section-local, so an FX device and post-FX device can legitimately share an ID; this helper would silently assign both the same latency and produce a golden for a graph other than the fixture intended. Please key latency by `OpKey`, `ChainNodePath`, or an equivalent track/section/device tuple.

tests/PlanGoldenFixtures.cpp
```diff
@@ -0,0 +1,274 @@
...
+    device.deviceType = DeviceType::Instrument;
+    device.isInstrument = true;
+    device.canReceiveMidi = true;
+    return device;
+}
+
+/// Where a device sits, as the key its process op will carry. The runner
+/// requires it to match one op exactly, so a fixture cannot name a location
+/// that does not exist and get silence.
+engine::OpKey deviceAt(TrackId trackId, DeviceId deviceId, RackId rackId = INVALID_RACK_ID,
```
lucaromagnoli, line 59: **[P2] `OpKey` still cannot distinguish section-local ID collisions**

This helper calls the key a device’s “whole location,” but `OpKey` has no FX/post-FX/mixer-analysis section field. `emitTrack()` passes the same `{trackId, INVALID_RACK_ID, INVALID_CHAIN_ID}` site to all three sections, and `emitDevice()` builds the key from that site plus `device.id`, so an FX D3 and post-FX D3 still produce identical keys. In that valid model, `deviceAt(1, 3)` matches both ops and the new `REQUIRE(matches == 1)` fails rather than allowing independent latencies. Please include the section in the fixture identity and, since plan keys must be unique for the differ, carry the same discriminator into `ChainSite`/`OpKey` (or use a section-aware path throughout).

PR #2090, opened 2026-08-13
Make device identity section-aware

### LINKED ISSUE #2089: OpKey has no section, so section-local device ids collide in the plan
Found while reviewing #2088.

`TrackManager` allocates device ids per section. `nextFxDeviceId_`, `nextPostFxDeviceId_` and `nextMixerAnalysisDeviceId_` are separate spaces, so id 3 legitimately exists on one track's FX chain and its post-FX chain at the same time. This is deliberate and already relied on: `test_midi_signal_routing_juce.cpp` asserts the two ids come out equal, and `test_remote_api_contract.cpp` calls the collision legitimate and addresses it by adding `devicePath`.

The plan compiler assumes the opposite. `ChainSite` records only the innermost rack and chain, `emitTrack` passes the same `{trackId, INVALID_RACK_ID, INVALID_CHAIN_ID}` site to all three sections, and `emitDevice` builds the op key from that site plus `device.id`. Both devices therefore compile to the same `OpKey`.

## What happens today

`validatePlan` catches it. Duplicate keys are an invariant there rather than a convention, because the differ hash-joins old and new plans on `OpKey` and a duplicate would carry one device's runtime state into another.

So the failure is loud rather than silent, which is the one good thing about it. It is still a failure: the app can build this model now, and the engine refuses to compile it.

Pinned as `"Section-local device ids compile to a plan the differ cannot key"` in `tests/test_render_plan_compiler.cpp`, so the gap is measured rather than argued about. The test asserts the current behaviour (the plan is rejected); it will need inverting when this is fixed.

## What the fix looks like

A section discriminator carried through `ChainSite` into `OpKey`, or a section-aware path used throughout in place of the bare device id. Either way it changes op identity, which is why it is not a change to make inside a validation slice.

## Why it matters before cutover

Op identity is what the differ carries runtime state on and what the crossfade pass keys its fades to. An identity scheme that cannot address every device the app can create is not one the engine can migrate onto, and #2081's real-project corpus will meet this the first time it loads a project with a post-FX device on it.

lucaromagnoli:
## Summary

- add a section-aware `DeviceKey` and include the chain segment in `OpKey`
- propagate FX, post-FX, and mixer-analysis identity through plan compilation, value resolution, bindings, execution, and runtime-state ownership
- add compiler, session-lifecycle, and plan-golden regression coverage for the same numeric device ID in all three sections

## Root cause

Device IDs are allocated independently for FX, post-FX, and mixer-analysis sections, but the render plan and runtime state used the bare numeric ID as a project-wide identity. Two valid devices such as FX D3 and post-FX D3 therefore compiled to colliding op keys and could not safely participate in plan diffing or runtime-state retention.

## Impact

Valid projects can now use the same numeric device ID in different sections without plan validation failures, incorrect value lookup, binding collisions, or one device inheriting another device's runtime state.

## Validation

- `cmake --build cmake-build-debug --target magda_tests -j2`
- `magda_tests "[engine]"`: 377 test cases and 394,978 assertions passed
- pre-commit formatting and repository checks passed

Closes #2089

### DIFF
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -294,14 +294,13 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
                 break;
 
             case OpKind::Device: {
-                const auto found = bindings.devices.find(op.key.deviceId);
+                const auto found = bindings.devices.find(op.key.deviceKey());
                 if (found == bindings.devices.end() || found->second == nullptr) {
                     // Passing audio through is what the current engine does
                     // with a plugin that failed to load, and it keeps the rest
                     // of the chain testable instead of silencing the track.
                     messages.push_back(describe(i) + "no device bound for device " +
-                                       std::to_string(op.key.deviceId) +
-                                       ", it passes audio through");
+                                       toString(op.key.deviceKey()) + ", it passes audio through");
                     break;
                 }
                 deviceForOp_[i] = found->second;
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -299,7 +299,7 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
                     // Passing audio through is what the current engine does
                     // with a plugin that failed to load, and it keeps the rest
                     // of the chain testable instead of silencing the track.
-                    messages.push_back(describe(i) + "no device bound for device " +
+                    messages.push_back(describe(i) + "no device bound for " +
                                        toString(op.key.deviceKey()) + ", it passes audio through");
                     break;
                 }
--- a/magda/engine/exec/RuntimeStateStore.cpp
+++ b/magda/engine/exec/RuntimeStateStore.cpp
@@ -35,7 +35,8 @@ template <typename Map> void prepareAll(Map& map, const RenderContext& context) 
         object->prepare(context);
 }
 
-void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceKey>& out);
+void collectDeviceIds(const std::vector<ChainElement>& elements, ChainSegment segment,
+                      std::set<DeviceKey>& out);
 
 void collectDeviceIds(const std::vector<PostFxChainElement>& elements, ChainSegment segment,
                       std::set<DeviceKey>& out) {
@@ -43,13 +44,14 @@ void collectDeviceIds(const std::vector<PostFxChainElement>& elements, ChainSegm
         out.emplace(segment, element.device.id);
 }
 
-void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceKey>& out) {
+void collectDeviceIds(const std::vector<ChainElement>& elements, ChainSegment segment,
+                      std::set<DeviceKey>& out) {
     for (const auto& element : elements) {
         if (isDevice(element)) {
-            out.emplace(ChainSegment::Fx, getDevice(element).id);
+            out.emplace(segment, getDevice(element).id);
         } else if (isRack(element)) {
             for (const auto& chain : getRack(element).chains)
-                collectDeviceIds(chain.elements, out);
+                collectDeviceIds(chain.elements, segment, out);
         }
     }
 }
@@ -221,7 +223,7 @@ RuntimeStateIds collectRuntimeStateIds(const std::vector<TrackInfo>& tracks,
         ids.tracks.insert(track.id);
         // Every section, and bypass is not consulted anywhere here: a bypassed
         // device is still a device the user owns.
-        collectDeviceIds(track.chain.fxChainElements, ids.devices);
+        collectDeviceIds(track.chain.fxChainElements, ChainSegment::Fx, ids.devices);
         collectDeviceIds(track.chain.postFxChainElements, ChainSegment::PostFx, ids.devices);
         collectDeviceIds(track.chain.mixerAnalysisElements, ChainSegment::MixerAnalysis,
                          ids.devices);
--- a/magda/engine/exec/RuntimeStateStore.cpp
+++ b/magda/engine/exec/RuntimeStateStore.cpp
@@ -35,17 +35,18 @@ template <typename Map> void prepareAll(Map& map, const RenderContext& context) 
         object->prepare(context);
 }
 
-void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceId>& out);
+void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceKey>& out);
 
-void collectDeviceIds(const std::vector<PostFxChainElement>& elements, std::set<DeviceId>& out) {
+void collectDeviceIds(const std::vector<PostFxChainElement>& elements, ChainSegment segment,
+                      std::set<DeviceKey>& out) {
     for (const auto& element : elements)
-        out.insert(element.device.id);
+        out.emplace(segment, element.device.id);
 }
 
-void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceId>& out) {
+void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceKey>& out) {
     for (const auto& element : elements) {
         if (isDevice(element)) {
-            out.insert(getDevice(element).id);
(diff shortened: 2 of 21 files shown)


lucaromagnoli (commented):
The identity change itself is right, and it lands everywhere it has to: compile, value resolution, bindings, runtime ownership and retention all move together, `fadeKey` copies the consumer's whole key so fades inherit the segment for free, and the fingerprint mixes it. The session test is the one I'd have asked for -- distinct values, distinct instances, and staged destruction across three sections sharing id 3.

Two things I would change before merge.

**The implicit `DeviceKey(DeviceId)` gives back exactly what the PR takes away.** The premise is that a bare device id is not an identity; a converting constructor means every site that still passes one keeps compiling and silently means `Fx`. That is not hypothetical here: `tests/test_parallel_executor.cpp` still writes `scene.bindings.devices[7]`, `[8]`, `[100 + id]` and `[200]`, and the file is untouched by this PR because it did not have to change. Making it `explicit` turns the remaining sites into compile errors, which is the only reason to introduce the type.

**`RenderPlan.hpp` now pulls JUCE into the engine's IR header.** `core/ChainNodePath.hpp` includes `<juce_core/juce_core.h>`, and `RenderPlan.hpp` was previously JUCE-free (`<cstdint>`, `<string>`, `<vector>`, `core/TypeIds.hpp`) and is included by every engine TU. That is a lot of header for a three-value enum. `ChainSegment` has no JUCE in it, and `ChainNodePath.hpp` already includes `TypeIds.hpp`, so moving the enum there costs nothing and leaves every existing user compiling unchanged.

The rest are small.

magda/engine/plan/RenderPlan.hpp
```diff
@@ -191,6 +195,42 @@ constexpr int crossfadeSlot(int index) {
...
+ * chains, but the same integer may name one device in each section. Runtime
+ * ownership therefore needs the section as well as the integer.
+ */
+struct DeviceKey {
+    ChainSegment segment = ChainSegment::Fx;
+    DeviceId deviceId = INVALID_DEVICE_ID;
+
+    DeviceKey() = default;
+    // Keeps the overwhelmingly common main-FX call sites compact.
+    DeviceKey(DeviceId id) : deviceId(id) {}
```
lucaromagnoli: Make this `explicit`.

The whole point of `DeviceKey` is that `DeviceId` alone does not identify a device. An implicit conversion means any site that still hands over a bare id compiles and quietly resolves to `ChainSegment::Fx` -- the exact failure this PR is fixing, now with a type in front of it.

It is already happening: `tests/test_parallel_executor.cpp:445`, `:446`, `:462` and `:484` write `scene.bindings.devices[7]` and friends, and that file needed no change in this PR precisely because the conversion took it. Those are FX devices so they are correct today, but nothing checked that and nothing would have said so if they were not.

`explicit` costs four call sites in one test file and buys a compiler that finds the next one.

magda/engine/plan/RenderPlan.hpp
```diff
@@ -1,9 +1,13 @@
 #pragma once
 
+#include <cstddef>
 #include <cstdint>
+#include <iosfwd>
 #include <string>
+#include <tuple>
 #include <vector>
 
+#include "core/ChainNodePath.hpp"
```
lucaromagnoli: This drags `<juce_core/juce_core.h>` into the engine's IR header, transitively through `ChainNodePath.hpp`, for the sake of a three-value enum. `RenderPlan.hpp` was JUCE-free before this and is included by every translation unit in `magda_engine`; `magda/engine/plan/` had exactly one file touching juce until now.

`ChainSegment` itself has no JUCE in it. Moving it to `core/TypeIds.hpp` -- which is JUCE-free, is already included here, and is already included by `ChainNodePath.hpp` -- keeps every current user compiling unchanged and leaves this header as light as it was.

magda/engine/plan/RenderPlan.hpp
```diff
@@ -191,6 +195,42 @@ constexpr int crossfadeSlot(int index) {
...
+    bool operator<(const DeviceKey& o) const {
+        return std::tie(segment, deviceId) < std::tie(o.segment, o.deviceId);
+    }
+};
+
+struct DeviceKeyHash {
+    std::size_t operator()(const DeviceKey& key) const noexcept {
+        const auto section = static_cast<std::size_t>(key.segment);
+        const auto device = static_cast<std::size_t>(key.deviceId);
+        return (section << (sizeof(std::size_t) * 4U)) ^ device;
```
lucaromagnoli: `sizeof(std::size_t) * 4U` is 32 bits on a 64-bit build but 16 on a 32-bit one, where `{Fx, 65536}` and `{PostFx, 0}` hash identically. Equality still decides, so nothing is wrong; it just creates a collision class for no reason. A fixed 32-bit shift, or folding the segment through a mix constant, says what is meant without depending on `sizeof`.

magda/engine/exec/RuntimeStateStore.cpp
```diff
@@ -35,17 +35,18 @@ template <typename Map> void prepareAll(Map& map, const RenderContext& context)
...
-        out.insert(element.device.id);
+        out.emplace(segment, element.device.id);
 }
 
-void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceId>& out) {
+void collectDeviceIds(const std::vector<ChainElement>& elements, std::set<DeviceKey>& out) {
     for (const auto& element : elements) {
         if (isDevice(element)) {
-            out.insert(getDevice(element).id);
+            out.emplace(ChainSegment::Fx, getDevice(element).id);
```
lucaromagnoli: The `PostFxChainElement` overload takes the segment as a parameter and this one hard-codes `Fx`. Correct today, because racks only exist in `fxChainElements` and the flat sections cannot hold one -- but the two overloads now disagree about whose job the segment is, and the reader has to go and establish the rack rule to know this line is safe. Taking the segment here too would make it locally obvious, and the recursive rack call would just pass it through.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -294,14 +294,13 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
                 break;
 
             case OpKind::Device: {
-                const auto found = bindings.devices.find(op.key.deviceId);
+                const auto found = bindings.devices.find(op.key.deviceKey());
                 if (found == bindings.devices.end() || found->second == nullptr) {
                     // Passing audio through is what the current engine does
                     // with a plugin that failed to load, and it keeps the rest
                     // of the chain testable instead of silencing the track.
                     messages.push_back(describe(i) + "no device bound for device " +
```
lucaromagnoli: Reads as "no device bound for device D7" now that `toString(DeviceKey)` prints its own `D` prefix. Either drop the word from the message or have the message say "no device bound for " and let the key speak.

magda/engine/plan/RenderPlan.hpp
lucaromagnoli: The implicit conversion means a bare integer compiles anywhere a DeviceKey is expected and silently means Fx, which is the exact ambiguity class this PR removes. Today every engine-side lookup goes through op.key.deviceKey() and the only users of the shorthand are the two latencyFor[7] test sites, so nothing is wrong now; the exposure is a future host-side RuntimeStateFactory or binding map written against bare ids that compiles without complaint. If you make it explicit those test sites become latencyFor[DeviceKey{7}], which still reads fine. Your call, the tradeoff is documented.

lucaromagnoli (commented):
Traced the identity through every surface it crosses and found no gaps. The compiler stamps the segment on the DeviceProcess key and the gain/meter/delay/midi-merge ops reuse that key, so they inherit it; track-level keys default to Fx, which is fine since their identity is role plus track. The only deviceId-bearing keys outside emitDevice are device meters, which come from the same inherited key, and SendTap ops are track-level with INVALID_DEVICE_ID, so the releaseDeleted device branch is segment-correct. Every OpKey container in the tree is ordered (PlanDiff and PlanCrossfade both join through std::map), so the updated operator< covers them and there is no OpKey hash anywhere that could ignore the new field. Nothing outside engine and tests implements RuntimeStateFactory yet, so the createDevice signature change lands clean. Existing goldens stay byte-identical since no prior fixture uses post-FX or mixer analysis, and the new golden's latency cascade checks out (16, 48, 112). The session test covering distinct instances, per-section value resolution and selective destruction is exactly the lifetime coverage this needed. One note inline.

PR #2092, opened 2026-08-14
Relative transport seek (#1987)

### LINKED ISSUE #1987: Lua API: relative transport seek for rewind / fast-forward controller buttons
Requested by Mike, working with an Arturia KeyLab Essential 49 mk3. The device sends its transport Rewind and Fast Forward buttons as MCU notes 0x5B and 0x5C, which the controller Lua script receives correctly, but he could not find an API to move the playhead with them.

## What already exists

There is a supported way to do this today, so the buttons are not blocked:

| Binding | Notes |
|---|---|
| `magda.transport.position_beats()` | current playhead position, in beats |
| `magda.transport.set_position_beats(beats)` | moves the playhead |

Both are documented in `manual/docs/reference/lua-scripting.md:78-79`. A rewind button can be implemented right now as a read-modify-write:

```lua
local function nudge_transport(deltaBeats)
  local pos = magda.transport.position_beats()
  magda.transport.set_position_beats(math.max(0, pos + deltaBeats))
end
```

So the immediate answer to Mike is yes, and pointing him at those two is the short-term unblock.

## The actual gap

The primitive is there but the shape is wrong for these buttons, and every script ends up rebuilding the same three things:

1. **No relative seek.** Each script writes the read-modify-write above and re-implements the clamp at zero.
2. **No bar-aware jump.** Rewind on a transport control usually means a bar or a beat, not an arbitrary float. Getting that from Lua means the script has to know the meter itself.
3. **Hold-to-scrub is not expressible.** MCU rewind and fast forward are normally held rather than tapped. With only an absolute setter, a script has to track button-down state and drive its own repeat from `on_tick`, picking a scrub rate by hand. That is a lot of boilerplate for a standard transport control, and every controller script will do it slightly differently.

## Proposal

Binding and documentation layer only, no engine change:

- `magda.transport.seek_beats(delta)` - relative, clamped at zero
- `magda.transport.seek_bars(delta)` - meter-aware, so "back one bar" is one call
- document the hold-to-scrub pattern in the controller scripting reference, so RW/FF land the same way across scripts

Beats stay the unit throughout, matching the rest of the transport API.

lucaromagnoli:
Closes #1987. Built to the shape agreed in [the comment on the issue](https://github.com/Conceptual-Machines/magda-core/issues/1987#issuecomment-5162727654): the primitive lands on `TransportApi`, and Lua, the remote API and OSC all delegate.

`seekBeats(delta)` and `seekBars(deltaBars)` are **concrete** on the facade, not virtual. Written as read-modify-write at each caller, the clamp at zero and the bar length get re-derived three times and drift three ways.

The single thing an implementation supplies is `beatsAtBarOffset(beats, deltaBars)` — how long a bar is, which is the project's business and not the caller's.

## The live implementation goes through bars-and-beats, not multiplication

A bar is only a fixed number of beats when the meter is. `TransportApiLive` converts to `tempo::BarsAndBeats`, adds the bar count, and converts back, so crossing a 4/4 into 7/8 lands where the bar line actually is and the offset within the bar carries across untouched.

## Surfaces

| | |
|---|---|
| **Lua** | `magda.transport.seek_beats(delta)` · `magda.transport.seek_bars(delta)` |
| **Remote** | `transport.seekRelative`, taking `deltaBeats` **or** `deltaBars` |
| **OSC** | `/magda/transport/seek` (beats) · `/magda/transport/seek/bars` |

One remote operation rather than two: a caller binding a rewind button is picking a unit, not an endpoint.

## OSC needed a new argument kind, and this is the interesting part

Every continuous value in the OSC namespace is coalesced latest-value-wins. That is correct for a *position* — the last one is where you want to be — and wrong for a *distance*: three presses of a rewind button are three bars, and one slot would keep one of them.

So `OscArgKind::Delta` rides the ordered ring alongside the triggers and toggles. The router's own comment already made this distinction for edges; a relative seek is an edge that happens to carry a magnitude. It also gets the ordering right for free: a seek sent in the same bundle as a locate lands *after* it, which is what a relative move needs.

`kOscUnindexedSlots` goes 8 → 10. The two seek kinds never publish into their slots, but keeping them means "a kind's slot is its ordinal" stays true of the whole enum — and the existing exhaustiveness test enforces that.

## Tests

- `test_transport_seek.cpp` — the shared implementation: relative movement, the clamp at zero for both units, and that `seekBars` follows a 7/8 meter rather than assuming four.
- `test_osc_router.cpp` — three repeated seeks all survive the drain (the coalescing point), a seek lands after the locate it was bundled with, and a bare or zero delta is declined without falling through to bindings.
- `test_osc_address.cpp` — both addresses parse, both are `Delta`, and `bars` is a unit under `seek` rather than a leaf of its own.

## Docs

The manual gets both calls, the KeyLab rewind/fast-forward example the issue was raised about, and the hold-to-scrub pattern — scaled by `on_tick`'s `dt` rather than a fixed amount per tick, since ticks are ~30 Hz and a beat per tick is thirty beats a second.

Verified: 2558 cases, 571849 assertions.

### DIFF
--- a/magda/daw/api/osc_command_sink_live.cpp
+++ b/magda/daw/api/osc_command_sink_live.cpp
@@ -1,5 +1,6 @@
 #include "osc_command_sink_live.hpp"
 
+#include <algorithm>
 #include <cmath>
 
 #include "../audio/controllers/ControllerParamWriter.hpp"
@@ -20,6 +21,14 @@ namespace magda {
 
 using osc::OscCommand;
 using osc::OscCommandKind;
+
+namespace {
+
+/// TransportApi's bound, as a float, so an OSC argument can be clamped before
+/// it is rounded rather than after.
+constexpr float kMaxSeekBars = static_cast<float>(TransportApi::kMaxBarOffset);
+
+}  // namespace
 
 OscCommandSinkLive::OscCommandSinkLive(MagdaApi& api, std::unique_ptr<ControllerParamWriter> writer)
     : api_(api), writer_(std::move(writer)) {
@@ -95,12 +104,23 @@ void OscCommandSinkLive::apply(const OscCommand& command, float value) {
             api_.transport().seekBeats(value);
             return;
         case OscCommandKind::TransportSeekBars:
+            // Clamped before it is rounded, not after. A float reaches 3.4e38
+            // and llround is a domain error past about 9.2e18: the result is
+            // unspecified, and on x86 and ARM it is LLONG_MIN for either sign,
+            // so a huge *positive* delta would seek to the start of the
+            // project. That is reachable from the network, since OSC is
+            // unauthenticated UDP. Clamping first makes the facade's own bound
+            // the thing that decides, on every platform.
+            //
             // Rounded rather than truncated, so a surface that sends its
             // integers as floats and lands on 0.999999 still moves a bar.
-            // llround rather than lround, and no narrowing: the facade bounds
-            // the count, so a surface sending 1e30 gets one end of the project
-            // rather than whatever the conversion happened to produce.
-            api_.transport().seekBars(std::llround(value));
+            //
+            // NaN first, because a clamp does not catch it: every comparison
+            // against it is false, so it would pass straight through to the
+            // same domain error the clamp is here to prevent.
+            if (!std::isfinite(value))
+                return;
+            api_.transport().seekBars(std::llround(std::clamp(value, -kMaxSeekBars, kMaxSeekBars)));
             return;
 
         case OscCommandKind::MasterVolume:
--- a/magda/daw/api/osc_command_sink_live.cpp
+++ b/magda/daw/api/osc_command_sink_live.cpp
@@ -97,7 +97,10 @@ void OscCommandSinkLive::apply(const OscCommand& command, float value) {
         case OscCommandKind::TransportSeekBars:
             // Rounded rather than truncated, so a surface that sends its
             // integers as floats and lands on 0.999999 still moves a bar.
-            api_.transport().seekBars(static_cast<int>(std::lround(value)));
+            // llround rather than lround, and no narrowing: the facade bounds
+            // the count, so a surface sending 1e30 gets one end of the project
+            // rather than whatever the conversion happened to produce.
+            api_.transport().seekBars(std::llround(value));
             return;
 
         case OscCommandKind::MasterVolume:
--- a/magda/daw/api/osc_command_sink_live.cpp
+++ b/magda/daw/api/osc_command_sink_live.cpp
@@ -1,5 +1,7 @@
 #include "osc_command_sink_live.hpp"
 
+#include <cmath>
+
 #include "../audio/controllers/ControllerParamWriter.hpp"
 #include "../core/ControlTarget.hpp"
 #include "../core/MixerStripOrder.hpp"
@@ -88,6 +90,14 @@ void OscCommandSinkLive::apply(const OscCommand& command, float value) {
             return;
         case OscCommandKind::TransportPosition:
             api_.transport().setPositionBeats(value);
+            return;
+        case OscCommandKind::TransportSeekBeats:
+            api_.transport().seekBeats(value);
+            return;
+        case OscCommandKind::TransportSeekBars:
+            // Rounded rather than truncated, so a surface that sends its
+            // integers as floats and lands on 0.999999 still moves a bar.
+            api_.transport().seekBars(static_cast<int>(std::lround(value)));
             return;
 
         case OscCommandKind::MasterVolume:
--- a/magda/daw/api/remote_api.cpp
+++ b/magda/daw/api/remote_api.cpp
@@ -1582,6 +1582,14 @@ OperationRegistry::OperationRegistry() {
             "required":["positionBeats"],"additionalProperties":false
         })json"),
         transportSchema());
+    add("transport.seekRelative", "Move the transport by beats or bars, clamped at zero",
+        OperationAccess::Write, &handlers::transportSeekRelative, operationInputSchema(R"json({
+            "type":"object",
+            "properties":{"deltaBeats":{"type":"number"},"deltaBars":{"type":"integer"}},
+            "oneOf":[{"required":["deltaBeats"]},{"required":["deltaBars"]}],
+            "additionalProperties":false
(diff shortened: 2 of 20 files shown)


magda/daw/api/remote_api.cpp
```diff
@@ -1582,6 +1582,14 @@ OperationRegistry::OperationRegistry() {
             "required":["positionBeats"],"additionalProperties":false
         })json"),
         transportSchema());
+    add("transport.seekRelative", "Move the transport by beats or bars, clamped at zero",
+        OperationAccess::Write, &handlers::transportSeekRelative, operationInputSchema(R"json({
+            "type":"object",
+            "properties":{"deltaBeats":{"type":"number"},"deltaBars":{"type":"integer"}},
+            "oneOf":[{"required":["deltaBeats"]},{"required":["deltaBars"]}],
```
lucaromagnoli, line 1631: This is the first oneOf in any schema, and validateValue does not implement it, only anyOf, and it ignores unknown keywords. So the constraint is advertised but not enforced: a request with neither field passes validation and quietly no-ops as seekBeats(0) with an ok response, and a request with both passes and silently prefers deltaBars. Switching to anyOf gets the missing-both case rejected by the existing validator (both-present would still pass, which the handler's deltaBars precedence can just document), or teach validateValue oneOf, which is a branch counter on top of the anyOf loop.

magda/daw/api/transport_api_live.cpp
```diff
@@ -81,4 +81,27 @@ void TransportApiLive::setPositionBeats(double beats) {
     e->getTransport().setPosition(t);
 }
 
+double TransportApiLive::beatsAtBarOffset(double beats, int deltaBars) const {
```
lucaromagnoli, line 87: The 7/8 test exercises the mock's fixed meter, so what it pins is the facade's delegation, not this walk. The headline behaviour, crossing a 4/4 to 7/8 change and landing on the real bar line with the offset carried, rests on this function and is untested against a real tempo sequence. A magda_juce_tests case with an Edit carrying a signature change would pin it, per the usual split of TE-Edit tests into the JUCE target.

lucaromagnoli (commented):
Checked the design against the tree and it holds together: TransportApiLive and MockTransportApi are the only two TransportApi implementations, so the new pure virtual lands everywhere it must; the OSC router coerces int32 args to float, so `/seek/bars` works from surfaces that send integers; OscCommandKind ordinals are never persisted (bindings serialize through the controller layer and the fixed namespace is parsed from address strings), so inserting the two kinds mid-enum is safe; and the manual's examples are real, on_tick does receive dt and the Lua MIDI event does expose type and number, and 0x5B/0x5C are the right MCU notes. The dead-slots-for-ordinal-integrity call and the values-before-edges drain ordering are both sound and both tested. Two findings inline, one real and one test-coverage.

magda/daw/api/transport_api_live.cpp
```diff
@@ -81,4 +81,27 @@ void TransportApiLive::setPositionBeats(double beats) {
...
+        return beats;
+
+    // Through bars-and-beats rather than by multiplying out a bar length,
+    // because a bar is only a fixed number of beats when the meter is. The
+    // sequence walks whatever time signatures lie between the two points, so
+    // crossing a 4/4 to 7/8 change lands where the bar line actually is, and
+    // the offset within the bar is carried across untouched.
+    const auto here = e->tempoSequence.toTime(te::BeatPosition::fromBeats(beats));
+    auto barsAndBeats = e->tempoSequence.toBarsAndBeats(here);
+    barsAndBeats.bars += deltaBars;
```
lucaromagnoli: **P1 — Bound bar offsets before signed arithmetic**

`deltaBars` reaches this addition after unchecked narrowing from external inputs. The remote schema has no range, and Lua and OSC likewise convert directly to `int`; even `INT_MAX` overflows here whenever the current bar is positive. Please validate or clamp before narrowing and use checked arithmetic here, with boundary tests for each input surface.

magda/daw/api/remote_api.cpp
lucaromagnoli, line 1631: **P1 — Do not rely on unsupported `oneOf` validation**

`validateValue()` only implements `anyOf`; it never evaluates `oneOf`. As a result, both `{}` and a payload containing both fields pass this schema. The handler then converts a missing `deltaBeats` to zero or silently prefers `deltaBars`. Please either add `oneOf` support to the validator or express and enforce this exclusivity using validation it supports, and cover the missing and both-present cases.

magda/daw/api/osc_command_sink_live.cpp
```diff
@@ -89,6 +91,17 @@ void OscCommandSinkLive::apply(const OscCommand& command, float value) {
...
+        case OscCommandKind::TransportSeekBeats:
+            api_.transport().seekBeats(value);
+            return;
+        case OscCommandKind::TransportSeekBars:
+            // Rounded rather than truncated, so a surface that sends its
+            // integers as floats and lands on 0.999999 still moves a bar.
+            // llround rather than lround, and no narrowing: the facade bounds
+            // the count, so a surface sending 1e30 gets one end of the project
+            // rather than whatever the conversion happened to produce.
+            api_.transport().seekBars(std::llround(value));
```
lucaromagnoli: llround with a float outside long long's range (anything past about 9.2e18, and a float reaches 3.4e38) is a domain error with an unspecified result; on x86 and ARM it comes back LLONG_MIN for both signs, so a huge positive delta seeks to the start, the wrong end, and the comment's promise only holds by the letter. Clamping the float before rounding makes it true everywhere: std::llround(std::clamp(value, -kMax, kMax)) with kMax as a float around TransportApi::kMaxBarOffset. Reachable from the network, since OSC is unauthenticated UDP.

lucaromagnoli (commented):
The follow-up holds up. The oneOf implementation gets the subtle part right, continuing to the sibling keywords on a match instead of returning, and the contract tests cover one, neither, both and the new bounds. The widening is layered coherently: schema bounds at the remote edge, a 64-bit facade that clamps once, and the live walk widening the addition and capping before the struct field. I verified the JUCE test's premise against the TE tempo code: bar length there is the numerator in beats, so 7/8 bars really are seven beats apart and the 4-8-15-22 expectations match TE's model. One residual edge inline.

PR #2093, opened 2026-08-14
OSC feedback: echo MAGDA's state back to the surface (#2091)

### LINKED ISSUE #2091: OSC feedback: echo MAGDA's state back to the surface
Section 4 of #1757, which shipped everything else. The input path (#2061), bindings and OSC-learn (#2062) are on `dev/0.19.0`; feedback is not, so a surface can drive MAGDA but never learns what MAGDA did. A motorised fader stays where the user left it, a TouchOSC page opens showing whatever its template defaulted to, and moving a fader in the app leaves the surface lying about the value.

## What exists

- `ControllerFeedbackSink` (`magda/daw/audio/controllers/ControllerFeedback.hpp`) is the hook, and `ControllerRouter::setFeedbackSink` takes it. An OSC sink is a new implementation, not new machinery.
- `OscAddress` already parses and formats the fixed namespace, so the addresses to send on are the ones already being received.
- `Config` has `oscEnabled` and `oscReceivePort`.
- `juce::OSCSender` is available; nothing in the tree constructs one yet.

## What this adds

- **The sink.** An OSC `ControllerFeedbackSink` owning a `juce::OSCSender`, echoing bound and fixed-namespace values to the configured destination on the same address they are received on: moving a MAGDA fader sends `/magda/track/3/volume 0.62`.
- **Transport state on change**, pushed rather than polled: `/magda/transport/play 1` and the rest of the fixed namespace.
- **Rate limiting**, around 30 Hz per address, latest-value-wins. A fader drag is a continuous stream and a surface does not need every sample of it.
- **Echo suppression.** A value that just arrived from a peer must not be sent straight back to that peer, or a surface and MAGDA will chase each other.
- **Snapshot on connect.** First message from a new peer optionally pushes full state, so a surface that joins late is correct rather than blank.
- **Config**: `oscFeedbackHost` and `oscFeedbackPort` (default empty / 9001), surfaced in the OSC section of `ControllersDialog` beside the existing enable and port fields.

## Share the coalescing layer

Per the note on #1757: rate-limited, latest-value-wins coalescing with echo suppression is the same machinery #1857 built for WebSocket subscriptions, over the same model listeners. #1857 has since shipped, so the shared piece exists. Build this on it rather than writing a second private one - two coalescing layers that drift is how the playhead ends up smooth on one transport and stuttering on the other.

lucaromagnoli:
Closes #2091. Section 4 of #1757, and the last of it.

The input path and bindings/learn made a surface able to drive MAGDA. Nothing made MAGDA able to answer, so every value a surface showed was whatever its template defaulted to, corrected only by the user's own finger.

## Three things the issue assumed that were not there

Each one moved the design rather than the estimate, so they are worth stating first.

**`OscAddress` did not format.** It parsed, and it mapped a command to a coalescing slot and back, but nothing turned an `OscCommand` into `/magda/track/3/volume`. The inverse already existed in the slot arithmetic, so this adds a formatter beside the parser, with a round-trip test over the whole slot space.

**JUCE will not say who sent a message.** `OSCReceiver`'s realtime listener hands over an `OSCMessage` and nothing else; the datagram's source address is consumed inside JUCE's own receive loop, and the parser is private to `juce_osc`. So "suppress the echo to *that peer*" and "snapshot on the first message from a *new peer*" are not implementable as written. Feedback goes to one configured destination, and the peer-shaped parts are re-derived against what is observable: a snapshot fires when the destination changes, and when inbound traffic resumes after five seconds of silence.

**There was no OSC section in `ControllersDialog`** to put the two new fields beside. Nothing in the UI read `oscEnabled`, `oscReceivePort` or `oscBindAddress`, so what shipped in #2061 and #2062 was unreachable for anyone who does not hand-edit JSON. This adds the tab, which is most of the value: it is what makes the whole of #1757 usable.

## What drives a send

Feedback is a projection of model state onto the fixed namespace, diffed against what was last sent. That framing answers three requirements at once. "On change" is what the diff means; the rate limit falls out of the flush cadence, since an address can be sent at most once per flush however often it was published in between; and latest value wins because a slot holds one value.

The projection splits the way `SubscriptionHub` splits its topics, for the same reason.

**Published:** track and master volume, pan, mute, solo and sends, and the focused device's macros. All of these notify a model listener however they moved, so `ChangeSource` says when to re-read them and a project in which nothing moves costs nothing per tick. That is the shared coalescing layer #1857 built, which already documents OSC feedback as its second consumer.

**Sampled:** play, record, loop, tempo and the playhead, on a 30 Hz tick. None of these reach `ChangeSource`. `ModelChangeBridge` listens to the model singletons and `ProjectManager`, none of which knows the transport is rolling, and `Topic::Transport` is marked only when a remote API request named `transport.*` executed. Pressing Play in MAGDA is invisible to it, which is a real gap in #1857's coverage as well; the honest fix for that is not this issue, and the playhead has to be sampled regardless, so the five scalars ride the same tick.

## Echo suppression is a bit, not a comparison

A value the surface just sent is a value it already shows, so sending it back is redundant, and for a control that reports its own motorised position it is a loop.

A slot is skipped when its projected value matches what arrived **and** the arrival belongs to this flush. The bit is what makes that correct rather than merely quiet: matching on value alone would suppress the send that brings a surface back from 0.9 after the user dragged MAGDA's own fader away from 0.62 and back. One consequence is kept on purpose, which is that the flush after a gesture ends sends the value once. That is a confirmation rather than an echo, and it is where the surface finds out what MAGDA rounded it to.

The same rule covers bound addresses, through a second tap on the router's binding drain. Without it a bound fader under a finger has its own position sent back on every tick of the drag.

## The budget defers rather than drops

A flush sends at most 192 messages. Only a snapshot of a large project reaches that, and a snapshot is also the one thing that has to finish, so what does not fit stays dirty and goes out on the next flush. A sixty-track snapshot is three ticks, not one long one.

## Bound addresses

A binding's address gets the position its target's value corresponds to, which needs the way back from a parameter to a fader. `BindingTransform` grows `invertCurve` and `invertRange`, total for what an OSC binding can be, since the relative modes already fall back to absolute on that path.

Reading the target back is `ControllerParamReader`, which mirrors `DefaultControllerParamWriter` branch for branch and is deliberately its neighbour rather than a private helper: the two only stay true if they are read side by side. Levels come from `TrackInfo` because that is what the writer wrote, not from the TE parameter, whose fader-position curve is a different domain and would not round-trip. `ModParam` has no reading and says why: its sync-division ordinal is a rounding that does not invert, and no surface displays a modifier rate.

## One shape worth flagging

The tick is an owned object rather than a base class. `juce::Timer` holds a `SharedResourcePointer<TimerThread>` as a member, so inheriting one starts JUCE's timer thread the moment a projector exists, including in a process with no message loop to drive it. Owning it means a MAGDA with feedback off does no per-tick work and a headless one brings no thread into existence. `SubscriptionHub` gives its sampler the same shape.

## Tests

`tests/test_osc_feedback.cpp`, 25 cases.

- The formatter round-trips every slot back through the parser.
- Diffing: a value sends once, an unchanged republish sends nothing, and a value that leaves and returns before a flush sends nothing.
- Echo suppression as the two cases that tell it apart from a comparison, plus the rounded-echo and snapshot-overrides cases.
- The budget sends the cap, counts the rest, and delivers every address exactly once across two flushes.
- The projection against `MockMagdaApi` and a capturing sink: a strip's members land on the addresses that name them at the position the writer's conversion implies, a position is the mixer's rather than the track's id, a shrinking project retires the strips it no longer has, focus going away retires the macros, and the echo rule end to end.
- The binding inverses over each curve and a range.

Built locally, full suite green: 2576 passed, 13 skipped, 0 failures. `make format` and `make lint-changed` clean.

### DIFF
--- a/magda/daw/api/osc_feedback_live.cpp
+++ b/magda/daw/api/osc_feedback_live.cpp
@@ -1,0 +1,498 @@
+#include "osc_feedback_live.hpp"
+
+#include <algorithm>
+#include <cmath>
+
+#include "../audio/osc/OscRouter.hpp"
+#include "../core/AutomationInfo.hpp"
+#include "../core/Config.hpp"
+#include "../core/ControlTarget.hpp"
+#include "../core/MixerStripOrder.hpp"
+#include "../core/ParameterUtils.hpp"
+#include "../core/TrackInfo.hpp"
+#include "../core/aliases/AliasRegistry.hpp"
+#include "../core/aliases/ChainContext.hpp"
+#include "../core/aliases/ResolverRegistry.hpp"
+#include "../core/aliases/TargetResolver.hpp"
+#include "../core/controllers/BindingRegistry.hpp"
+#include "../core/controllers/BindingTransform.hpp"
+#include "focused_api.hpp"
+#include "magda_api.hpp"
+#include "project_api.hpp"
+#include "track_api.hpp"
+#include "transport_api.hpp"
+
+namespace magda {
+
+using osc::OscCommand;
+using osc::OscCommandKind;
+
+namespace {
+
+/// A bound address is not a slot, so it carries its own change test. The same
+/// tolerance the slot table uses, and for the same reason: a normalized value
+/// that has been through a parameter's real range and back does not land on the
+/// bit pattern it started from.
+constexpr float kBoundEpsilon = 1.0e-4f;
+
+OscCommand unindexed(OscCommandKind kind) {
+    return OscCommand{kind, 0, 0};
+}
+
+/// Linear gain to the normalized fader position, the inverse of what
+/// `DefaultControllerParamWriter::writeTrackLevel` did on the way in.
+float normalizedFromGain(float gain, const ParameterInfo& info) {
+    const float dB = gain > 0.0f ? 20.0f * std::log10(gain) : info.minValue;
+    return ParameterUtils::realToNormalized(dB, info);
+}
+
+}  // namespace
+
+// ============================================================================
+// Tick and ControllerSink
+// ============================================================================
+
+/// The 30 Hz heartbeat, owned rather than inherited so that a projector with no
+/// destination — and a headless one, which has no message loop to drive a timer
+/// anyway — never brings JUCE's timer thread into existence.
+class OscFeedbackProjector::Tick : public juce::Timer {
+  public:
+    Tick(OscFeedbackProjector& owner, int intervalMs) : owner_(owner) {
+        startTimer(intervalMs);
+    }
+
+    ~Tick() override {
+        stopTimer();
+    }
+
+    void timerCallback() override {
+        owner_.tick();
+    }
+
+  private:
+    OscFeedbackProjector& owner_;
+};
+
+/**
+ * The `ControllerFeedbackSink` half. It carries a value, and deliberately
+ * ignores it: what a bound address should show is what its *target* holds now,
+ * and a `FeedbackEvent` names a binding rather than a target, so two OSC
+ * bindings on one parameter would each need looking up anyway. Marking the
+ * bindings dirty and re-reading on the next tick answers both, at the tick rate
+ * a surface can consume rather than at the rate a knob produces.
+ */
+class OscFeedbackProjector::ControllerSink : public ControllerFeedbackSink {
+  public:
+    ControllerSink(OscFeedbackProjector& owner, std::shared_ptr<std::atomic<bool>> alive)
+        : owner_(owner), alive_(std::move(alive)) {}
+
+    void send(const FeedbackEvent&) override {
+        if (alive_->load())
+            owner_.bindingsDirty_ = true;
+    }
+
+  private:
+    OscFeedbackProjector& owner_;
+    std::shared_ptr<std::atomic<bool>> alive_;
+};
+
+// ============================================================================
+// Lifecycle
+// ============================================================================
+
+OscFeedbackProjector::OscFeedbackProjector(MagdaApi& api, remote::ChangeSource& changes,
+                                           osc::OscRouter& router,
+                                           std::unique_ptr<ControllerParamReader> reader,
+                                           std::unique_ptr<osc::OscMessageSink> sink)
+    : api_(api), changes_(changes), router_(router), reader_(std::move(reader)) {
+    if (sink == nullptr) {
+        auto sender = std::make_unique<osc::OscSenderSink>();
+        sender_ = sender.get();
+        sink = std::move(sender);
+    }
+    feedback_ = std::make_unique<osc::OscFeedback>(std::move(sink));
+
+    // The router's drain is on the message thread, same as the tick, so both of
+    // these reach straight into the tables with nothing in between.
+    router_.setFeedbackTap([this](int slot, float value) { feedback_->noteReceived(slot, value); });
+    router_.setBindingFeedbackTap([this](const juce::String& address, float value) {
+        auto& entry = boundEntryFor(address);
+        entry.lastReceived = value;
+        entry.echoed = true;
+    });
+
+    changeToken_ = changes_.addListener(
+        [this](const std::vector<remote::ChangeSource::Change>& changed) { onChanges(changed); });
+
+    Config::getInstance().addListener(this);
+}
+
+OscFeedbackProjector::~OscFeedbackProjector() {
+    alive_->store(false);
+    Config::getInstance().removeListener(this);
+    tick_.reset();
+    changes_.removeListener(changeToken_);
+    router_.setFeedbackTap({});
+    router_.setBindingFeedbackTap({});
+}
+
+void OscFeedbackProjector::configChanged() {
+    applyConfig();
+}
+
+std::unique_ptr<ControllerFeedbackSink> OscFeedbackProjector::makeControllerFeedbackSink() {
+    return std::make_unique<ControllerSink>(*this, alive_);
+}
+
+// ============================================================================
+// Configuration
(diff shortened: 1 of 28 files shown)


magda/daw/api/osc_feedback_live.cpp
```diff
@@ -0,0 +1,498 @@
...
+/// bit pattern it started from.
+constexpr float kBoundEpsilon = 1.0e-4f;
+
+OscCommand unindexed(OscCommandKind kind) {
+    return OscCommand{kind, 0, 0};
+}
+
+/// Linear gain to the normalized fader position, the inverse of what
+/// `DefaultControllerParamWriter::writeTrackLevel` did on the way in.
+float normalizedFromGain(float gain, const ParameterInfo& info) {
```
lucaromagnoli: This is the third copy of the writer's inverse. ControllerParamReader.cpp:19 has the identical function with the identical comment, and projectStrip/projectMixer re-derive per strip what DefaultControllerParamReader::readTrackLevel already answers. The PR's own argument for the reader existing is that an inverse only stays true when it lives beside what it inverts; three sites now have to notice if writeTrackLevel's conversion ever changes, and this one is the file a future edit is least likely to grep for. One shared home, either ParameterUtils or a helper the reader exposes, closes that.

lucaromagnoli (commented):
Read the whole diff and traced the parts that could bite. Threading holds up everywhere I could break it: ControllerRouter emits feedback after its callAsync hop, so the ControllerSink's plain-bool dirty write is message-thread-only; the router's accepted counter is atomic; and both inbound paths bump it, so the silence-resume snapshot heuristic covers a bindings-only surface too. The echo bit-plus-value rule, the recomputed dirty on publish, the budget deferring inside the same walk that still runs the echo rule, and the retire-versus-snapshot interaction are all correct as written and each one has the test that would catch the plausible wrong version. Lifecycle is right at both ends: construction order in main gives the projector a live change source, and shutdown installs a NullSink and destroys the projector before everything it taps, with the alive flag covering the sink the ControllerRouter singleton keeps. Config semantics (empty host stays empty on load, idempotent reapply on unrelated saves) and the UI's commit-on-focus-loss are sound, and the tab matches the dialog's existing conventions. One structural point inline.

magda/daw/api/osc_feedback_live.cpp
```diff
@@ -0,0 +1,498 @@
...
+    if (!isSending())
+        return;
+
+    // A surface that has just started talking after a silence is most likely a
+    // surface that has just connected, which is the closest thing to "a new
+    // peer" that is observable: JUCE does not report who sent a datagram.
+    const auto accepted = router_.acceptedMessageCount();
+    if (accepted != lastAcceptedCount_) {
+        const auto now = juce::Time::currentTimeMillis();
+        if (lastInboundMs_ != 0 && now - lastInboundMs_ >= kResyncSilenceMs)
```
lucaromagnoli: **P1 — Snapshot on the first accepted message**

`lastInboundMs_` starts at zero, so the first accepted packet can never request a snapshot. The snapshot sent when the destination was configured may have gone out while the UDP surface was offline and been lost; when that surface later starts and sends its first valid packet, this branch only seeds the timestamp, leaving it stale until traffic resumes after another five-second silence. Please treat `lastInboundMs_ == 0` as a resync case too, and cover the late-starting-surface path.

magda/daw/api/osc_feedback_live.cpp
```diff
@@ -0,0 +1,498 @@
...
+    // The router's drain is on the message thread, same as the tick, so both of
+    // these reach straight into the tables with nothing in between.
+    router_.setFeedbackTap([this](int slot, float value) { feedback_->noteReceived(slot, value); });
+    router_.setBindingFeedbackTap([this](const juce::String& address, float value) {
+        auto& entry = boundEntryFor(address);
+        entry.lastReceived = value;
+        entry.echoed = true;
+    });
+
+    changeToken_ = changes_.addListener(
```
lucaromagnoli, line 117: **P1 — Invalidate feedback when bindings change**

`OscService` listens to `BindingRegistry` and updates the inbound routes, but the projector only listens to `ChangeSource` and `Config`. Adding or editing an OSC binding after the initial tick therefore makes the input live without projecting the current target value; it stays blank or stale until an unrelated model/controller event sets `bindingsDirty_`. This is especially visible after OSC learn, whose capture is consumed without moving the target. Please listen for registry changes here and dirty or reconcile the bound-address state.

magda/daw/api/osc_feedback_live.cpp
```diff
@@ -0,0 +1,498 @@
...
+                // nothing distinguishes them. Marking everything dirty is right
+                // for both: a swap changes the values and the diff sends them,
+                // a save changes nothing and the diff sends nothing. Resetting
+                // here instead would resend the whole namespace every time the
+                // user pressed save.
+                mixerDirty_ = macrosDirty_ = bindingsDirty_ = true;
+                break;
+            case remote::Topic::Clips:
+            case remote::Topic::Session:
+            case remote::Topic::Automation:
```
lucaromagnoli: **P2 — Reproject bound plugin parameters during automation**

For an ordinary `PluginParam`, `AutomationPlaybackEngine::currentValueChanged` emits `automationValueChanged` and `Topic::Automation` but does not emit `Topic::Devices`. Ignoring this topic leaves `bindingsDirty_` false, so a bound motor fader or OSC control never follows the automated base value. Marking bindings dirty for automation activity would preserve the existing 30 Hz coalescing while keeping the surface in sync.

Issue #2094, opened 2026-08-14
Post-FX pre/post-fader toggle has no UI


lucaromagnoli:
Follow-up to #2087, which added the fader boundary the post-FX stage sits on.

## What exists

`TrackChain::postFxPostFader` decides which side of the track fader the post-FX stage and the mixer-analysis rail compile to. Both engines read it: `PlanCompiler` partitions on it, and `ensureVolumePluginPosition` places the fader against it. It defaults to post-fader, it serializes, and `TrackManager::setPostFxPostFader` is the setter.

## What is missing

Nothing in the UI calls that setter, so the only way to put the stage back pre-fader is to hand-edit `postFxPostFader` in the project file.

That is the same trap the OSC settings were in until #2091: a shipped capability that is unreachable without a text editor. It matters more than usual here because the default changed. A project that had a processor in POST-FX was compiled pre-fader before #2087 and is compiled post-fader after it, and right now there is no in-app way to say "I wanted the old routing".

## What this needs

A per-chain switch. Where it goes is a UX call rather than an obvious one: `PostFxPanelContent` currently has no header, just the vertical "POSTFX" watermark down the add strip and the trailing "+" append zone, so there is no existing row to hang it off.

Worth deciding at the same time whether the panel should show which side it is on even when the switch is not the thing being looked at, since "is this meter reading before or after my fader" is the question the whole of #2087 came from.

## Not this issue

Per-device placement, so a meter can sit post-fader while a limiter in the same section stays pre-fader. That was considered for #2087 and deferred to v1. Both compilers already partition on a value rather than a constant, so widening the flag from per-chain to per-device is a widening rather than a rewrite, but it needs its own UI thinking and a serialization change.

Issue #730, opened 2026-02-18
Add post-FX section with dedicated bottom panel


lucaromagnoli:
## Description

Create a post-FX processing section that lives in a new hideable bottom panel area. This panel should allow users to add processing after the main FX chain (e.g. dithering, final limiting, metering).

## Requirements

- New bottom panel area (distinct from existing bottom panel) or a section within it
- Panel can be shown/hidden independently
- Post-FX chain editing UI
- Ties into the track signal flow after the main FX chain

## Design Considerations

- Consider whether this is a sub-panel of the existing bottom panel or a separate panel
- Should integrate with the panel collapse/expand system (see related issue)

PR #2095, opened 2026-08-14
Post-FX fader boundary: one contract both engines read (#2087)

### LINKED ISSUE #2087: Levels analyzer should reflect the track fader
Raised from discussion #2042: https://github.com/Conceptual-Machines/magda-core/discussions/2042

## Problem

The built-in Levels analyzer continues to show signal when the track mixer fader is reduced, including at `-inf`. Clip gain affects the reading, which indicates that the analyzer is currently receiving the pre-fader signal.

The UI/model describes this as post-fader analysis, so its reading should represent the signal leaving the track mixer.

## Reproduction

1. Place an audio clip on a track and start playback.
2. Open the built-in Levels analyzer on that track.
3. Pull the track mixer fader down to `-inf`.
4. Observe that the analyzer continues to report the clip signal.
5. Reduce the clip gain and observe that the analyzer reading changes.

## Expected behavior

The Levels analyzer should receive post-fader audio. Pulling the track fader to `-inf` should therefore remove the ongoing source signal from its reading.

## Engine migration note

This should be defined as an engine-neutral routing contract rather than fixed only in the Tracktion bridge. The native engine currently compiles `postFxChainElements` and `mixerAnalysisElements` before `TrackFader`, so cutover by itself would preserve the current behavior.

Apply the intended ordering to both engines while Tracktion remains active, and cover it in the parity suite for epic #1882.

## Acceptance criteria

- Levels analysis on an ordinary track receives post-fader audio.
- Changing the track fader changes the Levels reading accordingly.
- At `-inf`, a continuously playing dry source no longer registers in Levels.
- Tracktion and the native engine agree on the routing contract.
- Automated coverage asserts analyzer/fader ordering and observable level behavior.

lucaromagnoli:
Closes #2087.

The Levels analyzer kept showing a track's signal with the fader pulled to `-inf`, and clip gain still moved the reading, because the post-FX stage was compiled before the fader. In both engines.

## What was actually there

Worth stating, because the naming has been misleading for two releases.

`de7a0b7c`, the commit that introduced the section, ends with: "Post-fader audio routing (the fader boundary in PluginManagerSync) and the post-fx panel UI are **deferred**." #1335 stated the shipped routing as `flatten(fx) -> postFx -> VolumeAndPan -> LevelMeter`, and #730 asked for the section so users could put "dithering, final limiting, metering" after the FX chain. So "post-FX" meant *after the FX chain*, the fader boundary was a known gap, and the stage has always held real processing that renders.

What drifted is the wording. `PostFxPanelContent.hpp:18`, `TrackChain.hpp:15` and `TrackManager.cpp:2071/2094` all call it post-fader; only `TrackChain.hpp:34` said what was true. This PR is the deferred boundary, plus the switch that says which side.

## The contract

`TrackChain::postFxPostFader`, default true, read by both compilers rather than an ordering each hardcodes and drifts on:

```
post-fader (default):  fx -> preFaderSends -> Fader -> postFx -> mixerAnalysis -> postFaderSends -> Meter -> Mute
pre-fader:             fx -> postFx -> mixerAnalysis -> preFaderSends -> Fader -> postFaderSends -> Meter -> Mute
```

One switch per chain. Per-device placement is the v1 upgrade; both compilers already partition on a value rather than a constant, so widening it is a widening rather than a rewrite.

**Native.** `PlanCompiler` emits the post-FX stage and the mixer-analysis rail through one lambda called from one of two places, so the two sides cannot come to mean different things. Post-fader it runs between the fader and the post-fader sends, which keeps the sidechain tap and the mute downstream of it and leaves where a keyed compressor listens unchanged. A post-fader send therefore carries the post-FX processing and a pre-fader send does not, which is asserted both ways.

**Tracktion.** `ensureVolumePluginPosition` treated `LevelMeterPlugin` as the only thing that legitimately follows the fader and hoisted the fader past everything else. That is what made a post-fader stage unrepresentable. It now places the fader against a set of plugins that belong after it.

## A second thing that set was hiding

`PluginManagerMeasurement.cpp:62` appends the always-on `TrackMeasurementPlugin` at the end of the chain, commented "Post-fader: append at the very end of the chain (after volume/pan and TE's level meter)". The next chain sync hoisted the fader past it, so it was pre-fader from that moment on — and it is what the mixing agent and `MixAnalysisService` read.

It is in the post-fader set here, unconditionally rather than under the post-FX flag, because its placement should not depend on whether the user happens to own a post-FX device. Same root cause, one line, and leaving it would have left an inconsistency keyed on an unrelated fact.

## Behaviour change, for the release notes

**The post-FX stage.** A project saved before this compiled its stage pre-fader and loads with it below the fader. A meter there starts telling the truth. A processor there now takes the fader's output, so the track sounds different. `postFxPostFader: false` is how a project asks for the old routing back — though not yet from the UI, which is #2094.

**Post-fader sends.** `SendInfo::preFader` is read by `PlanCompiler` and by nothing else: on the TE side the flag is serialized, round-tripped through DAWproject, and then ignored when the `AuxSendPlugin` is placed. Sends were appended and never sequenced, which looked consistent only because the fader was hoisted past them all, so every send tapped pre-fader audio whatever its flag said. They are sequenced by the flag now, so a post-fader send follows the fader for the first time — which is what it has always claimed to do. Sends default to post-fader, so this reaches any project using one.

## Tests

- **Plan ordering** as reachability rather than op indices: which signal each stage reads, on both placements, plus the send tap on each side of the boundary and the mixer-analysis rail following the stage.
- **Tracktion plugin-list ordering** across a toggle, in the JUCE target, identifying devices by path rather than by name.
- **Observable level behaviour**, which is the acceptance criterion the ordering assertions stand in for: a block of full-scale DC walked through the synced plugin list in order, read off the `LevelsPlugin` snapshot at unity and at silence. Two things that shape it: the fader ramps its gain across a block rather than stepping, so the reading is taken after the ramp settles; and the buffer peak is asserted alongside the meter, because if the fader is not attenuating at all then the meter reading says nothing about which side of it the meter sits on.
- **The negative case**, which is what makes the rest mean anything: with the stage put back pre-fader, the meter reads the same whatever the fader does. The reported bug, reproduced on demand.
- The old pre-fader assertion in `test_midi_signal_routing_juce.cpp` updated to the new contract.
- Plan golden regenerated. The diff is the fader moving up four ops; total latency unchanged.

Built locally. Catch2 2580 passed / 13 skipped / 0 failed, JUCE target all passed, `make format` and `make lint-changed` clean.

## Follow-up

#2094 — the toggle has no UI, so putting the stage back pre-fader currently means hand-editing the project file.

### DIFF
--- a/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
+++ b/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
@@ -831,9 +831,11 @@ void PluginManager::syncTrackPlugins(TrackId trackId) {
         removeSidechainMonitor(trackId);
 
     // Create TE plugins for post-FX devices (flat list, no racks/instruments).
-    // Inserted at -1 (append); the reorder pass below places them after the fx
-    // tree and ensureVolumePluginPosition keeps the fader after them, so the
-    // final order is [fx..., postFx..., mixerAnalysis..., VolumeAndPan, LevelMeter].
+    // Inserted at -1 (append); the reorder pass below sequences them, and which
+    // side of VolumeAndPan they land on is TrackChain::postFxPostFader (#2087):
+    //
+    //   post-fader: [fx..., VolumeAndPan, postFx..., mixerAnalysis..., LevelMeter]
+    //   pre-fader:  [fx..., postFx..., mixerAnalysis..., VolumeAndPan, LevelMeter]
     auto loadFlatSection = [&](const std::vector<PostFxChainElement>& section, auto pathBuilder) {
         for (const auto& elem : section) {
             const auto& device = elem.device;
@@ -967,7 +969,7 @@ void PluginManager::syncTrackPlugins(TrackId trackId) {
 
     // Ensure VolumeAndPan is near the end of the chain (before LevelMeter)
     // This is the track's fader control - it should come AFTER audio sources
-    ensureVolumePluginPosition(teTrack);
+    ensureVolumePluginPosition(trackId, teTrack);
 
     // Ensure LevelMeter is at the end of the plugin chain for metering
     addLevelMeterToTrack(trackId);
@@ -1430,7 +1432,7 @@ void PluginManager::pollAsyncPluginLoad(const ChainNodePath& devicePath, te::Plu
     });
 }
 
-void PluginManager::ensureVolumePluginPosition(te::AudioTrack* track) const {
+void PluginManager::ensureVolumePluginPosition(TrackId trackId, te::AudioTrack* track) const {
     if (!track)
         return;
 
@@ -1466,26 +1468,91 @@ void PluginManager::ensureVolumePluginPosition(te::AudioTrack* track) const {
     if (volPanIndex < 0)
         return;
 
-    // Check if there are any non-utility plugins after VolumeAndPan.
-    // VolumeAndPan is the track fader — it must come AFTER all audio-producing
-    // plugins (instruments, racks, FX, sends) and only before LevelMeter.
-    bool needsMove = false;
-    for (int i = volPanIndex + 1; i < plugins.size(); ++i) {
-        if (!dynamic_cast<te::LevelMeterPlugin*>(plugins[i])) {
-            needsMove = true;
+    // What legitimately belongs after the fader (#2087). Until now the answer
+    // was "only LevelMeter", which is why the fader was hoisted past everything
+    // else and why the post-FX stage could never actually be post-fader.
+    const auto postFaderPlugins = collectPostFaderPlugins(trackId);
+
+    // Where the fader goes: immediately before the first plugin that belongs
+    // after it, or at the end when nothing does.
+    int firstPostFader = -1;
+    for (int i = 0; i < plugins.size(); ++i) {
+        if (plugins[i] == volPanRaw)
+            continue;
+        if (postFaderPlugins.count(plugins[i]) != 0) {
+            firstPostFader = i;
             break;
         }
     }
 
-    if (!needsMove)
-        return;
+    if (firstPostFader < 0) {
+        // Nothing post-fader: the fader is last but for the meter, which is the
+        // rule this function has always enforced.
+        bool needsMove = false;
+        for (int i = volPanIndex + 1; i < plugins.size(); ++i) {
+            if (!dynamic_cast<te::LevelMeterPlugin*>(plugins[i])) {
+                needsMove = true;
+                break;
+            }
+        }
+        if (!needsMove)
+            return;
 
-    // Move VolumeAndPan to the end of the list.
-    // addLevelMeterToTrack() runs right after this and ensures LevelMeter
-    // is always the very last plugin, so the final order will be:
-    // [instruments, FX, sends, ..., VolumeAndPan, LevelMeter]
+        // addLevelMeterToTrack() runs right after this and ensures LevelMeter is
+        // always the very last plugin, so the final order will be:
+        // [instruments, FX, sends, ..., VolumeAndPan, LevelMeter]
+        volPanPlugin->removeFromParent();
+        plugins.insertPlugin(volPanPlugin, -1, nullptr);
+        return;
+    }
+
+    if (volPanIndex == firstPostFader - 1)
+        return;  // already immediately before the post-fader run
+
+    // Recompute the anchor after the removal rather than adjusting the index by
+    // hand: removing the fader shifts everything after it down by one only when
+    // the fader was earlier in the list, and getting that wrong puts the fader
+    // one slot inside its own post-fader run.
+    auto* anchor = plugins[firstPostFader];
     volPanPlugin->removeFromParent();
-    plugins.insertPlugin(volPanPlugin, -1, nullptr);
+    const int insertAt = plugins.indexOf(anchor);
+    plugins.insertPlugin(volPanPlugin, insertAt, nullptr);
+}
+
+std::unordered_set<te::Plugin*> PluginManager::collectPostFaderPlugins(TrackId trackId) const {
+    std::unordered_set<te::Plugin*> result;
+
+    const auto* trackInfo = TrackManager::getInstance().getTrack(trackId);
+    if (trackInfo == nullptr)
+        return result;
+
+    // The always-on measurement tap, which the mixing agent and MixAnalysisService
+    // read. PluginManagerMeasurement appends it last and says "Post-fader" while
+    // doing so; before this it was hoisted above the fader on the next chain sync,
+    // so both of those were reading pre-fader audio. Included here rather than
+    // left to the post-FX flag because it is post-fader unconditionally and its
+    // placement should not depend on whether the user happens to own a post-FX
+    // device.
+    if (auto tap = trackMeasurementTaps_.find(trackId); tap != trackMeasurementTaps_.end()) {
+        if (tap->second)
+            result.insert(tap->second.get());
+    }
+
+    if (!trackInfo->chain.postFxPostFader)
+        return result;
+
+    auto addSection = [&](const std::vector<PostFxChainElement>& section, auto pathBuilder) {
+        for (const auto& elem : section) {
+            juce::ScopedLock lock(pluginLock_);
+            auto it = findSyncedDevice(pathBuilder(trackId, elem.device.id));
+            if (it != syncedDevices_.end() && it->second.plugin)
+                result.insert(it->second.plugin.get());
+        }
+    };
+    addSection(trackInfo->chain.postFxChainElements, &ChainNodePath::postFxDevice);
+    addSection(trackInfo->chain.mixerAnalysisElements, &ChainNodePath::mixerAnalysisDevice);
+
+    return result;
 }
 
 // =============================================================================
@@ -1742,7 +1809,7 @@ void PluginManager::syncMultiOutTrack(TrackId trackId, const TrackInfo& trackInf
     }
 
     // Ensure VolumeAndPan and LevelMeter are present
-    ensureVolumePluginPosition(teTrack);
+    ensureVolumePluginPosition(trackId, teTrack);
     addLevelMeterToTrack(trackId);
 
     if (pluginOrderChanged)
--- a/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
+++ b/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
@@ -892,58 +892,7 @@ void PluginManager::syncTrackPlugins(TrackId trackId) {
             }
         }
 
-        // The rest of the strip, in the order the native compiler emits it:
-        //
-        //   fx -> [postFx, mixerAnalysis if pre-fader] -> preFaderSends
-        //      -> fader -> [postFx, mixerAnalysis if post-fader] -> postFaderSends
-        //
-        // The fader is not in this list; ensureVolumePluginPosition slots it in
-        // afterwards, immediately before the first plugin that belongs after it.
-        // That only lands correctly if everything pre-fader precedes everything
-        // post-fader here, which is what the two groups below are for.
-        auto appendSection = [&](const std::vector<PostFxChainElement>& section, auto pathBuilder) {
-            juce::ScopedLock lock(pluginLock_);
-            for (const auto& elem : section) {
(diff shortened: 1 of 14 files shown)


magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -1466,26 +1468,91 @@ void PluginManager::ensureVolumePluginPosition(te::AudioTrack* track) const {
...
-    // VolumeAndPan is the track fader — it must come AFTER all audio-producing
-    // plugins (instruments, racks, FX, sends) and only before LevelMeter.
-    bool needsMove = false;
-    for (int i = volPanIndex + 1; i < plugins.size(); ++i) {
-        if (!dynamic_cast<te::LevelMeterPlugin*>(plugins[i])) {
-            needsMove = true;
+    // What legitimately belongs after the fader (#2087). Until now the answer
+    // was "only LevelMeter", which is why the fader was hoisted past everything
+    // else and why the post-FX stage could never actually be post-fader.
+    const auto postFaderPlugins = collectPostFaderPlugins(trackId);
```
lucaromagnoli: This changes which side of the fader every aux send sits on, as a side effect keyed on owning a post-FX device. AuxSendPlugins are not in the reorder pass's desiredOrder, so the contiguous packing leaves them after the postFx/MA run; placing the fader before the first postFx plugin therefore puts every send on that track downstream of both the fader and the post-FX stage. On a track without a post-FX device the anchor is the measurement tap, which sits past the sends, so they stay pre-fader. That is the same inconsistency-keyed-on-an-unrelated-fact this function's comment calls out for the tap, now applied to sends, and SendInfo::preFader is honoured nowhere on the TE side (the one grep hit is the struct field) while the native compiler splits sends by it and asserts both directions. The fix that matches the PR's own contract is to place sends by their flag, post-fader sends in this set and pre-fader ones out of it, which would make TE honour preFader for the first time; if that is out of scope here, the send movement at least belongs in the release-notes section and pinned in the JUCE test, because a send flipping across the fader is as audible as the meter fix.

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -1466,26 +1468,91 @@ void PluginManager::ensureVolumePluginPosition(te::AudioTrack* track) const {
...
+        if (tap->second)
+            result.insert(tap->second.get());
+    }
+
+    if (!trackInfo->chain.postFxPostFader)
+        return result;
+
+    auto addSection = [&](const std::vector<PostFxChainElement>& section, auto pathBuilder) {
+        for (const auto& elem : section) {
+            juce::ScopedLock lock(pluginLock_);
```
lucaromagnoli: The lock is taken and released once per element; hoisting it around the loop is cheaper and reads as one critical section. Also worth a look: trackMeasurementTaps_ is read a few lines up without any lock, so either that map is message-thread-only and the lock here guards only syncedDevices_, or the tap read is missing the same lock. A one-line comment saying which would settle it.

lucaromagnoli (commented):
The core of the PR is right: the compiler halves genuinely read one contract, the emitPostFx lambda makes the two placements structurally incapable of drifting, the reachability-style plan tests and the level-behaviour JUCE test with its negative case are exactly the right shape, the anchor-recompute in the fader move avoids the off-by-one it documents, the TrackChain copy function and serializer both carry the new flag with the right default-on-load semantics, and the golden diff is the fader moving with total latency unchanged. One real finding inline about aux sends, plus a locking nit.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -808,6 +819,17 @@ void Compiler::emitTrack(const TrackInfo& track) {
                          INVALID_DEVICE_ID, OpRole::TrackFader, 0};
     PortRef out{addOp(OpKind::Fader, faderKey, {signal.audio, noInput()}, {SignalKind::Audio}), 0};
 
+    // Post-fader, the stage runs between the fader and the post-fader sends, so
+    // a send tapped after the fader carries the post-FX processing with it and a
+    // meter dropped in here reads what leaves the track. The sidechain tap and
+    // the mute stay downstream of it either way, which is what keeps this from
+    // changing where a compressor keyed off this track listens.
+    if (track.chain.postFxPostFader) {
```
lucaromagnoli: **P1 — Keep the master chain on the same side in both engines**

This branch also runs for `MASTER_TRACK_ID`, so the native plan now puts master post-FX and mixer analysis after the master fader by default. Tracktion takes the separate `syncMasterPlugins()` path instead: those devices remain in `getMasterPluginList()`, and the Tracktion graph processes that whole list before its separate `getMasterVolumePlugin()`. Consequently `postFxPostFader=true` is ignored for the master in Tracktion, while native honors it; master limiters, meters, and the measurement tap change behavior at engine cutover. Please either implement the master boundary in Tracktion or scope the native/model contract until both engines can represent it, with a master parity test.

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -967,7 +969,7 @@ void PluginManager::syncTrackPlugins(TrackId trackId) {
 
     // Ensure VolumeAndPan is near the end of the chain (before LevelMeter)
     // This is the track's fader control - it should come AFTER audio sources
-    ensureVolumePluginPosition(teTrack);
+    ensureVolumePluginPosition(trackId, teTrack);
```
lucaromagnoli, line 902: **P1 — Preserve per-send fader placement in Tracktion**

`TrackInfo::sends` carries `preFader`, and the native compiler now emits the two classes on opposite sides of this boundary, but the Tracktion sync path only reads each send bus and level; neither this placement pass nor `collectPostFaderPlugins` classifies `AuxSendPlugin`s from `SendInfo::preFader`. Moving the fader here therefore moves it across every send as one group: with the default post-fader stage a pre-fader send can remain downstream, while with the stage switched pre-fader a post-fader send can remain upstream. Please position the AuxSend plugins from their per-send flag and add Tracktion parity coverage for both classes.

lucaromagnoli:
Both findings were right. Fixed in `85955728`.

**Sends.** Confirmed, and worse than the side effect it looked like. `SendInfo::preFader` is read by `PlanCompiler` and by nothing else on the TE side — the only other hits are the struct field, the serializer and the DAWproject adapter. Sends were appended to the plugin list and never sequenced, which was harmless only while the fader was hoisted past everything: every send ended up above it whatever the flag said. My change then made the side depend on whether the track owned a post-FX device, because the anchor on a track without one is the measurement tap, which sits past the sends.

So it gets the same answer as the tap rather than a release-note line: sends are sequenced by their flag, and the desired order is now the one the native compiler emits — fx, the stage if pre-fader, pre-fader sends, fader, the stage if post-fader, post-fader sends. That makes TE honour `preFader` for the first time, so a post-fader send follows the fader, which it has always claimed to do and never done. That is its own release-note line and I have added it to the PR body.

Pinned in the JUCE test with both flags, asserted both with and without a post-FX device present, since that presence is exactly what the sequencing exists to stop mattering.

**Locking.** Hoisted to one critical section for the whole walk. On the tap read: `trackMeasurementTaps_` is message-thread-only — every access in `PluginManagerMeasurement.cpp` is unlocked — and the lock is for `syncedDevices_`, which the async plugin-load path also touches. Said so in a comment where the tap is read.

Rebuilt: Catch2 2580 passed / 13 skipped / 0 failed, JUCE target all passed including the new send cases, format and lint clean.

lucaromagnoli:
Both right. Fixed in `7b235298`.

**Master.** Confirmed and scoped rather than implemented. `createMasterPluginNode` in `tracktion_EditNodeBuilder.cpp:1798-1806` builds the whole of `getMasterPluginList()` and only then wraps it in `getMasterVolumePlugin()`, so a plugin after the master fader is not expressible without patching TE's graph builder — that is a bigger piece of work than this PR, and doing it here would mean carrying a TE patch for a case #2087 does not ask about (its repro and acceptance criteria are both about an ordinary track).

So the master is pre-fader in both engines until TE can represent otherwise, which keeps the flag from becoming the one place they disagree. `compileTrack` takes `postFxPostFader && !isMaster`, with the reason at the call site, and a plan test asserts the master's post-FX device feeds the master fader rather than reading from it.

**Sends / the placement pass.** You were right that the previous commit did not finish the job, and the reason is worth recording: `syncMultiOutTrack` carried its own copy of the ordering pass, sequencing the post-FX stage and nothing else. So `85955728` fixed `syncTrackPlugins` and left multi-out tracks having the fader placed against a set the ordering had never arranged for — the same failure, one caller over.

Rather than fix the copy, both paths now call one `appendStripOrder`, which groups the strip as `fx -> [stage if pre] -> preFaderSends -> (fader) -> [stage if post] -> postFaderSends`. The duplication is what let the first fix look complete while missing half the callers, so removing it is the actual fix.

Catch2 2580 passed / 13 skipped / 0 failed, JUCE target all passed, format and lint clean.

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -1687,18 +1815,7 @@ void PluginManager::syncMultiOutTrack(TrackId trackId, const TrackInfo& trackInf
...
-                juce::ScopedLock lock(pluginLock_);
-                auto it = findSyncedDevice(devicePath);
-                if (it != syncedDevices_.end() && it->second.plugin &&
-                    teTrack->pluginList.indexOf(it->second.plugin.get()) >= 0)
-                    desiredOrder.push_back(it->second.plugin.get());
-            }
-        };
-        appendSection(trackInfo.chain.postFxChainElements, &ChainNodePath::postFxDevice);
-        appendSection(trackInfo.chain.mixerAnalysisElements, &ChainNodePath::mixerAnalysisDevice);
+        appendStripOrder(trackId, trackInfo, *teTrack, desiredOrder);
```
lucaromagnoli, line 1827: **P1 — Reconcile sends in the multi-output sync path before ordering them**

`appendStripOrder` can only move `AuxSendPlugin`s already present in `teTrack->pluginList`, but `syncMultiOutTrack` never runs the send-reconciliation block used by `syncTrackPlugins` (the special-path return happens before that block). Consequently a multi-output `TrackInfo` with sends reaches this call with no send plugins: both `appendSends` and `collectPostFaderPlugins` no-op, so Tracktion drops the sends entirely while the native compiler emits them. Please share the send reconciliation logic as well, and pin the multi-output case with a test that asserts both send plugins exist and straddle the fader according to `preFader`.

lucaromagnoli:
Fixed in `3fc42080`. One of the two was already addressed — evidence below rather than a claim.

**Multi-output sends — right, and it uncovered a second bug.**

`syncTrackPlugins` returns to the multi-out path at line 442, long before the send reconciliation at 732, so a multi-out track had no `AuxSendPlugin`s at all. Not misplaced: absent, while the native compiler emitted the sends from the same model. Ordering could not have surfaced that, since `appendStripOrder` only moves plugins that already exist.

The reconciliation is now a shared `reconcileSends` called by both paths, for the same reason `appendStripOrder` is shared.

That alone did not fix it, and the reason is its own bug. The rack instance carrying the multi-out routing was resolved through three early returns at 1705, 1709 and 1724, and **nothing after them depends on it** — it is used once, to insert the instance. So a source device that had not finished loading, or an output pair that did not resolve, cost the child track its fx devices, its post-FX stage, its sends, its ordering and its fader placement. It is a guarded block now and the generic tail always runs. That is what made the test go green; the shared helper on its own did not.

Pinned by a multi-out case built the way the model builds one, through `activateMultiOutPair` — a hand-made `multiOutLink` is not enough, which is how I found the early returns in the first place.

**Per-send placement (the comment at :957).** Already done in `85955728`, which the diff should show: `appendStripOrder` splits on `send.preFader` at `PluginManagerSync.cpp:1456-1475` and `collectPostFaderPlugins` classifies post-fader sends at `:1593`. Coverage for both classes has been in `test_post_fx_fader_order_juce.cpp` since that commit, asserted with and without a post-FX device present. I think that one is a re-anchor of the earlier finding rather than a new one — say if you are seeing something I am not.

Catch2 2580 passed / 13 skipped / 0 failed, JUCE target all passed, format and lint clean.

PR #2097, opened 2026-08-15
Own the OSC receive loop so feedback can answer whoever is talking (#2096)

### LINKED ISSUE #2096: Own the OSC receive loop so feedback can answer whoever is talking
Follow-up to #2091. Would delete a setting rather than add one.

## Why `oscFeedbackHost` exists

It is not a thing a user should have to know. Feedback needs a destination only because MAGDA cannot see where a message came from: `juce::OSCReceiver` hands a listener a parsed `OSCMessage` and nothing else, and the parser — `OSCInputStream` — is defined inside `juce_OSCReceiver.cpp` with no public header. The datagram's source address is consumed and dropped inside JUCE's own receive loop.

So the user is asked for their tablet's IP address to work around a JUCE limitation.

## Why this is closer than it sounds

`OscService` already binds its own `juce::DatagramSocket` and hands it to `connectToSocket`, because `OSCReceiver::connect(port)` binds every interface unconditionally and `oscBindAddress` had to mean something. We own the socket; we do not own the read loop.

`juce::DatagramSocket::read` reports the sender's IP and port. Reading the datagrams ourselves and parsing them is the whole of the gap.

## What it buys

- **`oscFeedbackHost` goes away.** Feedback answers whoever is talking. Zero configuration for the common case of one surface.
- **Snapshot on connect becomes real.** #2091 could only approximate it: a snapshot fires when inbound traffic resumes after five seconds of silence (`kResyncSilenceMs`), because "a surface appeared" was not observable. With a peer address it is exactly observable.
- **Echo suppression becomes per-peer.** Today the slot table has one destination, so a value from one surface suppresses the echo to every surface. With two surfaces on the same mixer that is wrong, and it is unfixable without knowing who sent what.
- **The OSC tab can show connected surfaces** instead of a message counter, which is the first thing anyone debugging a silent surface wants.

## What it costs

An OSC 1.0 parser: address pattern, type tag string, the argument types MAGDA accepts, and bundles including nesting. Roughly 300 lines with tests, and it has to be exactly as strict as `parseOscAddress` already is about what it will not accept.

Worth noting the receive thread is a real-time-ish path — `OscRouter::handleMessage` is documented as allocation-free apart from the drain hop — so the parser has to hold that line.

## What stays

`oscFeedbackPort`. Surfaces listen on a different port than they send from: TouchOSC sends from an ephemeral port and listens on 9001, so the reply port cannot be inferred from the sender.

## Not worth doing for tidiness

If one configured destination is fine in practice, this is a lot of code to delete one field. It earns itself on the second surface, or on wanting the peer list. Filing it so the reason `oscFeedbackHost` exists is written down somewhere other than a comment.

lucaromagnoli:
Closes #2096.

## Why

`oscFeedbackHost` asked the user for their tablet's IP address to work around a JUCE limitation. `juce::OSCReceiver` reads the datagram's source address into a local inside its own loop and drops it, and its parser has no public header, so taking the parse meant taking the loop that threw the peer away.

`OscService` already bound its own socket, because `OSCReceiver::connect(port)` binds every interface unconditionally and `oscBindAddress` had to mean something. Only the read loop was not mine. `DatagramSocket::read` reports the sender.

## What changed

**A parser.** `OscPacket` reads OSC 1.0: address pattern, type tag string, the argument types JUCE accepted plus the rest of the 1.0 set, and bundles including nesting.

It allocates nothing. `OscRouter::handleMessage` is documented as allocation-free apart from the drain hop, and a parser in front of it that allocated per datagram would move the problem rather than solve it, so an address is a `juce::StringRef` straight into the datagram (OSC strings are NUL-terminated in place) and arguments land in a fixed inline array. The one allocation left on that thread is JUCE's, in `DatagramSocket::read`, and the receive loop says so rather than pretending otherwise.

A malformed packet changes nothing: the bytes are walked twice, once to validate and once to deliver. Delivering as it went would apply the first two elements of a `[stop, position, play]` cue before discovering the third was corrupt, which is a stop and a locate the sender never asked for.

Addresses must be printable ASCII. `juce::StringRef`'s constructor asserts it, so without the check a hostile packet is a debug-build crash reachable from an unauthenticated UDP port. OSC 1.0 says the same thing and `parseOscAddress` would reject anything else a byte later.

**A peer table.** `OscPeers`, owned by `OscRouter`, bounded at eight with least-recently-heard eviction. An `OscPeerId` then travels with every message through the slot table, the ordered ring, the binding routes and both feedback taps.

Identity is the host, not the host and port. A surface sends from an ephemeral port and listens on a fixed one, which is exactly why `oscFeedbackPort` survives this issue, and it also means the sending port is not part of who a peer is: keying on it would mint a new peer, and a new snapshot, every time a surface recycled its socket. The cost is that two surfaces on one machine are one peer, which is the same limitation stated from the other end.

**Four things follow.**

- `oscFeedbackHost` is deleted, from `Config`, its JSON and the dialog. Enabling OSC is now the whole of the configuration for one surface. It was added in #2091, which has not shipped, so there is no migration to write.
- Snapshot on connect is exact. #2091 could only fire one when inbound traffic resumed after five seconds of silence, so `kResyncSilenceMs` and its timestamp bookkeeping are deleted rather than kept as a second path.
- Echo suppression is per peer. `OscFeedbackProjector` keeps one `OscFeedback` per surface, so a fader moved on one tablet still reaches the other, which one shared table could not express. What is not per surface is the expensive half: resolving a binding's target runs once per tick and the answer goes to everyone.
- The OSC tab lists the surfaces MAGDA has heard from, with datagrams in, feedback out and how long ago each was last heard, instead of one message counter.

## Tests

New `tests/test_osc_packet.cpp` builds packets from the wire format rather than from anything the reader shares, so a mistake in the reader cannot be a mistake both ends agree on, and so the malformed cases are expressible at all: truncated address, type tag string without its comma, an argument the tags promised but did not carry, an unknown type tag, a non-ASCII address, a bundle element whose size runs past the buffer, nesting past the depth cap, and more arguments than the inline array holds. Plus the two-pass guarantee, as a bundle whose last element is corrupt delivering none of the good ones.

`test_osc_feedback.cpp` gains the per-peer cases: a surface appearing is snapshotted and the one already there is not, a value from one surface is suppressed to it and sent to the other, a surface stops being answered when its peer goes, and a gesture landing before the first tick still counts as an echo.

`test_osc_service_juce.cpp` gains the end-to-end pair: a sender becomes a peer with the host it sent from and stays one peer across datagrams, and closing the socket forgets who was talking to it.

Full Catch2 suite 2602 passed / 13 skipped, JUCE target all passed, app builds clean.

## Two tooling fixes in here as well

`make lint` and `make lint-changed` exited 132 or 139 on every file and printed nothing. `Checks: *` pulls in `performance-noexcept-move-constructor` and `bugprone-pointer-arithmetic-on-polymorphic-object`, which on Homebrew LLVM 20.1.4 die with SIGILL and SIGSEGV on any file in this project, and a crashing check takes the whole run down. Disabled along with their aliases, since clang-tidy treats an alias as its own check name. Underneath that, the compile database is written by Apple's `/usr/bin/c++`, which finds the macOS SDK through `xcrun` and records no `-isysroot`, so Homebrew's clang-tidy could not find `<vector>` and every translation unit stopped at its first standard header. The lint targets now pass the SDK path.

Separately, the `compile_commands.json` at the repo root was months stale, which is why the editor reported "file not found" on JUCE includes in files that compile cleanly. Deleted, and `.clangd` names `cmake-build-debug` instead.


magda/daw/audio/osc/OscService.cpp
```diff
@@ -1,16 +1,102 @@
...
+            // port and listens on a fixed one, so this number is not where a
+            // reply goes and not part of who a peer is — see `OscPeers`. It is
+            // here because the overload that reports the host reports both.
+            int senderPort = 0;
+            const int bytes =
+                socket_.read(buffer_.getData(), kDatagramBufferSize, false, senderHost, senderPort);
+            if (bytes < kMinDatagramSize)
+                continue;
+
+            // Interned before the parse rather than after, so a datagram that
```
lucaromagnoli, line 71: This choice plus snapshot-on-new-peer makes the answering half a small UDP reflector: any four bytes of garbage with a spoofed source address mints a peer, and the next tick opens a sink to that host and streams a full snapshot at up to kMaxMessagesPerFlush per flush. Cycling nine or more spoofed sources through the eight-slot LRU sustains that indefinitely, at up to a few hundred surface creations a second, each owed a snapshot, aimed at whatever hosts the attacker names, and it evicts the real surface as a side effect, so the legitimate tablet gets churned and re-snapshotted too. The bind address being the whole access-control story contains this on 127.0.0.1, but 0.0.0.0 is the default and the tablet case. The fix that keeps what this comment wants: intern stays where it is so the UI still counts unparseable traffic, but the peer entry grows a hasValidMessage bit set by the loop after a successful parse, and syncSurfaces only builds surfaces for peers that have it. One bool, and answering requires having said at least one well-formed thing. Damping generation bumps to answerable peers would also stop the spoof churn from making the projector rebuild its fleet every tick.

lucaromagnoli (commented):
Read the whole diff with the parser under the microscope, since it faces an unauthenticated port. It holds: every read is bounds-checked before it happens, the arithmetic cannot overflow with sizes bounded by a datagram, recursion is depth-capped, the ASCII gate closes the StringRef assert that would otherwise be a remotely reachable debug crash, unknown tags and over-long argument lists reject rather than guess or truncate, and the two-pass walk being one function makes the all-or-nothing guarantee structural. The wire-format-built test corpus exercises each rejection path distinctly, which is the right way to test a reader. The peer table's LRU and generation design is sound, the router threads the peer id with the same value-before-bit discipline the slots already had, the receive loop gets the socket-outlives-thread ordering right in both the close path and the member order, and the projector's resolve-once-fan-out split keeps the per-surface cost to an array compare. The projector guarding surface identity by host as well as id closes the slot-reuse hole I went looking for. Config, UI (including the fail-closed bind combo and the lookAndFeelChanged handling), en.json, and both tooling fixes all check out. One security-shaped finding inline.

magda/daw/audio/osc/OscService.cpp
```diff
@@ -1,16 +1,102 @@
...
+            // turns out to be malformed still counts as the surface being
+            // there. "Something is arriving and none of it parses" is a
+            // different problem from silence, and the settings UI should be
+            // able to tell them apart.
+            peer_ = router_.peers().intern(senderHost, juce::Time::currentTimeMillis());
```
lucaromagnoli: **P1 — Do not enroll a feedback peer before accepting any OSC**

This interns the sender before the packet is parsed, and the receive callback also ignores whether `handleMessage` accepted anything. Therefore any UDP payload of at least four bytes—including malformed packets and valid-but-foreign addresses—creates a peer. On the next projector tick that peer gets a sender plus a full project snapshot; with the default all-interface bind and spoofable UDP source addresses, one tiny junk datagram can make MAGDA disclose state and emit a large feedback burst to a host that never sent a valid control message. The existing foreign-address service test only checks that no command was applied, so it misses this side effect. Please separate diagnostics for raw datagrams from feedback eligibility and enroll a peer only after at least one message is accepted, with malformed/foreign cases asserting that no surface is created.

magda/daw/api/osc_feedback_live.cpp
```diff
@@ -151,60 +165,145 @@ std::unique_ptr<ControllerFeedbackSink> OscFeedbackProjector::makeControllerFeed
...
+        const bool known =
+            std::any_of(surfaces_.begin(), surfaces_.end(), [&peer](const Surface& surface) {
+                return surface.peer == peer.id && surface.host == peer.host;
+            });
+        if (known)
+            continue;
+
+        auto sink = factory_(peer.host, feedbackPort_);
+        if (sink == nullptr)
+            continue;  // the socket would not open; the next sync tries again
```
lucaromagnoli: **P2 — Keep failed surface creation eligible for retry**

The comment says the next sync retries, but `syncSurfaces()` records `peersGeneration_ = generation` and clears `surfacesStale_` before reaching this branch. If the factory returns null once, the peer set has not changed, so every later tick exits at the generation fast path and this peer is never attempted again. A transient sender/socket failure therefore disables feedback for that surface until another peer joins or configuration changes. Please leave the projector stale when creation fails, or otherwise track pending peers, and add a fail-once/succeed-next-tick test.

magda/daw/audio/osc/OscPacket.cpp
```diff
@@ -0,0 +1,309 @@
...
+            std::size_t tagCount = 0;
+            if (!readString(pos, end, tags, tagCount))
+                return false;
+            if (tagCount == 0 || tags[0] != ',')
+                return false;
+            if (!readArguments(tags, tagCount, pos, end, arguments, count))
+                return false;
+        }
+
+        if (receiver_ != nullptr)
```
lucaromagnoli, line 217: **P2 — Reject unconsumed bytes at the end of a message**

After reading the arguments, `pos` is never required to equal `end`. A valid message followed by any four aligned junk bytes therefore passes both validation walks and is delivered; the current “not a whole number of words” test catches only a one-byte suffix. This also lets a bundle element whose declared size includes trailing garbage execute its leading command, contrary to the parser whole-packet/malformed-packet guarantee and JUCE checked-size behavior. Please require `pos == end` before delivery and pin an aligned trailing-word case.

magda/daw/audio/osc/OscService.cpp
lucaromagnoli: Fixed in 30ebe440.

Interning stays where it is, so the settings list still shows a host that is sending nothing MAGDA can parse. What is new is a second state: a peer becomes **answerable** only once `OscRouter::handleMessage` has accepted something from it, and `syncSurfaces` builds a surface only for answerable peers. The receive loop tracks acceptance across the whole packet, so one understood message in a bundle is enough and a bundle of nothing is not.

Two things beyond the one bool, because the flood still had teeth without them. The generation counter now moves on the answerable transition rather than on arrival, so spoofed traffic no longer makes the projector rebuild its fleet every tick. And eviction prefers peers that have never said anything usable, so cycling nine spoofed sources through the eight-slot table can no longer push the real tablet out of it and force a re-snapshot.

Tests: a peer heard but not understood gets no surface, and becomes one the moment it is understood; and end to end in the service test, junk plus a valid-but-foreign address plus one understood message is a single loopback peer with three datagrams counted and `answerable` true only because of the third.

magda/daw/api/osc_feedback_live.cpp
lucaromagnoli: Fixed in 30ebe440. Correct on both counts: the generation was recorded and the stale flag cleared before the factory ran, so the fast path meant one transient failure was permanent for that surface. `syncSurfaces` now sets `surfacesStale_` when creation fails, which is what makes the next tick look again. Test added: a factory that refuses once, then succeeds on the following tick.

magda/daw/audio/osc/OscPacket.cpp
lucaromagnoli: Fixed in 30ebe440. `message()` now requires `pos == end` after the arguments. The existing case only caught an unaligned suffix, so two are pinned: a message followed by an aligned junk word is rejected, and a bundle whose second element pads itself that way delivers none of the good elements either, which is the whole-packet guarantee it was leaking.

magda/daw/audio/osc/OscService.cpp
lucaromagnoli: Fixed in 30ebe440, taking the shape you suggested plus the two hardenings the same comment points at: generation bumps damped to the answerable transition, and eviction preferring peers that have never said anything usable so the churn cannot displace the real surface. Details in the reply on the P1 thread.

magda/daw/audio/osc/OscPeers.cpp
```diff
@@ -0,0 +1,131 @@
...
+    juce::int64 oldest = 0;
+
+    for (int i = 0; i < kMaxPeers; ++i) {
+        const auto& entry = entries_[static_cast<std::size_t>(i)];
+        if (!entry.used) {
+            chosen = i;
+            chosenAnswerable = false;
+            break;
+        }
+        const bool better = chosen < 0 || (chosenAnswerable && !entry.answerable) ||
```
lucaromagnoli: **P2 — Do not evict an answerable peer for unvalidated traffic**

When all eight slots are answerable, this selection necessarily chooses the oldest answerable entry, and `intern` overwrites it before the new datagram has parsed. A single spoofed junk packet can therefore still remove a real surface and bump the generation. When that surface speaks again it replaces the provisional noise entry and is re-snapshotted, returning the table to eight answerable entries so the attacker can repeat the churn indefinitely. The preference for non-answerable entries only protects the fleet once one such entry already exists. To make the stated hardening hold at capacity, please avoid admitting a new unvalidated host over an answerable slot—dropping/aggregating diagnostics at that bound is safer—or defer that replacement until acceptance, and pin eight answerable peers plus one junk sender.

lucaromagnoli: Fixed in 14e7e53b. You are right, and the previous pass only closed it once a noise entry already existed: at capacity `intern` still took the quietest answerable slot before the datagram had parsed.

Admission is two-phase now.

`intern` handles an unvalidated arrival. It will take a free slot or the quietest peer that has never said anything usable, and **never** one being answered; at capacity it returns `kNoOscPeer` and the datagram is routed with no peer at all. `admit` runs only after the router has accepted something. It marks an existing peer answerable, and it is the only path that can displace a surface, so a genuine ninth surface still takes the quietest slot while traffic that has proved nothing cannot.

Taking your point about the trade explicitly: the settings list stops counting new hosts once eight surfaces are connected. Losing a diagnostic line at that bound beats losing a surface, and it is stated in the header rather than left implicit.

`intern` returns the peer's prior answerable state so the hot path stays at one lock per datagram: every packet of a gesture skips `admit`.

Both shapes you asked for are pinned. Eight answerable peers plus twenty junk datagrams from a ninth host leaves the fleet size, the generation and the identity of the existing peers untouched. And a ninth sender that does say something usable takes the quietest slot, so refusing unvalidated hosts has not quietly become refusing real ones.

magda/daw/audio/osc/OscPeers.cpp
```diff
@@ -0,0 +1,168 @@
...
+                    if (chosen < 0 || entry.lastSeenMs < oldest) {
+                        chosen = i;
+                        oldest = entry.lastSeenMs;
+                    }
+                }
+            }
+
+            // Whatever was here belonged to another host, so its count goes
+            // with it. This datagram is the first from the one taking over.
+            auto& entry = entries_[static_cast<std::size_t>(chosen)];
```
lucaromagnoli, line 115: **P2 — Version peer identity across slot reuse**

This immediately gives the evicted host numeric id to the new surface, but router work already queued by the old host still carries only that id in the value table, ordered ring, and binding routes. If the projector observes this generation change and replaces the surface before that work drains, `surfaceFor(id)` attributes the old host value to the new host. Fixed-namespace snapshots happen to bypass the echo bit, but bound addresses do not: a held binding drain from the evicted peer can update the target and mark the replacement surface as having sent that value, causing `publishBinding` to suppress the first update even though the new surface still shows its pre-drain snapshot. Please carry a slot incarnation/token through the router queues and feedback taps, or otherwise reject taps whose id has been reused. A regression test can hold the drain, enqueue a bound value from the quietest peer, admit the ninth host, tick once to replace/snapshot surfaces, then drain and assert the ninth surface receives the resulting bound value.

lucaromagnoli: Fixed in d5f2ac7b, and you traced it exactly: the fixed namespace hid it because a snapshot ignores the echo bit, while a bound address did not.

I went for the stronger of your two options rather than a token, because it removes the failure mode instead of detecting it. An id is now minted from a counter that only ever climbs -- including across `clear`, since a rebind handing the old peers' numbers to whoever connects next is the same reuse one level up -- and the entry carries its id rather than being found by it. A stale id then matches nothing, which is what should happen to it, and there is no incarnation to keep in step or to wrap.

Two costs, both taken deliberately: `hostFor` is a scan of eight instead of an index, and the router carries a full id rather than a byte, so the per-slot peer column goes from 1.5 KB to 6 KB. That is a fair price for a class of bug that only appears under eviction.

Test is the one you described: hold the drain, queue a bound value from the peer about to go, replace it, tick so the replacement is snapshotted, then drain and assert the replacement still receives the next real update. It also asserts the two ids differ, which is the property the behaviour rests on.

magda/daw/audio/osc/OscPeers.cpp
```diff
@@ -0,0 +1,176 @@
...
+            return Arrival{.id = entry.id, .answerable = entry.answerable};
+        }
+    }
+
+    const int chosen = slotForUnvalidatedHost();
+    if (chosen < 0)
+        return {};  // every slot is a surface; this host waits for `admit`
+
+    auto& entry = entries_[static_cast<std::size_t>(chosen)];
+    entry.id = nextId_++;
```
lucaromagnoli: **P2 — Keep lifetime-unique peer IDs from overflowing**

`OscPeerId` is a signed `int`, but `intern()` advances `nextId_` for every new unvalidated host before the datagram has passed parsing/routing. Because the bounded table can continually replace an unanswerable entry, spoofed UDP source hosts can churn this counter without bound. Once it reaches `INT_MAX`, `nextId_++` has signed-overflow UB; if it wraps in practice, IDs are reused and the stale queued-work attribution bug this commit fixes returns. Please use a practically non-exhaustible unsigned ID (for example `uint64_t`, with a reserved no-peer value and an explicit exhaustion guard), or avoid allocating IDs to unadmitted traffic.

lucaromagnoli: Fixed in cfb324f6. Took both halves of the suggestion, because either alone still leaves a bound.

Ids are minted only by `admit` now, so traffic that has said nothing MAGDA understands cannot advance the counter at all -- a spoofed flood churns table slots and nothing else. And `OscPeerId` is a `std::uint64_t` with `kNoOscPeer` reserved as 0, so ids start at 1 and even a flood of *well-formed* OSC from spoofed hosts would need on the order of half a million years at a million admissions a second. There is an explicit guard at the maximum regardless, since the alternative at that bound is handing a live surface's id to somebody else.

A small simplification fell out: having an id and being answerable are the same fact now rather than two fields that have to agree, so `Entry::answerable()` is `id != kNoOscPeer`. An unvalidated host still takes a slot and is still counted, so the settings list can show that something is arriving and none of it parses; it just has no number.

Test: a real peer, then five hundred datagrams from two hundred and fifty distinct hosts none of which is understood, and the next admitted id is still `first + 1` with the real peer untouched.

Cost noted for the record: the router now carries 8 bytes per coalescing slot instead of 4, so `valuePeers_` is 12.5 KB.

magda/daw/audio/osc/OscService.cpp
```diff
@@ -1,16 +1,122 @@
...
+                continue;
+
+            // Interned before the parse rather than after, so a datagram that
+            // turns out to be malformed still counts as the surface being
+            // there. "Something is arriving and none of it parses" is a
+            // different problem from silence, and the settings UI should be
+            // able to tell them apart.
+            const auto now = juce::Time::currentTimeMillis();
+            const auto arrival = router_.peers().intern(senderHost, now);
+            peer_ = arrival.id;
```
lucaromagnoli, line 78: **P2 — Preserve the peer identity on the first accepted packet**

`intern()` now deliberately returns `kNoOscPeer` for every unadmitted host, so this assigns zero and every message in the first valid packet is published into the router with no peer. `admit()` only runs after parsing/routing has finished, and its returned ID is ignored. When that queued work drains, both feedback taps call `surfaceFor(0)` and cannot record the value as coming from the new surface; its first gesture is therefore echoed back instead of being suppressed. This also defeats the explicit first-drain synchronization in `surfaceFor()`, whose comment says it exists to retain that first gesture's echo bit. Please establish the admitted ID before publishing accepted messages (for example via a preflight/dispatch split), or keep allocating the practically non-exhaustible 64-bit ID in `intern()` so the first packet carries its eventual identity.

lucaromagnoli: Fixed in c2777485. You are right, and it is a regression I introduced with the previous fix rather than something that was always there: not minting ids in `intern` bought the exhaustion bound and quietly cost the thing the ids are for. `admit` ran after routing and I dropped its result, so the whole of a new surface's first accepted packet went in under `kNoOscPeer` -- and, as you say, that also defeats the `surfaceFor` sync whose only reason to exist is that gesture.

Took the preflight rather than moving allocation back into `intern`, since the second option gives the exhaustion vector back.

`handleMessage` takes a `Dispatch` now. `Preflight` answers the same question and changes nothing: nothing published, no drain scheduled, no learn session consumed, no counter advanced. The receive loop reads a packet from an unadmitted host twice -- decide, admit, apply under the id the host will keep -- and returns early if the first pass finds nothing, so junk costs one pass rather than two. An established peer pays none of it: `intern` reports it as already answerable and the packet goes straight to `Apply`, which is every packet of every gesture after the first.

I threaded the mode through the three branches instead of writing a separate `wouldAccept`, so there is still one definition of what acceptance means. The only place the passes can disagree is a learn session arming between them, which costs one message and leaves the session armed.

Test lives in the service suite where the real receive loop runs: a feedback tap records the peer of the first value to arrive, and it is the id of the peer that packet created rather than `kNoOscPeer`.

magda/daw/audio/osc/OscRouter.cpp
```diff
@@ -281,20 +300,19 @@ std::optional<bool> OscRouter::handleFixedNamespace(const juce::OSCMessage& mess
         return false;
 
     accepted_.fetch_add(1, std::memory_order_relaxed);
```
lucaromagnoli, line 310: **P1 — Make fixed-namespace preflight side-effect-free**

This function accepts `dispatch` but never checks it: after validating the value it always increments `accepted_` and calls `submit()`. The receive loop now parses an unadmitted host's first accepted packet twice, so fixed-namespace commands are applied during both `Preflight` and `Apply`. The new volume test does not expose this because two writes to an absolute-value slot coalesce before the drain; ordered trigger/toggle/delta commands do not, so a surface's first play/mute/relative-seek gesture is queued twice (toggles can cancel themselves and deltas move twice as far). Please return `true` for `Preflight` after payload validation but before the counter/submit side effects, and pin the first-packet case with an ordered command.

lang/en.json
```diff
@@ -480,15 +480,19 @@
     "upload_success": "Imported profile \"{0}\".",
     "osc.enable": "Listen for OSC",
     "osc.bind_address": "Listen on",
+    "osc.bind_all": "All interfaces (0.0.0.0)",
+    "osc.bind_loopback": "This computer only (127.0.0.1)",
```
lucaromagnoli: Use placeholders for IPs here so they dont get translated

lucaromagnoli: Done. Both keys take a `{0}` now and the address is substituted at the call site, the same way `osc.status_listening` already handles its host and port.

```
"osc.bind_all": "All interfaces ({0})",
"osc.bind_loopback": "This computer only ({0})",
```

Issue #2014, opened 2026-08-05
Engine core slice 5: plan differ, compatibility matrix and crossfades


lucaromagnoli:
Slice 5 of #1889.

The component that owns the rebuild-click problem.

- Hash-join old and new plans on `OpKey`; carry runtime state for ops that survive, build what is new, retire what is gone.
- A compatibility matrix says what may be carried into what: same kind and same ports is not always enough, and the cases where it is not are the ones that click.
- Crossfade only the ops that actually changed, over a swap the audio thread does not wait for.
- Delay-line contents are the open question this slice answers. They are the one piece of runtime state that is neither a device nor a source. Proposed rule: match on producer key, consumer key, slot and sample count, and start flushed on anything less than an exact match. Flushed is `latency` samples of silence; a mis-carry is `latency` samples of someone else's audio.
- Note the wrinkle: a delay's key uses the consuming slot index, and a variadic mix renumbers its slots when a send is added or removed. Under the rule above that costs a flush, not wrong audio.

Depends on #2009 landing.

PR #2098, opened 2026-08-16
Differ property tests over random edit sequences (#2077)

### LINKED ISSUE #2077: Validation slice 3: differ property tests over random edit sequences

lucaromagnoli:
Closes #2077. Slice 3 of #1896.

The differ (#2014) and the crossfade of changed edges across a swap (#2019) are pinned today by
pairs of plans I sat down and wrote, which pins them against my reading of the compatibility
matrix. The space of pairs is the space of projects squared, and the failures that matter are the
ones nobody thought to write down. This generates the pairs instead.

## The shape

A vocabulary of the edits a user can perform (devices, racks, chains, sends, sidechains, chain
power, tracks, routing, and a plugin changing the latency it reports), a seeded generator that
strings them together against the project as it stands, and every edit addressed by the id of
what it touches rather than by where that sits. That last part is what makes a failure usable:
any subsequence of a sequence still runs, so a failure is shrunk to the edits that caused it
before it is printed, with the seed beside it.

Nothing here needs a `te::Edit`, so it lives in `magda_tests` next to the other plan tests. The
harness is a session in every respect the differ can see, and deliberately not `EngineSession`:
the properties want the timeline to be a number they control, and they want the counts the
executor keeps about what it adopted. What it copies exactly is the order things happen in, so
every render is also the assertion that carried state outlived the epoch it was carried from.

## The properties

**Carry.** An op carries only when it is the same op computing the same thing from the same
places, which is checkable from the two plans without asking the differ: a signature over the
op's kind, its ports, its arity and the keys feeding each slot. Checked in both directions, so a
differ that carried nothing fails as surely as one that carried too much. Then the half a
signature cannot see: the delay lines and fade ramps the executor says it adopted are predicted
from the two plans and their resolved layouts and required to match.

**Retirement.** `carriedFrom` is a partial injection whose complement is exactly `retired`: every
old op is adopted once or retired once, never both and never neither. On the runtime side, the
store frees an instance only once neither the live plan nor the model names it, and always on the
publishing thread.

**PDC alignment.** Every fan-in has all its inputs arriving at one latency with at least one of
them waiting for nothing, and a track no edit reached reports the latency at its own output that
it did before. An edit in one branch that silently moves another is what that last one catches.

**The null, which catches the most.** Every track carries an analysis device that keeps what
passed through it. The sequence is rendered once whole and once with every edit that cannot reach
a chosen track removed, and the two recordings of that track have to be equal sample for sample.
Not close: equal. The two runs publish a different number of plans, swap at different moments and
carry different amounts of state over those swaps, and none of that may cost the track a sample.
A second variant runs the same assertion with the edits arriving faster than a fade can finish,
so what a swap carries is a running ramp rather than a settled plan.

**On the track an edit does change**, a bound and a destination rather than a null: no step where
the pass reported it faded every edge it moved, and, once the transient is over, exactly what a
session that had just opened the same project renders. Carried state may decide how the signal
got there, never where it arrives. Where an edge was refused a fade, or a delay line anywhere
upstream starts flushed, the step is a declared divergence and the bound does not apply, which
the property reads off the pass's own report rather than guessing.

A level is deliberately not asserted there, and the first draft of this got it wrong. A fade sits
on the edge the edit moved, which is usually inside the chain, and the rest of the chain then
scales a mixture of two internal signals: the result is routinely outside the two steady states
either side of it. A bound that said otherwise would be a bound on a project rather than on the
engine.

## Whether it can fail

A property test that cannot fail looks exactly like one that passes, so the engine was broken
four ways before any of this was believed:

| Mutation | Caught by | Shrunk to |
| --- | --- | --- |
| an op whose inputs moved carries anyway | carry, and audibly by the transient bound | 1 edit |
| a delay line never carries | carry, and audibly by the null | 1 edit |
| every compensation delay one sample too long | alignment | 0 edits |
| a fade takes a changed op as its old side | the transient bound | 5 edits |

Three harness tests carry the same weight: the generator has to cover every edit kind with most
of them landing, the shrinker has to reduce a planted failure to the one edit that plants it, and
dropping the edits a track *can* hear has to change what it renders, without which a null on the
untouched tracks would also pass on silence.

## What it found

No engine bug. The deep sweep is four thousand sequences of forty edits: 137k publishes, 7M carry
decisions, 458k retirements, and 47M samples nulled, all clean. What it did find, twice, was the
harness calling an edit unrelated when it was not, both fixed here: an `addTrack` addressed by
index moved where later tracks landed, and relevance had to be closed over the far end of a kept
edit as well as the near one.

`kRecordedSeeds` is therefore empty. The mechanism is in place and runs on every build, for the
first seed that fails.

## Cost

The ordinary run is a few hundred sequences of fourteen edits, about eight seconds inside a
forty-five second suite. The deep sweep is hidden behind a tag and takes eight minutes:

```
./cmake-build-debug/tests/magda_tests "[deep]"
MAGDA_DIFFER_SEQUENCE=60 MAGDA_DIFFER_PLAN_SEEDS=20000 ./cmake-build-debug/tests/magda_tests "[deep]"
```

It is what to reach for when the differ or the crossfade pass changes. Both sweeps print their
counts on every run rather than only on failure, because a property that has quietly stopped
exercising anything reads exactly like one that passes.

## Testing

- `./cmake-build-debug/tests/magda_tests` green: 2617 passed, 13 skipped, 577652 assertions.
- `./cmake-build-debug/tests/magda_tests "[deep]"` green.
- The four mutations above each verified to fail, and the engine reverted byte for byte.

### DIFF
--- a/tests/PlanEditSequence.cpp
+++ b/tests/PlanEditSequence.cpp
@@ -1,0 +1,1148 @@
+#include "PlanEditSequence.hpp"
+
+#include <algorithm>
+#include <sstream>
+#include <utility>
+
+#include "core/RackInfo.hpp"
+#include "plan/TrackRouting.hpp"
+
+/**
+ * @file PlanEditSequence.cpp
+ * @brief The edit vocabulary, and the generator that strings it together (#2077).
+ *
+ * Every edit here is one a user can perform, which is the only reason to have a
+ * vocabulary at all: a generator free to write arbitrary plans would spend its
+ * time on shapes the compiler never emits, and the differ's decisions about
+ * those are nobody's problem.
+ */
+
+namespace magda::edits {
+namespace {
+
+// --- reading and writing the project ----------------------------------------
+
+/// The elements a rack chain holds, wherever the rack is nested.
+std::vector<ChainElement>* chainElements(std::vector<ChainElement>& elements, ChainId chainId) {
+    for (auto& element : elements) {
+        if (!isRack(element))
+            continue;
+
+        auto& rack = getRack(element);
+        for (auto& chain : rack.chains) {
+            if (chain.id == chainId)
+                return &chain.elements;
+            if (auto* found = chainElements(chain.elements, chainId))
+                return found;
+        }
+    }
+    return nullptr;
+}
+
+/// The rack with this id, wherever it is nested, and the list holding it.
+struct RackAt {
+    std::vector<ChainElement>* container = nullptr;
+    std::size_t index = 0;
+    TrackId trackId = INVALID_TRACK_ID;
+};
+
+bool findRackIn(std::vector<ChainElement>& elements, RackId rackId, TrackId trackId, RackAt& out) {
+    for (std::size_t i = 0; i < elements.size(); ++i) {
+        if (!isRack(elements[i]))
+            continue;
+
+        auto& rack = getRack(elements[i]);
+        if (rack.id == rackId) {
+            out = RackAt{&elements, i, trackId};
+            return true;
+        }
+        for (auto& chain : rack.chains)
+            if (findRackIn(chain.elements, rackId, trackId, out))
+                return true;
+    }
+    return false;
+}
+
+bool findRack(Project& project, RackId rackId, RackAt& out) {
+    for (auto& track : project.tracks)
+        if (findRackIn(track.chain.fxChainElements, rackId, track.id, out))
+            return true;
+    return false;
+}
+
+/// The chain with this id, and the rack that owns it.
+struct ChainAt {
+    RackInfo* rack = nullptr;
+    std::size_t index = 0;
+    TrackId trackId = INVALID_TRACK_ID;
+};
+
+bool findChainIn(std::vector<ChainElement>& elements, ChainId chainId, TrackId trackId,
+                 ChainAt& out) {
+    for (auto& element : elements) {
+        if (!isRack(element))
+            continue;
+
+        auto& rack = getRack(element);
+        for (std::size_t i = 0; i < rack.chains.size(); ++i) {
+            if (rack.chains[i].id == chainId) {
+                out = ChainAt{&rack, i, trackId};
+                return true;
+            }
+            if (findChainIn(rack.chains[i].elements, chainId, trackId, out))
+                return true;
+        }
+    }
+    return false;
+}
+
+bool findChain(Project& project, ChainId chainId, ChainAt& out) {
+    for (auto& track : project.tracks)
+        if (findChainIn(track.chain.fxChainElements, chainId, track.id, out))
+            return true;
+    return false;
+}
+
+/// A device of the insert chain, wherever it sits, and the list holding it.
+struct DeviceAt {
+    std::vector<ChainElement>* container = nullptr;
+    std::size_t index = 0;
+    TrackId trackId = INVALID_TRACK_ID;
+};
+
+bool findDeviceIn(std::vector<ChainElement>& elements, DeviceId deviceId, TrackId trackId,
+                  DeviceAt& out) {
+    for (std::size_t i = 0; i < elements.size(); ++i) {
+        if (isDevice(elements[i])) {
+            if (getDevice(elements[i]).id == deviceId) {
+                out = DeviceAt{&elements, i, trackId};
+                return true;
+            }
+            continue;
+        }
+
+        for (auto& chain : getRack(elements[i]).chains)
+            if (findDeviceIn(chain.elements, deviceId, trackId, out))
+                return true;
+    }
+    return false;
+}
+
+bool findDevice(Project& project, DeviceId deviceId, DeviceAt& out) {
+    for (auto& track : project.tracks)
+        if (findDeviceIn(track.chain.fxChainElements, deviceId, track.id, out))
+            return true;
+    return false;
+}
+
+/// Where an edit adds something, or null when the project no longer has it.
+std::vector<ChainElement>* siteElements(Project& project, const Site& site) {
+    for (auto& track : project.tracks) {
+        if (track.id != site.trackId)
+            continue;
+        if (site.chainId == INVALID_CHAIN_ID)
+            return &track.chain.fxChainElements;
+        return chainElements(track.chain.fxChainElements, site.chainId);
+    }
+    return nullptr;
+}
+
+TrackInfo* findTrack(Project& project, TrackId trackId) {
+    for (auto& track : project.tracks)
+        if (track.id == trackId)
+            return &track;
+    return nullptr;
+}
+
+int trackIndex(const Project& project, TrackId trackId) {
+    for (std::size_t i = 0; i < project.tracks.size(); ++i)
+        if (project.tracks[i].id == trackId)
+            return static_cast<int>(i);
+    return -1;
+}
+
+std::size_t clampPosition(int position, std::size_t size) {
+    if (position <= 0)
+        return 0;
+    return std::min(static_cast<std::size_t>(position), size);
+}
+
+// --- what a track holds ------------------------------------------------------
+
+/// Every insert-chain device of a track, nested racks included.
+void collectDevices(const std::vector<ChainElement>& elements,
+                    std::vector<const DeviceInfo*>& out) {
+    for (const auto& element : elements) {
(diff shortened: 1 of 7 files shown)


tests/PlanEditSequence.cpp
```diff
@@ -0,0 +1,1148 @@
...
+constexpr int kMaxRacks = 4;
+constexpr int kMaxSendsPerTrack = 3;
+
+/// Latencies a generated device reports. Zero most of the time, because a plan
+/// with a delay on every edge is not the common shape, and never so large that
+/// a render property would need a longer settling window than the block budget.
+int pickLatency(Rng& rng) {
+    if (rng.chance(60))
+        return 0;
+    return 1 + rng.below(96);
```
lucaromagnoli: This hands out latencies up to 96, but kMaxDeviceLatency is 32 and TestDevice's ring is sized kMaxDeviceLatency + 1, with process() clamping the realized delay to capacity minus one. So a device that draws 33..96, which is about two thirds of the latent draws, reports the full number while delaying 32: the harness's own premise, latency in the plan is latency in the signal so a compensation error is audible rather than merely stated, silently fails for those devices. No property false-fails today because both legs of every comparison and the fresh session all share the same mismatch, but the audible half of the PDC coverage is quietly measuring a signal that is up to 64 samples early relative to what compensation assumed. One line either way: bound this by kMaxDeviceLatency, or raise kMaxDeviceLatency to 96 so the ring honours the whole range, and in either case derive this number from the constant so the two cannot drift again, which is the same class of harness bug the PR body says this work already caught twice.

lucaromagnoli (commented):
Read the whole thing. The design decisions that make property tests worth having are all present and correctly executed: id-addressed edits so any subsequence runs and deletion is the whole shrinking strategy, a generator that refuses to build what the compiler cannot express with the runner treating diagnostics as generator drift, signature-based carry checked in both directions so carry-nothing fails as surely as carry-too-much, adoption counts predicted from the plans rather than trusted, the retirement complement checked exactly, relevance closed over the far end of kept edits, and the three harness self-tests that keep the whole sweep falsifiable. The splitmix64-not-std-random call for cross-machine seed stability, the counts printed on every run, and the settle-window being asserted rather than assumed are each the right paranoia. checkAlignment asserting minimality (no fan-in where every path waits) goes further than the plan tests elsewhere do. One confirmed harness bug inline.

tests/PlanEditSequence.cpp
lucaromagnoli: **P1 — Make the simulated delay match the latency it reports**

This generates latencies through 96 samples, but `TestDevice::prepare()` allocates only `kMaxDeviceLatency + 1` entries with `kMaxDeviceLatency == 32`, and `process()` silently clamps `latency_` to that capacity. The executor/layout therefore compensate for the reported 33–96 samples while the device audio is delayed by only 32. That breaks the harness's core claim that declared latency is audible: the alignment property still passes on metadata, while the render properties exercise a physically misaligned graph and `pathStartsFlushed` can exempt transients based on latency the signal never incurred. It also defeats the 20-block settling bound, which only works with the documented 32-sample per-device maximum. Please derive generation and ring capacity from the same bound (or raise both and the settling window), and pin an impulse at the maximum generated latency.

tests/test_plan_edit_properties.cpp
```diff
@@ -0,0 +1,1202 @@
...
+        return std::nullopt;
+
+    const auto model = edits::deviceKeys(project);
+
+    std::set<engine::DeviceKey> playing;
+    for (const auto& op : published.faded.plan.ops)
+        if (op.kind == engine::OpKind::Device)
+            playing.insert(op.key.deviceKey());
+
+    for (const auto& key : destroyed) {
```
lucaromagnoli: **P2 — Assert retirement completeness, not only destruction safety**

This loop validates only keys that `releaseDeleted()` happened to destroy. A key that is no longer named by either the model or live plan but remains leaked in the store never appears in `destroyed`, so it is silently ignored; the suite-level `instancesDestroyed > 0` only proves that some instance was freed somewhere. For example, retaining every FX device while still freeing removed capture devices would satisfy this property. That leaves the runtime half of the stated retirement invariant untested. Please track the realized device-key set (or expose the store's owned keys) and require this publish's destroyed set to equal the previously owned keys minus the current model/live keep-set, while retaining the existing safety checks.

tests/test_plan_edit_properties.cpp
```diff
@@ -0,0 +1,1202 @@
...
+            return "track " + std::to_string(trackId) + " rendered something that is not a number";
+        if (std::abs(value) > kRunawayLevel)
+            return "track " + std::to_string(trackId) + " ran away at sample " +
+                   std::to_string(sample) + ": " + std::to_string(value);
+    }
+
+    // A path that starts from silence steps by whatever it was carrying, and
+    // an edge the pass refused steps by whatever it had left: both are declared
+    // divergences rather than failures, and both are counted where they happen.
+    if (!faded || flushed)
```
lucaromagnoli: **P2 — Prove that the transient bound is exercised**

The actual step-size assertion is skipped whenever the pass reports any unfaded edge or this track has a flushed upstream path. That is a legitimate exemption per step, but no coverage counter records that execution ever gets past this return: `fadesInserted > 0` can be satisfied by fades on exempt steps/tracks, and `steadyStatesCompared > 0` exercises only convergence. A generator drift—or a regression that starts reporting one unfaded edge on every pass—can therefore disable the whole discontinuity bound while the test remains green. Please count windows that actually reach the sample-delta loop and assert that count is nonzero (ideally alongside the existing exemption counts).

tests/test_plan_edit_properties.cpp
```diff
@@ -0,0 +1,1311 @@
...
+    // an edge the pass refused steps by whatever it had left: both are declared
+    // divergences rather than failures, and both are counted where they happen.
+    if (!faded || flushed) {
+        if (report != nullptr)
+            ++report->transientsExempt;
+        return std::nullopt;
+    }
+
+    if (report != nullptr)
+        ++report->transientsBounded;
```
lucaromagnoli: **P2 — Count bounded transients only on tracks the edit can affect**

`transientProperty` calls this for all three base tracks after every applied edit. An edit that changes track 1 can therefore exempt track 1 (because its path starts flushed) while unchanged tracks 2 and 3 increment `transientsBounded`; `CHECK(report.transientsBounded > 0)` still passes even if every track that actually changes is always exempt. In that case the discontinuity bound remains vacuous for real transients—the reported 360/867 are bounded track-windows, not necessarily transients. Please gate the coverage count/assertion on the edit reaching this track (using the same touched/feed relation already used by `planProperties`), or separately require a nonzero-range bounded window on an affected track.

PR #2099, opened 2026-08-16
Pre/post-fader toggle on the post-FX panel (#2094)

### LINKED ISSUE #2094: Post-FX pre/post-fader toggle has no UI

lucaromagnoli:
Closes #2094.

`TrackChain::postFxPostFader` landed with #2087 and both engines honour it, but nothing in the UI could set it. This adds the missing control.

A small clickable state tag on the post-FX add strip, in the mirror position below the "+" from the "POSTFX" watermark. It reads "POST FADER" or "PRE FADER" straight off `TrackManager::isPostFxPostFader` at paint time (no cached bool on the component) and flips the whole section across the fader on click. The setter already fires `notifyTrackDevicesChanged`, which lands in the panel's existing `trackDevicesChanged` override, so there is no extra refresh plumbing: the repaint there covers the tag, which is a child of the viewed container.

Per-device placement stays out of scope, as agreed on the issue.

Details:

- Hidden on the master track. The native compiler pins the master post-FX stage pre-fader whatever the flag says (`postFxPostFader && !isMaster`), so a toggle there would lie. Explicit `MASTER_TRACK_ID` check, since `getTrack` returns a valid track for it.
- Hidden on `INVALID_TRACK_ID`, mirroring how `addButton_` visibility is handled.
- Dropped rather than crowded when the strip is short: the tag centres in the gap between the bottom of the "+" and the strip's bottom inset, and only shows when that gap is at least 26px (so from roughly 92px of panel content height up). It never overlaps the "+".
- Secondary text colour at full alpha so it reads as state rather than decoration, accent colour on hover, pointing-hand cursor, and a state-dependent tooltip via `TooltipClient::getTooltip`.

UI wiring only; the model behaviour is covered by the #2087 tests. To see it: open the POST-FX side panel on the right of the device area in the bottom panel, the tag sits under the "+" on the add strip.

## Preference for the default

The tag changes one track at a time, which is the wrong grain if you want every track built the other way round. **New tracks run post-FX after the fader** now sits in the Behaviour section of Preferences, next to the other new-clip and new-track defaults, and `createTrack` seeds `TrackChain::postFxPostFader` from it. Existing tracks keep the side they were set to; master stays pre-fader regardless.

While adding the row I found the Behaviour block's preferred-height sums were already two rows behind the toggles laid out there, so the dialog opened shorter than its own content. They now count the rows that are actually there.

Tests: a model-level Catch2 test covers the seeding both ways, and the engine-order JUCE tests pin the preference for their duration, since they create tracks and would otherwise depend on how the machine running them has it set. Manual updated (`fx-chain.md` also still claimed post-FX ran before the fader, which stopped being true with #2087).

### DIFF
--- a/magda/daw/ui/panels/content/PostFxPanelContent.cpp
+++ b/magda/daw/ui/panels/content/PostFxPanelContent.cpp
@@ -130,6 +130,67 @@ class PostFxPanelContent::Container : public juce::Component, public juce::DragA
 };
 
 //==============================================================================
+// FaderSideTag - the clickable state tag sitting below the "+" on the add
+// strip, mirroring the "POSTFX" watermark above it. Reads and toggles
+// TrackChain::postFxPostFader, which moves the whole post-FX section (and the
+// sends with it) across the track fader. Per-device placement is out of scope.
+//==============================================================================
+class PostFxPanelContent::FaderSideTag : public juce::Component, public juce::TooltipClient {
+  public:
+    explicit FaderSideTag(PostFxPanelContent& owner) : owner_(owner) {
+        setMouseCursor(juce::MouseCursor::PointingHandCursor);
+    }
+
+    void paint(juce::Graphics& g) override {
+        // State, not decoration: full alpha, and accented while hovered.
+        g.setColour(hovered_ ? DarkTheme::getColour(DarkTheme::ACCENT_PRIMARY)
+                             : DarkTheme::getSecondaryTextColour());
+        g.setFont(FontManager::getInstance().getUIFont(9.5f));
+        const int lineH = getHeight() / 2;
+        g.drawText(isPostFader() ? "POST" : "PRE", juce::Rectangle<int>(0, 0, getWidth(), lineH),
+                   juce::Justification::centred);
+        g.drawText("FADER", juce::Rectangle<int>(0, lineH, getWidth(), getHeight() - lineH),
+                   juce::Justification::centred);
+    }
+
+    void mouseEnter(const juce::MouseEvent&) override {
+        hovered_ = true;
+        repaint();
+    }
+
+    void mouseExit(const juce::MouseEvent&) override {
+        hovered_ = false;
+        repaint();
+    }
+
+    void mouseDown(const juce::MouseEvent&) override {
+        if (owner_.trackId_ == magda::INVALID_TRACK_ID)
+            return;
+        // The setter notifies trackDevicesChanged, which relays out and repaints.
+        auto& trackManager = magda::TrackManager::getInstance();
+        trackManager.setPostFxPostFader(owner_.trackId_,
+                                        !trackManager.isPostFxPostFader(owner_.trackId_));
+    }
+
+    juce::String getTooltip() override {
+        return isPostFader() ? "Post-FX and sends run after the track fader. Click to move them "
+                               "before the fader."
+                             : "Post-FX and sends run before the track fader. Click to move them "
+                               "after the fader.";
+    }
+
+  private:
+    // Always read through the model - the tag caches no state of its own.
+    bool isPostFader() const {
+        return owner_.trackId_ != magda::INVALID_TRACK_ID &&
+               magda::TrackManager::getInstance().isPostFxPostFader(owner_.trackId_);
+    }
+
+    PostFxPanelContent& owner_;
+    bool hovered_ = false;
+};
+
+//==============================================================================
 // PostFxPanelContent
 //==============================================================================
 PostFxPanelContent::PostFxPanelContent() {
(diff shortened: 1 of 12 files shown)


magda/daw/ui/panels/content/PostFxPanelContent.cpp
```diff
@@ -129,6 +129,67 @@ class PostFxPanelContent::Container : public juce::Component, public juce::DragA
...
+        if (owner_.trackId_ == magda::INVALID_TRACK_ID)
+            return;
+        // The setter notifies trackDevicesChanged, which relays out and repaints.
+        auto& trackManager = magda::TrackManager::getInstance();
+        trackManager.setPostFxPostFader(owner_.trackId_,
+                                        !trackManager.isPostFxPostFader(owner_.trackId_));
+    }
+
+    juce::String getTooltip() override {
+        return isPostFader() ? "Post-FX and sends run after the track fader. Click to move them "
```
lucaromagnoli: The tooltip claims the sends move with the toggle, but they do not. In PluginManagerSync the sends are placed by their own per-send preFader flag (appendSends(true) lands before the fader, appendSends(false) after it, regardless of this setting); postFxPostFader only decides which side of the fader the post-FX stage and the analysis rail compile to. Flipping this tag will not move a post-fader send pre-fader.

The class comment up top makes the same claim ("and the sends with it").

Suggest dropping the sends from both, e.g. "Post-FX runs after the track fader. Click to move it before the fader."

magda/daw/ui/panels/content/PostFxPanelContent.cpp
```diff
@@ -129,6 +129,67 @@ class PostFxPanelContent::Container : public juce::Component, public juce::DragA
...
+        hovered_ = true;
+        repaint();
+    }
+
+    void mouseExit(const juce::MouseEvent&) override {
+        hovered_ = false;
+        repaint();
+    }
+
+    void mouseDown(const juce::MouseEvent&) override {
```
lucaromagnoli: Nit: mouseDown fires for every button, so a right-click toggles the routing too. A right-click on a small control usually means "show me options", not "do the thing", so I would bail on e.mods.isPopupMenu() to keep an instinctive right-click from silently moving the section across the fader.

lucaromagnoli (commented):
Implementation matches what was agreed on the issue: paint-time state read, no cached bool, refresh riding the existing trackDevicesChanged path, master hidden because the native compiler pins it pre-fader, and the tag dropping out cleanly when the strip is short. Two comments inline, one on the user-facing copy and one nit.

magda/daw/ui/panels/content/PostFxPanelContent.cpp
```diff
@@ -129,6 +129,71 @@ class PostFxPanelContent::Container : public juce::Component, public juce::DragA
...
+
+    void mouseExit(const juce::MouseEvent&) override {
+        hovered_ = false;
+        repaint();
+    }
+
+    void mouseDown(const juce::MouseEvent& e) override {
+        // A right-click on a small control means "show me options", not "do the
+        // thing"; there is no menu here yet, so it should just do nothing.
+        if (owner_.trackId_ == magda::INVALID_TRACK_ID || e.mods.isPopupMenu())
```
lucaromagnoli: `mouseDown` fires for every mouse button. `isPopupMenu()` excludes right-click/Ctrl-click, but middle/back/forward buttons still fall through and change live routing; acting immediately on mouse-down also removes the usual ability to cancel a click by dragging away. Please use a `juce::Button`, or restrict this to the primary button and commit on an in-bounds `mouseUp`.

PR #2102, opened 2026-08-16
Keep EQ curve readouts inside the plot, and the Enabled checkbox in sync (#2072)

### LINKED ISSUE #2072: EQ band readout is drawn outside the curve view at extreme frequencies
### What happened?

The live readout above the active EQ band is drawn past the edge of the curve view once the band is near either end of the range: below roughly 35 Hz, or above roughly 10 kHz. The text spills over the border instead of staying inside the plot.

### Steps to reproduce

1. Open an EQ device and show its curve.
2. Drag a band's frequency down below about 35 Hz, or up past about 10 kHz.
3. Watch the readout above the dot.

It runs outside the view's border.

### Expected behavior

The readout stays inside the plot area, sliding along the top edge as the dot approaches either end rather than following the dot past it.

The readout is placed in `CompiledEqCurveView.cpp`, in the active-band branch of the band drawing:

```cpp
g.drawFittedText(label,
                 juce::Rectangle<int>(static_cast<int>(dotX) - 60,
                                      static_cast<int>(plotArea_.getY()) + 2, 120, 14),
                 juce::Justification::centred, 1);
```

It is a fixed 120-wide box centred on the dot with nothing clamping it to `plotArea_`, so once `dotX` comes within 60 px of either edge the box hangs over the side. Constraining that rectangle to `plotArea_` before drawing should be the whole fix, and the same placement is worth checking in `CompiledMultibandCurveView.cpp`, which draws its readout the same way.

### MAGDA version

Current dev.

### Operating system

All.

lucaromagnoli:
Closes #2072.

### The readout ran outside the plot

The EQ's live readout was a fixed 120-wide box centred on the band dot with nothing clamping it, so once `dotX` came within 60 px of either edge the box hung over the border: below roughly 35 Hz or above 10 kHz the text spilled outside the view.

New `magda/daw/ui/utils/CurveLabelLayout.hpp` holds the one helper every curve view now positions its readouts with. `centredIn` centres a box on the tracked point but clamps it into the plot, so near an edge the box slides along the border instead of following the dot out of view. The clamp applies vertically too, and a box larger than the plot shrinks to fit rather than overflowing both sides.

Applied to:

- the EQ readout, and the per-band number, which rides 16 px above its dot and so was drawn above the view entirely at max gain
- `CompiledMultibandCurveView`, which the issue flagged as drawing its readout the same way: the crossover readout is clamped to the plot, the range readout to its own band

### The Enabled checkbox did not follow the curve

Separate defect found while documenting the double-click gesture. Double-clicking a band on the curve writes `Band N Enabled`, but the grid cell for it kept showing the old tick.

`ParamSlotComponent`'s toggle state was only ever set in `configureBoolToggle`, which runs when the widget is configured. Both value-only refresh paths — `ParamHostComponent::updateParameterValues` and the single-param echo in `DeviceParameterChangeHandler` — land in `setParamValue`, which re-pointed the slider and the discrete widgets but not the toggle. The comment already in that function describes the identical bug for discrete widgets; Boolean had been missed.

`setParamValue` now calls a new `syncBooleanToggle`, so any writer moves the checkbox: a device's own curve view, automation, a macro, a control surface.

### Also in here

- `CompiledEqCurveView::resampleFromDevice` fell back to `false` for `enabled` and `Bell` for `type` while `freq`/`gain`/`q` fell back to the cached value, so a `DeviceInfo` snapshot omitting those slots read every band as a disabled Bell. All five now fall back to what is cached.
- The EQ's catalog description mentions the double-click gesture. Worded as a toggle rather than an enable, since `mouseDoubleClick` flips the state — double-clicking an active band switches it off again.

### Tests

- `tests/test_curve_label_layout.cpp` (Catch2): centre tracking, both edges, the handover from tracking to sliding, narrow plots, the vertical clamp.
- `tests/test_param_slot_boolean_sync_juce.cpp` (JUCE target, where `ParamSlotComponent` links): initial tick, and the toggle following a value-only refresh in both directions. Verified it fails without the fix (`Enabling the band elsewhere must tick the grid checkbox`).

Both suites green locally: 2637 Catch2 cases, all JUCE suites.

### DIFF
--- a/magda/daw/ui/components/chain/compiled/CompiledEqCurveView.cpp
+++ b/magda/daw/ui/components/chain/compiled/CompiledEqCurveView.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cmath>
 
+#include "../../../utils/CurveLabelLayout.hpp"
 #include "audio/plugins/compiled/MagdaEqCompiledPlugin.hpp"
 #include "core/GestureRouter.hpp"
 #include "ui/themes/DarkTheme.hpp"
@@ -412,7 +413,7 @@ void CompiledEqCurveView::paint(juce::Graphics& g) {
         g.setColour(DarkTheme::getColour(DarkTheme::TEXT_PRIMARY).withAlpha(0.85f));
         g.drawFittedText(
             juce::String(b + 1),
-            juce::Rectangle<int>(static_cast<int>(dotX) - 5, static_cast<int>(dotY) - 16, 10, 12),
+            CurveLabelLayout::centredIn(plotArea_, dotX, dotY - 16.0f, 10.0f, 12.0f).toNearestInt(),
             juce::Justification::centred, 1);
 
         // Live readout above the dot for the active band — freq + gain (or
@@ -424,10 +425,11 @@ void CompiledEqCurveView::paint(juce::Graphics& g) {
                     : (juce::String(bandTypeShortName(band.type)) + "  " +
                        frequencyReadout(band.freq));
             g.setColour(DarkTheme::getColour(DarkTheme::TEXT_PRIMARY).withAlpha(0.9f));
-            g.drawFittedText(label,
-                             juce::Rectangle<int>(static_cast<int>(dotX) - 60,
-                                                  static_cast<int>(plotArea_.getY()) + 2, 120, 14),
-                             juce::Justification::centred, 1);
+            g.drawFittedText(
+                label,
+                CurveLabelLayout::centredIn(plotArea_, dotX, plotArea_.getY() + 2.0f, 120.0f, 14.0f)
+                    .toNearestInt(),
+                juce::Justification::centred, 1);
         }
     }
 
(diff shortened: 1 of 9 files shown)


magda/daw/ui/components/chain/compiled/CompiledEqCurveView.cpp
```diff
@@ -412,7 +417,7 @@ void CompiledEqCurveView::paint(juce::Graphics& g) {
         g.setColour(DarkTheme::getColour(DarkTheme::TEXT_PRIMARY).withAlpha(0.85f));
         g.drawFittedText(
             juce::String(b + 1),
-            juce::Rectangle<int>(static_cast<int>(dotX) - 5, static_cast<int>(dotY) - 16, 10, 12),
+            CurveLabelLayout::centredIn(plotArea_, dotX, dotY - 16.0f, 10.0f, 12.0f).toNearestInt(),
```
lucaromagnoli: **P2 — Keep the band number out of the active readout row.** At the view’s 90 px preferred height, a high-gain band’s number is clamped to the plot’s top edge here. When that band is hovered or dragged, the live readout is drawn at `plotY + 2` with height 14 and the same horizontal centre, so the two labels overpaint each other. Please hide the number while the band is active, or move it below the dot when the upper placement is clamped.

PR #2106, opened 2026-08-17
Keep test runs out of the developer's settings file (#2105)

### LINKED ISSUE #2105: Running the test suite overwrites the developer's real config.json with defaults
### What happens

Running the test suite replaces `~/Library/MAGDA/config.json` with defaults. Every preference goes: AI credentials, recent projects, preferred audio device, browser favourites and default directory, keyboard bindings, panel sizes. It is silent, and the file it overwrites is the developer's live settings.

This has actually happened, twice in one morning on my machine. The API keys were not recoverable.

### Why

`Config::save()` serialises the whole singleton, and **no test ever calls `Config::getInstance().load()`**. So in a test process the singleton is default-constructed, and any save writes those defaults over the real file.

Reaching a save takes nothing deliberate. `RemoteApiHost` persists client grants by calling `Config::save()`:

```
magda/daw/api/remote_api_host.cpp:273
    config.setRemoteApiClients(latest);
    config.save();
```

so constructing a `RemoteApiHost` in a test and setting a scope is enough. The evidence is unambiguous: the wrecked config's `remoteApi.clients` were named `cursor`, `someone-else` and `unknown` -- the fixture names from `tests/test_remote_api_host.cpp` -- with `firstSeenMs` timestamps from the test runs.

`ScopedRemoteApiConfig` in that file already anticipates the hazard:

```
// Grants too (#1860): a host persists them on every change, so a test
// that connects a client would otherwise leave its grant behind in the
// developer's real config file.
```

It restores the in-memory values on teardown, but it cannot undo the write to disk, and it does not stop the rest of the singleton being written as defaults.

### Fix

`paths::configFile()` honours a `MAGDA_CONFIG_FILE` override, and both test binaries link a translation unit that points it at a scratch file during static initialisation, before any test can run. `test_config_path_redirect` asserts the redirect took, so it cannot rot silently. `EnvScope` in `test_app_paths.cpp` snapshots and restores `MAGDA_CONFIG_FILE` as well, since a test that clears it would otherwise re-expose the real file to everything running after it.

Verified by checksumming the real `config.json` across full runs of both suites.

### Note

This is what #2104 was really about. That issue described a corrupted settings file being overwritten after a failed load; no corruption was ever involved, and the drive I originally blamed was a coincidence. #2104 still describes a genuine (if unobserved) weakness in `Config::load()`/`save()`, but it is not the cause of any reported data loss.

lucaromagnoli:
Closes #2105.

Running the test suite replaced `~/Library/MAGDA/config.json` with defaults -- AI credentials, recent projects, audio device, browser favourites, keyboard bindings, all of it. Silently, on the developer's live settings. It happened twice on my machine in one morning; the API keys were not recoverable.

`Config::save()` serialises the whole singleton, and no test calls `Config::getInstance().load()` first, so in a test process the singleton is default-constructed and any save writes those defaults over the real file. Reaching a save takes nothing deliberate: `RemoteApiHost` persists client grants through `Config::save()` (`remote_api_host.cpp:273`), so constructing one in a test is enough. The wrecked config's `remoteApi.clients` were named `cursor`, `someone-else` and `unknown` -- the fixture names from `tests/test_remote_api_host.cpp`.

`ScopedRemoteApiConfig` already anticipated the hazard in a comment, but restoring the in-memory values on teardown cannot undo a write to disk.

### The fix

`paths::configFile()` honours a `MAGDA_CONFIG_FILE` override -- the one way to move the settings file, which is otherwise deliberately anchored to the OS default because Config has to be read before the data-dir override is known.

Both test binaries link `TestConfigRedirect.cpp`, which points that variable at a scratch file during static initialisation, before any test can run and regardless of how the binary is invoked.

Two guards so this cannot rot:

- `test_config_path_redirect` asserts the redirect actually took, rather than trusting that it did.
- `EnvScope` now snapshots and restores `MAGDA_CONFIG_FILE` alongside the other knobs. Without that, the one test that legitimately clears it -- the data-dir anchoring case -- would leave every later test in the process pointed back at the real file.

`test_app_paths`'s anchoring case keeps testing what it was about (the data dir must not move config.json) and gains a sibling covering the new override.

### Verification

Checksummed the real `config.json` across full runs of both suites: unchanged. Before the fix, `magda_tests` rewrote it every run.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -454,8 +454,10 @@ target_sources(magda_juce_tests PRIVATE
     test_clip_nudge_juce.cpp
     test_multiband_curve_view_juce.cpp
     # Keep test runs out of the developer's real config.json (see the copy in
-    # magda_tests): RemoteApiHost persists grants via Config::save().
+    # magda_tests): RemoteApiHost persists grants via Config::save(). The guard
+    # test is duplicated per target because each uses its own framework.
     TestConfigRedirect.cpp
+    test_config_path_redirect_juce.cpp
     # Boolean param cells follow value-only refreshes (#2072 follow-up).
     # ParamSlotComponent builds into magda_daw_app, so the test target needs it
     # and the param helpers it calls into compiled in directly.
@@ -600,6 +602,9 @@ add_test(NAME arrangement_transport_loop_juce
 
 add_test(NAME multiband_curve_view_juce
          COMMAND magda_juce_tests "Multiband Curve View Tests")
+
+add_test(NAME config_path_redirect_juce
+         COMMAND magda_juce_tests "Config Path Redirect Tests")
 
 add_test(NAME param_slot_boolean_sync_juce
          COMMAND magda_juce_tests "Param Slot Boolean Sync Tests")
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -120,6 +120,10 @@ set(TEST_SOURCES
     test_arrangement_hit_tester.cpp
     # Curve-view readout stays inside the plot (#2072)
     test_curve_label_layout.cpp
+    # Keep test runs out of the developer's real config.json. The redirect must
+    # be linked into every test binary, and the guard test proves it took.
+    TestConfigRedirect.cpp
+    test_config_path_redirect.cpp
     # Keyboard clip nudging (#1957)
     test_clip_nudge.cpp
     test_mono_note_processor.cpp
@@ -449,6 +453,9 @@ target_sources(magda_juce_tests PRIVATE
     test_clip_paint_juce.cpp
     test_clip_nudge_juce.cpp
     test_multiband_curve_view_juce.cpp
+    # Keep test runs out of the developer's real config.json (see the copy in
+    # magda_tests): RemoteApiHost persists grants via Config::save().
+    TestConfigRedirect.cpp
     # Boolean param cells follow value-only refreshes (#2072 follow-up).
     # ParamSlotComponent builds into magda_daw_app, so the test target needs it
     # and the param helpers it calls into compiled in directly.
--- a/tests/TestConfigRedirect.cpp
+++ b/tests/TestConfigRedirect.cpp
@@ -1,0 +1,36 @@
+// Keep the test binaries away from the developer's real settings file.
+//
+// Config::save() persists the WHOLE singleton, and no test loads it, so a
+// single save from a test process replaces every preference in
+// ~/Library/MAGDA/config.json with defaults -- API keys included. That is not
+// hypothetical: RemoteApiHost persists client grants by calling Config::save(),
+// so simply constructing one in a test wrote defaults over the real file.
+//
+// This runs before main(), so the redirect is in place no matter which test
+// executes first or how the binary is invoked.
+
+#include <juce_core/juce_core.h>
+
+#include <cstdlib>
+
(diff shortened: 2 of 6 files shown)


tests/CMakeLists.txt
```diff
@@ -449,6 +453,9 @@ target_sources(magda_juce_tests PRIVATE
     test_clip_paint_juce.cpp
     test_clip_nudge_juce.cpp
     test_multiband_curve_view_juce.cpp
+    # Keep test runs out of the developer's real config.json (see the copy in
+    # magda_tests): RemoteApiHost persists grants via Config::save().
+    TestConfigRedirect.cpp
```
lucaromagnoli, line 459: The "cannot rot" guard only exists in magda_tests: test_config_path_redirect.cpp is Catch2, so magda_juce_tests gets the redirect but nothing that proves it took. If this line is ever dropped from the juce target, that binary silently goes back to writing the real config.json, and it is the target that hosts the shared-engine tests where a stray Config::save() is most likely. A tiny JUCE UnitTest sibling (same two assertions as the Catch2 guard) would close it.

tests/TestConfigRedirect.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <cstdlib>
+
+namespace {
+
+struct RedirectConfigFile {
+    RedirectConfigFile() {
+        auto dir = juce::File::getSpecialLocation(juce::File::tempDirectory)
+                       .getChildFile("magda-test-config");
+        dir.createDirectory();
+        const auto path = dir.getChildFile("config.json").getFullPathName();
```
lucaromagnoli: Nit: both test binaries point at the same scratch file, and it survives across runs. Nothing loads config today, so the worst case is two ctest processes interleaving writes into a disposable file, but a per-process name (append the pid, or use createTempFile) would keep the binaries from ever seeing each other's state if a test does start reading config back.

lucaromagnoli (commented):
The mechanism holds up across the seams I checked: envVar reads std::getenv and the redirect sets through setenv/_putenv_s, so both sides live in the same CRT environment on Windows; the guard test's reconstruction of the real path matches alwaysOSDefault exactly (userApplicationDataDirectory/MAGDA); and EnvScope snapshotting the redirect before the anchoring test unsets it closes the one in-process leak. Two comments inline: the rot-guard only covers one of the two binaries, and a nit on the shared scratch path.

tests/TestConfigRedirect.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+struct RedirectConfigFile {
+    RedirectConfigFile() {
+        auto dir = juce::File::getSpecialLocation(juce::File::tempDirectory)
+                       .getChildFile("magda-test-config");
+        dir.createDirectory();
+        // A name per process, not a shared one: two test binaries running at
+        // once (ctest does that) must never see each other's settings if a test
+        // ever starts reading config back.
+        const auto path = dir.getNonexistentChildFile("config", ".json").getFullPathName();
```
lucaromagnoli: **P2 — The scratch filename is not actually per-process.** `getNonexistentChildFile()` only checks whether a candidate currently exists; it does not create or reserve it. If the two test binaries start together while `config.json` is absent, both can select that same path before either writes it, so their settings are still able to collide. Please include a PID/UUID in the name, or atomically create a unique file or directory before exporting `MAGDA_CONFIG_FILE`.

PR #2107, opened 2026-08-17
Write the settings file atomically (#2104)

### LINKED ISSUE #2104: Config::save truncates config.json before writing it
### What happened

After unplugging an external drive, the entire `config.json` was silently reset to defaults: AI credentials gone, `aiPreset` back to `local_embedded` with every agent on `llama_local`, no recent projects, no preferred audio device, no browser favourites or default directory, no keyboard bindings. Only the missing API keys were noticed at the time; the rest of the file had gone too.

### Root cause

`Config::load()` returns early on any failure and leaves the default-constructed values in memory (`Config.cpp:370`):

```cpp
auto result = juce::JSON::parse(configFile.loadFileAsString(), parsed);
if (result.failed()) {
    DBG("Config::load - JSON parse error: " + result.getErrorMessage());
    return;                       // defaults kept; nothing records that the load failed
}
```

The same shape applies to the `existsAsFile()` and `getDynamicObject() == nullptr` branches just above and below it.

Nothing distinguishes "we never loaded" from "we loaded an empty config", so the next `Config::save()` -- any settings change, or app exit -- serialises the defaults straight over the user's file. The observed file matches a pure default save exactly, including `aiPreset`'s default of `"local_embedded"` (`Config.hpp:1481`).

Three things make it unrecoverable in practice:

- No backup is written before `save()` overwrites. `Config::save` goes straight to `configFile.replaceWithText(json)` (`Config.cpp:356`).
- The failure is reported through `DBG`, so it is invisible in a release build and leaves nothing in the log.
- A single bad parse is enough. The file does not have to be missing.

### Trigger

The drive is incidental to the cause but plausible as the trigger: pulling it while MAGDA was running can kill the app mid-write and leave a truncated file, which then fails to parse on the next launch. Worth noting the affected config had `browserDefaultDirectory` pointing at that drive.

### Suggested fix

- Track whether `load()` actually parsed a file. On a parse failure or unexpected root, refuse to `save()` over the existing file -- the in-memory state is not the user's config and must not be written back.
- Rename the unreadable file aside (`config.json.corrupt-<timestamp>`) rather than leaving it to be overwritten, so the contents remain recoverable.
- Write the settings file atomically: serialise to a temp file in the same directory and rename over the target, so an interrupted write cannot truncate the live file.
- Keep a rolling backup of the last known-good `config.json`.
- Surface the failure to the user instead of starting silently on defaults, and log it at a level that survives a release build.

### Impact

Total, silent loss of every user preference, including API keys, from one unclean shutdown. Nothing warns the user, and by the time it is noticed the original file has already been overwritten.

lucaromagnoli:
Closes #2104.

`Config::save()` calls `replaceWithText()` straight on config.json, which truncates the file before the new contents land. A crash, a force-quit or a yanked volume in that window leaves a half-written settings file, and every preference in it -- API keys included -- is gone.

The write now goes to a temporary file beside the target and is renamed over it, so the settings file is either untouched or wholly replaced. There is no window in which it is truncated. `juce::TemporaryFile` keeps the temporary in the same directory, which is what makes the rename atomic.

That is the whole change. `ConfigFileStore` exists so the write can be tested against a scratch file rather than the developer's own config, which is why nothing covered this before.

### What this deliberately does not do

An earlier version of this branch also refused to save over a settings file that had failed to parse, moved such a file aside as `config.json.corrupt-<stamp>`, kept a session backup, and reported the failure to the user.

All of that handled a corrupted settings file, and that scenario came from my own incorrect diagnosis of an API-key loss: the actual cause was the test suite writing a default `Config` over the real file (#2105, fixed in #2106). Checking the evidence afterwards, there is nothing to suggest a MAGDA config has ever failed to parse -- no `.corrupt` files, no parse errors anywhere in `magda.log`, and the config that was lost was valid JSON the whole time. It was pure defaults, not damaged.

So the recovery machinery, and a user-facing notification for a state that cannot be demonstrated, are out. Atomic writes need no such hypothesis: they remove the failure rather than cope with it.

One thing worth recording, since it cost real data. My first version refreshed `config.json.bak` on every save. That destroyed a live set of API keys during this work: the config was saved twice in the same second, so the second save's "backup" was the first save's damage and the last good copy was gone. If a backup is ever added here, it must be one a bad save cannot reach.

### Verification

Six cases covering creation, whole replacement, a shorter write leaving no tail of a longer one, no temporaries left behind, directory creation, and the result parsing back. Both suites green.

### DIFF
--- a/magda/daw/core/ConfigFileStore.cpp
+++ b/magda/daw/core/ConfigFileStore.cpp
@@ -1,0 +1,90 @@
+#include "ConfigFileStore.hpp"
+
+namespace magda::ConfigFileStore {
+
+namespace {
+
+/// Move an unreadable file out of the way so the next save cannot overwrite
+/// it. Returns the new location, or an invalid File if it could not be moved.
+juce::File quarantine(const juce::File& file, const juce::String& stamp) {
+    auto target = file.getSiblingFile(file.getFileName() + ".corrupt-" + stamp);
+    // Never clobber an earlier quarantine from the same stamp.
+    target = target.getNonexistentSibling();
+    return file.moveFileTo(target) ? target : juce::File();
+}
+
+}  // namespace
+
+ReadResult read(const juce::File& file, const juce::String& stamp) {
+    ReadResult out;
+
+    if (!file.existsAsFile()) {
+        out.status = ReadStatus::NoFile;
+        out.message = "no settings file at " + file.getFullPathName() + ", starting from defaults";
+        return out;
+    }
+
+    const auto text = file.loadFileAsString();
+    juce::var parsed;
+    const auto result = juce::JSON::parse(text, parsed);
+
+    juce::String problem;
+    if (result.failed())
+        problem = result.getErrorMessage().trim();
+    else if (parsed.getDynamicObject() == nullptr)
+        problem = "unexpected JSON root type";
+
+    if (problem.isEmpty()) {
+        out.status = ReadStatus::Loaded;
+        out.parsed = std::move(parsed);
+        out.message = "loaded " + file.getFullPathName();
+        return out;
+    }
+
+    // The file holds settings this build could not read. Preserve it: moving it
+    // aside keeps the contents recoverable AND leaves a clean slate, so the app
+    // can still save. Only if the move fails do we have to refuse to write.
+    const auto moved = quarantine(file, stamp);
+    if (moved != juce::File()) {
+        out.status = ReadStatus::Quarantined;
+        out.quarantinedAs = moved;
+        out.message = "could not read " + file.getFullPathName() + " (" + problem +
+                      "); moved it to " + moved.getFileName() + " and started from defaults";
+    } else {
+        out.status = ReadStatus::Unreadable;
+        out.message = "could not read " + file.getFullPathName() + " (" + problem +
+                      ") and could not move it aside; settings will NOT be saved over it";
+    }
+    return out;
+}
+
+bool mayWrite(ReadStatus status) {
+    // Everything except Unreadable: there, the file on disk still holds
+    // settings that were never loaded, so writing would destroy them.
+    return status != ReadStatus::Unreadable;
+}
+
+bool write(const juce::File& file, const juce::String& json) {
+    // createDirectory() is a no-op that reports success when the directory is
+    // already there.
+    if (!file.getParentDirectory().createDirectory().wasOk())
+        return false;
+
+    // Keep the previous contents as the last known good copy before replacing
+    // them.
+    if (file.existsAsFile()) {
+        auto backup = file.getSiblingFile(file.getFileName() + ".bak");
+        backup.deleteFile();
+        file.copyFileTo(backup);
+    }
+
+    // Write somewhere else and rename over the target, so a crash or a yanked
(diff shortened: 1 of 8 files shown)


magda/daw/core/ConfigFileStore.cpp
```diff
@@ -0,0 +1,21 @@
...
+    // createDirectory() is a no-op that reports success when the directory is
+    // already there.
+    if (!file.getParentDirectory().createDirectory().wasOk())
+        return false;
+
+    // Write somewhere else and rename over the target. juce::TemporaryFile puts
+    // the temporary beside the target so the rename stays within one volume,
+    // which is what makes it atomic.
+    juce::TemporaryFile temp(file);
+    if (!temp.getFile().replaceWithText(json))
```
lucaromagnoli, line 15: **P1 — This does not change the atomicity of the save.** In the vendored JUCE, `File::replaceWithText()` already constructs a same-directory `TemporaryFile` and calls `overwriteTargetFileWithTemporary()` (`juce_File.cpp:798-802`; the JUCE submodule is unchanged from the base). Wrapping that call in another `TemporaryFile` only adds a second temporary and rename, so the crash/force-quit behavior is unchanged and all six new tests would pass against the old call directly. It also inherits `replaceWithText()` ignoring the boolean from its internal `appendText()`, so this does not add write-completion validation either. Please either drop the no-op wrapper or write the temporary with an API whose result is checked before the single final rename.

tests/test_config_file_store.cpp
```diff
@@ -0,0 +1,110 @@
...
+namespace {
+
+// A scratch directory that cleans up after itself, so these tests never go
+// near the real config.json.
+class ScratchDir {
+  public:
+    ScratchDir() {
+        dir_ = juce::File::getSpecialLocation(juce::File::tempDirectory)
+                   .getChildFile("magda-config-store-test")
+                   .getNonexistentSibling();
```
lucaromagnoli, line 23: **P2 — The scratch directory is not reserved across discovered test processes.** `getNonexistentSibling()` only observes that a name is free; it does not create or reserve it. `catch_discover_tests()` exposes each case as its own CTest process, so parallel runs can select the same directory and then overwrite or delete one another’s files. This reproduces with the six config-store cases under `ctest -j 6 --repeat until-fail:25`: five failed with JSON from another case or an empty directory. Please use a PID/UUID name, or atomically create-and-retry until one process owns the directory.

lucaromagnoli:
Closing: the change is a no-op.

JUCE's `File::replaceWithText()` already does temp-and-rename internally (`juce_File.cpp:798`), so `Config::save()` was atomic before this branch. This wrapped that in a second temporary and rename and added nothing. The giveaway is that all six tests here pass against the old call unchanged -- they only ever proved the file ends up with the right contents, not that the write was atomic.

The premise was wrong twice: first a corrupted settings file (the real cause of the key loss was #2105, fixed in #2106), then a truncation window that does not exist.

One real gap surfaced while reading that code, recorded in #2104 for whenever it is wanted: JUCE discards `appendText()`'s return value, so a write that fails partway (disk full, quota) still gets renamed over the good config. That needs its own issue with disk-full as the stated motivation rather than inheriting this one's story.

PR #2110, opened 2026-08-17
Remember the browser view and panel tabs between sessions (#2103)

### LINKED ISSUE #2103: Browser and panel view selections are not saved between sessions
### What's missing

Browser and panel view selections reset to their defaults on every launch, so a user who has organised their plugins into folders has to re-pick the Folders view each time. A user asked for exactly this: "Is there a way to save a UI setup as default? For instance, I always want my right sidebar to open in Folder view in a new project, where I have my plugins all sorted nicely."

The folders themselves are safe -- `PluginBrowserContent::saveFolders` / `loadFolders` write to `magda::paths::pluginFoldersFile()`, and plugin favourites persist the same way. Only the *view selection* is lost.

### Gap 1: the plugin browser's view mode is never persisted

`PluginBrowserContent` hard-codes the selector at construction (`PluginBrowserContent.cpp:408`):

```cpp
viewModeSelector_.setSelectedId(1, juce::dontSendNotification);  // By Category
```

`currentViewMode_` is only ever written by the combo's `onChange`; nothing reads or writes `Config`. So By Category / By Manufacturer / By Format / Favorites / Folders always starts on By Category.

The Media Explorer already does this correctly, and is the pattern to copy: `Config::getBrowserLastView()` / `setBrowserLastView()` persists "filesystem" / "library", read at `MediaExplorerContent.cpp:1091` and written at 1305.

### Gap 2: panel tab selection and tab order are never persisted

`PanelController`'s state is only ever the defaults -- `PanelController.cpp:12` constructs from `getDefaultPanelStates()`, and line 212 resets to it. There is no load or save path, and no `activeTabIndex` or panel-layout key in `Config`.

Two things are lost as a result:

- Which tab is active in each panel. The left panel opens on `PluginBrowser` and the right on `Inspector` every time (`PanelState.hpp:104`, `:110`), regardless of what was in front at shutdown.
- The tab order. `PanelController::handleReorder` lets a user rearrange tabs within a panel and carefully preserves the active tab across the move, but the new order is gone on the next launch.

Panel *visibility*, *collapse state* and *sizes* are already persisted (`Config::getShowRightPanel`, `getRightPanelCollapsed`, `getRightPanelWidth`), which makes the tab state the odd one out.

### Suggested shape

- Add `pluginBrowserViewMode` to `Config` as a stable string ("category" / "manufacturer" / "format" / "favorites" / "folders"), not the combo's integer id, so reordering the menu later can't silently repoint an existing setting. Write it from `onChange`, read it in place of the hard-coded `setSelectedId(1)`.
- Persist `PanelController`'s per-panel active content type and tab order, keyed by content type for the same reason. Restore on construction, falling back to `getDefaultPanelStates()` for anything unknown or missing, so a config from an older build still opens.

Both are user-level preferences rather than project state, so they belong in `Config` and apply to new projects, which is what was asked for.

### Out of scope

Search text and scroll position -- those reasonably reset.

lucaromagnoli:
The plugin browser reopened on By Category every launch, so a user who had sorted their plugins into folders had to re-pick Folders each time. The panel tab layout was worse off: PanelController only ever held getDefaultPanelStates() and had no load or save path at all, so both the front tab and the tab order were gone on the next start.

Config now carries a pluginBrowserViewMode string and a panelTabs object holding each panel's front tab and tab order. Both store stable content-type ids ("pluginBrowser", "folders") rather than combo box item numbers or tab indices, so reordering a menu or the tab enum later cannot silently repoint an existing setting. Restoring is deliberately forgiving: an id this build does not know is dropped, and a tab the saved order never mentions is appended in its default position, so a config from another build still opens with every tab this build has.

The bottom panel restores its tab order but not its front tab. That one is derived from the selection, so restoring it would put a clip editor in front with no clip behind it until the user next clicked something.

Also flushes Config when MainComponent goes away. It was already writing panel collapse state and sizes there, but nothing saved on the way out and the app ends in _exit(), so those writes only ever reached the in-memory singleton -- config.json on disk still had every panel size at 0 after months of use.

### DIFF
--- a/magda/daw/core/Config.cpp
+++ b/magda/daw/core/Config.cpp
@@ -81,6 +81,25 @@ void Config::save() {
     root->setProperty("rightPanelWidth", rightPanelWidth);
     root->setProperty("bottomPanelHeight", bottomPanelHeight);
 
+    // Panel tab layout (front tab + left-to-right order, by content-type id)
+    {
+        auto writeTabs = [](const PanelTabsConfig& tabs) {
+            auto* panelObj = new juce::DynamicObject();
+            panelObj->setProperty("active", toJuceString(tabs.activeTab));
+            juce::Array<juce::var> orderArray;
+            for (const auto& id : tabs.tabOrder)
+                orderArray.add(toJuceString(id));
+            panelObj->setProperty("order", orderArray);
+            return juce::var(panelObj);
+        };
+
+        auto* panelTabsObj = new juce::DynamicObject();
+        panelTabsObj->setProperty("left", writeTabs(leftPanelTabs));
+        panelTabsObj->setProperty("right", writeTabs(rightPanelTabs));
+        panelTabsObj->setProperty("bottom", writeTabs(bottomPanelTabs));
+        root->setProperty("panelTabs", juce::var(panelTabsObj));
+    }
+
     // Language
     root->setProperty("language", toJuceString(language));
 
@@ -201,6 +220,7 @@ void Config::save() {
     root->setProperty("browserTreeView", browserTreeView);
     root->setProperty("browserDefaultDirectory", toJuceString(browserDefaultDirectory));
     root->setProperty("browserLastView", toJuceString(browserLastView));
+    root->setProperty("pluginBrowserViewMode", toJuceString(pluginBrowserViewMode));
     root->setProperty("sampleTaggerModelsDir", toJuceString(sampleTaggerModelsDir));
     root->setProperty("commandModelModelsDir", toJuceString(commandModelModelsDir));
     root->setProperty("loadSampleTaggerOnStartup", loadSampleTaggerOnStartup);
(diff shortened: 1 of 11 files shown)


lucaromagnoli (commented):
Re-reviewed at `cff6903f`. The active-content preservation through a persisted reorder is correct, including the bottom-panel `Empty` case and unknown active-id fallback. Focused panel tests pass locally (10 cases, 17 assertions); no further code findings.

CI note: macOS failed before checkout because codeload.github.com returned HTTP 429 while downloading `actions/checkout`, so that failure is unrelated to this patch. Windows is still running; the remaining checks pass.

lucaromagnoli (commented):
Re-reviewed at `602459a1`. The switch-based stable-ID mapping and reverse lookup are consistent, and the new coverage verifies every listed content type has a unique, non-empty ID that round-trips. Local build passes; the focused panel suite passes all 12 cases and 61 assertions. No new code findings.

CI note: macOS again failed before checkout because codeload.github.com returned HTTP 429 while downloading `actions/checkout`, so that failure is unrelated to this patch. Linux, JUCE Linux, and Windows are still running; quality, security, and release checks pass.

PR #2112, opened 2026-08-18
Gate the corpus on block-size invariance at 64, 512 and 4096 (#2078)

### LINKED ISSUE #2078: Validation slice 4: block-size invariance gate at 64, 512 and 4096
Slice 4 of #1896, from the block-size comment on it. Depends on #2075 for the whole-project
case.

`RenderContext` requires that output is a function of timeline position and nothing else, and
#2040 proved it over real material for two block sizes on the clip corpus. That check earned
its place immediately: it is what caught a rate resolved from a block's own two ends and a
stretcher framing whatever sizes it was handed, both of which made output a function of how the
callback was cut up.

This slice makes it a permanent gate at the sizes that matter.

Render the corpus at 64, 512 and 4096 samples. Projects with internal devices only must be
strictly bit identical across all three. Projects with external plugins are compared within an
epsilon with the diffs attributed to the plugin rather than absorbed, which is the external
tier from #2075 doing its job.

4096 is the size that carries the guarantee users actually depend on, which is a large-buffer
offline render that sounds like the small-buffer one. 64 is where anything that assumes it has
room to work fails. A path that behaves at 512 and misbehaves at both ends is a path that was
tested at one size.

The gate belongs in CI rather than in a nightly, and it costs three renders instead of one on
a corpus already budgeted at thirty seconds. If it will not fit, cases get shorter rather than
sizes getting dropped.

lucaromagnoli:
Closes #2078. Slice 4 of #1896.

`RenderContext` requires that output is a function of timeline position and nothing else.
#2040 proved it over real material at two sizes; this makes it a permanent gate at the sizes
that matter, in CI rather than in a nightly.

## The gate

`tests/test_null_diff_block_size.cpp`. Every non-MIDI corpus case renders at 64, 96, 512 and
4096, and the other three are compared against 512. It runs in `magda_tests` because the claim
is the native engine's alone: the incumbent owes nobody block-size invariance and could not be
held to it here if it did.

**A project of internal devices is held to bit identity**, not to a floor. There is no mechanism
by which a deterministic graph fed the same timeline produces anything else, and a floor would
let a real dependence hide under it until it grew. Written as a floor of negative infinity so it
goes through the same comparator everything else does, rather than adding a second one to keep
honest.

**A project hosting an external plugin is compared within an epsilon it declares**, with the
plugins named beside the residual. The naming is the point: the difference is attributed to what
can account for it rather than absorbed into a wider number. The epsilon is refused outright on
a project with no plugin in it, so it can only ever be bought by something there is to attribute
the difference to, which is the one way this tier could become a way to make a failing case
pass. Nothing hosts a plugin yet, so every corpus case is on the strict side today and the
external path is covered by hand-built cases: #1893 should find a gate to land in rather than a
gate to widen.

Per-rung assertions, so a failure says which size and by how much rather than just naming the
case. `knownDependent` is empty and asserted in both directions, the same rule the corpus's
calibration list lives by.

## Why 96 is here

Not one of the three sizes the issue names, and kept anyway, because it is what this check ran
at before it was a gate. A voice drives a stretcher in 128-sample cells anchored to where its
event begins, so at 512 and 4096 every block boundary is also a cell boundary and the holdover
buffer, the head-drop and the mid-cell resume never run for a clip that starts at zero. 64 puts
a boundary at half a cell, always the same half. 96 rotates through every phase the cell grid
has. Dropping it to reach three sizes would have traded coverage for a round number.

## What it found

Adding 4096 caught a bug on its first run, in a path that had been green at 64, 96 and 512 all
along. `SoundTouchClipStretcher::preRollSamples` sized its cushion from the block, so the
stretcher was primed with 4096 samples of surplus at 4096 where it had 512 at 512. Priming is
what sets a phase vocoder's state, so the same clip rendered a decibel apart at the two sizes:
peak -1.1 dB on `stretch.soundtouch.normal` and -0.5 dB on `.better`. Completely different
audio, not rounding.

The cushion is a cell now, which is the unit `ClipVoice::renderThroughCells` actually drives it
in, and a block was never the right unit here. Nothing changes at the sizes that already agreed,
which the gate itself confirms: 64 and 96 are still bit identical to 512.

Signalsmith takes its pre-roll from the library's own seek length and never had the problem.

## Also

The corpus's three stand-in device builders declared `PluginFormat::VST3` by struct default.
Nothing read the field before, so it went unnoticed; the gate does read it, and left alone it
would have handed the whole corpus an external project's allowance.

## Verification

- Full Catch2 suite: 2648 passed, 13 skipped, 0 failed.
- Full JUCE suite: 0 failed.
- The two-engine corpus runner is unchanged: the same eight calibrating cases, list still exact
  in both directions.
- Gate runtime is about 26 s over the corpus in a debug build. Per #2040's rule, restated by
  #2078: when it will not fit, cases get shorter rather than sizes getting dropped.


tests/test_null_diff_block_size.cpp
```diff
@@ -0,0 +1,408 @@
...
+    const std::set<std::string> knownDependent{};
+
+    std::set<std::string> failed;
+
+    for (const auto& value : sharedCorpus(scratch())) {
+        // A MIDI case renders no audio at all: the device standing in for a
+        // synth records what arrives instead of sounding. What block size does
+        // to a stream of captured events is the runner's question, asked of
+        // both engines, not this one's.
+        if (value.capturesMidi())
```
lucaromagnoli: **P2 — Keep mixed audio/MIDI projects in the audio gate**

`capturesMidi()` says that MIDI streams are compared; it does not mean the project has no audio. `project.mixed` deliberately sets `AudioTier::Exact` and contains an audio clip, so this condition removes the only mixed project from every block-size comparison. Please skip cases based on the absence of an audio assertion (for example `tier == AudioTier::None`) rather than on MIDI capture.

tests/test_null_diff_corpus.cpp
```diff
@@ -95,7 +96,15 @@ TEST_CASE("Every allowance carries a mechanism", "[nulldiff][corpus]") {
...
-                            value.declaredMidiShiftBeats != 0.0;
+                            value.declaredMidiShiftBeats != 0.0 ||
+                            // A raised block-size epsilon is the same kind of
+                            // thing, against the engine's own second render
+                            // rather than against the incumbent (#2078). The
+                            // gate itself refuses one on a project with no
+                            // plugin to pin it on; this is the other half, so
+                            // that a project which does host one still has to
+                            // say what about it frames its own work.
+                            std::isfinite(value.blockSizeEpsilonDb);
```
lucaromagnoli: **P2 — Reject positive infinity as an epsilon**

`std::isfinite(+inf)` is false, so a case with `blockSizeEpsilonDb = +inf` does not need a mechanism here. The gate then uses that value as `floorDb`, and every finite residual satisfies `peakDb <= +inf`, silently disabling the check for an external-plugin project. Please reject non-finite raised bounds explicitly, preserving only `-inf` as the strict default.

PR #2114, opened 2026-08-18
Legacy corpus: project and preset migrators (#2079)

### LINKED ISSUE #2079: Validation slice 5: project and preset migrators, with a legacy corpus
Slice 5 of #1896, the migration half. Unblocked, and independent of the render corpus: this is
serialization work with no audio in it.

The requirement from #1896 is the whole of it. Old projects and presets must load bit
identically, or through a migration that is documented and tested. A user's project opening
different after an engine swap is the failure that ends the swap, and it will not be discovered
by a null-diff corpus whose cases are built in code.

Three migrators:

- **Internal device `pluginState` v1 to v2.** #1887 defined the engine-neutral schema. Every
  device that shipped with a v1 state needs its v1 read and its v2 written, and the read has to
  survive states written by versions that are already in the wild.
- **`paramIndex` mapping where ordering changes.** A parameter index that moves silently
  repoints automation and macro links at the wrong parameter, which is the worst kind of
  corruption because it loads without complaint and sounds wrong.
- **Pinned enum values.** Anything an enum serializes as an ordinal is a hazard the moment a
  value is inserted. Pin the values, and add a test that fails when someone inserts one.

**A legacy corpus is the deliverable, not just the code.** Real `.mgd` projects and real
presets saved by released versions, checked in, each loaded and asserted. This is the corpus
that has to come from disk, and it is the reason #2040's build-cases-in-code rule does not
apply here: the load is the thing under test, so a case that skips it tests nothing.

Coverage per released version that can still be opened, since a project saved by 0.14 and a
project saved by 0.18 do not exercise the same path.

Where a migration changes what the project means rather than only how it is written, it is
documented in the manual and asserted here, and it renders identically before and after
through whichever engine is current. A migration nobody can point at in writing is a behaviour
change users discover on their own.

lucaromagnoli:
Closes #2079. Slice 5 of #1896, the migration half.

## The corpus

22 real .mgd projects and 2 presets saved by released builds from 0.4.8 through 1.0.0, checked in under tests/corpus/legacy and asserted on load. The load is the thing under test, so every case comes off disk; nothing is built in code. The files are excluded from git line-ending translation and from the pre-commit fixers, because two presets carry CRLF from Windows saves and the corpus only proves anything while it stays byte for byte what those versions wrote.

## The three migrators

- **pluginState v1 to v2**: the JUCE-suite test converts every pre-v2 device state the corpus carries on a live plugin and asserts the conversion is idempotent. That check caught the bridge re-capturing Tracktion's parameter-shadow properties, including the three whose property name is spelled differently from their paramID (chosusSpeed, chrousWidth, damp; typos frozen in released TE files). Capture now drops all of them, so a load-save round trip is stable.
- **paramIndex remapping** (DeviceParamMigrations): where a device renumbered its parameters, saved automation lanes, macro and mod links, and the device's own saved parameters follow the move. Runs after the retired-device aliases so a rewritten device is already its successor when indices are looked up. Ships the EQ band shift and the limiter's dropped parameters.
- **Pinned enum ordinals**: every enum that serializes as an ordinal is pinned, with a test that fails the moment someone inserts a value.

## Also in here

- Parameters saved by builds that wrote paramIndex -1 get their index restored from array position on load; without it nothing addressed by index on those files can resolve.
- Clips serialize sorted by id, so re-saving an opened project no longer rewrites the clips array in whatever order the unordered_map iterated.

Both suites green locally: magda_tests 2670 passed, magda_juce_tests all passed.


magda/daw/core/DeviceParamMigrations.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+    migrateIndices(device.visibleParameters);
+    migrateIndices(device.miniMixerParameters);
+    migrateIndices(device.aiSoundDesignerParameters);
+}
+
+/// Which migration each device needs, collected BEFORE any of them run: a
+/// migration is matched on the saved parameter count, and migrating a device
+/// changes that count.
+using MigrationsById = std::map<DeviceId, const ParamIndexMigration*>;
```
lucaromagnoli, line 137: **P1 — Key migrations by the full device path**

`DeviceId` is not global: main FX, post-FX, and mixer-analysis deliberately use separate ID counters, so a valid project can contain the same numeric ID in multiple sections. This map merges those devices. During apply, a colliding non-migrating device can then be run through another device's mapping (or two migrations overwrite each other), silently dropping or repointing parameters and automation on load. Please key the project pass by `ChainNodePath` (as `LegacyDeviceAliases` does); ID-only matching is safe only for preset fragments.

tests/test_persisted_enum_pins.cpp
```diff
@@ -0,0 +1,383 @@
...
+static_assert(static_cast<int>(ControlTarget::Kind::ModParam) == 2);
+static_assert(static_cast<int>(ControlTarget::Kind::TrackVolume) == 3);
+static_assert(static_cast<int>(ControlTarget::Kind::TrackPan) == 4);
+static_assert(static_cast<int>(ControlTarget::Kind::SendLevel) == 5);
+static_assert(static_cast<int>(ControlTarget::Kind::Tempo) == 6);
+
+static_assert(static_cast<int>(ChainStepType::Rack) == 0);
+static_assert(static_cast<int>(ChainStepType::Chain) == 1);
+static_assert(static_cast<int>(ChainStepType::Device) == 2);
+static_assert(static_cast<int>(ChainStepType::Segment) == 3);
```
lucaromagnoli, line 141: **P2 — Pin the persisted `ChainSegment` ordinals**

`ChainStepType` is pinned here, but a segment step also persists `ChainSegment` in `ChainPathStep::id` (`PostFx`/`MixerAnalysis`), and `AutomationModSerializer` writes that integer unchanged. Inserting or reordering `ChainSegment` would redirect old automation, macro, and modulation paths while this gate stays green. Please pin `Fx == 0`, `PostFx == 1`, and `MixerAnalysis == 2` here too.

PR #2115, opened 2026-08-19
DAWproject round trip as an engine-facing cross-check (#2080)

### LINKED ISSUE #2080: Validation slice 6: DAWproject round-trip as an engine-facing cross-check
Slice 6 of #1896, the DAWproject cross-check. Depends on #2075.

Most of this already exists. `tests/test_project_document_adapters.cpp` round-trips transport,
tracks and arrangement clips, track volume and pan automation, group hierarchy, session clip
scenes, master role and send routing, autoTempo warped audio, embedded audio, VST3 device state
and the builtin device mappings, and `DawProjectValidator` checks the vendored schemas. What is
missing is the engine-facing half, which is the only part #1896 wants.

**Export each corpus project to DAWproject, reimport it, and require the reimported model to
compile to the same plan and render to the same null.** Plan equality uses the goldens
machinery from #2076; the render uses the corpus from #2075.

What this catches is not a serialization bug in the ordinary sense, since the existing tests
already assert field by field that the round trip preserves what it was asked to preserve. It
catches the facts the engine depends on that nobody thought to assert, because the assertion
list was written by reading the model rather than by rendering it. A property the plan compiler
reads and the adapter drops passes every test in that file today and fails here.

It is a semantic cross-check rather than a parity test, so it compares native against native,
not native against the fork. The fork has no part in it.

Cases whose DAWproject representation is genuinely lossy declare it, with the field and the
reason, and assert the loss is exactly that and nothing more. A lossy mapping is a fine thing
to have written down and a bad thing to discover.

lucaromagnoli:
Closes #2080. Slice 6 of #1896.

Every corpus project is exported to a `.dawproject`, reimported, and required to compile to the same plan and render to the same samples.

`tests/test_project_document_adapters.cpp` already asserts, field by field, that the round trip preserves what it was asked to preserve, and it does. What it cannot assert is the facts the engine depends on that nobody thought to put on the list, because the list was written by reading the model rather than by rendering it. This asks the model to play instead.

## What it compares

Native against native, so the fork has no part in it and the corpus's tiers do not apply: two renders of what is supposed to be one model owe each other bit identity rather than a floor.

- **The plan**, through `dumpPlan` and the first-difference reporting from the plan goldens (#2076), which is now shared between the two runners rather than copied into both.
- **The audio**, through the corpus comparator (#2075) at a floor of `-inf`, which is bit identity written as a level so it goes through the same comparator everything else does.
- **The MIDI**, literally and per track. `compareMidi` exists to hold a grid sampler against a curve; there is no slack to allow for between two runs of one compiler.
- **The diagnostics**, because a snapshot that dropped a clip and a render that matched it anyway are two bugs, not none.

The trip is the real path, archive included. Audio reaches the importer as a path inside a zip, and it is `DawProjectArchive` that extracts it, repoints the event and rescales its source-domain anchors for whatever rate the extracted file turns out to run at. A round trip that stopped at the XML would skip all three and then render whatever the original still had pooled, which is a test of the corpus rather than of the export.

The importer renumbers, so ids are mapped back by name before anything is compiled. By name and not by position, which is what keeps the mapping from hiding a reordering: the rebuilt case keeps the imported document's order, so a project whose tracks came back in a different order compiles to a plan whose ops are in a different order and the comparison says so. A name that matches nothing, or twice, is a refusal rather than a residual.

## Declared losses

Eighteen cases declare one: fades, reverse, speed ratio, stretch mode, transposition, warp markers, takes and comping, controllers and per-note expression, groove templates, internal devices, and the tempo map past its first point. Each names the field, says why the format cannot hold it, and restores it from the original before anything is compiled.

Restoring rather than erasing from both sides is the choice worth reading. A fades case with its fades erased stops asserting anything about its fades; one with its fades put back still asserts everything else the format was supposed to carry, which is the "and nothing more" half.

Every declaration has to fire. A restore that finds nothing to put back means the format carried the field after all, so a mapping that stops being lossy fails here rather than leaving a paragraph that is quietly no longer true. A second test refuses a declaration against a case name the corpus no longer has.

## On it passing first time

It found no engine-facing bug, which I did not take at face value. Each limb was sabotaged to confirm it has teeth:

| sabotage | what it produced |
|---|---|
| drop the `fades.curves` declaration | audio residual peak −6.0 dB, first difference at sample 0 |
| drop the `midi.cc` declaration | `T1 events=15` against `T1 events=2` |
| drop the `midi.fold` declaration | note-off at sample 9923 against 8820 |
| drop the `project.mixed` declaration | `ops=45` against `ops=23`, and the MIDI capture gone |
| declare a loss on `placement.grid`, which has none | "nothing had to be restored, so the format carries this field after all" |

There is no known-lossy suppression list, because nothing needs suppressing and a list written before it does is somewhere for a real finding to be filed.

## Limitation, stated in the code

`internalDevices` restores a track's chain wholesale. Every device in this corpus is internal and none survives the trip, so that is exact today. The day a case carries a VST3 alongside an internal device, restoring the chain would put the exported device back too and stop comparing what the format did carry. #1893 brings the first hosted plugin and will need that restore to become surgical.

## Testing

`make test` green locally: 2687 cases, 588129 assertions, 0 failures. The new gate is 6.5 s and runs in `magda_tests`, so it is in CI rather than in a nightly.

### DIFF
--- a/tests/DawProjectRoundTrip.cpp
+++ b/tests/DawProjectRoundTrip.cpp
@@ -1,0 +1,604 @@
+#include "DawProjectRoundTrip.hpp"
+
+#include <algorithm>
+#include <map>
+#include <set>
+#include <utility>
+
+#include "core/SourcePool.hpp"
+#include "magda/daw/project/serialization/DawProjectArchive.hpp"
+#include "magda/daw/project/serialization/ProjectDocument.hpp"
+
+namespace magda::nulldiff {
+
+// =============================================================================
+// The losses
+// =============================================================================
+//
+// Each entry names a field, says why DAWproject cannot hold it, and puts it
+// back. Nothing here is a tolerance: a restore either copies the original's
+// value over an imported default or does nothing, and doing nothing is what the
+// runner fails on.
+//
+// The bar is in the header. A field belongs here when the case sets it and the
+// trip does not bring it back, whether or not anything reads it.
+
+namespace {
+
+/// The clip of @p value named @p name, or nullptr. Names are unique per case
+/// and the runner asserts it, so a second match cannot silently win.
+const ClipInfo* clipNamed(const Case& value, const juce::String& name) {
+    for (const auto& clip : value.clips)
+        if (clip.name == name)
+            return &clip;
+    return nullptr;
+}
+
+/// Walk both cases' primary audio events in matched pairs, applying @p apply.
+///
+/// Matched by clip name rather than by index: the importer's clip order follows
+/// the file, and a restore that paired by position would put one clip's fade on
+/// another the first time that order changed.
+bool forEachAudioEvent(Case& imported, const Case& original,
+                       const std::function<bool(AudioEvent&, const AudioEvent&)>& apply) {
+    bool restored = false;
+
+    for (auto& clip : imported.clips) {
+        auto* event = clip.primaryEvent();
+        if (event == nullptr)
+            continue;
+
+        const auto* source = clipNamed(original, clip.name);
+        if (source == nullptr)
+            continue;
+
+        const auto* sourceEvent = source->primaryEvent();
+        if (sourceEvent == nullptr)
+            continue;
+
+        restored = apply(*event, *sourceEvent) || restored;
+    }
+
+    return restored;
+}
+
+bool forEachClip(Case& imported, const Case& original,
+                 const std::function<bool(ClipInfo&, const ClipInfo&)>& apply) {
+    bool restored = false;
+
+    for (auto& clip : imported.clips)
+        if (const auto* source = clipNamed(original, clip.name))
+            restored = apply(clip, *source) || restored;
+
+    return restored;
+}
+
+Loss fades() {
+    return {.field = "AudioEvent::fadeInSeconds, fadeOutSeconds, fadeInType, fadeOutType, "
+                     "fadeInBehaviour, fadeOutBehaviour",
+            .reason = "a DAWproject clip has no fade of its own. The format puts a level "
+                      "envelope on an automation lane instead, which is a different object "
+                      "with different semantics: it cannot say that this clip's edge is a "
+                      "speed ramp rather than a gain ramp, and it outlives the clip being "
+                      "moved",
+            .restore = [](Case& imported, const Case& original) {
+                return forEachAudioEvent(imported, original,
+                                         [](AudioEvent& event, const AudioEvent& source) {
+                                             if (event.fadeInSeconds == source.fadeInSeconds &&
+                                                 event.fadeOutSeconds == source.fadeOutSeconds &&
+                                                 event.fadeInType == source.fadeInType &&
+                                                 event.fadeOutType == source.fadeOutType &&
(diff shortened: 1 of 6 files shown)


tests/DawProjectRoundTrip.cpp
```diff
@@ -0,0 +1,604 @@
...
+    std::set<SourceId> seen;
+
+    for (const auto& clip : clips) {
+        const auto* event = clip.primaryEvent();
+        if (event == nullptr || event->sourceId == INVALID_SOURCE_ID)
+            continue;
+        if (!seen.insert(event->sourceId).second)
+            continue;
+
+        const auto* source = SourcePool::getInstance().get(event->sourceId);
```
lucaromagnoli, line 510: **P2 — Verify that the render uses the extracted archive media**

This accepts any source already present in `SourcePool`. If the exporter regresses to writing an absolute/external reference for one of these corpus files, the importer will legitimately leave that path alone, this lookup will recover the original source, and the plan/audio comparison can still null without exercising archive embedding or extraction. Please pass the expected `media` directory into this check and refuse audio sources whose resolved file is not beneath it.

tests/DawProjectRoundTrip.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    std::map<TrackId, TrackId> mapped;
+    for (const auto& track : imported.tracks) {
+        const auto found = byName.find(track.name);
+        if (found == byName.end()) {
+            refusal = "a track came back named '" + track.name.toStdString() +
+                      "', which nothing went out under";
+            return {};
+        }
+        mapped[track.id] = found->second;
```
lucaromagnoli: **P2 — Require a one-to-one name mapping on the imported side**

Only duplicate names in the original are rejected. If an import returns tracks `A, A` for originals `A, B`, both imported IDs map to A's original ID and `mapped.size()` can still equal the original track count, so the harness normalizes a missing/duplicated track instead of refusing it. `takeClips` has the same issue because it checks only the final count. Please consume each original name once (and reject reused or missing names) for both tracks and clips, matching the stated “matches twice is a refusal” contract.

tests/DawProjectRoundTrip.cpp
```diff
@@ -0,0 +1,678 @@
...
+
+    NameClaims<TrackId> claims("track", std::move(byName));
+    std::map<TrackId, TrackId> mapped;
+
+    for (const auto& track : imported.tracks) {
+        const auto id = claims.claim(track.name, refusal);
+        if (!id.has_value())
+            return {};
+
+        mapped[track.id] = *id;
```
lucaromagnoli: **P2 — Refuse duplicate imported track IDs before remapping**

The name claims are now one-to-one, but this assignment still silently overwrites when two uniquely named imported tracks carry the same `track.id`. The previous `mapped.size() != original.size()` check happened to catch that case; after replacing it with `allClaimed()`, both names can be claimed successfully and `takeTracks()` maps both tracks through the last value stored for the shared ID. That normalizes another invalid import instead of refusing it (and can also redirect every clip using that ID). Please use `emplace` and reject a duplicate imported ID, or retain an explicit uniqueness check alongside the name claims.

Issue #2121, opened 2026-08-19
Parameters slice 6: macros at track, rack and device scope


lucaromagnoli:
Slice 6 of #1891. Depends on slices 1 and 2.

Macros at track, rack and device scope, at parity with the uniform macro behaviour every device
has today through its rack wrapper.

- Scope resolution: which macro array a `DeviceMacro` target lands on is decided by the owner of
  its path, and a device inside a rack inside a rack resolves to the nearest one.
- Macro to mod-param links, so a macro can drive an LFO's rate. That is a `ModParam` target and
  therefore a second hop through the link graph.
- Nested racks, where the same macro index exists at more than one level.

Issue #2118, opened 2026-08-19
Parameters slice 3: automation bake


lucaromagnoli:
Slice 3 of #1891. Depends on slices 1 and 2.

Automation playback as a bake target for the MAGDA-owned curve model. Nothing is ported here: the
curves are already MAGDA's, and this compiles them into the segment lanes slice 1 defined.

- Absolute lanes and automation clips, with clip loop, length and occlusion resolved the way
  `ClipLaneFlattener` resolves them today.
- Linear, bezier, step and hard corner, with tension.
- Stepped targets quantised at the spec rather than at the curve.
- Baked against the tempo map, so a tempo edit republishes the lanes the way it republishes the
  clip snapshot.
- The dB to fader-position curve has to match the fork bit-exactly. It is baked into the persisted
  value domain of every volume lane, so a difference here is a difference in every project ever
  saved. `faderGainFromVolume` in `PlanValues.hpp` is the same curve seen from the other end, and
  the comparison belongs against it.

Carried over from #2117: track volume, pan and send level are addressable in the model but are
resolved into `PlanValues`, per op, off the audio thread, and the parameter table does not carry
them. Nothing can move them faster than a publish today, so it costs nothing until this slice: a
lane over a fader moves it per block, which a table resolved by the publisher cannot do.

So they cross over here. The mixer ops read their gains out of the parameter table, `PlanValues`
keeps what is not a parameter (mute and solo, which no lane targets), and the fader and pan law
move to where the block resolves rather than where the publish does. The diagnostic #2117 emits
for those targets goes away with them.

Issue #2117, opened 2026-08-19
Parameters slice 2: publication, addressing and the link graph


lucaromagnoli:
Slice 2 of #1891. Depends on slice 1.

Where the resolved values come from, and how a link finds its parameter.

- Parameter values resolved off the audio thread against the plan they were resolved for,
  fingerprint-checked and swapped the way `PlanValues` already is. A parameter move happens at
  mixer speed, so it must never compile a plan.
- `ControlTarget` resolved once into a plan-side parameter address, the way the compiler resolves
  a model location into an `OpKey`. Every consumer already speaks `ControlTarget`, so this is the
  one translation, and the place where a target naming something the plan no longer has is
  reported rather than dropped.
- The link graph: a named source contributes to a target with its own depth and bipolarity,
  sources resolve in a defined order, and a cycle is rejected at resolve time rather than
  discovered on the audio thread. Sources are constants at this point; the engines arrive in
  slices 4 and 5.
- A canonical dump, so parameter state joins the goldens in #2076. Excludes addresses and
  allocation order, like the plan dump.

PR #2124, opened 2026-08-19
The parameter lane and what a device reads (#2116)

### LINKED ISSUE #2116: Parameters slice 1: the value lane and what a device reads
Slice 1 of #1891.

The engine has no parameter concept today. `EngineDevice::process()` takes audio, MIDI and a
sidechain, and `PlanValues` covers only what a mixer move can touch. This slice adds the lane
everything else in #1891 writes into, and settles the contract the first comment on #1891 asked
for.

Four things, named separately, resolved in this order:

- base: the stored value, which is what a UI move, a control surface or a host write sets;
- automation: segments baked from the MAGDA curve model (slice 3);
- modulation: the sum of link contributions (slices 4 to 6);
- the resolved stream, which is the only thing a device ever reads.

The decisions:

- Automation replaces the base wherever a Reading lane covers the block. The base is what the
  parameter is where no lane covers it. The engine honours the authority state it is handed
  rather than deciding it; `AutomationStateMachine` already owns that decision on the MAGDA
  side, and read mode is what ships here.
- Modulation is additive in the normalised domain, with per-link depth and bipolarity applied at
  the link, as they are today.
- One clamp, at the end, then the spec's range and step quantisation. A device never clamps a
  value it was handed.
- Within a block a lane is constant or linear-ramp segments. A device opts into segment accuracy;
  otherwise it reads one value for the block. "Resolved once per block" and a sample-accurate
  ramp are the same thing when the block carries segments rather than a scalar.
- Parameter events resolve before MIDI at the same sample offset, so a note-on sees the value
  automation put there.

What lands here is the parameter spec, the resolved buffer and the device read API, with base
values and host writes only. The host-write-dropped-under-modifier bug class dies here: a host
write goes into a lane modulation does not write, so nothing can overwrite it.

lucaromagnoli:
Closes #2116. Slice 1 of #1891.

The native engine had no parameter concept at all. `EngineDevice::process()` took audio, MIDI and a sidechain, and nothing in the plan or in the value table said what a device's cutoff was. This is the lane the rest of #1891 writes into, and it settles the contract the first comment on #1891 asked for.

## Three lanes, not one value

The base is the stored value: a knob, a control surface, a preset load, a host write. Automation is a curve covering as much of a block as it covers. Modulation is what the links reaching the parameter add on top.

They stay separate all the way to the device. The incumbent engine has one value and several writers, which is why a host write on a parameter under an active modifier disappears: the modifier writes next and there was only ever one place to write. Here the write goes into a lane modulation never touches, so nothing can overwrite it. That case is in the tests by name rather than as a side effect of the design.

## The precedence, in one function

`resolveParam` is the only place any of it is implemented. A device is shown the resolved stream and has no way to ask where a value came from.

- Automation replaces the base wherever a lane covers the block, and the base holds everywhere else, which is what an automation clip starting and stopping inside a block does.
- Links are summed in the normalised domain, each with its own depth and polarity. A parameter that does not take modulation adds nothing rather than dropping its links: the links are the model's, and a block is not where a model edit happens.
- One clamp, after the sum, then the parameter's own scale. Clamped per link, a +0.8 and a -0.6 against a base of 0.5 land at 0.4; clamped once they land at 0.7, which is where the parameter actually is.
- A stepped parameter is quantised by the scale and never ramped, because there is nothing between two of its values to ramp through.
- Authority state is not decided here. The model owns it (`AutomationStateMachine`), a lane reaching a block is one the model has already decided is playing, and read mode is what the port ships.

## Segments, and why nothing opts in yet

Within a block a parameter is constant or linear-ramp segments rather than a scalar, which is what lets "resolved once per block" and "a sample-accurate ramp" both be true instead of contradicting each other.

`ParamSpec::segmentAccurate` is off and nothing sets it during the port. The reason is parity rather than cost: the fork settles a parameter at the block boundary, so a device resolving inside the block would differ from it by however much the curve moves across a block, on every automated parameter, in every project. The corpus would report that as an engine bug and it would be right.

## One curve, not two

Normalised to real is the model's conversion. `ParameterUtils` grows a `ParameterDomain`, the flat part of a `ParameterInfo` the curve actually reads, and `normalizedToReal` / `realToNormalized` become that one under another name. A `ParameterInfo` carries strings, a choice list and a shared display provider and cannot go near a block, and a second implementation of the curve is a difference nobody sees until a project sounds different in one engine than in the other.

`modulationOffset` comes out of `applyModulation` for the same reason: neither existing helper can be called from a block, since one clamps per link and the other wants a `std::vector`.

## Storage

One table per plan, indexed by parameter id, each parameter's segments in a fixed-width slot of an arena allocated off the audio thread. A device reads a contiguous window of it and indexes its own parameters from zero, the way `ParameterInfo::paramIndex` does.

The width is a budget, like the MIDI one in `EngineDevice.hpp`: a curve with more breakpoints than fit keeps the ones there is room for and rolls the rest into its last segment, which then ramps to where the curve actually ends. Coarser on a very busy curve, never a wrong destination, and it is a case.

## What is not here

The table is not published, so the executor hands every device an empty window. Resolution against a live plan, the fingerprint check, `ControlTarget` addressing and the link graph are #2117. The sources that fill the lanes are #2118 through #2121.

## Testing

`make test` green locally: 2703 cases, 588065 assertions, 0 failures. `make test-juce` green. The 16 new cases are model-level Catch2 in `magda_tests` and cost no measurable time.

### DIFF
--- a/magda/engine/param/ParamResolve.cpp
+++ b/magda/engine/param/ParamResolve.cpp
@@ -1,0 +1,130 @@
+#include "param/ParamResolve.hpp"
+
+#include <juce_core/juce_core.h>
+
+#include <algorithm>
+
+namespace magda::engine {
+
+namespace {
+
+/// Everything the links add, before the clamp. A parameter that does not take
+/// modulation adds nothing rather than dropping the links: the links are the
+/// model's, and a block is not where a model edit happens.
+float modulationSum(const ParamSpec& spec, std::span<const ModContribution> links) {
+    if (!spec.modulatable)
+        return 0.0f;
+
+    float offset = 0.0f;
+    for (const auto& link : links)
+        offset += magda::ParameterUtils::modulationOffset(link.value, link.amount, link.bipolar);
+
+    return offset;
+}
+
+}  // namespace
+
+void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
+                  const ParamSources& sources) {
+    auto* slot = out.slotFor(param);
+    if (slot == nullptr)
+        return;
+
+    const int numSamples = out.numSamples();
+    const int capacity = out.segmentCapacity();
+    const float offset = modulationSum(spec, sources.modulation);
+    const bool stepped = magda::ParameterUtils::isStepped(spec.domain);
+
+    // The one clamp, and the one conversion out of the normalised domain the
+    // lanes share into the units the device reads.
+    const auto toReal = [&](float normalised) {
+        return magda::ParameterUtils::normalizedToReal(
+            juce::jlimit(0.0f, 1.0f, normalised + offset), spec.domain);
+    };
+
+    // A stepped parameter holds its value across a segment: there is nothing
+    // between two of its values for a ramp to pass through.
+    const auto write = [&](int index, int startSample, float startNormalised, float endNormalised) {
+        const float startValue = toReal(startNormalised);
+        slot[index] =
+            ParamSegment{startSample, startValue, stepped ? startValue : toReal(endNormalised)};
+    };
+
+    const auto& automation = sources.automation;
+
+    const int coverStart =
+        automation.empty() ? 0 : std::clamp(automation.front().startSample, 0, numSamples);
+    // Zero is the common case spelled shortest: an absolute lane covers the
+    // whole block, and only a lane that stops inside one says where.
+    const int coverEnd = automation.empty()
+                             ? 0
+                             : (sources.automationEnd <= 0
+                                    ? numSamples
+                                    : std::clamp(sources.automationEnd, coverStart, numSamples));
+
+    int covered = 0;
+    for (const auto& segment : automation) {
+        if (segment.startSample >= coverEnd)
+            break;
+        ++covered;
+    }
+
+    const bool coversFirstSample = covered > 0 && coverStart == 0;
+
+    // The value the block opens with, which is all a device that reads once for
+    // the block ever sees, and all there is to write when nothing moves.
+    const float openingValue = coversFirstSample ? automation.front().startValue : sources.base;
+
+    if (covered == 0 || !spec.segmentAccurate) {
+        write(0, 0, openingValue, openingValue);
+        out.setSegmentCount(param, 1);
+        return;
+    }
+
+    // The virtual list: the base before the lane, the lane, the base after it.
+    // Never built anywhere, because building it would allocate; walked instead,
+    // which is what the index arithmetic below is for.
+    const int leading = coverStart > 0 ? 1 : 0;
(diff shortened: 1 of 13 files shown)


magda/engine/param/ParamResolve.cpp
```diff
@@ -0,0 +1,130 @@
...
+        return magda::ParameterUtils::normalizedToReal(
+            juce::jlimit(0.0f, 1.0f, normalised + offset), spec.domain);
+    };
+
+    // A stepped parameter holds its value across a segment: there is nothing
+    // between two of its values for a ramp to pass through.
+    const auto write = [&](int index, int startSample, float startNormalised, float endNormalised) {
+        const float startValue = toReal(startNormalised);
+        slot[index] =
+            ParamSegment{startSample, startValue, stepped ? startValue : toReal(endNormalised)};
```
lucaromagnoli: **P2 — Preserve normalized interpolation through nonlinear scales**

This converts only the two endpoints and `ParamValues::valueAt()` then interpolates linearly between the resulting real-unit values. That commutes only for a linear domain with no interior clamp. For example, a logarithmic 20–20000 Hz parameter on a normalized 0→1 segment reads about 10,010 Hz at the midpoint here instead of `normalizedToReal(0.5) ≈ 632 Hz`; a positive modulation offset that reaches 1 partway through a ramp likewise misses the required clamp plateau. Please keep enough normalized-domain information to apply clamp + scale at the sampled position, or subdivide the resolved stream so the real-unit segments approximate those transformations correctly.

magda/engine/param/ParamResolve.cpp
```diff
@@ -0,0 +1,130 @@
...
+    int written = 0;
+    int index = 0;
+
+    // True when the segment just offered was the last one that fits and more
+    // were coming: it is written as a ramp to where the lane ends instead, and
+    // the rest are rolled into it.
+    const auto offer = [&](int startSample, float startNormalised, float endNormalised) {
+        const bool more = index < total - 1;
+        if (more && written == capacity - 1) {
+            write(written++, startSample, startNormalised, finalNormalised);
```
lucaromagnoli: **P2 — Preserve the final choice when a stepped lane exceeds its budget**

The overflow path relies on `finalNormalised` becoming this last segment's `endValue`, but `write()` deliberately replaces every stepped `endValue` with its start value. With capacity 2 and stepped choices 0→1→2, the second slot starts at choice 1, discards the requested final choice 2, and holds 1 through the block, violating the documented “never a wrong destination” guarantee. Please coalesce stepped lanes by retaining a final step to the actual destination (for example, reserve the last slot for it) rather than using the continuous-ramp fallback.

Issue #2120, opened 2026-08-19
Parameters slice 5: ADSR, random and the envelope follower


lucaromagnoli:
Slice 5 of #1891. Depends on slices 1, 2 and 4.

The three remaining modifier engines.

- ADSR: attack, decay, sustain and release with per-segment curvature, gated from MIDI, stage
  reported for the display.
- Random: stepped and noise, with shape, smoothing and step depth, sharing the rate and trigger
  fields with the LFO.
- Envelope follower: gain, attack, hold and release over its source's amplitude, with the
  pre-detection high-pass and low-pass that let it track part of the spectrum.

The follower is what needs the source the plan does not carry yet. `PlanCompiler.cpp` reports a
rack sidechain as feeding rack modulation and passes it over; this slice is where that diagnostic
goes away and the edge becomes real.

lucaromagnoli:
A timing requirement from the #2119 review, so it is not lost when the trigger and gate feeds get wired here.

The native engine resolves parameters once at the top of the block, then walks the ops. A monitor is an op, so a trigger it fires mid-block is only spent at the next block's resolve: a hit in block N produces the gap in block N+1. The fork's monitor pokes the live TE modifier mid-block, so when the source track is ordered before the destination the gap lands in block N itself; otherwise the fork is also N+1. The forced-zero latch (added in the #2119 review pass) makes the retrigger correct at block granularity, but the feed decides the latency.

Correction to the first version of this comment: detecting against the prior block's signal does not recover the fork's timing, it is the same N+1 as the latch. The accurate requirement splits by trigger type:

- MIDI triggers: block N's MIDI is known before the resolve, so run note detection at block top and the gap lands in block N. Same block as the fork's best case, deterministic, and achievable with no structural change.
- Audio triggers: block N's audio does not exist until the ops render it, so under resolve-at-top an audio retrigger is structurally one block behind the fork's best case. Either accept the deterministic one-block lag and document it, or resolve the modifier's targets after the source subgraph renders (split resolution). That is a design decision this slice has to make explicitly, not a scheduling detail.

Parity cases should assert the trigger-to-gap distance per trigger type, not just the gap itself.

Issue #2119, opened 2026-08-19
Parameters slice 4: the LFO


lucaromagnoli:
Slice 4 of #1891. Depends on slices 1 and 2.

The LFO, at parity with what `ModInfo` describes and `PluginManagerModifiers` drives today.

- Sine, triangle, square, saw, reverse saw, and the custom curve: `CurvePointData` with its
  tension and per-point curve type, and the MSEG loop region.
- Rate in Hz, or tempo synced off the transport snapshot. The persisted rate ordinal is pinned
  (`ModRateType`), so the sync divisions are a mapping rather than a decision.
- Phase offset, one-shot with its hold at the end value, and `invertOutput`, which the Sidechain
  device depends on: an inactive modifier outputs 0, so a level curve is flipped on the way out.
- Trigger modes and both retrigger paths, including skip-resync. The lessons in the fork patches
  are the specification here.
- Per-link depth and bipolarity stay where they are, on the MAGDA side, applied at the link by
  slice 2.

PR #2125, opened 2026-08-19
Publish a plan's parameters, and say what links to what (#2117)

### LINKED ISSUE #2117: Parameters slice 2: publication, addressing and the link graph

lucaromagnoli:
Closes #2117. Slice 2 of #1891.

#2116 built the lane a device reads and left nothing filling it. This is where the values come from, how a link finds the parameter it drives, and how both reach the audio thread.

## Addressing

The model addresses a parameter with a `ControlTarget`: a path through the track, rack and chain hierarchy. Every consumer in MAGDA speaks it, which makes it the right thing to translate from and the wrong thing to carry into a block, since it holds a vector of steps and says where a parameter is by describing the walk to it.

`ParamKey` is that address flattened, the way `OpKey` is a model location flattened for the plan: no heap, comparable, hashable, and it names the owner rather than the route. `paramKeyFor` is the one translation between the two, so a target that resolves differently in two places is not a thing that can happen.

## Publication

The table is compiled off the audio thread beside the op values, over the same model, against the same plan, and it travels **inside** `PlanValues` rather than beside it. A fader move and a knob move are then one publish, one fingerprint and one thing that can go out of step with a plan swap rather than two. It is a `shared_ptr`, so a fader move that changed no parameter republishes the pointer it already had.

The executor sizes its per-block room and its per-op device windows when the plan is prepared, and resolves the whole table at the top of every block, before any op runs. Both executors, through one call: the reference one and the parallel one have to agree about when a parameter is settled, and a second answer would be a race.

An end-to-end case publishes a project through `EngineSession` and asserts what the device saw on the audio thread: a macro at full, a link at half depth, a parameter reading 0 to 100, and 50 arriving inside `process()`.

## Links, and the order they resolve in

A link is what a macro or a modifier does to a parameter, with the depth and polarity MAGDA keeps per link. Sources resolve in a topological order baked with the table, so a macro driving a macro driving a device parameter is one pass rather than a fixed point, and the golden pins the order rather than the arithmetic.

A cycle is broken where it is found: the links inside it are dropped, with a diagnostic, and every parameter still resolves. A parameter is worth more than the link that made it impossible.

## What is worth a slot

Every scope in the model comes with sixteen macros whether or not anything is wired to them. A project of any size would spend most of its table on knobs with nothing behind them, so a macro is carried when it drives something or something drives it — the second half matters, because a macro a modifier writes to needs somewhere to be written.

Modifiers are sources with an identity and the model's own reading of their output, constant until the engines land in #2119 and #2120. Two things that are already invariants rather than implementation are applied here: an inactive modifier outputs nothing, and a curve drawn as a level is applied as its complement.

## Reported rather than dropped

Three things the model can ask for and this cannot do:

- **the mixer values**, which `PlanValues` resolves per op. Nothing can move them faster than a publish yet, so it costs nothing today; #2118 is where a lane makes them move per block, and the crossover is written into that issue.
- **the tempo**, which belongs to the transport.
- **a modifier's own parameters**, which arrive with the engines that define them. The rate's scale is the modifier's knowledge, and duplicating it here to be replaced in two slices is how two answers to one question start.

## Goldens

The dump joins the plan goldens as a third section (#2076), with a `parameter-links` fixture carrying the parameters, macros, modifier and macro-to-macro chain the graph fixtures have no reason to have. Existing goldens gain five lines each.

## Testing

`make test` green locally: 2719 cases, 588139 assertions, 0 failures. `make test-juce` green. Twenty new cases across `test_param_table.cpp` and the goldens.

### DIFF
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -644,10 +644,19 @@ PlanExecutor::BlockStart PlanExecutor::beginBlock(const PlanValues& values,
 }
 
 void PlanExecutor::resolveParameters(const PlanValues& values, int numSamples) {
-    if (!appliesValues(values) || values.params == nullptr) {
-        // Empty rather than stale: a device asking for a parameter gets a
-        // window of nothing, and nothing is what it had before a table was
-        // ever published.
+    // Two shapes have to match, and neither is something appliesValues can
+    // see: it compares the plan's fingerprint and its op count, and a link
+    // edit changes neither. A macro gaining its first link grows the table by
+    // a slot, and a link added to the parameter that already had the most
+    // widens the room one parameter's contributions are gathered in. Both
+    // arrive on a table that belongs to this plan and neither fits what was
+    // allocated for it.
+    //
+    // Empty rather than stale, and empty rather than partly applied: a device
+    // gets a window of nothing, which is what it had before a table was ever
+    // published, and the session escalates a publish of this shape into a
+    // structural one so the emptiness lasts a block rather than for ever.
+    if (!appliesValues(values) || values.params == nullptr || !fitsParameters(values)) {
         paramValues_.beginBlock(numSamples);
         return;
     }
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -223,7 +223,8 @@ void PlanExecutor::reset() {
 
 std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const PlanBindings& bindings,
                                                const RenderContext& context,
-                                               const PlanExecutor* previous) {
+                                               const PlanExecutor* previous,
+                                               const ParamTable* params) {
     reset();
 
     std::vector<std::string> messages;
@@ -245,6 +246,22 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
     midiSourceForOp_.assign(numOps, nullptr);
     meterForOp_.assign(numOps, nullptr);
     midiTapForOp_.assign(numOps, nullptr);
+    paramWindowForOp_.assign(numOps, ParamTable::DeviceWindow{});
+
+    // The parameters of the table this plan is published with (#2117). Sized
+    // here, on this thread, so the block that resolves them allocates nothing,
+    // and looked up here so the block that reads them hashes nothing. A plan
+    // published without a table renders devices with no parameters, which is
+    // what every render did before there was a table at all.
+    if (params != nullptr) {
+        paramValues_.prepare(params->size());
+        paramScratch_.assign(static_cast<std::size_t>(std::max(params->maxLinksPerParam, 0)),
+                             ModContribution{});
+
+        for (std::size_t i = 0; i < numOps; ++i)
+            if (plan.ops[i].kind == OpKind::Device)
+                paramWindowForOp_[i] = params->windowFor(plan.ops[i].key.deviceKey());
+    }
 
     const auto describe = [&plan](std::size_t index) {
         return "op " + std::to_string(index) + " (" + toString(plan.ops[index].kind) + " " +
@@ -626,9 +643,22 @@ PlanExecutor::BlockStart PlanExecutor::beginBlock(const PlanValues& values,
     return start;
 }
 
+void PlanExecutor::resolveParameters(const PlanValues& values, int numSamples) {
+    if (!appliesValues(values) || values.params == nullptr) {
+        // Empty rather than stale: a device asking for a parameter gets a
+        // window of nothing, and nothing is what it had before a table was
+        // ever published.
+        paramValues_.beginBlock(numSamples);
+        return;
+    }
+
+    resolveParams(*values.params, paramValues_, paramScratch_, numSamples);
+}
+
 void PlanExecutor::process(const PlanValues& values, const BlockInfo& requestedBlock,
                            juce::AudioBuffer<float>& output) {
     const auto start = beginBlock(values, requestedBlock, output);
+    resolveParameters(values, start.block.numSamples);
     if (!start.render)
         return;
 
@@ -791,11 +821,12 @@ void PlanExecutor::renderOp(OpId id, const OpValue& value, const BlockInfo& bloc
             // No parameters yet: the table a device reads is resolved against a
             // plan and published with it, which is #2117. Until then a device
             // is handed an empty window rather than a stale one.
+            const auto window = paramWindowForOp_[static_cast<std::size_t>(i)];
             DeviceBlock deviceBlock{.audio = audio,
                                     .midiIn = &midiIn(op.inputs[1]),
                                     .midiOut = deviceMidiOut,
                                     .sidechain = {},
-                                    .params = {},
+                                    .params = paramValues_.device(window.first, window.count),
                                     .block = block};
             if (op.inputs[2].valid())
                 deviceBlock.sidechain = audioIn(op.inputs[2], numSamples);
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -219,6 +219,10 @@ void PlanExecutor::reset() {
     meterForOp_.clear();
     boundMeterCount_ = 0;
     midiTapForOp_.clear();
+    paramWindowForOp_.clear();
+    paramScratch_.clear();
+    paramValues_.prepare(0);
+    paramLayout_ = 0;
 }
 
 std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const PlanBindings& bindings,
@@ -247,6 +251,7 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
     meterForOp_.assign(numOps, nullptr);
     midiTapForOp_.assign(numOps, nullptr);
     paramWindowForOp_.assign(numOps, ParamTable::DeviceWindow{});
+    paramLayout_ = 0;
 
     // The parameters of the table this plan is published with (#2117). Sized
     // here, on this thread, so the block that resolves them allocates nothing,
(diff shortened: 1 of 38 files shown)


lucaromagnoli (commented):
Read the whole slice. The addressing, the window arithmetic, the Kahn order and the golden all check out; I traced the parameter-links golden's order by hand and it is right. Four findings, two of them on one theme: publishValues() can deliver a table whose shape the executor was not prepared for, and both fallback paths handle that worse than their comments claim.

magda/engine/param/ParamResolve.cpp
```diff
@@ -147,4 +147,53 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
     out.setSegmentCount(param, written);
 }
 
+void resolveParams(const ParamTable& table, ResolvedParams& out, std::span<ModContribution> scratch,
+                   int numSamples) {
+    // Indexed by the same ParamId, so a size that disagrees is two things
+    // prepared against different plans. Refused whole: half a table of
+    // parameters is a project with some of its values from somewhere else.
+    if (table.size() != out.size())
```
lucaromagnoli, line 161: This early return happens before `beginBlock`, so a refused table leaves `out` holding the previous block's resolved values, and every device reads that frozen block until the next structural publish. The comment in `PlanExecutor::resolveParameters` promises "empty rather than stale" for exactly this case, and the test only sees empty because it calls `beginBlock` by hand before resolving. Moving `out.beginBlock(numSamples)` above the size check makes the refusal actually empty.

This path is hotter than a defensive check: `appliesValues` compares the plan fingerprint and op count only, and a link edit does not change the plan. A macro gaining its first link grows the table by a slot, so a same-plan `publishValues` carrying that table passes `appliesValues`, fails this size check, and the project's parameters silently freeze at the previous block's values.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -245,6 +246,22 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+    paramWindowForOp_.assign(numOps, ParamTable::DeviceWindow{});
+
+    // The parameters of the table this plan is published with (#2117). Sized
+    // here, on this thread, so the block that resolves them allocates nothing,
+    // and looked up here so the block that reads them hashes nothing. A plan
+    // published without a table renders devices with no parameters, which is
+    // what every render did before there was a table at all.
+    if (params != nullptr) {
+        paramValues_.prepare(params->size());
+        paramScratch_.assign(static_cast<std::size_t>(std::max(params->maxLinksPerParam, 0)),
```
lucaromagnoli, line 264: Same theme as the size check in `resolveParams`, second shape dimension: the scratch is sized from the table the plan was published with, but `publishValues` can legitimately deliver a same-plan table with a larger `maxLinksPerParam` (a link added to the parameter that already had the most). The resolver's `used >= scratch.size()` guard then drops the newest link at block time with no diagnostic, which contradicts the table's own "never a silent drop" rule.

Since neither shape change is visible to `appliesValues`, I would have `resolveParameters` check both (`params->size()` against `paramValues_.size()` and `maxLinksPerParam` against `paramScratch_.size()`) and fall back to the empty block, and have the session or the bridge escalate a shape-changing values publish to a full publish. The freeze then lasts until the escalation instead of being permanent, and no link is half-applied.

magda/engine/param/ParamTableCompiler.cpp
```diff
@@ -0,0 +1,476 @@
...
+        auto& links = perParam_[target];
+        const auto removed = std::remove_if(links.begin(), links.end(), [&](const ParamLink& link) {
+            return link.source.kind == ParamSourceRef::Kind::Parameter &&
+                   static_cast<std::size_t>(link.source.index) < count &&
+                   stuck[static_cast<std::size_t>(link.source.index)] != 0;
+        });
+
+        if (removed != links.end()) {
+            diagnose(toString(table_.keys[target]) +
+                     ": part of a modulation cycle; the links inside it are dropped");
```
lucaromagnoli: The stuck set is wider than the cycle: it is the cycle's members plus everything downstream of them, because a parameter waiting on a cycle member never reaches in-degree zero either. This pass then drops any link whose source is stuck, so a parameter that merely reads a cycle member, with no cycle through itself, loses that link and gets a "part of a modulation cycle" diagnostic it does not deserve. Once the cycle's own links are dropped its members resolve at their base, and the downstream link could have been honoured against that.

Breaking only the links inside strongly connected components of more than one node (or a self link), then re-running the Kahn pass over the survivors, keeps the collateral links and confines the diagnostic to the parameters actually in the cycle. Cycles are rare user error, so this is about the blast radius rather than the common case, but the current reading turns one bad link into a silent loss of every chain hanging off it.

magda/engine/exec/PlanValues.cpp
```diff
@@ -480,7 +481,22 @@ void applyLinearPanLaw(float gain, float pan, float& left, float& right) {
...
                                            const TrackInfo& master, PlanValues& values) {
-    return Resolver(plan, tracks, master).run(values);
+    auto messages = Resolver(plan, tracks, master).run(values);
+
+    // The parameters travel with the values (#2117), so one call resolves the
+    // model into everything the plan reads and the two cannot be published out
+    // of step with each other. Rebuilt whole rather than patched, the way the
+    // op values are: a fader move re-resolves a project's parameters as well,
+    // which is a vector and a topological sort off the audio thread.
+    auto params = std::make_shared<ParamTable>(compileParamTable(plan, tracks, master));
```
lucaromagnoli, line 491: Minor: the PlanValues doc says a fader move that changed no parameter republishes the pointer it already had, but this compiles a fresh table on every `resolvePlanValues` call, so every mixer-speed publish pays the full walk, the link resolution and the toposort, and the pointer is always new. The shared_ptr makes the reuse possible; nothing implements it yet. Fine to leave for a later slice, but worth either wiring the reuse (hand the previous table in and return it when the compile would be identical, which the fingerprint plus a model revision could decide) or softening the comment so it describes what happens.

lucaromagnoli (commented):
Re-checked the fix commit against all four findings. The resolve path now empties before refusing, and the test pins it. fitsParameters asks both shape questions in one place, both executors expose it, and the session escalates a shape-changing values publish into a structural one on the live plan, with the end-to-end test proving the relinked macro reaches the device on the next block. I walked the iterative Tarjan line by line: init on first visit is guarded by the edge counter, the counter increments before the child push so no reference into the work vector is read after a reallocation, back edges fold index rather than lowlink, and the component closes at the root. Running it over the consumers graph is sound because strongly connected components do not care about edge direction. The drop is confined to components of more than one node plus self links, the diagnostic names only actual members, and the fresh Kahn afterwards is what the downstream test pins. One cosmetic leftover, inline.

magda/engine/exec/PlanExecutor.hpp
```diff
@@ -347,6 +364,23 @@ class PlanExecutor {
...
+     *
+     * The other half of applicable, and the half a fingerprint cannot answer.
+     * A link edit changes no op and no plan, so a table that gained a
+     * parameter or a link travels on a values publish, matches the plan, and
+     * does not fit the room allocated for it. Asked here so the publisher and
+     * the block get one answer: the publisher escalates such a publish into a
+     * structural one, and the block that meets one anyway renders empty rather
+     * than stale.
+     */
+    bool fitsParameters(const PlanValues& values) const {
```
lucaromagnoli: Nit: fitsParameters landed between appliesValues' doc block and appliesValues itself, so the "Whether process() would apply @p values" comment now dangles above this function's own doc and appliesValues below is left undocumented. Moving the stray block down to sit on appliesValues restores both.

magda/engine/exec/PlanExecutor.hpp
```diff
@@ -347,6 +364,23 @@ class PlanExecutor {
...
+        return values.params == nullptr ||
+               (values.params->size() == paramValues_.size() &&
```
lucaromagnoli: **P2 — Include cached window layout in the shape check**

This accepts any table with the same total size and scratch width, but `paramWindowForOp_` was cached from the previous table's exact `deviceWindows` offsets. A same-plan model update can remove the only used macro in an earlier scope and add one in a later scope (or redistribute device parameter counts) while keeping both checked numbers unchanged; `publishValues()` then takes the fast path, and `renderOp()` indexes the new resolved array through stale windows, handing affected devices another parameter's slots. Please make the prepared shape include the key/window layout (for example a layout fingerprint), and require that to match here so such updates trigger the structural republish.

lucaromagnoli:
Re-checked the layout commit. The third shape dimension was a real miss in my earlier review: I asked for count and link width, and neither can see a same-size table whose parameters moved, which the cached per-op windows turn into devices reading their neighbour's slots. The layout fingerprint closes it correctly: FNV over the keys in order and nothing else, so values, links and the resolution order still travel at mixer speed, and it is stamped and required in the same prepare that caches the windows. An epoch prepared without a table holds layout 0, which no real table hashes to, so the first table to arrive escalates rather than fast-pathing onto empty windows. reset() clearing the param state also closes the stale-size leftover from a prepare without a table. The two-track test pins the failure mode end to end, and the appliesValues doc is back where it belongs. Nothing further from me.

Issue #1087, opened 2026-04-25
Clip-based automation: UI to author lanes inside clips


lucaromagnoli:
## Motivation
The data model already supports clip-based automation lanes (`AutomationLaneInfo::isClipBased()`, the clip-based branch in `AutomationPlaybackEngine::bakeLane`). What's missing is the UI to actually author and edit lanes that live inside a clip and move with it.

Track-based (absolute) automation is the day-one feature; clip-based is advanced — not all DAWs have it (Bitwig and Live's per-clip envelopes do, REAPER's per-take envelopes do).

## Scope
- UI affordance to "convert this lane to clip-based" or "add automation inside this clip"
- Lane editing inside the clip editor (piano roll / waveform area), not on the track timeline
- Move/copy/paste/duplicate clip carries the lane with it
- Per-clip lane visibility toggle in the clip editor

## What already exists
- `AutomationLaneInfo` has both absolute and clip-based modes
- `AutomationPlaybackEngine::bakeLane` reads from `lane.clipIds` and bakes the union range
- `AutomationManager::getValueAtTime` resolves clip-based lanes via stored points

## Out of scope
- Linked vs independent clip lanes (same clip vs duplicate-with-its-own-lane) — separate decision
- MIDI CC as clip-based automation — separate flow

PR #2126, opened 2026-08-19
Play a lane over a parameter, and over a fader (#2118)

### LINKED ISSUE #2118: Parameters slice 3: automation bake

lucaromagnoli:
Closes #2118. Slice 3 of #1891.

#2116 built the lane, #2117 filled it with stored values and links. This is the third writer: the curve.

## One curve, read in three places

The shape of a lane was inside `AutomationManager`, which the engine cannot reach. It moves to `core/AutomationCurve.hpp` as two pure functions: the value of a breakpoint list at a beat, and the value of a whole lane at one, clips and loops and gaps included. The manager delegates to them, so the editor drawing a curve, the DAW playing it and the engine baking it are one function rather than three readings that agree until they do not.

Nothing about the curve itself is ported. Bezier through the parametric cubic solved for the queried beat, linear through the shared `curvemath::evalSegment`, step held, hard corner as two runs meeting at its apex: all of it is the model's, moved rather than rewritten.

## Beats in the table, samples in the block

A lane rides in the parameter table beside the base and the links, as breakpoints in beats. Not baked into segments at publish time, because a block is what decides which stretch of a curve is being asked about and how long that stretch lasts: a curve resolved into samples by the publisher would have to be republished every time the tempo moved.

A clip-based lane arrives already flattened by the model's own flattener (#1087), loops unrolled and gaps held at the nearest clip edge. The engine plays what MAGDA plays rather than forming a second opinion about what a looping clip means.

Which lanes play is the model's decision and is asked rather than second-guessed: read mode is playback, and disabled, touching and writing are all the user's hand on the parameter, where the base is what a parameter is worth.

## What a block reads

One value, held, unless the device asked for more. That is the port's shape everywhere: the incumbent engine settles a parameter at the block boundary, and a curve read more finely than that would differ from it on every automated parameter in every project.

A parameter that opted into segment accuracy gets a segment per breakpoint the block contains. The bend between two breakpoints reads as a straight line inside a block; over a few milliseconds that is a difference no device has asked to hear, and subdividing is what to do when one does.

## The mixer values cross over

As the issue said they would, and this is the slice that needs them: a lane over a fader moves it inside a block, which a table resolved by the publisher cannot do.

A track's volume, its pan and its send levels become parameters **when something reaches them**. The fader op then reads its gain from the block rather than from the value table, through the same TE curve `PlanValues` has always used (`faderGainFromDecibels`, then the linear pan law through the same deadband). Mute and solo stay where they are: nothing plays a lane over them, and a fader that moved does not make a muted track audible.

When nothing reaches them they are not carried at all. A project with no automation on its faders allocates nothing, renders through the value table exactly as it did before, and the null-diff corpus goes on comparing the same arithmetic it did yesterday. That is deliberate: the crossover is opt-in by use, not a rewrite of every render.

## Testing

`make test` green locally: 2735 cases, 588729 assertions, 0 failures. `make test-juce` green.

Twenty-seven new cases in `test_automation_bake.cpp`: a lane replacing a base, modulation adding to a lane, a lane over a macro driving what the macro drives, a stepped parameter quantised by its own scale rather than by the curve, a looping clip lane holding in its gaps, the block-rate default, the segment-accurate opt-in, and a fader lane ramping from silence to unity through a live `EngineSession`, read back off the master's output.

The `parameter-links` golden now carries a device lane and a fader lane, so the table's third writer is pinned as text alongside the first two.

### DIFF
--- a/magda/daw/core/AutomationManager.cpp
+++ b/magda/daw/core/AutomationManager.cpp
@@ -4,6 +4,7 @@
 #include <cmath>
 #include <iterator>
 
+#include "AutomationCurve.hpp"
 #include "ClipLaneFlattener.hpp"
 #include "CurveMath.hpp"
 #include "GridDivision.hpp"
@@ -1101,41 +1102,8 @@ double AutomationManager::getValueAtBeat(AutomationLaneId laneId, double beatPos
     if (!lane)
         return 0.5;
 
-    if (lane->isAbsolute()) {
-        return interpolatePoints(lane->absolutePoints, beatPosition);
-    }
-
-    // Clip-based: find clip containing beat position
-    for (auto clipId : lane->clipIds) {
-        const auto* clip = getClip(clipId);
-        if (clip && clip->containsBeat(beatPosition)) {
-            double localBeatPosition = clip->getLocalBeat(beatPosition);
-            return interpolatePoints(clip->points, localBeatPosition);
-        }
-    }
-
-    // Gap between clips: hold the nearest clip edge — the previous clip's
-    // final value, or before the first clip, the first clip's initial value.
-    const AutomationClipInfo* prevClip = nullptr;  // greatest end <= beat
-    const AutomationClipInfo* nextClip = nullptr;  // smallest start > beat
-    for (auto clipId : lane->clipIds) {
-        const auto* clip = getClip(clipId);
-        if (!clip)
-            continue;
-        if (clip->getEndBeats() <= beatPosition) {
-            if (!prevClip || clip->getEndBeats() > prevClip->getEndBeats())
-                prevClip = clip;
-        } else if (clip->startBeats > beatPosition) {
-            if (!nextClip || clip->startBeats < nextClip->startBeats)
-                nextClip = clip;
-        }
-    }
-    if (prevClip)
-        return interpolatePoints(prevClip->points, prevClip->getEndLocalBeat());
-    if (nextClip)
-        return interpolatePoints(nextClip->points, 0.0);
-
-    return 0.5;  // Lane has no clips at all
+    return automation::laneValueAtBeat(
+        *lane, [this](AutomationClipId clipId) { return getClip(clipId); }, beatPosition);
 }
 
 double AutomationManager::getClipValueAtBeat(AutomationClipId clipId,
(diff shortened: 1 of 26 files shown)


magda/engine/param/ParamResolve.cpp
```diff
@@ -147,8 +149,61 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
...
+
+    // A segment per breakpoint inside the block, from where the block opens to
+    // where it ends. The bend between two breakpoints reads as a straight line
+    // here; over a few milliseconds that is a difference no device has asked to
+    // hear, and subdividing is what to do when one does.
+    int written = 0;
+    int startSample = 0;
+    float startValue = opening;
+
+    for (const auto& point : curve) {
```
lucaromagnoli: The per-breakpoint bake linearizes the shape between breakpoints using their endpoint values, which the header honestly documents for a bend. But Step and HardCorner are not bends. A step segment whose closing breakpoint falls inside a segment-accurate block is baked as a ramp from the held value to the next one, so a device hears a glide through values the drawn curve never has; a hard corner's apex between two breakpoints disappears entirely. The closing segment gets this right, because its end samples valueAt, which respects the curve type; it is only the intra-block breakpoint segments that lose the shape.

Dormant, since nothing opts into segmentAccurate during the port, but the first parameter that does inherits it silently and the difference is exactly the kind the null-diff corpus cannot see (nothing opted in there either). Cheap fix while it is fresh: for a Step opening point write the flat segment and let the jump land at the breakpoint, which ParamSegment already expresses, and for a HardCorner emit the apex as one extra knot. Bezier and tension can stay endpoint-linearized as documented.

magda/daw/core/AutomationManager.cpp
```diff
@@ -1153,117 +1121,16 @@ double AutomationManager::interpolateLinear(double t, double v1, double v2) cons
...
-// drawing and the modulator engine outputs — so the played value sits exactly
-// on the drawn curve. (The old local quadratic/power copies here had drifted
-// from the renderer.) t is the x-fraction along the segment.
-static double evalLinearSegment(double t, const AutomationPoint& p1, const AutomationPoint& p2) {
-    const bool hasShaper = !p1.outHandle.isZero() || !p2.inHandle.isZero();
-    const double controlY = p1.value + p1.outHandle.value;
-    return static_cast<double>(curvemath::evalSegment(
-        static_cast<float>(p1.value), static_cast<float>(p2.value), static_cast<float>(controlY),
-        static_cast<float>(p1.tension), hasShaper, static_cast<float>(t)));
+    // Kept as a member because callers outside this file ask for it; the curve
```
lucaromagnoli: This comment says callers outside this file ask for interpolateBezier, but grepping the tree finds none: the only references left are this definition and the header declaration, and interpolateLinear is in the same position. Since the shim also changed semantics (it now dispatches on p1.curveType instead of forcing the bezier evaluation the name promises) and allocates a vector per call, I would delete both members rather than keep them. If something does still call it that I could not find, the dispatch change is worth a second look for that caller.

magda/engine/param/ParamResolve.cpp
```diff
@@ -147,8 +149,61 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
     out.setSegmentCount(param, written);
 }
 
-void resolveParams(const ParamTable& table, ResolvedParams& out, std::span<ModContribution> scratch,
-                   int numSamples) {
+int bakeCurve(std::span<const magda::AutomationPoint> curve, const ParamSpec& spec,
```
lucaromagnoli, line 152: Scaling note rather than a bug: bakeCurve walks the curve from its first point every block, and valueAt does the same inside valueAtBeat, so an automated parameter costs O(lane length) per block on the audio thread. An unrolled looping clip lane over a long arrangement is thousands of points, and the walk repeats for every automated parameter at several hundred blocks a second, almost all of it skipping points behind the playhead. The points are in beat order, so a lower_bound on beatPosition in both places makes it O(log n) plus the points actually inside the block. Fine to leave until a real project shows it, but the shape of the cost is worth knowing about since it grows with arrangement length, not with block size.

lucaromagnoli (commented):
Read the whole slice. The parts I was most suspicious of all check out: the fader conversion chains agree between the publish path and the block path (faderGainFromVolume is faderGainFromDecibels of gainToDecibels, which is exactly what the table's normalizedFromGain plus the FaderDB round trip reproduces, for sends as well as volumes), the parallel executor renders through core_.renderOp so the mixer override covers both executors, mute and solo stay outside the override because they are separate ops, the capacity arithmetic in bakeCurve reserves the closing slot correctly, and I re-derived the regenerated golden's resolution order by hand. Three findings, none blocking: one contract bug that is dormant until the first segment-accurate opt-in, one piece of dead code with a comment that claims otherwise, one audio-thread scaling note.

magda/engine/param/ParamResolve.cpp
```diff
@@ -147,8 +149,88 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
...
+    };
+
+    const auto inside = [&block](double beat) {
+        return beat > block.startBeat && beat < block.endBeat;
+    };
+
+    // Every knot the block contains, in beat order: each breakpoint, and the
+    // apex of a hard corner, which is where that curve changes direction
+    // between two breakpoints rather than at one.
+    for (std::size_t i = 0; i < curve.size(); ++i) {
```
lucaromagnoli: The commit message says the bake walks only the knots the block contains, but this loop still starts at index 0 every block: everything behind the playhead is iterated just to fail the inside() check, with hardCornerOf called per pre-block pair along the way. valueAtBeat got the binary search; this walk did not, so the per-block cost is still O(lane length) and grows with arrangement position. Since the points are in beat order, starting i at the lower_bound of block.startBeat minus one (clamped to zero, so the corner of the segment the block opens inside is still seen) makes it O(log n) plus the knots actually in the block, which is what the message describes.

lucaromagnoli (commented):
Re-checked the fix commit. The step and hard-corner semantics are right: I traced both new tests by hand, the hold-versus-travel decision reads the curve type through the same segmentOpening the evaluator uses, a step's jump lands exactly on its knot sample the way ParamSegment expresses a step, same-sample knots coalesce by advancing the run's value without writing an empty segment, and an apex belonging to a breakpoint before the block still gets closed because the loop checks corners of pre-block pairs. The binary-searched valueAtBeat is behaviorally identical to the old walk including the duplicate-beat and boundary cases, and the dead shim is gone. One piece of the third finding did not land, inline.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -684,11 +711,39 @@ void PlanExecutor::process(const PlanValues& values, const BlockInfo& requestedB
                  output);
 }
 
-void PlanExecutor::renderOp(OpId id, const OpValue& value, const BlockInfo& block,
+OpValue PlanExecutor::mixerValueFor(std::size_t op, const OpValue& published) const {
+    const auto params = mixerParamForOp_[op];
+    if (params.gain == INVALID_PARAM_ID)
```
lucaromagnoli, line 716: **P1 — Carry fader volume and pan as a pair**

`allocateMixer` registers volume and pan independently, but this path only applies a mixer override when gain exists and substitutes centered pan when pan does not. As a result, a pan-only lane/link returns `published` here and never runs, while a volume-only lane/link on an already-panned track sets both channels to the same gain and recenters it. Please make either targeted fader parameter carry both base values through the table (or otherwise preserve the untouched component), and cover both pan-only and volume-on-panned cases.

magda/engine/param/ParamResolve.cpp
```diff
@@ -147,8 +149,88 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
...
+        if (curve[i].beatPosition >= block.endBeat)
+            break;
+        if (written + 1 >= static_cast<int>(out.size()))
```
lucaromagnoli: **P2 — Preserve the final step when the bake arena fills**

For a segment-accurate stepped curve with more knots than `out`, this break leaves `holds` true at the last accepted knot. The unconditional final segment at line 225 then holds that old value through the block end, dropping every later step—including the curve's destination. That bypasses `ResolvedParams`' stepped-overflow policy of spending its last slot on the ending step. Please reserve/coarsen the final slot to the end value and add a regression with more than 16 in-block steps.

magda/engine/param/ParamResolve.cpp
```diff
@@ -147,8 +149,127 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
...
+    // never has.
+    const auto holdsFrom = [&curve](double beat) {
+        const auto* opener = magda::automation::segmentOpening(curve, beat);
+        return opener != nullptr && opener->curveType == magda::AutomationCurveType::Step;
+    };
+
+    // Where the curve ends up over this block, and where it gets there. Both
+    // are needed before the walk, because a walk that runs out of room has to
+    // spend its last segment on arriving rather than on where it had got to.
+    const float ending = valueAt(block.endBeat);
```
lucaromagnoli, line 191: **P2 — Keep a step at `endBeat` in the next block**

When the bake overflows and a step lies exactly at `block.endBeat`, these two values describe different knots: `valueAt(block.endBeat)` returns the boundary step's new value, while `lastKnot - 1` is the previous in-block knot. The `overflowed && holds` branch therefore writes that next-block value at the previous knot's sample, pulling the step earlier into this block. The normal non-overflow path correctly holds the old value because the end boundary is exclusive. Please derive the stepped overflow destination from the last knot strictly inside the block (and add a >16-step case whose final point is exactly `endBeat`).

lucaromagnoli:
Re-checked the third commit. All three review findings are closed, and the volume/pan coupling is a genuine catch my review missed: I verified mixerValueFor against the old allocation and both failure directions were real, a pan lane alone never ran the override, and a volume lane alone recentred a panned track through the pan.empty() branch. The together-or-not-at-all allocation closes both, the send case still correctly has no pan, and the parity math holds when only one of the two is automated because the base round trip reproduces the published gain exactly. The bake now starts at the run the block opens inside; I walked the lower_bound edges (block at a knot, block past the lane, block before it) and they all land right, with the pre-block opener's corner still considered and its own knot correctly outside. The overflow arrival fix is sound: the spurious-overflow paths cannot trigger with any real capacity, the final flat segment reads correctly even from samples before its start because both its values are equal, and endingSample is guarded by max() against the one case where the last knot sits behind the block. The regenerated golden's order re-derives correctly with the pan slot in. Nothing further.

magda/engine/param/ParamResolve.cpp
```diff
@@ -147,8 +149,135 @@ void resolveParam(ResolvedParams& out, int param, const ParamSpec& spec,
...
+            }
+    }
+
+    // A run that holds cannot arrive by ramping, so a bake that ran out of room
+    // inside one spends its last segment on the step the curve ends on, at the
+    // sample it lands. What is lost is the steps in between, which is the same
+    // coarsening ResolvedParams makes when a curve outgrows its slot; the run
+    // before it simply holds a little longer. A run that does not hold ramps to
+    // the same place, which arrives on time whether or not it overflowed.
+    if (overflowed && holds && hasKnotInside)
```
lucaromagnoli: **P2 — Choose overflow shape from the final run**

`holds` describes the run where the walk exhausted its slots, not the run containing `block.endBeat`. Curve type is per point, so a skipped later knot can change that: if overflow happens in a Step and the last in-block knot opens a Linear/Bezier run, this branch holds `endingHeld` instead of ramping to the boundary value; in the reverse direction the `else` branch ramps through what should be a final step. Please derive the last-slot shape from the final run (and, for a travelling final run, preserve its start value at `endingSample`) and add a mixed Step/Linear overflow regression.

PR #2127, opened 2026-08-20
Make the LFO turn (#2119)

### LINKED ISSUE #2119: Parameters slice 4: the LFO

lucaromagnoli:
Slice 4 of #1891. Closes #2119.

A modifier has been a source with an identity and a value the model last saw,
which is a value that does not move between publishes. This is the engine that
moves it.

## The shape stays the model's

On the terms slice 3 took the automation curve: the modulator's cycle moves out
into `core/ModCurve.hpp`, and the editor's preview, the fork's lock-free audio
snapshot and the native LFO all read one function rather than three copies of
one. `ModulatorEngine` and `CurveSnapshot` become callers.

The Rate lane's two descriptions move into `ParameterPresets` for the same
reason. A lane read back through a second copy of its own numbers is a lane the
engine and the editor quietly disagree about.

## The run is the engine's

Six waveforms and the drawn curve with its tension, its per-point kind and its
sustain loop. A rate in Hz, or a musical division that is a fraction of a bar
and therefore three beats long in three four. Phase offset, one-shot holding
what the cycle ended on, and `invertOutput`, without which an idle Sidechain
reads as silence rather than as no attenuation.

Trigger modes and both retrigger paths, including the skip-resync a cross-track
LFO needs so the track it ducks stops retriggering it.

Block rate, from the block's first sample, which is where the fork settles a
modifier and therefore what a parity case is measured against.

## Where the phase lives

The settings travel in the published table, which is republished every time a
knob moves. The phase stays in the executor, keyed by the modifier's own address
and shared with the epoch being replaced rather than copied out of it, the way a
delay line is. An LFO goes on turning through a knob turn, a link edit and a
device insert somewhere else in the project.

## Modifiers join the resolution order

They depend on parameters in both directions: a link reads a modifier, and a
modifier reads the Rate parameter driving it. One order over both kinds of node
means a macro driving an LFO's rate while the LFO drives a cutoff is one pass,
with no fixed point, and a cycle through a modifier's rate is broken where it is
found. That is the diagnostic slice 2 left pointing here.

`modifierFingerprint` protects the runtime's indexing the way `layoutFingerprint`
protects the parameters': a scope gaining a page of modifiers moves every one
after it without touching a key, and a values publish of that shape escalates
into a structural one rather than handing one LFO's phase to another.

## What is not wired, and why

Two of the three trigger modes have nothing feeding them yet. A note-triggered
modifier needs the MIDI reaching its scope; an audio-triggered one needs a level
detector. Neither exists in the native engine, so the trigger calls are the
shape those feeds will use and are exercised directly rather than end to end.

The transport mode needs no feed and is wired: a timeline-locked modifier is a
function of where the block is. The rest arrive with the slice that cannot avoid
the same problem for the ADSR's gate (#2120).

## One defect fixed on the way

A modifier the model had switched off carried its links with an output of zero,
and a bipolar link reading zero pushes its parameter down by the link's own
depth: a switched-off modifier doing something. The fork creates no modifier for
one at all, so now neither does the table.

## Testing

26 cases in `tests/test_mod_lfo.cpp`, from the shape at a phase through to an
LFO moving a device parameter out of the master of a live session and going on
turning across a structural republish. Full suite green (588,461 assertions),
app builds.

### DIFF
--- a/magda/engine/param/ModLfo.cpp
+++ b/magda/engine/param/ModLfo.cpp
@@ -151,9 +151,18 @@ void restartLfo(LfoState& state, const LfoSettings& settings) {
 float advanceLfo(LfoState& state, const LfoSettings& settings,
                  std::span<const magda::CurvePointData> curve, const BlockInfo& block,
                  const ModTiming& timing) {
-    if (!state.started) {
+    // A fresh state takes its gate from the settings, and so does one whose
+    // trigger mode has changed underneath it: the gate belongs to the triggers
+    // of the mode that opened it, and a mode change retires all of them. An
+    // audio-triggered LFO sits shut between hits, and switching it to free
+    // running would otherwise leave it shut for ever, because nothing in the
+    // new mode ever opens a gate. The held-note count goes with it, for the
+    // same reason.
+    if (!state.started || state.trigger != settings.trigger) {
         state.started = true;
+        state.trigger = settings.trigger;
         state.gated = settings.startGated;
+        state.heldNotes = 0;
     }
 
     // A latch belongs to the setting that made it. Turning one-shot off clears
--- a/magda/engine/param/ModLfo.cpp
+++ b/magda/engine/param/ModLfo.cpp
@@ -100,8 +100,12 @@ double barFractionOf(int rateType) {
     return 1.0;
 }
 
-double cycleBeats(int rateType, int numerator) {
-    return barFractionOf(rateType) * std::max(numerator, 1);
+double cycleBeats(int rateType, int numerator, int denominator) {
+    // A bar in quarter notes, which is what a beat is here. The signature's
+    // own note value is the denominator's, so six eight is three quarter notes
+    // rather than six.
+    const auto barBeats = 4.0 * std::max(numerator, 1) / std::max(denominator, 1);
+    return barFractionOf(rateType) * barBeats;
 }
 
 int rateTypeFromLaneValue(float laneValue) {
@@ -123,13 +127,16 @@ ModTiming modTimingFor(const BlockInfo& block, double sampleRate) {
     // what it gets is what a session that has never seen a transport renders
     // at: 120 in four four, which is the same default TransportState holds.
     if (block.tempo != nullptr) {
+        const auto signature = block.tempo->barsAndBeatsAt(block.startBeat);
         timing.bpm = block.tempo->bpmAt(block.startBeat);
-        timing.numerator = block.tempo->barsAndBeatsAt(block.startBeat).numerator;
+        timing.numerator = signature.numerator;
+        timing.denominator = signature.denominator;
     }
 
     if (!(timing.bpm > 0.0))
         timing.bpm = 120.0;
     timing.numerator = std::max(timing.numerator, 1);
+    timing.denominator = std::max(timing.denominator, 1);
 
     return timing;
 }
@@ -174,7 +181,7 @@ float advanceLfo(LfoState& state, const LfoSettings& settings,
 
     const double hz = std::max(static_cast<double>(settings.rate.hz), kMinHz);
     const double beatsPerCycle =
-        std::max(cycleBeats(settings.rate.rateType, timing.numerator), 1.0e-6);
+        std::max(cycleBeats(settings.rate.rateType, timing.numerator, timing.denominator), 1.0e-6);
 
     // A timeline-locked LFO is a function of where the block is rather than of
     // how many blocks have gone by, which is what puts two of them at one rate
--- a/magda/engine/param/ModLfo.cpp
+++ b/magda/engine/param/ModLfo.cpp
@@ -215,13 +215,24 @@ float advanceLfo(LfoState& state, const LfoSettings& settings,
     if (state.gated)
         value = 0.0f;
 
+    // A trigger asked for one block of nothing, and this is the first block a
+    // device can see (LfoState::forceZero). The phase is published where the
+    // restart put it, so the editor's dot is already back at the top, and it
+    // is not advanced, so the shape resumes from its start on the next block
+    // rather than a block into itself.
+    const bool zeroed = state.forceZero;
+    if (zeroed) {
+        state.forceZero = false;
+        value = 0.0f;
+    }
+
     state.phase = phase;
     state.value = value;
 
     // Moved on for the next block, after this one's value has been settled,
     // which is the order the fork's timer uses: the value a block renders with
     // is the value at its first sample.
-    if (settings.sync != ModSync::Transport && !holding) {
+    if (settings.sync != ModSync::Transport && !holding && !zeroed) {
         const double blockSeconds =
             static_cast<double>(std::max(block.numSamples, 0)) / timing.sampleRate;
         const double periodSeconds =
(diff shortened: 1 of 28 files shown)


magda/engine/param/ModRuntime.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if (state == nullptr || index >= static_cast<int>(table.modifiers.size()))
+        return;
+
+    const auto& settings = table.modifiers[static_cast<std::size_t>(index)].lfo;
+
+    // A cross-track sidechain LFO follows its source track and nothing else.
+    // Retriggering it from the track it ducks is the bug the fork's flag
+    // exists to prevent: the phase restarts against the sidechain rhythm and
+    // the gate flaps on every note the destination happens to play.
+    if (settings.skipNativeResync && !settings.gateOnTrigger)
```
lucaromagnoli: With skipNativeResync and gateOnTrigger both set, a note from the track the modifier lives on falls through this guard and opens the gate (and counts a held note). The fork's flag means "driven externally, skips its own MIDI" and suppresses the whole own-track path, gate included; the header doc above noteOn says the same ("refused when the modifier belongs to another track's monitor"). The combination is unreachable until the trigger feeds are wired, but since this function is the shape those feeds will use, I'd make the guard just `if (settings.skipNativeResync) return;` so the destination track's notes can never touch a cross-track LFO's gate.

magda/engine/param/ParamTable.cpp
```diff
@@ -28,7 +28,34 @@ std::uint64_t paramLayoutFingerprint(const std::vector<ParamKey>& keys) {
...
+    Fingerprint hash;
+    hash.mix(keys.size());
+
+    for (const auto& key : keys)
+        hash.mix(key);
+
+    return hash.value();
+}
+
+std::uint64_t paramModifierFingerprint(const std::vector<ParamModifier>& modifiers) {
```
lucaromagnoli, line 51: This hashes only the keys, and TrackManager::setModType flips a modifier's type in place at the same address. So a kind change travels as a values publish: fitsParameters passes, the runtime is never re-prepared, and its recorded kinds go stale. Nothing breaks today because advance() reads the kind off the table, but switching away and back resumes the old phase where the fork restarts (it recreates the TE modifier), and when #2120 adds per-kind state the prepare-time kind check only protects structural publishes. Mixing modifier.kind into the hash here makes a type switch structural by construction, for one line.

magda/engine/param/ModRuntime.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    state->lfo.gated = false;
+    state->lfo.started = true;
+    restartLfo(state->lfo, settings);
+
+    // The forced zero stands in for the gap a gated retrigger leaves, so the
+    // device sees a transition rather than a continuation. Never for a level
+    // curve: there, zero output is full level, and forcing one in the middle
+    // of a duck pops the gain up for a block, which is a click on every hit.
+    if (forceZero && !settings.invertOutput) {
```
lucaromagnoli: I don't think this forced zero can ever reach a device. resolveParams reads modifier outputs at the top of the block, and a monitor firing trigger() does so mid-block, after that read; by the next block's read, advance() has already overwritten values_ with the restarted shape. In the fork the zero persists on the TE modifier until its next timer tick, which is what makes the gap audible. For shapes that do not start at zero (a sine opens at 0.5) the gap will silently not happen once the feeds are wired. A forcedZero latch on LfoState that the next advanceLfo consumes and publishes for that one block would reproduce the fork's behaviour. Fine to leave for the slice that wires the feeds, but worth a note so it is not lost.

lucaromagnoli (commented):
Read the whole slice against the fork: the sync fold matches ModifierHelpers::mapSyncType, barFractionOf matches TE's ModifierCommon table with the ordinals lining up name for name, the loop-region and one-shot semantics match CurveSnapshotHolder::evaluateCallback (including the latch surviving a rewind, which the fork's positive-delta integration also produces, and the offset-invariant loop position), and the dedup into core/ModCurve.hpp is faithful to all three copies it replaces. The golden's resolution order is right; I walked it by hand. Three minor findings inline, none of them audible today.

third_party/tracktion_engine
```diff
@@ -1 +1 @@
-Subproject commit c8724e69d74282662eb219fd8fa7533d4468e3d7
+Subproject commit 77fd45613abe5ef1708970ec29705a26592c6b7c
```
lucaromagnoli: **P1 — Do not apply the denominator twice in Tracktion transport sync**

The referenced commit changes the transport branches to divide `tempoSequence.getBeats()` by `4 * numerator / denominator`. That beat count already uses the signature note value because `TempoSequence` is configured through `lengthOfOneBeatDependsOnTimeSignature()` (and `MagdaEngineBehaviour` keeps the default `true`). In 6/8, one written bar therefore reaches 6 here; dividing by 3 makes the LFO, Random, Step, and Breakpoint modifiers run two cycles per bar. The duration branch does need the denominator correction, but transport should derive its position from `tempoSequence.getBarsBeats().getTotalBars()`, which also preserves the bar origin across meter changes.

magda/engine/param/ModLfo.cpp
```diff
@@ -0,0 +1,275 @@
...
+    const double hz = std::max(static_cast<double>(settings.rate.hz), kMinHz);
+    const double beatsPerCycle =
+        std::max(cycleBeats(settings.rate.rateType, timing.numerator, timing.denominator), 1.0e-6);
+
+    // A timeline-locked LFO is a function of where the block is rather than of
+    // how many blocks have gone by, which is what puts two of them at one rate
+    // in phase with each other and with the bar however playback got there.
+    if (settings.sync == ModSync::Transport) {
+        state.cycles =
+            settings.tempoSync ? block.startBeat / beatsPerCycle : block.startSeconds * hz;
```
lucaromagnoli: **P2 — Anchor transport phase to the bar grid, not beat zero**

This still loses the bar origin when the signature changes. For example, after two 4/4 bars a change to 6/8 at beat 8 starts a new bar (`TempoMap` explicitly guarantees that), but this computes `8 / 3`, so a one-bar LFO opens at phase 2/3 and keeps that offset on every following 6/8 bar. Use `block.tempo->barsAndBeatsAt(block.startBeat)` to form the continuous bar position (`bar + beat / numerator`) and divide that by the requested bar fraction; that preserves both denominator-aware length and signature-change alignment.

PR #2129, opened 2026-08-20
The last three modifiers, and what feeds them (#2120)

### LINKED ISSUE #2120: Parameters slice 5: ADSR, random and the envelope follower

lucaromagnoli:
Slice 5 of #1891. Closes #2120.

The three remaining modifier engines, on the terms slice 4 set for the LFO:
settings in the published table, state in the runtime carried by the modifier's
own address, one block at a time read from the block's first sample.

## The envelope

The fork's stage machine, curvature included. A segment's bend is a persisted
number between -0.5 and 0.5, so the quadratic bezier behind it is transcribed
rather than reasoned out again.

Two things differ from the LFO and both are the fork's rather than choices made
here. An envelope is nothing but a gate, so where the gate comes from is what
the sync mode selects, and tempo sync stays out of that fold because it scales
the stages rather than deciding them. And the fork's ADSR timer advances before
it publishes, where its LFO timer publishes before it advances: the two are one
block apart and the second is what a parity case is measured against.

## The walk

The fork's too, one difference aside. Its type parameter is written and never
read, so a project set to noise sounds stepped there. Implemented here rather
than reproduced as the same omission, because a control that does nothing is
worse than a difference that is written down.

The generator is seeded from the modifier's own address, which the fork cannot
promise: it seeds from the clock, so two of its modifiers can walk together and
a project draws different numbers on its second run. A project that renders
differently twice cannot be null-diffed against anything, itself included.

## The follower

The one modifier that is not a function of time, and the reason this slice
exists. Its input is audio the modifier does not own, so the plan has to carry
it. The detector is the fork's: per-follower input gain, then the band limits on
JUCE's own coefficients so both engines listen to the same part of the spectrum,
then the block's peak. The envelope is the fork's one-pole attack, hold and
release over that peak.

## The edge becomes real

A modifier listens to the sidechain source of the scope it lives on, and to the
track it lives on when that scope has none. One rule, in one header, because two
compilers need the answer: the plan compiler to emit the edge and the table
compiler to record which modifier is on the far end of it. A rule stated twice
is a rule that will eventually be two rules, and the failure would be silent.

The plan gains one op per listened-to track, a second sink fed the source's
post-fader tap and its chain-head MIDI. The rack sidechain `PlanCompiler` used
to report as unrepresentable is now an ordering constraint and an op, and the
diagnostic is gone.

## Where the block hears it

Parameters resolve at the top of a block, so a trigger fired by an op is not
spent until the next block's resolve. That is fine for a level and not for a
note, and the two are settled differently.

A note is known before the block resolves: the ops that produce MIDI read clips
and input queues rather than audio, so the executor renders that prefix first
and a note-gated modifier is resolved with the note that arrived in this block.
The fork's ordinary gate is a block later than that and only its monitor path is
this early, so this is the fork's best case made unconditional. Both executors
share the prefix; the parallel one still schedules those ops so their consumers
release as usual, and skips re-running them the way it already skips its
outputs. MIDI a device makes is not in the prefix and cannot be, so a modifier
triggered from an arpeggiator is a block late.

A level is not known, because it is what the ops are about to produce. The
detector runs during the walk and the resolve that reads it is the next block's:
one block, always exactly one, documented as a consequence of resolving at the
top rather than as an accident. The alternative is to resolve a modifier's
targets after its source's subgraph renders, which splits resolution into two
passes and makes the order a plan-shaped question. That is not this slice's.

`test_mod_feeds.cpp` asserts the distance per trigger type end to end through a
session rather than only the gap, which is what the timing note on #2120 asked
for.

## What is here

Forty-six cases across four files, and a `modulation-sources` golden that shows
the edge from both ends: the tap in the plan, and the modifier that reads it in
the table.

### DIFF
--- a/magda/daw/core/ModInfo.hpp
+++ b/magda/daw/core/ModInfo.hpp
@@ -249,6 +249,43 @@ enum class LFOTriggerMode {
 };
 
 /**
+ * @brief Where in a source track's chain a modifier listens.
+ *
+ * The two points the engines actually have, named for what they are rather than
+ * for a fader they sit either side of. A third point between them (post-FX,
+ * pre-fader) is not something either engine taps today, so it is deliberately
+ * not offered: adding it is a change to both engines and its own decision.
+ *
+ * Per modifier rather than per sidechain, because a modifier listening to its
+ * own track has no sidechain to carry the setting, and because two listeners on
+ * one source can reasonably want different points, which is the same reason the
+ * follower's band limits are per modifier.
+ */
+enum class ModTapPoint {
+    /// After whatever makes the track's sound and before its effects: the chain
+    /// head on a track with no instrument, and immediately past the instrument
+    /// on one that has it. What an audio trigger keys off, so a hit is detected
+    /// on what the track played rather than on what survived its chain.
+    PreFx,
+
+    /// The far end: post-FX, post-fader, before the muting node. What a
+    /// follower tracks, so riding the source fader moves the modulation, and
+    /// before the mute so a muted source still ducks what it is ducking.
+    PostFader,
+};
+
+/**
+ * @brief Where a modifier of this type listens unless it says otherwise.
+ *
+ * The fork's own split, which is what makes the default pure parity: its
+ * follower tap is post-fader and its trigger monitor is pre-FX, so a project
+ * carrying neither setting sounds the same in both engines.
+ */
+inline ModTapPoint defaultModTapPoint(ModType type) {
+    return type == ModType::Follower ? ModTapPoint::PostFader : ModTapPoint::PreFx;
+}
+
+/**
  * @brief A single mod-to-parameter link with its own amount.
  *
  * The target is a ControlTarget which addresses any parameter the system can
@@ -360,6 +397,12 @@ struct ModInfo {
     float followerHoldMs = 0.0f;       // Envelope hold time (ms)
     float followerReleaseMs = 500.0f;  // Envelope release time (ms)
 
+    // Where in the source track's chain this modifier listens. Only means
+    // anything for a modifier that listens to audio at all: a follower, or one
+    // triggered by level. Defaulted per type on creation, on a type change and
+    // on loading a project that predates the field (defaultModTapPoint).
+    ModTapPoint tapPoint = ModTapPoint::PreFx;
+
     // Band-limit detection: filter the raw source audio BEFORE peak detection so
     // the follower can track just part of the spectrum (e.g. only the bass).
     // Applied by the post-FX FollowerSourceTapPlugin, not TE's own post-rectify
@@ -384,7 +427,19 @@ struct ModInfo {
 
     // Constructor with index (for initialization)
     explicit ModInfo(int index)
-        : id(index), name(getDefaultName(index, ModType::LFO)), type(ModType::LFO) {}
+        : id(index),
+          name(getDefaultName(index, ModType::LFO)),
+          type(ModType::LFO),
+          tapPoint(defaultModTapPoint(ModType::LFO)) {}
+
+    /// Become @p t, and take the tap point that goes with it. What
+    /// TrackManager::setModType does, for the places that build a ModInfo
+    /// directly: assigning the type on its own leaves the point at the one the
+    /// old kind wanted, which for a follower is the wrong end of the chain.
+    void setType(ModType t) {
+        type = t;
+        tapPoint = defaultModTapPoint(t);
+    }
 
     bool isLinked() const {
         return !links.empty();
--- a/magda/daw/core/ModInfo.hpp
+++ b/magda/daw/core/ModInfo.hpp
@@ -261,17 +261,21 @@ enum class LFOTriggerMode {
  * one source can reasonably want different points, which is the same reason the
  * follower's band limits are per modifier.
  */
-enum class ModTapPoint {
+enum class ModTapPoint : int {
     /// After whatever makes the track's sound and before its effects: the chain
     /// head on a track with no instrument, and immediately past the instrument
     /// on one that has it. What an audio trigger keys off, so a hit is detected
     /// on what the track played rather than on what survived its chain.
-    PreFx,
+    ///
+    /// Ordinals are pinned: this is written into a project as its integer
+    /// (AutomationModSerializer), so an insert here would move an existing
+    /// project's modifiers to the other point (test_persisted_enum_pins.cpp).
+    PreFx = 0,
 
     /// The far end: post-FX, post-fader, before the muting node. What a
     /// follower tracks, so riding the source fader moves the modulation, and
     /// before the mute so a muted source still ducks what it is ducking.
-    PostFader,
+    PostFader = 1,
 };
 
 /**
--- a/magda/engine/exec/ParallelPlanExecutor.cpp
+++ b/magda/engine/exec/ParallelPlanExecutor.cpp
@@ -128,7 +128,7 @@ OpId ParallelPlanExecutor::runOp(OpId op) {
     // add into a buffer they all share, so they are left for the thread that
     // drove the block to run in plan order. Counted here anyway, because what
     // waits on what is the plan's business and not this decision's.
-    if (plan_->ops[static_cast<std::size_t>(op)].kind != OpKind::Output)
+    if (plan_->ops[static_cast<std::size_t>(op)].kind != OpKind::Output && !core_.inMidiPrefix(op))
         core_.renderOp(op, valueOf(op), block_, *output_);
 
     OpId carryOn = INVALID_OP_ID;
@@ -189,8 +189,16 @@ void ParallelPlanExecutor::process(const PlanValues& values, const BlockInfo& re
     if (!start.render || plan_ == nullptr)
         return;
 
-    // Before any worker starts: a device reads its parameters, and every op
-    // that could be one is about to become runnable.
+    // Before any worker starts, and in this order. The MIDI a block has before
+    // it has audio is rendered on this thread, so the notes in it reach the
+    // modifiers of this block rather than the next one; then the parameters
+    // resolve, and every op that could read one is about to become runnable.
+    //
+    // The prefix ops are still scheduled below, so their consumers are released
+    // exactly as they would be otherwise; what the schedule skips is running
+    // them a second time, the same way it skips the outputs it renders at the
+    // end. Anything else would consume a live input queue twice.
+    core_.renderMidiPrefix(values, start.block);
     core_.resolveParameters(values, start.block);
 
     block_ = start.block;
--- a/magda/engine/exec/ParallelPlanExecutor.cpp
+++ b/magda/engine/exec/ParallelPlanExecutor.cpp
@@ -56,6 +56,7 @@ std::vector<std::string> ParallelPlanExecutor::prepare(const RenderPlan& plan,
 
     plan_ = nullptr;
     outputOps_.clear();
+    modSourceOps_.clear();
 
     auto messages = core_.prepare(plan, bindings, context,
                                   previous != nullptr ? &previous->core_ : nullptr, params);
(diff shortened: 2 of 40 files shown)


magda/engine/exec/PlanExecutor.cpp
```diff
@@ -683,7 +734,159 @@ PlanExecutor::BlockStart PlanExecutor::beginBlock(const PlanValues& values,
...
+        const bool all = !on && !off && (message.isAllNotesOff() || message.isAllSoundOff());
+        if (!on && !off && !all)
+            continue;
+
+        for (const auto index : modSourceForOp_[op]) {
+            if (mods_.listensFor(index, *blockTable_) != ModListen::Midi)
+                continue;
+
+            if (on)
+                mods_.noteOn(index, *blockTable_);
```
lucaromagnoli, line 852: Cross-track MIDI-triggered modifiers never fire. Every listener of a source track's tap that listens for MIDI is fed through noteOn/noteOff/allNotesOff, and all three refuse a modifier that is drivenFromElsewhere, which is exactly what a cross-track listener is: the tap of track B feeds the modifiers whose source is B, and the ones not living on B carry skipNativeResync. So the MIDI branch of a rack or device sidechain is dead end to end.

The fork's MIDI monitor routes this through the other path: on a note-on it calls triggerSidechain (SidechainMonitorPlugin.cpp:77), which is the native trigger(), the one call skipNativeResync does not refuse, and it deliberately does nothing on note-off so the modifier runs its full cycle. The unit test ("A cross-track envelope hears its source and nothing else") passes because it calls mods.trigger() by hand; no executor path makes that call for a Midi listener.

Fix shape: in feedNoteTriggers, a listener the runtime reports as driven from elsewhere gets trigger() on note-on and nothing on note-off, matching the fork; same-track listeners keep the note-count gate they have now. That needs drivenFromElsewhere (or an equivalent answer) exposed from ModRuntime, since the executor currently cannot tell the two apart.

magda/engine/exec/ParallelPlanExecutor.cpp
```diff
@@ -128,7 +128,7 @@ OpId ParallelPlanExecutor::runOp(OpId op) {
     // add into a buffer they all share, so they are left for the thread that
     // drove the block to run in plan order. Counted here anyway, because what
     // waits on what is the plan's business and not this decision's.
-    if (plan_->ops[static_cast<std::size_t>(op)].kind != OpKind::Output)
+    if (plan_->ops[static_cast<std::size_t>(op)].kind != OpKind::Output && !core_.inMidiPrefix(op))
```
lucaromagnoli: Data race: ModSource ops run on worker threads, and two of them can run concurrently (two listened-to tracks have disjoint subgraphs, so both become ready together). renderModSource writes the shared detectMono_ on every call, and detectFollowerSource writes the runtime's shared detectScratch_ when a follower has band limits. Two concurrent taps corrupt each other's detection.

ModSource is a sink, same as Output, so the cheapest correct fix is the mechanism this line already uses: skip it here and let the driving thread render the ModSource ops after the walk, in plan order. That also cannot stall anything, since nothing consumes them. Per-op scratch would work too but buys nothing over serializing two ops that each cost a block scan.

No test runs a mod feed under the pool, which is why nothing caught it.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -255,6 +272,10 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
     midiTapForOp_.assign(numOps, nullptr);
     paramWindowForOp_.assign(numOps, ParamTable::DeviceWindow{});
     mixerParamForOp_.assign(numOps, OpMixerParams{});
+    modSourceForOp_.assign(numOps, std::span<const int>{});
+    triggerForOp_.assign(numOps, TriggerDetector{});
```
lucaromagnoli: The trigger detectors reset on every prepare while ModRuntime carries its states across. A structural publish during playback (a link edit escalation is enough) zeroes envelope and open; if the source is above threshold at that moment, the next block reads a rising edge and fires a spurious trigger, forceZero gap included, so every such edit is an audible blip on a playing sidechain. The fork's monitor plugin survives those edits with its envLevel_ and gateOpen_ intact.

Carrying TriggerDetector from previous, keyed by the op's track id the way ModRuntime keys ModState by address, restores the fork's behavior.

magda/engine/param/ParamTable.hpp
```diff
@@ -84,8 +93,22 @@ struct ParamModifier {
...
+    /// left at their defaults and never read.
     LfoSettings lfo;
+    AdsrSettings adsr;
+    RandomSettings random;
+    FollowerSettings follower;
+
+    /**
+     * @brief The track whose audio this modifier follows, or INVALID_TRACK_ID.
+     *
+     * Only a follower has one, and only a follower whose scope names a source:
```
lucaromagnoli: Stale: the compiler sets source for any listening modifier, MIDI- and audio-triggered ones included (modifierSourceTrack is called for every kind, and buildListenerRouting and the feeds depend on that). Only the first sentence's follower claim is wrong; the rest of the comment holds.

lucaromagnoli (commented):
Read the whole slice against the fork. The engines themselves are faithful transcriptions (I checked the ADSR bezier, the random step/shape/smooth arithmetic, the follower filters, and the trigger constants against the fork sources; all match). Four findings, one of them functional.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -881,6 +894,45 @@ void Compiler::emitTrack(const TrackInfo& track) {
...
 }
 
+void Compiler::emitModulationTaps() {
+    for (const auto sourceId : modulationSourceTracks_) {
+        // Post-fader and pre-mute, which is the point the fork taps as well:
+        // its follower tap sits at the end of the source track's plugin list
+        // and its trigger monitor keys off the same signal, so a muted source
+        // still ducks what it is ducking.
+        PortRef audio;
+        if (const auto found = trackSidechainTap_.find(sourceId); found != trackSidechainTap_.end())
```
lucaromagnoli: P1 — Split the audio-trigger and follower tap points. This single post-fader/pre-mute input is later used both for follower detection and for the peak that drives every non-follower Audio listener. The fork does not use one point: FollowerSourceTapPlugin is post-FX/post-fader, while ensureAudioSidechainMonitor inserts the trigger monitor at the front of the chain (or immediately after the instrument). Lowering the source fader below the threshold, or adding gain-changing FX, therefore makes the native engine miss or fire triggers differently from the fork. Please carry separate trigger/follower audio refs, or add separate taps, and dispatch each listener from its matching signal point.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -320,6 +329,13 @@ std::vector<const TrackInfo*> Compiler::computeTrackOrder() {
         // contributes no edge either.
         std::set<TrackId> upstream;
         collectSidechainSources(track, std::nullopt, upstream);
+
+        // A modifier listening to another track reads that track's signal, so
+        // it is an ordering constraint exactly as a sidechain is. It is also
+        // where the modifiers of a sidechained rack come in, which the plan
+        // used to report as something it could not carry (#2120).
+        collectModulationSources(track, upstream);
```
lucaromagnoli: P1 — Do not make a modulation feed a track-order dependency. ModSource ops are emitted after every track, and the audio feed is consumed on the next block; the MIDI prefix also renders independently before resolution. This edge can therefore invent a cycle with real routing. Repro: declare tracks in order {2, 1}, route track 1 audio into track 2, and put a rack on track 1 whose audio-triggered modifier listens to track 2. This adds 2→1 against the real 1→2 edge; the cycle breaker compiles track 2 first, reports that the incoming connection arrived after compilation, and drops the track 1→2 connection. Keep source collection for tap/MIDI emission, but leave it out of the track topology.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -255,6 +273,34 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+    // A modulation tap's envelope is state the way an LFO's phase is state, and
+    // a structural republish during playback must not zero it: a source above
+    // the threshold at that moment would read as a rising edge on the next
+    // block and fire a trigger nothing played. The fork's monitor plugin
+    // survives the same edits with its own level and gate intact.
+    std::map<TrackId, TriggerDetector> carriedTriggers;
+    if (previous != nullptr)
+        for (std::size_t i = 0; i < previous->plan_->ops.size(); ++i)
+            if (previous->plan_->ops[i].kind == OpKind::ModSource)
+                carriedTriggers[previous->plan_->ops[i].key.trackId] = previous->triggerForOp_[i];
```
lucaromagnoli: P1 — This copies triggerForOp_ while the previous executor may still be rendering; renderModSource mutates the same envelope/open fields on the audio thread. EngineSession prepares the replacement before the published swap, so this is a data race and can also carry an incoherent detector state, recreating the spurious edge this carryover is meant to prevent. Share detector state across epochs, keyed by track, as ModRuntime and the delay lines do, so only the retiring or replacement executor mutates it after the swap; otherwise synchronize the handoff.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -255,6 +273,34 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+
+    // The detectors of the epoch being replaced, by the track each watches.
+    // A modulation tap's envelope is state the way an LFO's phase is state, and
+    // a structural republish during playback must not zero it: a source above
+    // the threshold at that moment would read as a rising edge on the next
+    // block and fire a trigger nothing played. The fork's monitor plugin
+    // survives the same edits with its own level and gate intact.
+    std::map<TrackId, TriggerDetector> carriedTriggers;
+    if (previous != nullptr)
+        for (std::size_t i = 0; i < previous->plan_->ops.size(); ++i)
```
lucaromagnoli, line 297: P2 — Guard previous->plan_ before dereferencing it. prepare accepts a previous pointer without a documented prepared-state precondition, and a default-constructed or reset executor has plan_ == nullptr. Passing one here crashes, while the later plan-diff carry code already handles this case with previous->plan_ != nullptr (and previous != this). Apply the same guard here.

lucaromagnoli (commented):
Latest-head code review: four actionable findings.

magda/daw/core/ModInfo.hpp
```diff
@@ -384,7 +427,19 @@ struct ModInfo {
...
+        : id(index),
+          name(getDefaultName(index, ModType::LFO)),
+          type(ModType::LFO),
+          tapPoint(defaultModTapPoint(ModType::LFO)) {}
+
+    /// Become @p t, and take the tap point that goes with it. What
+    /// TrackManager::setModType does, for the places that build a ModInfo
+    /// directly: assigning the type on its own leaves the point at the one the
+    /// old kind wanted, which for a follower is the wrong end of the chain.
+    void setType(ModType t) {
```
lucaromagnoli, line 443: P1 — The normal construction path does not use this helper. TrackManager::addMod still does newMod.type = type after constructing an LFO-default ModInfo, so addMod(..., ModType::Follower, ...) leaves tapPoint at PreFx. Those are the UI add paths, which means every newly created follower listens at the wrong point; loaded followers and later type changes are the only ones that get PostFader. Please change addMod to newMod.setType(type) and cover addMod(Follower) in a test.

magda/daw/core/ModInfo.hpp
```diff
@@ -248,6 +248,43 @@ enum class LFOTriggerMode {
...
+ * for a fader they sit either side of. A third point between them (post-FX,
+ * pre-fader) is not something either engine taps today, so it is deliberately
+ * not offered: adding it is a change to both engines and its own decision.
+ *
+ * Per modifier rather than per sidechain, because a modifier listening to its
+ * own track has no sidechain to carry the setting, and because two listeners on
+ * one source can reasonably want different points, which is the same reason the
+ * follower's band limits are per modifier.
+ */
+enum class ModTapPoint {
```
lucaromagnoli: P2 — Pin these ordinals now that AutomationModSerializer writes and reads tapPoint as an integer. This repository already keeps persisted enums in test_persisted_enum_pins.cpp; leaving these values implicit means a later insertion or reorder silently redirects existing projects to another tap. Please make PreFx = 0 and PostFader = 1 explicitly and add the corresponding persisted-enum assertions.

lucaromagnoli (commented):
Second-pass review: two follow-up findings in the tap-point fix.

Issue #2123, opened 2026-08-19
Parameters slice 8: parity cases for parameters, modifiers, macros and automation


lucaromagnoli:
Slice 8 of #1891. Depends on everything above.

The parity slice this issue carries, the way #2040 was slice 7 of #1890.

Cases through the #2075 oracle for what was ported here: a parameter under automation, a parameter
under a modifier, a parameter under both, a host write arriving under each, every modifier engine
against the fork's, macros at each scope, and the fader curve at the ends of its range, where the
clamp is.

Written here rather than in #1896, and written now rather than later. A case asserts behaviour
while the person writing it still knows where the feature can lie, and asserts the code once that
knowledge has to be read back out of it.

lucaromagnoli:
One requirement from the #2119 review that belongs to this slice: the triggered-modifier parity cases must assert the trigger-to-gap distance in blocks, per trigger type (MIDI, audio), not just that the gap and retrigger happen. The native engine resolves parameters at block top, so trigger latency is a property of where the feeds run, and it can silently differ from the fork by a block while every shape assertion still passes. The timing target per type is defined by the decision recorded on #2120 (same-block via block-top MIDI detection and split resolution, or an explicitly accepted deterministic one-block lag for audio); whichever is chosen, the parity case pins it.

PR #2131, opened 2026-08-21
Read a value back, without simulating one (#2122)

### LINKED ISSUE #2122: Parameters slice 7: value read-back taps
Slice 7 of #1891. Depends on slices 4, 5 and 6.

What the UI reads back.

The current engine is polled. `TrackManager::updateAllMods` reads modifier output off TE, and the
random modulator carries a MAGDA-side simulation of its own so the editor keeps animating between
audio-thread updates. Neither survives the port.

- A tap for modifier output, in the shape of `LevelTap` and `MidiTap`.
- A tap for the resolved parameter value, which is what a device UI draws when a parameter is
  modulated, and what touch and latch would read when they ship.
- The simulation in `TrackManager` goes away rather than being ported.

lucaromagnoli:
Closes #2122.

What the UI reads back: the two numbers a window wants and cannot work out for itself. What a modifier's output did this block, and what a parameter resolved to once its base, its lane and its links had been settled against each other. The second is what a device UI draws when something is modulating a knob, and what touch and latch will read at the moment a gesture starts.

## The tap

`magda/engine/tap/ValueTap.hpp`, the sibling of `LevelTap` and `MidiTap`, with the one difference that matters: a read is not destructive. A meter reports the loudest thing since it was last looked at because a transient falling between two frames must not go missing. A position has nothing of the kind to miss, so the value stands until a block replaces it, any number of readers get the same answer rather than taking it from each other, and a reader arriving late gets the current position rather than a zero.

It carries a count of blocks published beside the value, which is what tells a value that is holding from a value that is not moving. The two are one 64-bit word rather than two atomics, and that is not an optimisation: the count is what a reader gates a repaint on, so a reading pairing one block's count with another block's value is not a stale frame but a value that never arrives. Read the value first and the reader draws the old one, then finds the count unmoved at the next poll, decides nothing changed, and never draws the new one. Read the count first and the mirror image happens. One store on the audio thread, one load off it, no sequence lock and nothing to retry.

Bound by `ParamKey` rather than by `OpKey`, which is the one difference from the two existing taps and follows from what is being read: a meter is a place in the signal and this is a number in the parameter system. One map for both numbers, because the key already says which: a key with a parameter index is a parameter's resolved position, and a modifier key with none is that modifier's output.

## How a host gets one

`RuntimeStateFactory::valuesToTap()` returns the keys the host wants read back; the store makes the taps and `EngineSession::valueTap(key)` hands them out. One question rather than a `createValueTap` beside its neighbours, and the asymmetry is a matter of how many there are: a plan has hundreds of ops and a host with a mixer open wants a meter on a good fraction of them, so asking op by op costs about what the answers are worth. A table has a parameter for every parameter of every device in the project, and a host wants the few dozen it has on screen. Asking one by one would be thousands of declines per edit to find them.

The store makes them rather than the factory handing back instances, because a `ValueTap` is not a thing the engine needs the host's help to build. The other factory methods exist because the engine has no idea what a device or a file reader is; it knows exactly what a tap on a number is.

Eviction takes the live `ParamTable` alongside the live plan, because a plan does not name a parameter and nothing else could say which taps the audio thread can reach. A rack scope is retained by its track, which is coarser than the rest: `RuntimeStateIds` names devices and tracks, a rack is neither, and what the coarser rule costs is one word per deleted rack held until the track goes.

## Where it publishes

At the end of `PlanExecutor::resolveParameters`, which both executors call, and which is the one moment both numbers exist and the last moment they are still the block's own rather than something a device has since been handed. Bound taps are held in compact lists resolved at prepare, so the block hashes nothing and walks tens of entries rather than thousands of nulls.

A block whose table does not fit publishes nothing at all rather than the zeros an empty table answers with. The watched values then hold, which is what they do whenever nothing is rendering; slamming every watched knob to zero for the one block a mismatched table takes to be replaced would be a visible fault reporting an invisible one.

## Nothing simulates a modifier

The third bullet of the issue. The incumbent engine keeps a MAGDA-side copy of the random modulator turning inside `TrackManager::updateAllMods`, so its editor preview animates between audio-thread updates: there the value is polled at frame rate and a poll landing between two callbacks has nothing to report. Here the block publishes what it produced, so a tap that is not moving is an engine that is not rendering, and that is the honest answer rather than a gap to paper over.

That simulation stays exactly where it is. It drives the engine that still ships, and removing it now would break the modifier editors in the live product; what the issue asks is that it not be ported, and it is not.

## The tests

`tests/test_value_taps.cpp`, twelve cases:

- the tap itself: holding the last block's value, surviving repeated reads, counting blocks rather than changes, reading as nothing before anything publishes
- the count belonging to the value beside it across successive writes, which is the pair that must not come apart
- a modifier tap publishing the modifier's own output, shut before the note and open after it
- a parameter tap publishing what the block resolved and not what is stored: a quarter stored plus half a link's worth of envelope reads 0.75 while the modifier beside it reads 1.0, so the link's depth is visibly on one and not the other
- a value nobody asked about having no tap at all
- a key the table does not carry bound to nothing rather than refused, which is a window that outlived what it was opened on
- a block with no table it fits leaving the taps holding, value and count both
- a tap surviving an unrelated structural edit as the same object with the same reading, and going when the device it names is deleted while the track's own value stays

## Verified

`make test`: 2843 cases, 589,811 assertions, all passing.

## Notes

- `docs/architecture/native-engine.md` section 8 still says automation curves, modifiers and macros do not exist in the engine. Stale since the parameter lane slice, left alone here for the same reason the last slice left it: #2123 is still open and the doc reads like it gets its update at the end of the arc.

### DIFF
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -199,15 +199,30 @@ class EngineSession {
      * @brief Where @p key's value is published, or nullptr (#2122).
      *
      * What a host collects after asking for a value through
-     * RuntimeStateFactory::valuesToTap(). The lookup is on the publishing
-     * thread; the tap it hands back is readable from any thread, and is what a
-     * knob under modulation, a modifier editor, and a touch or latch gesture
-     * all draw from.
+     * RuntimeStateFactory::valuesToTap(). What it hands back is where a knob
+     * under modulation, a modifier editor, and a touch or latch gesture all
+     * draw from.
      *
-     * Good until the next publish. See RuntimeStateStore::valueTap().
+     * Both the lookup and the pointer belong to the publishing thread, and that
+     * is a lifetime rather than a convention: the next publish can destroy the
+     * tap outright, on that thread, with nothing deferring the reclamation. A
+     * cached pointer read from a paint or animation thread is a use-after-free
+     * waiting for an edit. Fetch it here, use it before returning to the
+     * message loop, and ask again after every publish; read() is safe from
+     * wherever it is called for exactly as long as the pointer is.
+     *
+     * See RuntimeStateStore::valueTap() for what null means, which is not the
+     * same as the tap being unbound.
      */
     ValueTap* valueTap(const ParamKey& key) const {
         return store_.valueTap(key);
+    }
+
+    /// Values the live plan found somewhere to publish from, which is not the
+    /// number of taps the store holds: a key the parameter table does not carry
+    /// has a tap and is never written through it. On the publishing thread.
+    int boundValueTapCount() const {
+        return live_ == nullptr ? 0 : live_->executor.boundValueTapCount();
     }
 
     /// The plan currently published, or null. On the publishing thread.
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -195,6 +195,21 @@ class EngineSession {
         return store_.size();
     }
 
+    /**
+     * @brief Where @p key's value is published, or nullptr (#2122).
+     *
+     * What a host collects after asking for a value through
+     * RuntimeStateFactory::valuesToTap(). The lookup is on the publishing
+     * thread; the tap it hands back is readable from any thread, and is what a
+     * knob under modulation, a modifier editor, and a touch or latch gesture
+     * all draw from.
+     *
+     * Good until the next publish. See RuntimeStateStore::valueTap().
+     */
+    ValueTap* valueTap(const ParamKey& key) const {
+        return store_.valueTap(key);
+    }
+
     /// The plan currently published, or null. On the publishing thread.
     std::shared_ptr<const RenderPlan> livePlan() const {
         return livePlan_;
--- a/magda/engine/exec/PlanBindings.hpp
+++ b/magda/engine/exec/PlanBindings.hpp
@@ -5,9 +5,11 @@
 
 #include "core/TypeIds.hpp"
 #include "exec/EngineDevice.hpp"
+#include "param/ParamKey.hpp"
 #include "plan/RenderPlan.hpp"
 #include "tap/LevelTap.hpp"
 #include "tap/MidiTap.hpp"
+#include "tap/ValueTap.hpp"
 
 namespace magda::engine {
 
@@ -66,6 +68,28 @@ struct PlanBindings {
      * observing, and an unbound op is not reported.
      */
     std::map<OpKey, MidiTap*> midiTaps;
+
+    /**
+     * @brief Where a value is published for whoever is drawing it (#2122).
+     *
+     * Keyed by ParamKey rather than by OpKey, which is the one difference from
+     * the two above and follows from what is being read: a meter is a place in
+     * the signal and this is a number in the parameter system. One map for both
+     * of the numbers, because the key already says which: a key with a
+     * parameter index is a parameter's resolved position, and a modifier key
+     * with none is that modifier's output (ParamKey.hpp).
+     *
+     * Optional in the same way and for the same reason. A host binds what it is
+     * drawing, which is a device editor's worth of parameters out of a
+     * project's thousands, and a value nobody has asked about publishes nothing
+     * and is not reported.
+     *
+     * Resolved against the table the plan is prepared with, so a key the table
+     * does not carry binds nothing. That is not an error either: a host may ask
+     * for a parameter that has since been deleted, and the answer is that
+     * nothing writes to it.
+     */
+    std::map<ParamKey, ValueTap*> valueTaps;
 };
 
 }  // namespace magda::engine
--- a/magda/engine/exec/PlanBindings.hpp
+++ b/magda/engine/exec/PlanBindings.hpp
@@ -85,9 +85,13 @@ struct PlanBindings {
      * and is not reported.
      *
      * Resolved against the table the plan is prepared with, so a key the table
-     * does not carry binds nothing. That is not an error either: a host may ask
-     * for a parameter that has since been deleted, and the answer is that
-     * nothing writes to it.
+     * does not carry binds nothing. That is not an error either, and it is not
+     * rare: the table carries a mixer value, a macro or a modifier's rate only
+     * while something reaches it, so an unautomated fader with nothing linked
+     * to it has no publisher at all. The tap reads no writes, which is what
+     * tells a host the engine does not move this value and the model's own is
+     * the answer (ValueTap.hpp). A parameter that has since been deleted lands
+     * in the same place by the same route.
      */
     std::map<ParamKey, ValueTap*> valueTaps;
 };
--- a/magda/engine/exec/PlanExecutor.hpp
+++ b/magda/engine/exec/PlanExecutor.hpp
@@ -316,6 +316,14 @@ class PlanExecutor {
         return boundMeterCount_;
     }
 
+    /// Value taps this plan's table found a home for (#2122): a parameter the
+    /// table carries, or a modifier the runtime holds. A key the bindings
+    /// offered and the table does not have is not one of these and is not an
+    /// error; it is a host asking about something this project no longer has.
+    int boundValueTapCount() const {
+        return static_cast<int>(paramTaps_.size() + modTaps_.size());
+    }
+
     /// Modulation taps that took over the detector of the executor they
     /// replaced rather than starting again. What that buys is the same thing
     /// ModRuntime's carry buys: an edit during playback does not restate a
@@ -560,6 +568,27 @@ class PlanExecutor {
     /// PlanBindings::midiTaps.
     std::vector<MidiTap*> midiTapForOp_;
 
+    /** @brief One value the block publishes for whoever is drawing it (#2122). */
+    struct BoundValueTap {
+        /// A ParamId, or an index into the modifier runtime.
+        int index = 0;
+        ValueTap* tap = nullptr;
+    };
+
+    /// Compact rather than one slot per parameter, which is the difference
+    /// between these and the per-op bindings above. A plan has hundreds of ops
+    /// and a host binds a good fraction of the meters; a table has a parameter
+    /// per parameter of every device in the project and a host draws the few
+    /// dozen it has on screen. A vector the length of the table would be
+    /// thousands of nulls walked every block to find them.
+    ///
+    /// Resolved at prepare against the table the plan was prepared with, so the
+    /// block never looks a key up, and safe to hold by index for exactly as
+    /// long as the windows are: a table whose layout differs is refused whole
+    /// (see fitsParameters).
+    std::vector<BoundValueTap> paramTaps_;
+    std::vector<BoundValueTap> modTaps_;
+
     /// This block's parameter values, and the room the resolver gathers one
     /// parameter's links in. Sized at prepare from the table the plan was
     /// published with, so the block that fills them allocates nothing.
@@ -643,6 +672,22 @@ class PlanExecutor {
     /// Read one modulation tap: hand the source's level to the followers
     /// listening to it, and its rises and falls to the triggered ones.
     void renderModSource(OpId id, const BlockInfo& block);
+
+    /**
+     * @brief Publish what this block settled, to whoever is watching (#2122).
+     *
+     * On the audio thread, at the end of the resolve and nowhere else: that is
+     * the moment both numbers exist and the last moment they are still the
+     * block's own rather than something a device has since been handed.
+     *
+     * A block that resolved no table publishes nothing at all, rather than
+     * publishing the zeros an empty table would answer with. The values a host
+     * is drawing then hold, which is what they do whenever nothing is
+     * rendering; slamming every watched knob to zero for the one block a
+     * mismatched table takes to be replaced would be a visible fault reporting
+     * an invisible one.
+     */
+    void publishValueTaps();
 
     /// Spend the notes in @p midi on the modifiers listening to @p op's track.
     void feedNoteTriggers(std::size_t op, const juce::MidiBuffer& midi);
--- a/magda/engine/exec/PlanExecutor.hpp
+++ b/magda/engine/exec/PlanExecutor.hpp
@@ -319,10 +319,30 @@ class PlanExecutor {
     /// Value taps this plan's table found a home for (#2122): a parameter the
     /// table carries, or a modifier the runtime holds. A key the bindings
     /// offered and the table does not have is not one of these and is not an
-    /// error; it is a host asking about something this project no longer has.
+    /// error; it is a host asking about a value this project does not move.
     int boundValueTapCount() const {
         return static_cast<int>(paramTaps_.size() + modTaps_.size());
     }
+
+    /**
+     * @brief Take back what this plan no longer publishes (#2122).
+     *
+     * Off the audio thread, and only after the swap that made this plan live.
+     * Every tap the bindings offered and this table found no home for is
+     * cleared, so a host reads no writes at all rather than a value frozen
+     * where the last plan left it.
+     *
+     * That is not a rare case. A lane deleted off a fader takes that fader out
+     * of the parameter table, because a mixer value is carried only while
+     * something reaches it; without this the tap would sit at the last
+     * automated position under a count that never moves again, which the header
+     * tells hosts to read as an engine that has stopped.
+     *
+     * After the swap rather than at prepare, because until the swap the epoch
+     * being replaced is still rendering: a tap cleared while it is still being
+     * written would be counting from one again by the next block.
+     */
+    void clearUnboundValueTaps();
 
     /// Modulation taps that took over the detector of the executor they
     /// replaced rather than starting again. What that buys is the same thing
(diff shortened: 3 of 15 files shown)


magda/engine/exec/RuntimeStateStore.cpp
```diff
@@ -212,6 +260,20 @@ std::size_t RuntimeStateStore::releaseDeleted(const RenderPlan& livePlan,
         ++removed;
     }
 
+    // The live table first and unconditionally, on the same reading the plan
+    // gets above: a tap the table carries is one the executor holds a pointer
+    // to and the audio thread writes through every block, whatever the caller
+    // believes the model still holds.
+    for (auto entry = valueTaps_.begin(); entry != valueTaps_.end();) {
+        if ((liveTable != nullptr && carries(*liveTable, entry->first)) ||
```
lucaromagnoli: A tap whose key drops out of the live table while its model object survives is retained by isNamed() but never re-bound or written again: it freezes the last value under a stalled write count that ValueTap.hpp tells hosts to read as "engine stopped".

Repro: tap TrackVolume while an automation lane plays it (targeting is the only way mixer keys enter the table, ParamTableCompiler.cpp:592), then delete or disable the lane. The structural publish drops the key, prepare() binds nothing, but the tap survives via isNamed and valueTap() keeps returning it. A UI following the documented writes-gate keeps drawing the old automated fader position, or concludes the engine stopped, indefinitely while audio renders. Same trap for link-less macros and untargeted modifier rates.

magda/engine/exec/RuntimeStateStore.hpp
```diff
@@ -147,8 +179,33 @@ class RuntimeStateStore {
...
      * run on the calling thread.
+     *
+     * @p liveTable is the parameter table that plan is rendering with, and it
+     * is to the value taps what the plan is to everything else: the one thing
+     * that says which of them the audio thread can reach. A plan does not name
+     * a parameter, so nothing else here could answer it. Null is a session
+     * rendering without a table, where no value tap is reachable at all.
+     */
+    std::size_t releaseDeleted(const RenderPlan& livePlan, const RuntimeStateIds& modelIds,
+                               const ParamTable* liveTable = nullptr);
```
lucaromagnoli: The `const ParamTable* liveTable = nullptr` default makes the only guard keeping audio-thread-reachable value taps alive opt-in, and tests/PlanEditHarness.cpp:270 already calls the two-arg form. Harmless today only because no harness factory overrides valuesToTap(). Any future caller compiles cleanly without the table; eviction then falls back to model IDs alone, and a tap the live table carries but the caller's ID set does not name is destroyed while the executor still holds its pointer in paramTaps_/modTaps_, so the next block's publishValueTaps() writes through freed memory on the audio thread. The doc here says the plan-side keep cannot be waived; the table-side keep deserves the same treatment. Make it a required parameter.

magda/engine/exec/PlanBindings.hpp
```diff
@@ -66,6 +68,28 @@ struct PlanBindings {
...
+     * of the numbers, because the key already says which: a key with a
+     * parameter index is a parameter's resolved position, and a modifier key
+     * with none is that modifier's output (ParamKey.hpp).
+     *
+     * Optional in the same way and for the same reason. A host binds what it is
+     * drawing, which is a device editor's worth of parameters out of a
+     * project's thousands, and a value nobody has asked about publishes nothing
+     * and is not reported.
+     *
+     * Resolved against the table the plan is prepared with, so a key the table
```
lucaromagnoli, line 87: A live, model-held key the table never carries (unautomated TrackVolume/TrackPan/SendLevel, link-less macro) taps as permanently silent writes==0, and the docs frame an uncarried key only as "a parameter that has since been deleted". ValueTap.hpp advertises the tap for liveness checks and for what touch/latch would read at gesture start; both fail for exactly the parameters a first gesture is about (a mixer strip tapping an unautomated fader reads writes==0 forever while audio plays). One sentence here and in ValueTap.hpp naming the writes==0 model-value fallback closes it.

magda/engine/exec/EngineSession.hpp
```diff
@@ -195,6 +195,21 @@ class EngineSession {
         return store_.size();
     }
 
+    /**
+     * @brief Where @p key's value is published, or nullptr (#2122).
+     *
+     * What a host collects after asking for a value through
+     * RuntimeStateFactory::valuesToTap(). The lookup is on the publishing
+     * thread; the tap it hands back is readable from any thread, and is what a
```
lucaromagnoli: "Readable from any thread" overpromises: the tap is destroyed synchronously on the publishing thread inside releaseDeleted with no deferred reclamation, so a host that polls a cached ValueTap* from a paint/GL or background animation thread can be inside read() when the message thread's next publish erases the unique_ptr. Use-after-free. The intended contract (fetch and read on the publishing thread, re-ask after every publish) is safe but is not what this says; scope the pointer lifetime to the publishing thread and restrict "any thread" to read() under that lifetime.

magda/engine/tap/ValueTap.hpp
```diff
@@ -0,0 +1,164 @@
...
+     * Unconditional rather than only on a change, because the count is of
+     * blocks published and a tap that fell silent on a value that stopped
+     * moving would be indistinguishable from one whose engine had stopped.
+     *
+     * The count comes off the word this is about to replace rather than out of
+     * a counter of its own, which is what makes the pair inseparable. Loading
+     * it needs no ordering and no compare-and-swap: there is one writer, and it
+     * is the only thing that ever changes this.
+     */
+    void write(float value) {
```
lucaromagnoli, line 159: write() lets the 32-bit count wrap through 0, re-arming the documented "writes == 0 means bound but never published" sentinel with a live value beside it. 2^32 blocks is roughly 33 days of continuous uptime at 96kHz with 64-sample buffers, reachable for a live rig or installation; at the wrap a host following this header's contract classifies a live parameter as never-published. Skipping 0 on wrap is one branch and keeps delta comparisons valid.

magda/engine/exec/RuntimeStateStore.hpp
```diff
@@ -147,8 +179,33 @@ class RuntimeStateStore {
...
+
+    /**
+     * @brief The tap publishing @p key, or nullptr (#2122).
+     *
+     * The other half of valuesToTap(): the host says which values it wants read
+     * back and this is where it collects them. Off the audio thread; what it
+     * hands back is readable from any thread, which is the whole point of it.
+     *
+     * Null until a publish has been through, because that is when the taps are
```
lucaromagnoli: "Null until a publish has been through, because that is when the taps are made" is false after a refused publish: realise() creates the taps before both refusal exits in EngineSession::publish, so valueTap() returns non-null although result.published was false. The behavior is deliberate (realising early costs nothing) and harmless, but a host treating non-null as "my plan is live and this key is bound" per this doc draws a frozen zero. Say taps exist after any publish attempt and that writes==0 is the real "not yet bound" signal.

magda/engine/exec/RuntimeStateStore.cpp
```diff
@@ -212,6 +260,20 @@ std::size_t RuntimeStateStore::releaseDeleted(const RenderPlan& livePlan,
         ++removed;
     }
 
+    // The live table first and unconditionally, on the same reading the plan
```
lucaromagnoli, line 263: This comment is wrong for keys the host stopped asking for: prepare() binds only the current valuesToTap() list, so a key the table carries is not necessarily one the executor holds a pointer to. The carries() branch also retains every key ever watched until its device or track dies, so a long session accumulates a map node plus atomic per parameter ever displayed, and every subsequent releaseDeleted does a ParamTable::find per retained tap. Growth is bounded by live-parameter count so the cost is modest; the false comment is the part that will mislead the next editor of this loop.

tests/test_value_taps.cpp
```diff
@@ -0,0 +1,488 @@
...
+    const auto before = session.writes(param);
+    session.render();
+
+    CHECK(session.read(param) == approx(0.75f));
+    CHECK(session.writes(param) == before);
+}
+
+// --- what the store does with them
+
+TEST_CASE("A value tap outlives the plans that reference it", "[engine][tap][value]") {
```
lucaromagnoli, line 467: This proves pointer identity and the held value across a republish but never renders afterwards, so the re-binding half of the carry (realise re-offering the tap, prepare re-resolving it against the new table) is never exercised end-to-end. A regression that keeps the store's map intact but breaks re-binding on republish (bindings.valueTaps not repopulated, or the prepare loop skipped for a carried epoch) freezes every open window from the first project edit onward while this stays green. Add one render plus a writes-advanced check after the second publish.

magda/engine/exec/RuntimeStateStore.cpp
```diff
@@ -212,6 +260,20 @@ std::size_t RuntimeStateStore::releaseDeleted(const RenderPlan& livePlan,
...
+    // gets above: a tap the table carries is one the executor holds a pointer
+    // to and the audio thread writes through every block, whatever the caller
+    // believes the model still holds.
+    for (auto entry = valueTaps_.begin(); entry != valueTaps_.end();) {
```
lucaromagnoli: Third hand-rolled iterate-and-erase in this file (after eraseUnnamed and the meters_ loop). C++20 `std::erase_if` over std::map is behavior-identical, returns the erased count, and is already used at about a dozen sites in the codebase:

```cpp
removed += std::erase_if(valueTaps_, [&](const auto& e) {
    return !((liveTable && carries(*liveTable, e.first)) || isNamed(e.first, keep));
});
```

tests/test_value_taps.cpp
```diff
@@ -0,0 +1,488 @@
...
+    return Catch::Approx(value).margin(1e-4);
+}
+
+constexpr double kSampleRate = 48000.0;
+constexpr int kBlock = 512;
+
+constexpr TrackId kTrack = 1;
+constexpr DeviceId kDevice = 7;
+
+TrackInfo makeTrack(TrackId id) {
```
lucaromagnoli: About 82 lines of scaffolding here (makeTrack/makeMaster/makeDevice plus the ScriptedMidi/ScriptedMidiSource/Factory/Session quartet) are verbatim copies from test_mod_feeds.cpp, with looser variants in test_mod_lfo.cpp and test_engine_session.cpp. A publish-protocol or collectRuntimeStateIds change now has to be applied in lockstep across all copies; this second verbatim copy is the point where the quartet earns a shared tests/ header.

magda/engine/exec/PlanExecutor.hpp
```diff
@@ -316,6 +316,14 @@ class PlanExecutor {
         return boundMeterCount_;
     }
 
+    /// Value taps this plan's table found a home for (#2122): a parameter the
+    /// table carries, or a modifier the runtime holds. A key the bindings
+    /// offered and the table does not have is not one of these and is not an
+    /// error; it is a host asking about something this project no longer has.
```
lucaromagnoli: boundValueTapCount() is added to both executors with zero callers, unlike boundMeterCount which the parallel-parity test consumes (tests/test_parallel_executor.cpp:741). As it stands the bind-count is asserted nowhere and the parity suite has no tap equivalent. Either use it in the new tests or the parity suite (which also closes that coverage gap) or drop it.

magda/engine/tap/ValueTap.hpp
```diff
@@ -0,0 +1,164 @@
...
+        state_.store(pack(value, writes + 1), std::memory_order_release);
+    }
+
+  private:
+    /// The count in the high half, the float's bits in the low one. A zeroed
+    /// word is therefore no writes and a value of zero, which is what a tap
+    /// nothing has published reads as.
+    static std::uint64_t pack(float value, std::uint32_t writes) {
+        std::uint32_t bits = 0;
+        std::memcpy(&bits, &value, sizeof(bits));
```
lucaromagnoli: pack()/unpack() type-pun float to uint32 via memcpy into mutable locals; the project is C++20 and already uses std::bit_cast for exactly this (magda/daw/media_db/MelSpectrogram.cpp:50). `(uint64_t(writes) << 32) | std::bit_cast<std::uint32_t>(value)` and `std::bit_cast<float>(bits)` drop the locals, are constexpr, and let <cstring> become <bit>.

tests/test_value_taps.cpp
```diff
@@ -0,0 +1,488 @@
+#include <catch2/catch_approx.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <memory>
+#include <vector>
+
+#include "core/TrackInfo.hpp"
+#include "exec/EngineSession.hpp"
+#include "exec/PlanValues.hpp"
+#include "exec/RuntimeStateStore.hpp"
+#include "param/ParamTableCompiler.hpp"
```
lucaromagnoli: Unused include: ParamTableCompiler.hpp declares only compileParamTable, which this test never calls; ParamKey/ParamTable arrive transitively via the other includes. Carried over with the test_mod_feeds.cpp scaffolding; delete the line.

lucaromagnoli (commented):
Reviewed the value-tap surface end to end. The core mechanism is sound: I verified the pack/unpack atomics, single-writer ordering across the farbot swap, key-shape consistency between the table compiler, ModRuntime and carries(), fingerprint-protected index stability, and that eviction never frees an executor-reachable tap on any production path. 13 inline comments below, most severe first: the top ones are behavioral/API-contract gaps in the new host-facing surface, the rest are doc corrections and cleanup.

One finding has no diff line to anchor to, so it goes here: ParamKey's default `index = 0` (ParamKey.hpp:73) spells the modifier's Rate parameter (kModRateParamIndex == 0), while the modifier-as-source contract valuesToTap() depends on requires `index = -1`, and nothing in the new API docs warns hosts or points at modifierKeyFor(). A host that hand-builds `ParamKey{ModParam, Track, trackId, modId}` and leaves index defaulted will tap the rate if anything targets it (the macro-to-rate path from the modifier-feeds work carries the identical key), or bind nothing and stall if nothing does. The tests only avoid this by hand-writing `key.index = -1` (test_value_taps.cpp:113). Worth a doc line on valuesToTap(), or a dedicated helper that builds the source-shaped key.

lucaromagnoli:
All fourteen addressed, in three commits. One pushback at the end.

## Behavioural

**The freeze (comment on RuntimeStateStore.cpp:268), and its sibling.** Correct, and the worst of the set. The table carries a mixer value, a macro or a modifier's rate only while something reaches it, so deleting the lane off an automated fader drops the key: the tap stopped being written and sat at the last automated position under a stalled count, which my own header tells hosts to read as an engine that has stopped.

A count of zero now means one thing everywhere, including after the fact: nothing in the engine publishes this value, so the model's own is the answer. A publish that binds a tap to nothing clears it, after the swap rather than at prepare, because until the swap the epoch being replaced is still writing it and a tap cleared under it would be counting from one again by its next block.

While closing that I found the same hole one path further out: binding the taps ran inside the guard that skips everything when a plan is published with no parameter table, so nothing reached the unbound list and the clear found nothing to clear. A plan with no table publishes no values at all, which makes every offered tap one of these. The loop is now outside the guard and consults the table where there is one.

**The wrap (ValueTap.hpp:138).** Correct, and the clear above makes it load-bearing rather than cosmetic: the count now skips zero on the way round, since reaching it would hand a host a live parameter wearing the one sign that says to ignore the tap. `nextWriteCount` is public for that reason, and because 2^32 blocks is not a number a test can render its way to; the rule is asserted where it is decided.

**The defaulted table (RuntimeStateStore.hpp:190).** Agreed and required now. Eviction falling back to model IDs alone would free a tap the executor still points at, and the next block writes through it on the audio thread. A keep that cannot be waived must not be possible to leave out. `PlanEditHarness` passes `nullptr` explicitly, which is what it means: it prepares without a table.

## Docs

- **valuesToTap and `index = -1`** (the finding with no line to anchor to): confirmed against the compiler, `kModRateParamIndex == 0` and `ParamKey` defaults `index = 0`, so a hand-built modifier key asks about the Rate parameter. The doc now says so and points at `modifierKeyFor`.
- **EngineSession::valueTap (:203):** the "any thread" claim is gone. The pointer belongs to the publishing thread as a lifetime, not a convention, since the next publish destroys the tap there with nothing deferring reclamation. `read()` is safe from wherever it is called for exactly as long as the pointer is.
- **The store's valueTap (:199):** taps exist after any publish attempt, not a successful one, and non-null does not mean bound. What answers that is the count.
- **The eviction comment (RuntimeStateStore.cpp:263):** no longer claims every key the table carries is one the executor holds, and says what the model rule retains and what it costs.
- **PlanBindings (:87):** an uncarried key is now framed as the common case it is rather than only as a deleted parameter.

## Cleanup

`std::erase_if` for the eviction loop, `std::bit_cast` for the packing (which lets `<cstring>` become `<bit>` and makes both helpers constexpr), and the unused `ParamTableCompiler.hpp`.

## Tests

- the freeze, end to end through a lane that arrives and goes
- a plan published with no table at all, which is the sibling case
- a render after the republish, so the re-binding half of the carry is exercised rather than only the store's map surviving
- the wrap, asserted where it is decided
- `boundValueTapCount`, which had no caller: it is reachable through the session now and asserted in two cases, including the one where a key the table does not carry binds nothing

`make test`: 2846 cases, 591,492 assertions, all passing.

## One pushback

On the shared scaffolding (test_value_taps.cpp:50): "about 82 lines verbatim" overstates it. The actual verbatim overlap with `test_mod_feeds.cpp` is around forty lines, and the `Factory`/`Session` pair differs substantively between the two files (different device, different factory surface, different idea of an edit) rather than being copied.

I extracted it anyway, because the part that is genuinely shared is the part the finding is really about: the publish protocol is what has to be edited in lockstep when `collectRuntimeStateIds` or the publish signature changes. `tests/EngineSessionScaffold.hpp` now holds the track and master builders, the scripted-MIDI pair and `publishProject`, and both files use it. I left `Factory` and `Session` in their own files rather than inventing a base class that would have to guess at what each test wants around the session.

Issue #2138, opened 2026-08-22
Rack graph slice 4: pins, channel counts and implicit summing


lucaromagnoli:
Slice 4 of #1892. Depends on the engine core.

Pins and channel counts, which is what #1892's "pin-level connection graph with implicit summing
at shared pins" turns into once the chain model is what it is.

The plan is stereo everywhere. `RenderContext::numChannels` is 2, with the comment saying so
outright: the model has no mono. The incumbent is not, and the difference is audible rather than
cosmetic.

`RackSyncManager` wires a chain by asking each plugin how many channels it has
(`getAudioInputCount`, `getAudioOutputCount`) and then:

- a device reporting one input takes only pin 1, and a device reporting one output feeds pin 1 to
  both sides of the bus downstream;
- a device reporting no audio input is not connected to the incoming bus at all, so the bus
  survives past it;
- an instrument pushes itself onto the audio bus sources without clearing them, so its output
  sums with whatever audio was already flowing rather than replacing it;
- several sources feeding one destination pin sum implicitly, which is the whole shared-pin
  behaviour.

`Compiler::emitDevice` does none of this. It feeds `signal.audio` into every device op and takes
that op's output as the new chain audio, whatever the device reports. An instrument placed after
an audio source in a chain therefore drops the audio the incumbent would have summed, and a mono
device is handed and read as stereo.

What this slice settles: channel counts on plan ports, a device declaring its audio input and
output width, an injector shape for devices with no audio input, and summing where more than one
source lands on the same input.

PR #2140, opened 2026-08-22
Give an output pair the track that opened it (#2135)

### LINKED ISSUE #2135: Rack graph slice 1: aux outputs and multi-out tracks
Slice 1 of #1892. Depends on the engine core.

Rack aux outputs and the multi-out tracks they feed. Two diagnostics today, one shape underneath.

`PlanCompiler.cpp:614` drops every rack chain whose `outputIndex` is not 0, and says so rather
than folding it into the main mix. `PlanCompiler.cpp:722` does the same for `track.multiOutLink`,
where a track renders silence because the output pair of the instrument it is fed by is not
routed. `RackSyncManager::ensureRackOutputPair` is what the incumbent does instead: it grows the
rack's output list until the pair exists, and wires the chain to pins `1 + index * 2` and
`2 + index * 2`.

What this slice settles:

- A rack instance has more than one audio output. The plan's rack ops carry an output pair rather
  than a single port, and `RackMix` becomes one mix per pair rather than one per rack.
- `MultiOutTrackLink` resolves to an edge from the source device's pair to the linked track's
  chain head, and the compile order carries it. The comment at `PlanCompiler.cpp:366` records the
  current state honestly ("multi-out is not compiled at all, so it contributes no edge either");
  the edge lands here and the ordering has to follow.
- The aux pair a chain leaves by is part of its op key, so moving a chain between outputs is a
  differ change rather than a rebuild.
- A chain routed to a pair no track links to renders nowhere, and that stays a diagnostic rather
  than becoming silent main-mix signal.

lucaromagnoli:
Closes #2135. Slice 1 of #1892.

The issue asked for aux outputs and multi-out tracks as one thing. They are two, and only one of them was work.

## The one that was work

A multi-out instrument's further output pairs are now ports on its own `Device` op, after the main audio and any MIDI. There is one port per pair the device declares rather than one per pair a track opened, so a port's position is its pair number and the device writes its outputs where its own layout says they go without having to ask what is being listened to.

That is what the current engine does. `InstrumentRackManager::wrapMultiOutInstrument` puts the instrument in a rack, the instrument writes every pin of that rack, and it is the `RackInstance` made for a pair, with its own `leftFrom`/`rightFrom`, that decides whether anyone reads those pins. A pair nobody opened is rendered and dropped in both engines.

The MultiOut track reading a pair picks it up through `pendingInputs_`, the same route a send or a group child takes into a track compiled after it, and the link is an ordering edge so the source is always compiled first. The comment in `computeTrackOrder` saying multi-out contributes no edge because it is not compiled goes with it.

Each pair port is cleared before the device is called and before anything that can leave the op early. A pair the device has nothing to play on this block therefore goes silent instead of handing the track reading it the block it last had material for.

## The one that was not

`PlanCompiler.cpp:614` reported a rack chain routed to an aux output as "not routed yet", and its comment said those chains feed multi-out tracks. They do not, and it is not the same mechanism: a `MultiOutTrackLink` names a device, `ChainInfo::outputIndex` names a chain of a user rack, and nothing connects the two.

Such a chain is silent in the current engine too. `RackSyncManager` wires it to output pins 3 and up via `ensureRackOutputPair`, and the rack instance it makes for the track never sets `leftFrom`/`rightFrom`, so it reads pins 1 and 2 and nothing is on the other end. `InstrumentRackManager` is the only caller that sets those properties, and only for multi-out instruments.

Compiling nothing is therefore already parity. The message now says the chain reaches nothing rather than promising to route it later. The path is also reachable only from the agent DSL: `setChainOutput` has no UI caller.

## The second commit

The first pass judged a link by whether the device that owns the pair reached emission. Everything that keeps an instrument out of the plan then read as a broken link. Bypass I had handled by resolving ahead of the bypass check; chain power was the other way in, and `emitTrack` only walks `fxChainElements` when `chain.enabled`, so powering off the source track's insert chain told the MultiOut track its pair does not exist.

Patching the second case would have left the third. Whether a link names a pair that exists is a question about the model, so it is answered from the model now, once, before emission. Whether the device is compiled is a separate question, and a pair whose instrument is bypassed or whose chain is powered off is silent for the same reason everything else on that chain is.

Found by the stop-time review, not by me.

## Cases

Nine, five on the compiler and four on the executor:

| Case | What it pins |
| --- | --- |
| pair feeds the track that opened it | routing, and a port per declared pair |
| a link the device cannot honour | a missing pair, a device that is not multi-out, pair 0, two tracks on one pair |
| a bypassed instrument | silent, and not reported |
| a link survives its source not being compiled | chain power off; a link to a missing track still reported |
| multi-out that also writes MIDI | pairs sit after the MIDI port |
| an aux-routed rack chain | reaches nothing, and says so |
| each pair reaches its track | 0.1, 0.2 and 0.3 read off three track meters |
| a pair nobody opened | rendered, dropped, and the master carries the rest |
| an ordinary device | handed no extra outputs at all |
| a pair the device stops writing | goes silent rather than repeating its last block |

## Not in this slice

Channel counts. `MultiOutOutputPair` carries `numChannels`, and the current engine honours a mono pair by pointing `leftFrom` and `rightFrom` at the same pin, but `RenderContext::numChannels` is 2 and says so in its own comment. Every pair here is stereo. That is #2138, where the plan gets a channel model and mono devices and instrument injection come with it.

## Two findings from review

**The ordering edge was gated on arithmetic, not on connectivity.** It read the link's pair number and nothing else, while emission asks four further questions: does the pair resolve, did this track win it, is the source chain running, is the instrument bypassed. An edge with no connection behind it can close a loop that is not there, and `computeTrackOrder`'s cycle breaker resolves a loop by dropping a connection, so a real one would have gone to give an imaginary one its ordering. Both sides ask the same question now, through `multiOutPairIsConnected`.

**The pair cap cited the wrong code.** I justified `kMaxMultiOutPairs = 31` by `RackSyncManager`'s pin clamp, which is the user-rack chain mechanism this PR spends a section proving is unrelated. `wrapMultiOutInstrument` adds a pin per channel with no clamp, and a device's pairs are read off the plugin's own buses. The only real bound is the executor's stack budget, so the constant is that now, at 64 pairs, and a device with more is reported rather than quietly shortened. A link past the budget is told which limit it hit instead of being called a broken link.

Both are pinned by cases: a bypassed multi-out source whose reader routes back into it, which is the shape that used to invent a cycle, and a device declaring more pairs than the budget.

## One unrelated commit

Gitleaks commented on this PR about two tokens in `tests/test_remote_permissions.cpp`, a file it does not touch: the scan runs `--no-git --source .` over the whole tree and uploads SARIF, so every PR gets told about every finding in the repository. Neither is a secret. One is sixteen hex characters, the other is the hex alphabet in order, and both are handed to `registerRemoteSecret` so the test can prove the value never reaches a log line. Marked with gitleaks' own inline signature, which is where a reader of the test will see it. Verified against gitleaks 8.18.1, the version CI installs: two findings before, none after, and the tree is clean.

## Verification

`[compiler]`, `[exec]`, `[goldens]`, `[diff]`, `[crossfade]`, `[layout]`, `[properties]`, `[session]` and `[taps]` pass.

https://claude.ai/code/session_012wuYCTJYYofVqPPy81cJmS

### DIFF
--- a/magda/engine/plan/PlanCompiler.cpp
+++ b/magda/engine/plan/PlanCompiler.cpp
@@ -171,6 +171,12 @@ class Compiler {
     void diagnose(std::string message);
 
     const TrackInfo* findTrack(TrackId id) const;
+
+    /// The top-level FX device @p deviceId on @p trackId, which is where a
+    /// multi-out instrument sits and the only place a MultiOutTrackLink can
+    /// name: the link carries no section, and TrackManager::getDevice resolves
+    /// it against the track's own chain rather than inside its racks.
+    const DeviceInfo* findMultiOutSource(TrackId trackId, DeviceId deviceId) const;
     TrackId resolveSendDestination(const SendInfo& send) const;
     std::vector<const TrackInfo*> computeTrackOrder();
 
@@ -254,10 +260,6 @@ class Compiler {
     /// instrument is an FX-section device, which is the one space those ids can
     /// come from.
     std::map<std::pair<TrackId, DeviceId>, std::map<int, TrackId>> multiOutTargets_;
-    /// MultiOut tracks whose link named a pair the model actually has. A link
-    /// that resolved to nothing is what emitTrack reports; a resolved pair on a
-    /// bypassed instrument is silent without being wrong.
-    std::set<TrackId> multiOutResolved_;
     /// Tracks whose MIDI another track reads, through a MIDI sidechain or an
     /// internal MIDI route. Their MIDI clips have to be compiled even when
     /// their own chain has nothing that consumes MIDI.
@@ -301,6 +303,17 @@ const TrackInfo* Compiler::findTrack(TrackId id) const {
     const auto found =
         std::ranges::find_if(tracks_, [id](const TrackInfo& track) { return track.id == id; });
     return found == tracks_.end() ? nullptr : &*found;
+}
+
+const DeviceInfo* Compiler::findMultiOutSource(TrackId trackId, DeviceId deviceId) const {
+    const auto* track = findTrack(trackId);
+    if (track == nullptr)
+        return nullptr;
+
+    for (const auto& element : track->chain.fxChainElements)
+        if (isDevice(element) && getDevice(element).id == deviceId)
+            return &getDevice(element);
+    return nullptr;
 }
 
 bool Compiler::carriesClips(const TrackInfo& track) const {
@@ -524,16 +537,8 @@ ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site
                                        kMaxMultiOutPairs)
                           : 0;
 
-    // Resolved before the bypass check below, because a bypassed instrument
-    // makes its pairs silent rather than unrouted: the link is still the one
-    // the model meant, and reporting it would put a diagnostic in front of an
-    // ordinary edit.
     const auto targets = multiOutTargets_.find({site.trackId, device.id});
     const bool hasTargets = ownsMultiOutPairs && targets != multiOutTargets_.end();
-    if (hasTargets)
-        for (const auto& [pairIndex, targetTrack] : targets->second)
-            if (pairIndex <= multiOutPairCount)
-                multiOutResolved_.insert(targetTrack);
 
     if (device.bypassed)
         return signal;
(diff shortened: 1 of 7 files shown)


magda/engine/plan/PlanCompiler.cpp
```diff
@@ -612,12 +709,18 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
...
+            // too: RackSyncManager wires it to output pins 3 and up, and the
+            // rack instance on the track reads pins 1 and 2, so nothing is on
+            // the other end of them. Compiling nothing is therefore parity,
+            // and it is reported because the model says something the signal
+            // flow does not: routing a chain to an aux output loses it, and
+            // silently dropping it here would look like the plan's doing.
             diagnose("rack " + std::to_string(rack.id) + " chain " + std::to_string(chain.id) +
-                     ": aux output " + std::to_string(chain.outputIndex) + " is not routed yet");
+                     ": aux output " + std::to_string(chain.outputIndex) +
+                     " reaches nothing, the chain is silent");
```
lucaromagnoli, line 723: **P2 — Keep the aux-output work in scope or stop closing #2135**

This branch still skips every chain with `outputIndex != 0`, so the rack-aux acceptance criteria in #2135 remain unimplemented: the rack op does not expose the pair, no per-pair `RackMix` is built, and the pair is not represented in the op key. The revised diagnostic accurately describes current-engine parity, but it does not deliver the behavior this PR says it closes. Please either implement that half here or split/update #2135 and remove `Closes #2135`.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -1043,6 +1148,57 @@ RenderPlan Compiler::run() {
     // head, and a source with no MIDI ops compiled has no chain head to read.
     midiSourceTracks_.insert(noteSourceTracks_.begin(), noteSourceTracks_.end());
 
+    // Which multi-out pairs anyone reads, before any device declares its ports.
+    // A device emits a port per pair a track links to and no others, so a pair
+    // nobody opened costs nothing and a dead port never reaches the plan.
```
lucaromagnoli: **P3 — Make this comment match the emitted layout**

This says only linked pairs become ports, but `emitDevice` intentionally appends every declared pair and the executor test asserts that unopened pairs are rendered and dropped. Stable port position is the core invariant here, so please describe this pass as collecting readers; port declaration remains independent of which pairs are opened.

PR #2153, opened 2026-08-22
Let the transport measure itself (#2071)

### LINKED ISSUE #2071: Transport overflow menu always shows: the width budget is a second, hardcoded copy of the layout
### What happened?

The transport's overflow ("burger") menu shows at 1280x1024 even though the full layout should fit. It is not really a 1280 problem: at the shipped defaults it is on from the first frame at any window width up to about 1288.

`TransportPanel::resized` decides what fits by summing hardcoded per-section width estimates into `required`, then dropping tiers until `W >= required`. Most of those estimates scale off `buttonSize`, which is derived from the panel's **height**:

```cpp
const int buttonSize = juce::jmax(24, H - buttonMargin * 2);   // H - 6
```

At `LayoutConfig::defaultTransportHeight = 48` that is 42 px per button, and the totals come out as:

| window width | required | fits |
| --- | --- | --- |
| 1200 | 1228 | no |
| 1280 | 1288 | no |

MainWindow sizes itself `jmin(1200, workArea.getWidth() - margin * 2)`, so on a 1280 display the window is 1200 wide and needs 1228. The transport spans the full window width (`layoutTransportArea` runs before `layoutSidePanels`), so the side panels are not involved.

### Steps to reproduce

1. Run MAGDA at its default window size on a 1280-wide display.
2. Look at the right of the transport bar.

The overflow button is there, and CPU/QWERTY are hidden, with plenty of visible empty space.

### Expected behavior

The full transport at the default window size on a 1280 display, with the overflow reserved for widths that genuinely cannot fit it.

### The underlying problem, which is worth fixing rather than re-tuning

The numbers above can be nudged until 1280 passes, and the next change to a button or a padding will put it back. The real fault is that `required` is a **second description of the layout**, written in hardcoded pixels, sitting beside the code that actually places the components:

```cpp
const int metroSectionW = 112;
const int gridSectionW  = 6 + 30 + 4 + 44;
const int cursorGroupW  = 10 + boxWidth + 24;
const int cpuW          = 70;
```

Those constants have to agree with the real `setBounds` calls further down, and nothing checks that they do. They have already drifted, which is this bug. Suggested direction:

- **Measure, do not estimate.** Each section reports its own preferred width from the same constants that place it, so the fit decision and the placement cannot disagree. One source of truth per section.
- **Declare the tiers instead of hand-rolling them.** The cascade is currently six copies of `visible = (W >= required); if (!visible) required -= sectionW;`. A list of sections in priority order, each with a width provider, collapses that to one loop and makes the drop order reviewable in one place.
- **Pin it with a test.** The layout decision is pure arithmetic over (W, H), so it can be asserted without a GUI: at the app's own default window size and default transport height, nothing is collapsed. That is the regression that would have caught this on day one, and it is what stops the constants drifting again.

### MAGDA version

Current dev.

### Operating system

All. Reported at 1280x1024.

### LINKED ISSUE #2074: Transport layout should measure itself instead of estimating pixels
Comes out of #2071, which is the bug this would prevent.

### The problem

`TransportPanel::resized` decides what fits by summing hardcoded per-section width estimates:

```cpp
const int metroSectionW = 112;
const int gridSectionW  = 6 + 30 + 4 + 44;
const int cursorGroupW  = 10 + boxWidth + 24;
const int cpuW          = 70;
```

then dropping tiers until `W >= required`. Those numbers are a second description of a layout the placement code below already describes, and nothing checks that the two agree. They have drifted, which is #2071: at the shipped `defaultTransportHeight` the estimate exceeds the window width the app chooses for itself, so the overflow button is on from the first frame at any width up to about 1288.

Re-tuning the constants fixes that one display and leaves two other faults untouched:

- **They ignore density.** `LayoutConfig` has a user-facing spacing density (`applyDensityScale`, `densityScaled()`), and `TransportPanel.cpp` does not call it once. Every constant above is a normal-density guess.
- **They ignore text.** The labels those widths have to contain are set in points (`getUIFont(8.0f)`, `getMonoFont(11.0f)`), and no estimate measures a string. `cpuW = 70` is a guess about how wide "100%" renders; the elastic `lerpi(100, 130, 1100, 1280, W)` on the time boxes is a guess about how wide a timecode is, expressed as breakpoints against window width. A different font, a longer string or a larger density silently overflows.

### The shape of the fix

A measure/arrange pass, with each section measuring itself, so the fit decision and the placement cannot disagree:

```cpp
int  measure(const Metrics&) const;      // intrinsic width
void arrange(juce::Rectangle<int>);      // place children in exactly that
```

`resized()` measures every section, sums, drops by declared priority while the total exceeds the available width, then arranges the survivors. Drift stops being possible because `arrange` is handed the rectangle `measure` produced, and each child's width is written once inside its own section.

Widths then come from what actually determines them:

- `cpuW` becomes `font.getStringWidth("100%") + densityScaled(padding)`
- the time boxes' `lerpi(100, 130, 1100, 1280, W)` becomes `monoFont.getStringWidth("000|00|000") + padding`, with leftover width distributed by flex rather than lerped against magic breakpoints
- button groups become `n * buttonSize + (n - 1) * densityScaled(spacing)`, which is nearly what they are already

`juce::FlexBox` is a good fit for distributing the slack once the surviving set is known, but it has no notion of dropping a section entirely, so the priority cascade stays. It should be a declared ordered list rather than the six hand-written copies of `visible = (W >= required); if (!visible) required -= sectionW;` that exist today, so the drop order is reviewable in one place.

### Why this is worth doing rather than re-tuning

The decision becomes a pure function of (width, height, font metrics, density) to the set of visible sections, which makes it testable without a GUI:

- nothing collapses at the app's own default window size and transport height
- the drop order is the one declared
- both still hold at a non-default density

That last group is what stops #2071 recurring. Today there is no test that can express it, because the answer depends on constants that live apart from the layout they describe.

lucaromagnoli:
Fixes #2071. Closes #2074.

## What was wrong

`TransportPanel::resized` decided what fitted by summing hardcoded per-section width estimates, then dropping tiers until the total came under the panel width. Those estimates were a second description of a layout the `setBounds` calls below already described, and nothing checked that the two agreed.

They had drifted, in both directions:

| section | counted | placed |
| --- | --- | --- |
| grid cluster | `6 + 30 + 4 + 44` = 84 | `6 + 56 + 4 + 44` = 110 |
| right cluster | `cpuW + 8 + qwertyW + 4` | no 8px gap in the placement |
| metronome | flat 112 | a 92px box centred in it |

At the shipped `defaultTransportHeight` the sum came out above the width MainWindow chooses for itself, so the overflow button was on from the first frame at any width up to about 1288.

Two further faults that re-tuning the constants would have left in place: `TransportPanel.cpp` never called `densityScaled()` once, so every width was a normal-density guess while spacing density is a user-facing preference; and the labels those widths had to contain are set in points, with no estimate measuring a string.

## What replaces them

**`TransportLayout`** is a measure/arrange pass. Each section is one function that places its children from an origin and returns the width it consumed, and measuring a section is running that same function into a throwaway layout:

```cpp
int measure(Builder build, const Metrics& m) {
    Layout scratch;
    return build(scratch, m, 0);
}
```

The width the fit decision counts and the width the placement occupies are therefore the same number by construction. The six hand-written copies of `visible = (W >= required); if (!visible) required -= sectionW;` become one loop over a declared `kDropOrder`, so the priority is reviewable in one place.

**`TransportTextWidths`** measures each string in the font that draws it: the CPU readout from `"100%"`, the timecode boxes from the widest bar number and the share of the box the bars segment gets, AUTO/SNAP and the division button from their captions. It is a separate translation unit so the arithmetic stays font-free and testable without a graphics context.

Spacing tokens are density-scaled. Leftover width is shared out among the readouts up to a cap, instead of `lerpi(100, 130, 1100, 1280, W)` against magic window-width breakpoints.

## Tests

`compute()` is a pure function of (width, height, measured text, density), so `test_transport_layout.cpp` asserts it headlessly:

- nothing collapses at `LayoutConfig::defaultWindowWidth` and the default transport height, and keeps not collapsing with strings ~15% wider than the shipped font draws
- the drop order is the declared one, and a section that is on stays on as the panel widens
- the arranged flow never crosses the right cluster, at every width from 600 to 1600
- the overflow button appears exactly when something is hidden
- it still fits at every spacing density the preference offers

`test_transport_layout_juce.cpp` runs the same check against the widths the shipped fonts really produce, which is the half a headless test cannot speak for.

## Also in here

- The CPU frame's header separator was a second guess at where the title label ends; it now reads the boundary the layout drew.
- Arming automation write ignored whether the banner had room, so it could put a zero-width label on screen. One method decides that now.
- The section widths come from the fonts, so a live font change has to re-measure them. `applyFontFromConfig` broadcasts a look-and-feel change, which repaints but does not relayout; `TransportPanel::lookAndFeelChanged` now relayouts too. That same function only watched the font *family*, so changing the font *scale* broadcast nothing at all.

## Two things worth knowing

- The time band's trailing pad was 24px against a 10px lead pad. I balanced it to 10, which moves the grid cluster 14px left. Without it, spacious density on a 1200px window missed by 6px.
- The icon buttons are square, so a taller transport is a wider one. Dragged to `maxTransportHeight` on a 1200px window, the right cluster still collapses. There is a test asserting that only that section gives, and nothing else. Capping button growth above the default height would close it, but that is a design call rather than a bug fix.

https://claude.ai/code/session_012wuYCTJYYofVqPPy81cJmS

### DIFF
--- a/magda/daw/ui/components/common/BarsBeatsTicksLabel.cpp
+++ b/magda/daw/ui/components/common/BarsBeatsTicksLabel.cpp
@@ -152,7 +152,7 @@ void BarsBeatsTicksLabel::paint(juce::Graphics& g) {
     }
 
     g.setColour(getTextColour());
-    g.setFont(FontManager::getInstance().getUIFont(10.0f));
+    g.setFont(FontManager::getInstance().getUIFont(kTextFontSize));
 
     if (textOverride_.isNotEmpty()) {
         // Draw override text centred in bounds
@@ -251,7 +251,7 @@ juce::String BarsBeatsTicksLabel::SegmentLabel::formatDisplay() const {
 void BarsBeatsTicksLabel::SegmentLabel::paint(juce::Graphics& g) {
     if (!isEditing_ && owner_.textOverride_.isEmpty()) {
         g.setColour(owner_.getTextColour());
-        g.setFont(FontManager::getInstance().getUIFont(10.0f));
+        g.setFont(FontManager::getInstance().getUIFont(kTextFontSize));
         g.drawText(formatDisplay(), getLocalBounds(), juce::Justification::centred, false);
     }
 }
@@ -351,7 +351,7 @@ void BarsBeatsTicksLabel::SegmentLabel::startEditing() {
 
     editor_ = std::make_unique<juce::TextEditor>();
     editor_->setBounds(getLocalBounds());
-    editor_->setFont(FontManager::getInstance().getUIFont(10.0f));
+    editor_->setFont(FontManager::getInstance().getUIFont(kTextFontSize));
     editor_->setText(formatDisplay(), false);
     editor_->selectAll();
     editor_->setJustification(juce::Justification::centred);
--- a/magda/daw/ui/components/common/BarsBeatsTicksLabel.cpp
+++ b/magda/daw/ui/components/common/BarsBeatsTicksLabel.cpp
@@ -203,22 +203,59 @@ bool BarsBeatsTicksLabel::isDragging() const {
            (ticksSegment_ && ticksSegment_->isDragging());
 }
 
-void BarsBeatsTicksLabel::resized() {
-    auto bounds = getLocalBounds().reduced(2, 0);
-    int totalWidth = bounds.getWidth();
+std::array<int, 3> BarsBeatsTicksLabel::segmentWidthsFor(double maxValue, int minBeatsPerBar,
+                                                         int maxBeatsPerBar, bool isPosition) {
+    const int offset = isPosition ? 1 : 0;
+    const int lowBeatsPerBar = juce::jmax(1, minBeatsPerBar);
+    const int highBeatsPerBar = juce::jmax(lowBeatsPerBar, maxBeatsPerBar);
 
-    // Proportions: bars ~25%, dot ~5%, beats ~25%, dot ~5%, ticks ~40%
-    int dotWidth = 6;
-    int availableWidth = totalWidth - dotWidth * 2;
-    int barsWidth = static_cast<int>(availableWidth * 0.25);
-    int beatsWidth = static_cast<int>(availableWidth * 0.25);
-    int ticksWidth = availableWidth - barsWidth - beatsWidth;
+    // Fewest beats per bar gives the most bars; most beats per bar gives the
+    // highest beat number. Ticks are zero-padded to three digits either way.
+    const int maxBars = static_cast<int>(maxValue / lowBeatsPerBar) + offset;
+    const int maxBeats = highBeatsPerBar - 1 + offset;
+
+    const auto font = FontManager::getInstance().getUIFont(kTextFontSize);
+    // Measure a string of the widest digit rather than the number itself, so a
+    // value made of narrow digits does not under-size the segment.
+    const auto widthOfDigits = [&font](int digits) {
+        return juce::GlyphArrangement::getStringWidthInt(
+                   font, juce::String::repeatedString("8", digits)) +
+               kSegmentPad;
+    };
+
+    return {widthOfDigits(juce::String(maxBars).length()),
+            widthOfDigits(juce::String(maxBeats).length()), widthOfDigits(3)};
+}
+
+int BarsBeatsTicksLabel::preferredWidthForRange(double maxValue, int minBeatsPerBar,
+                                                int maxBeatsPerBar, bool isPosition) {
+    const auto needed = segmentWidthsFor(maxValue, minBeatsPerBar, maxBeatsPerBar, isPosition);
+    return needed[0] + needed[1] + needed[2] + (2 * kDotWidth) + (2 * kEdgeInset);
+}
+
+std::array<int, 3> BarsBeatsTicksLabel::segmentWidths() const {
+    return segmentWidthsFor(maxValue_, beatsPerBar_, beatsPerBar_, barsBeatsIsPosition_);
+}
+
+void BarsBeatsTicksLabel::resized() {
+    auto bounds = getLocalBounds().reduced(kEdgeInset, 0);
+
+    // Share the strip out in proportion to what each segment has to show. A
+    // fixed quarter for the bar number clipped long positions while leaving
+    // the three-digit tick segment half the box.
+    const auto needed = segmentWidths();
+    const int total = juce::jmax(1, needed[0] + needed[1] + needed[2]);
+    const int available = juce::jmax(3, bounds.getWidth() - (2 * kDotWidth));
+
+    const int barsWidth = available * needed[0] / total;
+    const int beatsWidth = available * needed[1] / total;
+    const int ticksWidth = available - barsWidth - beatsWidth;
 
     int x = bounds.getX();
     barsSegment_->setBounds(x, bounds.getY(), barsWidth, bounds.getHeight());
-    x += barsWidth + dotWidth;
+    x += barsWidth + kDotWidth;
     beatsSegment_->setBounds(x, bounds.getY(), beatsWidth, bounds.getHeight());
-    x += beatsWidth + dotWidth;
+    x += beatsWidth + kDotWidth;
     ticksSegment_->setBounds(x, bounds.getY(), ticksWidth, bounds.getHeight());
 }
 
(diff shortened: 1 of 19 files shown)


magda/daw/ui/windows/MainWindow.cpp
```diff
@@ -367,15 +367,18 @@ void MainWindow::applyDensityFromConfig() {
 }
 
 void MainWindow::applyFontFromConfig() {
-    const auto& family = Config::getInstance().getUIFontFamily();
-    if (family == appliedFontFamily_)
+    const auto& config = Config::getInstance();
+    const auto& family = config.getUIFontFamily();
+    const auto scale = config.getUIFontScale();
```
lucaromagnoli, line 372: This now runs during Config::save(), before PreferencesDialog::refreshFontScaleForOpenWindows(). TransportPanel::lookAndFeelChanged installs fonts at the new scale, then scaleExplicitFonts recursively multiplies every juce::Label by new/old again. At 125%, the CPU labels and automation banner end up at 156.25% while measureTextWidths assumes 125%. Please coordinate or remove the second scaling path and test the actual Preferences apply sequence.

magda/daw/ui/panels/TransportTextWidths.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    TextWidths text;
+    // The bars segment is the widest of the three: beats stop at the time
+    // signature and ticks at three digits.
+    text.barNumber = widthOf(timecodeFont, "0000");
+    text.tempo = widthOf(readoutFont, juce::String(MAX_VALID_BPM, 2));
+    text.timeSigNumerator = widthOf(readoutFont, juce::String(MAX_TIME_SIGNATURE_VALUE) + "/");
+    text.timeSigDenominator = widthOf(readoutFont, juce::String(MAX_TIME_SIGNATURE_VALUE));
+    text.cpuTitle = widthOf(cpuTitleFont, tr("transport.cpu.cpu"));
+    text.cpuValue = widthOf(cpuValueFont, "100%");
```
lucaromagnoli: setCpuUsage() also emits retained-peak values such as 97/100% whenever the peak exceeds the average. Measuring only 100% shrinks the new CPU slot to an inner width of about 35 px at the default font, while the seven-character monospace value needs about 40 px, so normal peak decay clips. Please measure the widest reachable avg/peak form.

magda/daw/ui/panels/TransportTextWidths.cpp
```diff
@@ -0,0 +1,53 @@
...
+    const auto divisionFont = fonts.getUIFontBold(GridDivisionButton::kFontSize);
+    const auto toggleFont =
+        fonts.getUIFontBold(SmallButtonLookAndFeel::getInstance().getFontSize());
+    const auto cpuTitleFont = fonts.getUIFont(kCpuTitleFontSize);
+    const auto cpuValueFont = fonts.getMonoFont(kCpuValueFontSize);
+
+    TextWidths text;
+    // The bars segment is the widest of the three: beats stop at the time
+    // signature and ticks at three digits.
+    text.barNumber = widthOf(timecodeFont, "0000");
```
lucaromagnoli: These labels accept up to 100000 beats and display positions one-indexed. That can render 25001 in 4/4 and 100001 in 1/1, not four digits. Because intrinsicTimeBoxWidth gives the bars segment exactly this measured width at tight sizes, long positions clip. Please derive the sample from the configured range and minimum beats-per-bar, or expose the maximum display string from BarsBeatsTicksLabel.

magda/daw/ui/components/common/BarsBeatsTicksLabel.cpp
```diff
@@ -203,22 +203,59 @@ bool BarsBeatsTicksLabel::isDragging() const {
...
+
 void BarsBeatsTicksLabel::resized() {
-    auto bounds = getLocalBounds().reduced(2, 0);
-    int totalWidth = bounds.getWidth();
+    auto bounds = getLocalBounds().reduced(kEdgeInset, 0);
+
+    // Share the strip out in proportion to what each segment has to show. A
+    // fixed quarter for the bar number clipped long positions while leaving
+    // the three-digit tick segment half the box.
+    const auto needed = segmentWidths();
```
lucaromagnoli, line 260: The segment geometry now depends on maxValue_, beatsPerBar_, and barsBeatsIsPosition_, but setRange(), setBeatsPerBar(), and setBarsBeatsIsPosition() still only update the text and repaint. For example, TransportPanel::setTimeSignature can change 4/4 to 16/16 without changing this component bounds, leaving the beat segment at its old one-digit width until some later parent resize; two-digit beats can clip and the interaction regions stay stale. Please trigger a relayout from those setters and test changing the meter without an outer resize.

PR #2154, opened 2026-08-22
Give the things that drive MAGDA a section of their own (#2142)

### LINKED ISSUE #2142: Remote clients are not an AI setting
**AI Settings → Clients** governs permissions for every remote API client, not
just AI ones. The same `RemoteClientRegistry` backs the WebSocket transport, so
a Lua script, a bespoke controller, or anything else holding the bearer token
appears in that list and is gated by those checkboxes. None of it is AI.

It sits there because that is where the feature arrived from: the remote API was
built as the MCP surface, **Remote** turns the listener on, so **Clients** landed
beside it. The grouping now says something untrue about its scope — a user
looking for "what is allowed to drive my DAW" has no reason to open AI Settings,
and a user who does open it is told, by placement, that these grants only concern
models.

The giveaway is OSC. Control surfaces are configured in **Controllers**, so two
kinds of "things that drive MAGDA from outside" live in two unrelated dialogs,
split by which feature shipped first rather than by what they are. The permission
vocabulary is shared — `read`, `edit`, `transport`, `session`, `hardware-midi`
are about what a *client* may do, not about what a *model* may do.

## Suggested shape

A **Remote** (or **Connections**) section holding:

- the listener toggle and token controls, currently AI Settings → Remote
- the client list and grants, currently AI Settings → Clients
- OSC, currently in Controllers

leaving AI Settings for model and provider configuration, which is what its name
promises.

A UI move only — no change to `RemoteClientRegistry`, the scope vocabulary, the
transports, or the discovery record. Nothing in the API depends on where the tab
lives.

## Noticed while

Verifying #2059 against a running build. The tab itself renders correctly, which
was one of that issue's open items — the layout is fine and the activity log is
genuinely useful, showing `tracks.create denied (edit)` next to the checkbox that
would allow it. One small nit worth folding in if this is touched: each client's
name sits above its own checkbox row with no separator or grouping, so when the
list is scrolled mid-row it is easy to read a row of checkboxes as belonging to
the name below it. A hairline divider between clients would settle it.

lucaromagnoli:
Closes #2142.

**AI Settings → Clients** governed every remote API client, not just AI ones. The same `RemoteClientRegistry` backs the WebSocket, so a Lua script or a bespoke controller holding the token was gated by checkboxes filed under AI. OSC made the split obvious: two kinds of "things that drive MAGDA from outside" living in two unrelated dialogs, divided by which feature shipped first.

**Settings → Connections**, four tabs: `MCP | WebSocket | OSC | Clients`. AI Settings keeps model and provider configuration, which is what its name promises; Controllers keeps MIDI profiles and Lua scripts.

## Why this is more than the UI move the issue asked for

The move exposed that the page was lying about its own scope. "Remote" gated **two** listeners while its copy named only MCP, never showed the WebSocket port at all, and rendered ws-alive/mcp-dead as a failure that `RemoteApiHost`'s own header calls a supported state. Leaving that wording in place — and freezing it into Crowdin source strings — would have undercut the point of the move, so the transports are now genuinely independent.

- **Config** carries `remoteApi.websocket` and `remoteApi.mcp` in place of one `enabled`. A file predating the split resolves to both, and `enabled` is still written as `(ws || mcp)` so an older build reads it the way it always did. That rule lives in `remoteApiSwitchesFrom()`, lifted out of `Config::load` because `Config` is a private-ctor singleton that reads from disk — untestable in place, and this is the one rule where getting it wrong silently switches off a transport somebody was already using.

- **`RemoteApiHost`** gained `startTransport` / `stopTransport` / `isRunning(Transport)` and `rotateToken(Transport)`. MCP-only was not expressible before: the WebSocket was the anchor for both `isRunning()` and the discovery record.

- **One credential per transport**, published as `token`/`url`/`port` and `mcpToken`/`mcpUrl`/`mcpPort`, each group present only while its listener is up. This buys **rotation granularity, not isolation** — both tokens sit in the same owner-only file and both doors open onto the same operations and the same grants — but it means re-credentialling a misbehaving script no longer drops an AI host mid-conversation. The bridge and the harness prefer `mcpToken` and fall back to `token`, so a client written before the split keeps working.

## Also in here

- **Localisation.** Every string on the two remote pages now comes from `lang/en.json`; they were hardcoded English inside AI Settings. English only — translations go through Crowdin. Scope names (`read`, `edit`, …) are deliberately left untranslated: they are wire identifiers, and a translated checkbox beside an untranslated `tracks.create denied (edit)` would break the link the page exists to make.
- **The nit from the issue.** Client rows are outlined, not just tinted — a name sits directly above its own checkboxes and the next name directly below them, so a list scrolled mid-row read as checkboxes belonging to the name underneath.
- **Docs.** `remote-api-mcp.md` gained a "Two transports, one record" section; the contract and permissions docs no longer claim one shared token; new `manual/docs/interface/connections.md` leads with the two-listener model and states that OSC is not one of them.

## Verification

Run against a **pre-split config** on a real install: it migrated to both transports and the record carried two distinct tokens.

```
python3 tools/transport_check --all   →  48 passed, 4 skipped
make test                             →  exit 0
```

The harness exercises MCP over the new `mcpToken`, the WebSocket over `token`, and the rebuilt `magda-mcp` bridge.

New coverage in `tests/test_remote_api_host.cpp`: either transport alone, per-transport rotation leaving the other untouched, stopping one keeping the other's record entry, and the config migration table.

One regression the suite caught while I was writing this: the old `start()` deleted a stale same-pid record *before* the enabled check, and my early-returning rewrite skipped it. Fixed by publishing at the end of `start()`, which deletes the record when nothing came up. The test failed before the fix and passes after.

## Not verified

The four tabs on screen. macOS reset the mic permission when I restarted the app and `osascript` has no assistive access in this environment, so I could not drive the menu. The 53 strings resolve and are present in the built bundle, but the layout is unseen.


lucaromagnoli:
Verified on a running build in all three transport configurations, by toggling the switches in the dialog:

| Config | Harness | Discovery record |
|---|---|---|
| both on | 48 passed, 4 skipped | both groups, two distinct tokens |
| MCP off | 17 passed, 9 skipped | `port` / `token` / `url` only |
| WebSocket off | 35 passed, 5 skipped | `mcpPort` / `mcpToken` / `mcpUrl` only |

MCP-only is the case that could not be expressed before this branch — the WebSocket was the anchor for both `isRunning()` and the record — and the `magda-mcp` bridge passes there.

The MCP-off run initially failed four checks, fixed in a442500c. All four predate the split honestly: while one flag gated both listeners, "the record has no `mcpUrl`" could only mean the endpoint failed to bind, so `run_discovery` asserted on it and the bridge suite ran regardless. The record can't distinguish "switched off" from "port was taken" — that lives in config, which the harness deliberately doesn't read — so those checks now skip and name both causes, in both directions. If *neither* group is present it still fails, since nothing is reachable.

The bridge's dead-end message also told the user to enable the remote API, which in that configuration they already had; it now names `Settings > Connections > MCP`.

Clients tab confirmed visually — rows outlined so each name is bound to its own checkboxes (the nit in #2142), `read` ticked and locked, per-row Disconnect live only for the connected client, and the activity tail showing both rejection paths. MCP and WebSocket tab layouts are still unseen.

PR #2157, opened 2026-08-22
Let a window's background follow the theme it is shown under

### LINKED ISSUE #2073: Top border keeps its old colour after switching themes
### What happened?

After switching themes, the top border keeps the colour from the theme that was active before. Everything around it repaints in the new theme, so the border is left as a stripe of the old one until the app is restarted.

### Steps to reproduce

1. Open MAGDA on one theme.
2. Switch to a theme with a visibly different border colour.
3. Look at the top border.

It is still the old colour.

### Expected behavior

The border follows the theme with everything else, without a restart.

This has the shape of the cached-colour case rather than the paint-time one: a role read during `paint()` picks the new palette up on the next repaint by itself, but a colour pushed into a component once with `setColour` keeps whatever it was given until something calls `lookAndFeelChanged` on it. If the border is drawn from a stored colour or a `setColour` made at construction, that would explain it surviving a theme switch while its neighbours update.

Worth confirming whether anything else set the same way is also stale, since one missed refresh usually is not the only one.

### MAGDA version

Current dev.

### Operating system

All.

lucaromagnoli:
Fixes #2073.

## What was actually stale

Every border MAGDA *draws* resolves its colour inside `paint()`, so the repaint that a look-and-feel change triggers already heals it. That covers all of `MainView`, `TabbedPanel`, `FooterBar`, `PanelTabBar`, `TimelineHeaderPanel`, `MarkerLaneComponent` and the piano-roll lanes. None of them was the strip in the report.

The strip is not drawn at all. `LookAndFeel_V4::drawMenuBarBackground` fills at 40% alpha, so what shows through the in-window menu bar is the window's own background. That background is a per-component colour override handed to `DocumentWindow` once at construction, and JUCE keeps such overrides across a look-and-feel change, so it stays on whichever theme was active when the window was built.

Sampling the reported screenshot pins it exactly. The strip measured `#979A9D`; the app body measured `#0C0F14`, the dark theme's `BACKGROUND`:

| | value |
|---|---|
| dark `BUTTON_NORMAL` at 40% over **light** `BACKGROUND` (`#F4F6F8`) | `(151, 154, 157)` |
| measured strip | `(151, 154, 157)` |
| after this change (over live dark `BACKGROUND`) | `(12, 15, 20)` |

So the window was constructed under Light and never let go of it, while everything around it repainted in Dark.

## The change

`refreshHostWindowBackground()` in `DarkTheme.hpp`, and a `lookAndFeelChanged()` on each place that froze the same colour:

- **Windows**: `MainWindow` (the reported one), `FloatingHostWindow`, `LFOCurveEditorWindow`, `FaustCodeEditorWindow`, `SessionClipEditorWindow`, `DebugDialog`, `MediaDbBrowserContent::PopOutWindow`, `SplashScreen`
- **Dialog contents** that hand a background to their `DialogWindow`: AI settings, audio settings, audio export, MIDI export, gain staging, parameter config, project settings. `PreferencesDialog` had already grown this inline and now shares the helper.
- **LookAndFeels**: `MixerLookAndFeel` and `FileBrowserLookAndFeel` cached theme colours in their constructors. A `LookAndFeel` is not a `Component`, so no broadcast reaches it. Both gained a `refreshThemeColours()`; the mixer one is called by its three owners, the file browser singleton by `refreshThemedLookAndFeels()`.

The window-level fix is what the issue reported; the rest is the sibling sweep it asked for. The `setColour`-on-child-widget cases (search boxes, combos, inspector labels) are a separate and much larger family and are not touched here.

## Verification

- `make debug` clean
- `make test`: 2877 cases, 599511 assertions, 0 failures (13 pre-existing skips)
- `make test-juce-build` clean
- `make format` clean

Not verified on a live theme switch on Windows; the arithmetic above is what identifies the defect, and the mechanism is platform-independent (the strip is only visible where the menu bar is in-window, so Windows and Linux).

### DIFF
--- a/magda/daw/ui/components/chain/custom_ui/FaustCodeEditorWindow.cpp
+++ b/magda/daw/ui/components/chain/custom_ui/FaustCodeEditorWindow.cpp
@@ -126,4 +126,9 @@ void FaustCodeEditorWindow::closeButtonPressed() {
     setVisible(false);
 }
 
+void FaustCodeEditorWindow::lookAndFeelChanged() {
+    juce::DocumentWindow::lookAndFeelChanged();
+    setBackgroundColour(DarkTheme::getColour(DarkTheme::BACKGROUND));
+}
+
 }  // namespace magda::daw::ui
--- a/magda/daw/ui/components/chain/custom_ui/FaustCodeEditorWindow.hpp
+++ b/magda/daw/ui/components/chain/custom_ui/FaustCodeEditorWindow.hpp
@@ -20,6 +20,11 @@ class FaustCodeEditorWindow : public juce::DocumentWindow {
 
     void closeButtonPressed() override;
 
+    // The background is a per-component colour override, so it outlives the
+    // look-and-feel change that repaints everything else. Re-resolve it, or a
+    // window left open across a theme switch keeps the old palette.
+    void lookAndFeelChanged() override;
+
   private:
     class Content;
     std::unique_ptr<Content> content_;
--- a/magda/daw/ui/components/chain/modulation/LFOCurveEditorWindow.cpp
+++ b/magda/daw/ui/components/chain/modulation/LFOCurveEditorWindow.cpp
@@ -587,4 +587,9 @@ void LFOCurveEditorWindow::closeButtonPressed() {
     }
 }
 
+void LFOCurveEditorWindow::lookAndFeelChanged() {
+    juce::DocumentWindow::lookAndFeelChanged();
+    setBackgroundColour(DarkTheme::getColour(DarkTheme::BACKGROUND));
+}
+
 }  // namespace magda::daw::ui
--- a/magda/daw/ui/components/chain/modulation/LFOCurveEditorWindow.hpp
+++ b/magda/daw/ui/components/chain/modulation/LFOCurveEditorWindow.hpp
@@ -87,6 +87,11 @@ class LFOCurveEditorWindow : public juce::DocumentWindow {
 
     void closeButtonPressed() override;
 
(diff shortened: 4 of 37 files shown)


lucaromagnoli:
**P2 — Refresh the AI Prompt dialog host too**

`ParameterConfigDialog::lookAndFeelChanged()` only reaches the window hosting `ParameterConfigDialog`. `showAiPromptEditor()` creates a separate `AIPromptEditorComponent` dialog with `dialogBackgroundColour` resolved from the current theme, but that content component has no `lookAndFeelChanged()`. If the prompt editor remains open during a theme switch, its host window retains the old background. Please add the same `refreshHostWindowBackground(*this)` callback to `AIPromptEditorComponent`.

https://github.com/Conceptual-Machines/magda-core/blob/3b645b31b5a90ab88811bd4c72dbd96f988262b3/magda/daw/ui/dialogs/ParameterConfigDialog.cpp#L112-L128

lucaromagnoli:
Confirmed and fixed in `Refresh the AI prompt editor's host window too`.

`AIPromptEditorComponent` is launched with `launchAsync()` into a `DialogWindow` of its own with its own frozen `dialogBackgroundColour`, so it can be on screen when the theme changes under it, and `ParameterConfigDialog::lookAndFeelChanged()` never reaches it. It now calls `refreshHostWindowBackground(*this)`.

I re-audited the rest of the family rather than fixing just this one: every `dialogBackgroundColour` site in the tree, resolved to the content class it actually owns. The other two launches in this file host `ParameterConfigDialog` itself, and the remaining seven are the dialog contents already covered by this PR. This was the only gap.

PR #2161, opened 2026-08-22
Subtract the dry signal a device or rack was handed (#2136)

### LINKED ISSUE #2136: Rack graph slice 2: delta solo at device and rack scope
Slice 2 of #1892. Depends on the engine core.

Delta solo, at device and rack scope. Both sites are already named in the compiler and both say
the same thing: the alignment is a delay like any other now, and what is missing is the edge
carrying the dry signal past the processing and the op that subtracts it.

- `PlanCompiler.cpp:500`, device scope: subtract the device's dry input, aligned to its output,
  from that output.
- `PlanCompiler.cpp:588`, rack scope: the same one level up, around a rack instance. The current
  engine keeps a dry edge around every rack instance for exactly this and delays it to match the
  wet path.

What this slice settles:

- A subtract op, with its own kind rather than a gain of -1 into a mix, so the differ and the
  dump can name it.
- Where the dry edge is taken from. For a device it is the input the device was handed, after any
  sidechain and MIDI wiring, so a bypassed device's delta is silence rather than the chain.
- That the alignment uses the existing PDC delay rather than a second latency concept.
- Delta solo on a device inside a rack inside a rack, where the dry edge crosses no boundary the
  op keys do not already carry.

lucaromagnoli:
Slice 2 of #1892. Closes #2136.

Delta solo, at device and rack scope. Both sites were already named in the compiler and both said the same thing: the alignment is a delay like any other now, and what was missing was the edge carrying the dry signal past the processing and the op that subtracts it.

## What it settles

**A subtract with its own kind.** `OpKind::Subtract`, arity two (wet, dry), rather than a gain of -1 into a mix, so the differ, the dump and the layout can name it. Roles: `deviceDelta`, `rackDelta`, and `subtractInputDelay` for its fan-in.

**Where it sits.** For a device, between the process op and the slot's gain and meter — where the current engine subtracts it, since those are MAGDA's own and sit outside the plugin. For a rack, around the instance's own fader, so the rack's volume and pan are inside what is measured: they are the `RackInstance`'s output levels and the current engine applies them on the wet path whether or not a delta is heard. A rack with no chains gets one too, for the same reason its fader still applies there. A transparent tap gets none: its output is its input, so its delta is silence by construction.

**Where the dry edge is taken from.** The chain audio in front of whatever aligned the device's own input, not the device's input slot. Taken from the slot, a sidechained device produces a delay reading a delay — double-counting that edge's compensation, which `validatePlan` already rejects. Taken from in front of it, the subtract's own dry delay resolves to the whole distance instead: the input alignment and the device's latency together.

**The alignment is the existing PDC delay.** Both sides go through `alignInputs`. The wet delay resolves to nothing and is elided; the dry one holds exactly the wet path's distance. No second latency concept.

**It is in every plan, and the flag is a value.** The dry edge's delay is a delay line, and a delay line is history: one created at the moment the button was pressed would hand back its own length in silence, so a device with any real latency would leak its wet signal for that long and then step, well past what a 5 ms graph fade can cover. The current engine keeps the same line running for the same reason — `PluginNode` builds `deltaLatencyProcessor` from the plugin's latency alone and writes into it every block, gating only the read. So the subtract and its delay are compiled unconditionally and the model decides only whether the subtraction happens, through `OpValue::subtractsDry`. Toggling delta solo compiles nothing: the plan is identical either way (pinned by fingerprint), so it publishes values only and the line keeps its contents.

**Nesting.** A device inside a rack inside a rack keys its delta `T1/R5/C11/D7:deviceDelta` — the dry edge crosses no boundary the op keys do not already carry.

## What it costs

A device slot is four ops rather than three, and a device no longer writes its output over its input buffer, because the chain audio now has a second reader — one block copy per device, which is the copy `PluginNode` makes unconditionally anyway. The dry delay elides to nothing where the device reports no latency, which is the rule TE uses to decide whether to build its line at all. Every golden shows this.

## One judgement call worth reviewing

The issue says "a bypassed device's delta is silence rather than the chain". This keeps structural bypass compiling to no delta at all, which is parity: MAGDA's bypass reaches the current engine as `Plugin::setEnabled(false)`, and `PluginNode` skips `applyDeltaSolo` for a plugin that is not enabled, handing back the chain. The reading here is that the delta of a device that does nothing is silence because subtracting its input from its output leaves nothing — which is now literally true for any transparent processing — not because the flag is on. Discussed inline; say the word and it is three lines the other way.

## Tests

- Compiler: the flag changing nothing about the plan, both scopes, the sidechain case, the empty rack, the doubly nested device, transparent taps, bypass.
- Layout: what each dry delay resolves to — including 140 samples for a device with 100 of its own behind a key that arrives 40 late.
- Executor: the difference itself; a unity device netting to silence; a pure-delay device netting to silence, which only passes if the compensation is right; and **the toggle mid-render**, where a device that only delays is silent on the very first block after the flag flips, which is reachable only if the dry line was already running.
- A `delta solo` scene joins the parallel executor's comparison: this is the one op that reads two live buffers and writes over one of them.
- A `deltaSoloDevice` edit joins the random sequence generator. It found that a value-only publish steps rather than ramps; that is exempt in the transient property now, with the reason at `Transient::valueOnly` — nothing moved, so there was nothing to fade, and mute and every fader position have the same shape today. The deep sweep still bounds 1603 transients that did move.
- A `delta-solo` golden pinning a device's delta inside a rack's, and what their dry edges wait for.

Negative checks: removing `alignInputs` from `emitDelta` fails exactly the five alignment assertions; taking the dry edge off the device's input slot trips `validatePlan`'s delay rules; compiling the subtract conditionally fails the mid-render toggle on every sample of the block after the flip.

Full suite green locally (2874 passed, 13 skipped), property sweeps included. Not run under TSan: `make test-tsan` cannot link on this machine.

### DIFF
--- a/magda/engine/plan/PlanCompiler.cpp
+++ b/magda/engine/plan/PlanCompiler.cpp
@@ -213,6 +213,20 @@ class Compiler {
     ChainSignal emitDevice(const DeviceInfo& device, const ChainSite& site, ChainSignal signal);
     ChainSignal emitRack(const RackInfo& rack, const ChainSite& site, ChainSignal signal);
 
+    /// Delta solo: @p wet minus @p dry, which is what the device or rack the
+    /// key names added to the signal. Both sides go through alignInputs, so
+    /// the dry edge meets the wet one through the same PDC delay every other
+    /// fan-in uses rather than through a second latency concept. What that
+    /// delay resolves to is the whole distance the wet path travelled, the
+    /// processing's own latency included.
+    PortRef emitDelta(const OpKey& key, PortRef wet, PortRef dry);
+
+    /// The same thing around a rack instance, or @p wet where the rack is not
+    /// soloing its delta. Two call sites, because a rack with no chains still
+    /// has an output of its own to measure: its fader is applied on the wet
+    /// path whether or not there are chains under it.
+    PortRef emitRackDelta(const RackInfo& rack, const ChainSite& site, PortRef wet, PortRef dry);
+
     /// Sums `sources` into one audio port, always through an op so the op's
     /// identity survives sources appearing and disappearing.
     PortRef emitMix(const OpKey& key, const std::vector<PortRef>& sources);
@@ -496,6 +510,26 @@ PortRef Compiler::emitMix(const OpKey& key, const std::vector<PortRef>& sources)
                    0};
 }
 
+PortRef Compiler::emitDelta(const OpKey& key, PortRef wet, PortRef dry) {
+    return PortRef{addOp(OpKind::Subtract, key, alignInputs(key, OpKind::Subtract, {wet, dry}),
+                         {SignalKind::Audio}),
+                   0};
+}
+
+PortRef Compiler::emitRackDelta(const RackInfo& rack, const ChainSite& site, PortRef wet,
+                                PortRef dry) {
+    if (!rack.deltaSolo)
+        return wet;
+
+    // The dry edge the current engine keeps around every rack instance,
+    // delayed to match the wet path. One level up from a device's, and the
+    // same shape: the rack's output, its own volume and pan included, minus
+    // what reached it.
+    const OpKey key{site.trackId,      rack.id, INVALID_CHAIN_ID, INVALID_DEVICE_ID,
+                    OpRole::RackDelta, 0,       site.segment};
+    return emitDelta(key, wet, dry);
+}
+
 std::vector<PortRef> Compiler::alignInputs(const OpKey& key, OpKind kind,
                                            std::vector<PortRef> inputs) {
     // One input is its own maximum, so its compensation is zero in every
@@ -519,6 +553,8 @@ std::vector<PortRef> Compiler::alignInputs(const OpKey& key, OpKind kind,
                 return OpRole::DeviceInputDelay;
             case OpKind::Fader:
                 return OpRole::FaderInputDelay;
+            case OpKind::Subtract:
+                return OpRole::SubtractInputDelay;
             default:
                 jassertfalse;  // nothing else fans in
                 return OpRole::MixInputDelay;
@@ -572,16 +608,6 @@ ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site
 
     if (device.bypassed)
         return signal;
-
-    // Delta solo subtracts the device's dry input, aligned to its output, from
-    // that output. The alignment is a delay like any other now; what is still
-    // missing is the edge carrying the dry signal past the device and the op
-    // that subtracts it.
-    if (device.deltaSolo)
-        diagnose("device " + std::to_string(device.id) + " on track " +
-                 std::to_string(site.trackId) +
-                 ": delta solo needs a dry edge past the device and an op to subtract it, "
-                 "which the plan does not carry yet");
 
     const auto node = routing::makeRoutingNode(device);
 
@@ -639,6 +665,22 @@ ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site
     ChainSignal out;
     out.audio = PortRef{processOp, 0};
 
+    // Delta solo: what the device added, which is its output minus the dry
+    // signal it was handed. The dry edge is that signal rather than the device's
+    // own input slot, which may already carry a delay lining the audio up with a
+    // sidechain: a delay reading a delay would count that edge's compensation
+    // twice, and taken from in front of it the subtract's own delay resolves to
+    // the whole distance instead, the input alignment and the device's latency
+    // together.
+    //
+    // Ahead of the slot's gain and meter, which is where the current engine
+    // subtracts it too: those are MAGDA's own and sit outside the plugin, so
+    // what they trim and report is the delta once there is one.
+    if (device.deltaSolo) {
+        key.role = OpRole::DeviceDelta;
+        out.audio = emitDelta(key, out.audio, signal.audio);
+    }
+
     if (!isTransparentTap(device)) {
         key.role = OpRole::DeviceGain;
         out.audio = PortRef{addOp(OpKind::Gain, key, {out.audio}, {SignalKind::Audio}), 0};
@@ -677,15 +719,6 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
     // that track is the edge, and collectModulationSources is what puts the
     // source ahead of this rack's own track in the compile order (#2120).
 
-    // Same dry path a device's delta solo needs, one level up: the rack
-    // instance subtracts its own input, aligned to its wet output, from that
-    // output. The current engine keeps a dry edge around every rack instance
-    // for exactly this, and delays it to match the wet path.
-    if (rack.deltaSolo)
-        diagnose("rack " + std::to_string(rack.id) +
-                 ": delta solo needs a dry edge around the rack and an op to subtract it, "
-                 "which the plan does not carry yet");
-
     // A rack with no chains passes signal rather than silence: compiling it to
     // a zero-input mix would silence the track. It is not fully transparent
     // though, because rack volume and pan land on the instance's output gains,
@@ -698,10 +731,9 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
     if (rack.chains.empty()) {
         const OpKey faderKey{site.trackId,      rack.id, INVALID_CHAIN_ID, INVALID_DEVICE_ID,
                              OpRole::RackFader, 0,       site.segment};
-        return {
-            PortRef{addOp(OpKind::Fader, faderKey, {signal.audio, noInput()}, {SignalKind::Audio}),
-                    0},
-            signal.midi};
+        const PortRef faded{
+            addOp(OpKind::Fader, faderKey, {signal.audio, noInput()}, {SignalKind::Audio}), 0};
+        return {emitRackDelta(rack, site, faded, signal.audio), signal.midi};
     }
 
     std::vector<PortRef> chainAudio;
@@ -765,6 +797,7 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
     const OpKey faderKey{site.trackId,      rack.id, INVALID_CHAIN_ID, INVALID_DEVICE_ID,
                          OpRole::RackFader, 0,       site.segment};
     out.audio = PortRef{addOp(OpKind::Fader, faderKey, {mixed, noInput()}, {SignalKind::Audio}), 0};
+    out.audio = emitRackDelta(rack, site, out.audio, signal.audio);
 
     if (chainMidi.empty()) {
         out.midi = signal.midi;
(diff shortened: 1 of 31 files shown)


magda/engine/plan/PlanCompiler.cpp
```diff
@@ -639,6 +665,22 @@ ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site
...
+    // own input slot, which may already carry a delay lining the audio up with a
+    // sidechain: a delay reading a delay would count that edge's compensation
+    // twice, and taken from in front of it the subtract's own delay resolves to
+    // the whole distance instead, the input alignment and the device's latency
+    // together.
+    //
+    // Ahead of the slot's gain and meter, which is where the current engine
+    // subtracts it too: those are MAGDA's own and sit outside the plugin, so
+    // what they trim and report is the delta once there is one.
+    if (device.deltaSolo) {
```
lucaromagnoli: P1: Keep the aligned dry path alive while delta solo is disabled. Because this branch only emits the Subtract and its PDC delay when deltaSolo is true, enabling it during playback creates a cold delay with no pre-toggle samples. The downstream graph crossfade lasts only 5 ms, so for a device whose latency is longer than that it finishes before valid dry history appears, producing wet leakage or a discontinuity until the delay fills. The existing-engine regression in tests/test_delta_solo.cpp explicitly advances this history while monitoring is off. Please preserve the dry delay independently of the enabled state and make only the subtraction conditional; the same issue applies to the rack-level conditional in emitRackDelta.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -573,16 +609,6 @@ ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site
     if (device.bypassed)
```
lucaromagnoli, line 615: P1: This early return makes a bypassed device with delta solo enabled pass the full incoming chain through unchanged. Issue #2136 explicitly settles the opposite behavior: a bypassed device output equals its dry input, so its delta should be silence. The new compiler test currently codifies the contradiction. Please either retain enough structure to subtract the bypass passthrough from the dry input, or revise the issue acceptance criterion before this PR closes it.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Agreed, and fixed — this was a real gap and the reasoning holds exactly as you set it out.

Checked against the incumbent first. `PluginNode` builds `deltaLatencyProcessor` from `latencyNumSamples > 0` alone (`tracktion_PluginNode.cpp:119`), writes the input into it every block, and gates only the *read* (`:175-179`). `tests/test_delta_solo.cpp:73` pins that the history stays aligned while the read is off. So the line running whether or not anything is soloed is the contract, not an implementation detail.

What changed:

- **The subtract and its dry delay are now in every plan**, for every device slot (bar transparent taps, which have no difference to take) and around every rack instance. `emitDelta` no longer looks at the flag, and `emitRackDelta` lost its early return.
- **Only the subtraction is conditional**, and it moved to the value layer: `OpValue::subtractsDry`, resolved from `DeviceInfo::deltaSolo` / `RackInfo::deltaSolo`. Its own field rather than a gain on the dry edge, because `kUnityValue` — the reading of a value table that does not apply — has to mean "unchanged", and for a difference that is taking nothing away rather than taking all of it.
- **Delta solo therefore compiles nothing.** The plan is byte-identical with the flag on and off (pinned by a fingerprint comparison in the compiler tests), so toggling it publishes values only: no swap, no fade, and the delay line keeps its contents.

New test, which is the one this is about: `Delta solo turned on mid-render reads a dry line that has been running`. A `LatentDevice` delaying by a full block renders one block with the delta unheard, the model flag flips, values are re-resolved, and the very next block is silence sample for sample — only reachable if the dry line already held what the wet side is now arriving with. Reverting `emitDelta` to conditional fails it on every sample of that block.

Costs, since they are real and visible in the goldens: every device slot is four ops rather than three, and a device no longer writes its output over its input buffer, because the chain audio now has a second reader. That is one block copy per device — the same copy `PluginNode` makes unconditionally (`copyIfNotAliased`). The dry delay elides to nothing when the device reports no latency, which is the same rule TE uses to decide whether to build the line at all.

Two follow-ons worth naming:

- The random edit-sequence generator gained a `deltaSoloDevice` edit, which promptly found that a value-only publish steps rather than ramps. It is exempt in the transient property now, with the reason spelled out at the `Transient::valueOnly` field: nothing moved, so there was nothing to fade, and the model asked to hear something else. Mute and every fader position have the same shape today; smoothing published values is an engine-wide question rather than this slice's.
- `A reorder fades the one edge that can be faded` counts 4 unfaded edges rather than 2: each device is fed twice now, once into its processing and once into its delta's dry edge.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Taking the second option you offer — revising the criterion — because the behaviour the issue asks for is not the behaviour the engine we are reaching parity with has. Flagging rather than deciding, since it is your issue text.

What the current engine does with a bypassed, delta-soloed device:

- `PluginManagerSync.cpp:2218` turns bypass into `plugin->setEnabled(!device.bypassed && …)`.
- `tracktion_PluginNode.cpp:301` guards the subtraction with `shouldProcessPlugin && plugin->isEnabled() && numInputChannelsToCopy > 0`, so a disabled plugin never reaches `applyDeltaSolo`.
- The node has already copied its input to its output by then, so what comes out is the chain, unchanged.

So "bypassed → silence" would be a deliberate divergence, and the null-diff harness is the thing that would eventually notice it.

My reading of the sentence in #2136 is that it states the consequence of where the dry edge is taken from rather than a rule about the flag: because the dry side is the device's own input, a device that passes its input through nets to silence. That is now literally true in the plan — subtract(x, x) is zero — for any device whose processing is transparent. Structural bypass is the one case where MAGDA removes the processing from the graph entirely rather than making it transparent, and there the incumbent hands back the chain.

The code as it stands keeps parity, and the test says why at the assertion. If you would rather have the issue's reading, it is three lines — move the delta emission ahead of the `device.bypassed` early return and let wet and dry be the same port — and I will also want to know what a bypassed *rack* should do, since it takes the same early return. Say which and I will push it.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Re-review found one regression in this fix: DeviceDelta is now inside the !isTransparentTap(device) branch, making delta solo a no-op for every DeviceType::Analysis device. These plugins are transparent, so processed output and dry input are both x; the delta is x - x = 0, but omitting the subtract leaves x on the chain. DeviceSlotComponent exposes the delta button for analysis devices, and the first version emitted the subtract before this condition. Please emit DeviceDelta for every non-bypassed device and keep only gain/meter conditional, or make analysis devices ineligible for delta solo throughout the model and UI.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Re-review: this remains unresolved against the issue being closed. The reply chose the option of revising the criterion, but issue #2136 has not been edited and still explicitly says a bypassed device delta is silence rather than the chain. The parity argument may be the right product decision, but the implementation and acceptance criterion still disagree. Please either implement the stated behavior or update the issue before this PR closes it.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Checked the premise before changing anything, and it does not hold: the delta button is not exposed for analysis devices.

`DeviceSlotHeaderSpec.cpp` sets `visibility.delta = device.deviceType == magda::DeviceType::Effect`, and that one flag drives both `getExpandedVisibility` and `getCollapsedVisibility`, so neither header shows it for `DeviceType::Analysis`. `setDeviceDeltaSoloByPath` has exactly one caller in the tree — that button's `onClick` — with no API, DSL or OSC route to it. So delta solo cannot be turned on for an analysis device, and the `!isTransparentTap(device)` gate lines up with the UI rather than diverging from it. `isTransparentTap` and `visibility.delta` read the same field, so the two cannot disagree.

What is real is smaller and worth fixing. `deltaSolo` is serialised on every device (`TrackSerializer.cpp:444/594`) and read straight back into the struct, so a stored project can carry the flag on an analysis device whatever the setters allow, and the compiler ignored it in silence. That is the one thing the compiler is not meant to do.

So it is reported now, and the plan is unchanged: `delta solo on an analysis device has nothing to subtract, the chain passes through unchanged`. Taking option (a) instead would put a `Subtract` and its dry delay on every analysis device in every plan, permanently, to compute a difference no UI can ask for — and the mixer-analysis rail means most tracks carry one. Option (b) is already true in the UI; making it true "throughout the model" would mean removing a serialised field, which costs a migration for the same outcome the diagnostic gives.

Pushed on `fix/delta-solo-unreachable-flags`, along with the bypassed-device and bypassed-rack cases, which are the same hole. The random edit generator was also building the unreachable pair — `BypassDevice` and `DeltaSoloDevice` wrote their fields independently while `TrackManager` does not — so it mirrors the setters now; reverting that leaves four sequences reporting the new diagnostic, which is how it turned up.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: Withdrawing the earlier answer — it argued parity when it should have checked reachability, and the reachability answer settles it without either side having to give ground.

**A bypassed device with delta solo enabled is not a state the model can be in.** Every setter keeps the two apart, in both directions and at both scopes:

- `TrackManagerDevices.cpp:1092` `setDeviceInChainBypassedByPath` — `if (bypassed) device->deltaSolo = false;`
- `TrackManagerDevices.cpp:1101` `setDeviceDeltaSoloByPath` — `if (deltaSolo) device->bypassed = false;`
- `TrackManager.cpp:2257` `setDeviceBypassed` and `:2266` `setDeviceBypassedByPath` — same clear
- `TrackManager.cpp:2409` `setRackBypassed` and `:2417` `setRackBypassedByPath`, `:2427` `setRackDeltaSoloByPath` — same, for racks

The UI enforces it a second time in both button handlers (`DeviceSlotComponent.cpp:399` and `:423`). Bypass and delta solo are one switch with three positions, not two switches.

So the issue text and the implementation are not in conflict about anything reachable, and #2136 does not need editing: the sentence describes the consequence of where the dry edge is taken from — subtract(x, x) is silence for any device whose processing is transparent — and structural bypass is the one case where MAGDA removes the processing from the graph rather than making it transparent, which the model then refuses to combine with the flag.

**What the comment did catch** is that the pair is representable even though it is unreachable. `bypassed` and `deltaSolo` are serialised independently (`TrackSerializer.cpp:444/594` for devices, `771/821` for racks) and read straight back into the struct, so a stored project can hold both and the compiler ignored the flag in silence. The compiler test that codified this built exactly that state by hand and asserted `diagnostics.empty()`, which was the wrong assertion to make.

It now reports `delta solo on a bypassed device measures nothing, the chain passes through unchanged`, and the rack case alongside it. Behaviour is unchanged — the chain still passes through, which is parity — but the model no longer says something the plan drops without a word. On `fix/delta-solo-unreachable-flags`, together with the analysis-device case from the sibling thread.

PR #2165, opened 2026-08-23
Report a delta solo the plan cannot take (#2136)

lucaromagnoli:
Follow-up to the review on #2161. Both comments name a defect that is not reachable, and both
are circling one that is.

## What the code already does

**The delta button is `DeviceType::Effect` alone.** `DeviceSlotHeaderSpec.cpp` sets
`visibility.delta = device.deviceType == magda::DeviceType::Effect`, and that one flag drives
both `getExpandedVisibility` and `getCollapsedVisibility`. `setDeviceDeltaSoloByPath` has exactly
one caller in the tree — that button's `onClick` — with no API, DSL or OSC route to it. So delta
solo cannot be turned on for an analysis device, and the `!isTransparentTap(device)` gate lines
up with the UI rather than diverging from it. Both read the same field, so they cannot disagree.

**Bypass and delta solo are one switch with three positions.** Every setter clears each when the
other is set, in both directions and at both scopes: `setDeviceInChainBypassedByPath`,
`setDeviceDeltaSoloByPath`, `setDeviceBypassed`, `setDeviceBypassedByPath`, `setRackBypassed`,
`setRackBypassedByPath`, `setRackDeltaSoloByPath`. The UI enforces it a second time in both
button handlers. A bypassed device with delta solo enabled is not a state the model can be put
into, so the implementation and #2136 are not in conflict about anything reachable and the issue
text stands.

## What is real

`deltaSolo` and `bypassed` are serialised independently and read straight back into the struct
(`TrackSerializer.cpp:444/594` for devices, `771/821` for racks), so a stored project can hold
either combination whatever the setters allow. The compiler ignored both in silence, which is
the one thing it is not allowed to do — a model configuration it cannot express belongs in
`diagnostics`. The delta button would be lit over a device passing its input through, and
nothing anywhere would say why.

The compiler test that codified this built exactly that state by hand and asserted
`diagnostics.empty()`, which was the wrong assertion to make.

## The change

All three cases are reported and none of them changes the plan:

- a bypassed device — `delta solo on a bypassed device measures nothing, the chain passes through unchanged`
- a bypassed rack — the same, one level up
- a transparent tap — `delta solo on an analysis device has nothing to subtract, the chain passes through unchanged`

Emitting the subtract for analysis devices instead — the other option offered — would put a
`Subtract` and its dry delay on every analysis device in every plan, permanently, to compute a
difference no UI can ask for; the mixer-analysis rail means most tracks carry one. Making them
ineligible "throughout the model" is already true in the UI, and doing it in the model would mean
removing a serialised field, which costs a migration for the same outcome the diagnostic gives.

## The generator was building it too

`BypassDevice` and `DeltaSoloDevice` wrote their fields independently while `TrackManager` does
not, so random sequences produced the unreachable pair. It mirrors the setters now, which is what
"every edit here is one a user can perform" already claimed. Reverting that leaves four sequences
reporting the new diagnostic — that is how it was found.

## Testing

Each diagnostic verified negatively: reverting `PlanCompiler.cpp` alone fails all three new
assertions. Full suite green — 2889 cases, ~600k assertions, deep property sweep included.

### DIFF
--- a/magda/engine/plan/PlanCompiler.cpp
+++ b/magda/engine/plan/PlanCompiler.cpp
@@ -613,19 +613,10 @@ ChainSignal Compiler::emitDevice(const DeviceInfo& device, const ChainSite& site
     const bool hasTargets = ownsMultiOutPairs && targets != multiOutTargets_.end();
 
     if (device.bypassed) {
-        // Bypass removes the device from the plan, so there is no output to
-        // measure a difference against and the chain passes through unchanged,
-        // which is what the current engine does: bypass reaches it as
-        // `Plugin::setEnabled(false)` and PluginNode's delta subtraction is
-        // guarded on the plugin being enabled.
-        //
-        // The model's own setters cannot produce this pair - bypassing clears
-        // delta solo and enabling delta solo clears bypass, for devices and for
-        // racks (TrackManager::setDeviceBypassed and the rest) - so a project
-        // holding both came in through deserialisation, which writes the two
-        // fields straight into the struct. Reported rather than ignored,
-        // because the delta button will be lit over a device that is passing
-        // its input through, and nothing else would say why.
+        // The model's setters keep bypass and delta solo apart, so a project
+        // holding both arrived through deserialisation, which writes the fields
+        // straight into the struct. The chain passes through, which is parity,
+        // and the flag is reported rather than dropped in silence.
         if (device.deltaSolo)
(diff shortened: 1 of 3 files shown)


lucaromagnoli:
Folded into #2163 — not worth its own CI run for nine lines of compiler change. Same commit, same tests, and the two touch the same two files anyway.

PR #2167, opened 2026-08-23
Dev/0.19.0

lucaromagnoli:
Release merge of `dev/0.19.0` into `main`. 81 commits.

## Native engine

The bulk of the release. A render plan IR with a compiler, canonical dump and goldens in CI, then the pieces that execute it: the transport in beats, disk streaming, offline render and tap ops, the parallel executor, latency compensation and buffer assignment, runtime-state ownership with epoch retirement, the differ, and crossfades on the edges an edit moved.

Clip playback on top of that: rate conversion, looping and reverse, time stretch and pitch, warp markers, beat detection and loop info, MIDI clips, and what a track's clips sum to.

Parameters: the parameter lane and what a device reads, publishing a plan's parameters and link table, lanes over parameters and faders, the LFO and the last three modifiers.

Rack graph slices: device identity made section-aware, saying which rack a nested device is in, giving an output pair the track that opened it, subtracting the dry signal a device or rack was handed, refusing a rack that contains itself, and channel counts with implicit summing.

Correctness work alongside it: the null-diff corpus, whole-project cases and the tiered oracle, block-size invariance at 64, 512 and 4096, differ property tests over random edit sequences, node id uniqueness, legacy project and preset migrators, and a DAWproject round trip as an engine-facing cross-check.

## Remote API

Authenticated JSON-RPC 2.0 over WebSocket, with subscriptions, revisioned events and backpressure. Message-thread dispatch, revisions and undo transactions. Permissions, audit and cross-transport conformance. Device control surface and automation enumeration, racks and chains addressed by path, and canonical ControlTarget serialization. Exposed as MCP tools and resources.

## OSC control surfaces

Input path, bindings and learn, feedback echoing MAGDA's state back to the surface, and a receive loop that can answer whoever is talking.

## Editing and UI

Overlapping arrangement clips, clip container separated from content, relative transport seek, a pre/post-fader toggle on the post-FX panel, EQ curve readouts kept inside the plot, both arrangement columns given the same idea of a row, browser view and panel tabs remembered between sessions, and window backgrounds following the theme they are shown under.

## Fixes

Moved clips leaving a ghost outline, imported clips losing their colour, juce::String decoding UTF-8 source literals as ASCII, the MCP server losing a refusal it had already sent, discrete cells writing on rebuild, Faust instrument index metadata, and test runs writing to the developer's settings file.

## Notes

Required checks on this PR ignore workflow_dispatch runs, so CI has to be dispatched to validate and the merge done past the ruleset.


magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,1055 @@
...
+                if (outgoing.ownsSlot)
+                    connection->release();
+                if (!sent)
+                    break;
+            }
+        });
+
+        std::string message;
+        while (socket.is_open() && running.load() && !connection->isClosed()) {
+            if (socket.read(message) == httplib::ws::ReadResult::Fail)
```
lucaromagnoli, line 905: **stop() and disconnect never complete against an idle client that answers pings.** cpp-httplib's `WebSocket::read()` loops `while (!closed_)` and on a Pong does `unacked_pings_ = 0; continue;`, so it only returns to this loop on Text/Binary/Fail. The server pings every second (`kPingIntervalSeconds = 1`) and every real client library (browsers, Node `ws`, Python `websockets`, httplib's own `WebSocketClient` while in `read`) answers pongs automatically, so the read timeout restarts on every pong and `running`/`isClosed()` are never re-checked. `stop()` then blocks forever in `listener.join()` (`listen_internal` ends with `task_queue->shutdown()`, which joins the worker stuck here), and `stop()` runs on the message thread from `ConnectionsDialog` (toggle off, rotate token) and from `~RemoteApiHost` at quit, so the UI hangs until the client goes away. The same mechanism makes `disconnectByHandle` and the hub's backpressure drop inert: `markClosed()` only sets a flag, so a "disconnected" client stays registered, listed and pinged until it happens to send a frame, which contradicts the permissions doc ("the socket itself goes within one read timeout") and the rotation doc ("every client connected over it is disconnected"). The tests do not see it because their clients are not sitting in `read()` when `stop()` runs, so no pongs flow. The reader needs a wake-up that does not depend on the peer: keep the accepted fd and `shutdown(2)` it from `stop()`/`disconnect` (safe cross-thread, fails the blocked `recv` immediately), or have the heartbeat thread do the close (that is what `set_websocket_max_missed_pongs` is for, and `close()` only reads for `CPPHTTPLIB_WEBSOCKET_CLOSE_TIMEOUT_SECOND`).

magda/mcp_bridge/main.cpp
```diff
@@ -0,0 +1,669 @@
...
+        // status and content type before deciding what it was.
+        //
+        // Only `subscriptions/listen` streams. The transport permits any request
+        // to be answered with SSE, so the non-streaming path below still handles
+        // an SSE body it did not expect — it just gets to do it after the fact.
+        if (method == "subscriptions/listen") {
+            // A stream never completes on its own, so no read timeout may apply
+            // to it. Cancellation and shutdown both stop the client instead,
+            // which is prompt regardless.
+            client->set_read_timeout(0, 0);
```
lucaromagnoli: **`subscriptions/listen` through the bridge dies on the first idle read.** `set_read_timeout(0, 0)` is not "no timeout" in cpp-httplib 0.52: `ClientImpl::set_read_timeout` stores 0/0, `SocketStream::wait_readable()` calls `select_read(sock_, 0, 0)` which is `poll(&pfd, 1, 0)`, and `SocketStream::read` turns a false `wait_readable` into `Error::Timeout`. A listen stream is quiet by design (nothing between the acknowledgment and the next update or the keep-alive comment), so the first read with no bytes queued fails, `Post` returns null, `isCancelled(id)` is false, `send` returns false, the outer loop re-resolves and retries three times (re-emitting the acknowledgment to the host each time) and then answers "Could not reach a running MAGDA". No bridge test covers listen. Use a long timeout (hours) and keep relying on `Client::stop()` for cancellation, which is what the comment says is the real mechanism.

magda/daw/api/remote_mcp.cpp
```diff
@@ -0,0 +1,950 @@
...
+
+    if (call.method == "resources/read") {
+        readResource(call, std::move(onComplete));
+        return;
+    }
+
+    // MCP pins the status for this one: `404` with a JSON-RPC body is what
+    // separates "this server does not have that method" from "there is no MCP
+    // endpoint here", which is the probe a dual-era client uses.
+    onComplete(McpReply::fail(MCP_METHOD_NOT_FOUND, "Unknown method: " + call.method, 404));
```
lucaromagnoli: **404 for an unknown method tells a legacy-era client its session is gone.** In the 2025-03-26 / 2025-06-18 / 2025-11-25 Streamable HTTP transport, 404 on a request carrying `Mcp-Session-Id` is the defined "session terminated" signal and the client MUST re-`initialize`. This fires for any method not implemented here (`prompts/list`, `logging/setLevel`, `completion/complete`; Claude Desktop calls `prompts/list` regardless of advertised capabilities), and the same 404 is used for `server/discover` on a legacy connection a few lines up. Through `magda-mcp` it cascades: `Bridge::send` treats `404 && !currentSession().empty()` as a dead session, clears it, replays `initialize` (minting a new server session), returns false, and the three-attempt loop does that three times, so one such call leaves three orphaned sessions for `sessionIdleSeconds` (300 s) and answers the host "Could not reach a running MAGDA". `maxSessions` is 8, so after the third such call `createSession` returns 503, the bridge is left with no session, every later request goes out without `Mcp-Session-Id`, gets the 400 "-32602 must carry _meta" refusal, and nothing re-handshakes because that path needs a 404. Return method-not-found with HTTP 200 in the legacy era (keep 404 only where the 2026-07-28 revision pins it), and make the bridge re-handshake on the "Unknown or expired session" body rather than on any 404.

magda/daw/api/remote_model_bridge_live.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // ---- TrackManagerListener -------------------------------------------
+
+    void tracksChanged() override {
+        // Session slots are keyed by track, so a track appearing or going away
+        // adds or removes a whole column of the grid.
+        service_.noteModelChanged({Topic::Tracks, Topic::Session});
+    }
+    void trackPropertyChanged(int) override {
```
lucaromagnoli, line 76: **Volume/pan/send automation playback bumps the revision on every tick.** `AutomationPlaybackEngine::currentValueChanged` writes every automated TrackVolume/TrackPan/SendLevel value back through `TrackManager::setTrackVolume/setTrackPan/setSendLevel(..., fromAutomation=true)` under an `AutomationWriteScope`, and those call `notifyTrackPropertyChanged` unconditionally (`fromAutomation` only skips the `manualVolume` write). This handler routes that to `noteModelChanged(Topic::Tracks)`, which does `revision_.fetch_add(1)`. So with any track volume/pan/send lane in the project the revision advances tens of times per second while the transport rolls, and every remote write carrying `expectedRevision` is refused with `conflict`, which is exactly the failure the header says `noteModelActivity` exists to avoid ("bumping the revision for them would leave expectedRevision permanently stale whenever the transport is rolling"). `masterChannelChanged` has the same problem via master volume automation. Route to `noteModelActivity` when `AutomationManager::getInstance().isApplyingAutomationWrite()` is set, the way the device-parameter callbacks below already do.

magda/daw/api/remote_api.cpp
```diff
@@ -208,28 +212,58 @@ const juce::var& clipSchema() {
...
+            "type":"object",
+            "properties":{
+                "id":{"type":"integer","minimum":0},
+                "trackId":{"type":"integer","minimum":0},
+                "rackId":{"type":["integer","null"]},
+                "chainId":{"type":["integer","null"]},
+                "devicePath":{},
+                "name":{"type":"string"},
+                "type":{"type":"string","enum":["instrument","effect","midi","analysis"]},
+                "format":{"type":"string","enum":["vst3","au","vst","internal"]},
```
lucaromagnoli: `format` enum here is `["vst3","au","vst","internal"]`, but `pluginFormatName()` in `remote_api_projection.cpp` emits `"lv2"` for `PluginFormat::LV2` and never emits `"vst"`. `deviceCatalogEntrySchema()` below already has the right set (`vst3, au, lv2, internal`). As it stands `devices.list` and the `devices` snapshot for any track holding an LV2 plugin violate the output schema that `system.describe` and the MCP tool `outputSchema` advertise (a host that validates `structuredContent` will reject every result), and `deviceFromJson()` refuses it on the round trip. Same enum in both, ideally from one shared definition.

magda/daw/api/remote_service.cpp
```diff
@@ -0,0 +1,599 @@
...
+        cache_.erase(cache_.begin(),
+                     cache_.begin() + static_cast<long>(cache_.size() - cacheCapacity_));
+}
+
+juce::String RemoteApiService::idempotencyKey(const RequestContext& context) {
+    if (context.requestId.isEmpty())
+        return {};
+    // Scoped by client: two clients are free to number their own requests from
+    // 1, and one must never replay the other's response.
+    return context.clientId + "\n" + context.requestId;
```
lucaromagnoli, line 568: The idempotency cache is keyed by connection + request id only; `cachedResponse()` never compares the operation or its input. Over WebSocket the request id *is* the JSON-RPC `id` (`idempotencyKeyFor`), and JSON-RPC 2.0 does not require ids to be unique for the life of a connection. A script that sends `"id": 1` for every call (very common with websocat/curl-style tooling) gets its first write's response replayed for every later write: `tracks.create {name:"A"}` then `tracks.create {name:"B"}` returns ok with A's id and never creates B, and a `clips.delete` with a reused id returns a stale `tracks.create` result, silently. Store the operation name plus an input fingerprint with the entry and answer a same-key request that differs with `conflict`, or only cache when an explicit idempotency key is supplied, as the MCP side already does.

magda/daw/api/remote_mcp_server.cpp
```diff
@@ -0,0 +1,1496 @@
...
+        if (stream == nullptr)
+            return;
+
+        // Deregister before closing: removeClient takes the hub's lock, so it
+        // returns only once a publish already inside the sink has finished.
+        if (subscriptions != nullptr && stream->subscriber != 0) {
+            subscriptions->removeClient(stream->subscriber);
+            stream->subscriber = 0;
+        }
+        stream->close();
```
lucaromagnoli: A listen stream the *server* ends (Disconnect in the settings UI via `disconnectByHandle`, the hub dropping a slow consumer, `stop()` on toggle/rotate/quit) closes with no JSON-RPC response for the listen request: `EventStream::close()` sets `closed` and clears the outbox, `take()` then reports `Closed`, and the provider calls `sink.done()`. `McpEndpoint::listenClosedResult` exists for exactly this and is referenced nowhere in the server. Through the bridge the host's listen id therefore stays outstanding forever while it believes it is still subscribed. Enqueue `listenClosedResult(subscriptionId)` as a terminal frame that `take()` drains before reporting `Closed`, instead of clearing the outbox.

magda/daw/api/remote_mcp_server.cpp
```diff
@@ -0,0 +1,1496 @@
...
+                closed.push_back(session.id);
+            impl_->sessions.clear();
+        }
+        impl_->noteSessionsClosed(closed);
+    }
+
+    impl_->server.stop();
+
+    if (impl_->listener.joinable())
+        impl_->listener.join();
```
lucaromagnoli, line 1607: `stop()` runs on the message thread (ConnectionsDialog toggle -> `stopTransport`, `rotateToken`, `~RemoteApiHost`) and this join waits for every pool thread, but `executeAndReply` parks its pool thread on `waiter->ready.wait_for(lock, limit, ...)` until `RemoteApiService::dispatch` completes, and `dispatch` hops via `MessageManager::callAsync`, which cannot run while the message thread is sitting in this join. Any request whose job was posted but not yet run when the click landed (or one accepted in the gap before the listening socket closes) therefore only unblocks when `wait_for` expires after `defaultDeadlineMs + 2 s` (12 s by default), during which the UI is frozen. The WebSocket server does not have this because its completion is asynchronous. Keep the live waiters under a mutex and have `stop()` mark them done with a `Cancelled` reply and `notify_all()` before joining.

magda/mcp_bridge/main.cpp
```diff
@@ -0,0 +1,669 @@
...
+        if (result->status == 202 || result->body.empty())
+            return true;
+
+        if (isEventStream(*result)) {
+            magda::mcp::SseParser complete;
+            complete.consume(result->body, [this](const std::string& m) { emitRaw(m); });
+            return true;
+        }
+
+        emitRaw(result->body);
```
lucaromagnoli: `handlePost` refuses rate-limited (429), over-concurrency (503) and oversized (413) requests before parsing the body and writes `jsonRpcError({}, ...)`, i.e. `"id": null`, even though the body is fully read and the id is recoverable. The bridge then `emitRaw()`s that body as-is, so the host receives an error with `id: null` that matches no pending request and the original request hangs until the host's own timeout. The bucket bursts to `maxConcurrentRequests` (8) at 50/s, so an agent issuing a burst of parallel tool calls hits this in normal use. Parse the envelope before `admit()` so refusals carry the real id, and in the bridge, when a non-2xx body has a null id and the request had one, answer `fail(id, ...)` instead of forwarding it unchanged.

magda/daw/api/remote_model_bridge_live.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // state that no longer exists. This is the trigger the service's
+        // cancellation path exists for, and it already bumps the revision and
+        // publishes Topic::Project — bumping again here would advance it twice
+        // for one swap.
+        service_.projectReplaced();
+    }
+    void projectClosed() override {
+        service_.projectReplaced();
+    }
+    void projectSaved(const ProjectInfo&) override {
```
lucaromagnoli, line 212: `transport` and `project` subscribers never hear about changes made in the MAGDA UI. `Topic::Transport` is marked only by `RemoteApiService::topicsFor` when a remote `transport.*` request ran (grep: `remote_service.cpp:44` is the only site), and `Topic::Project` only by remote `project.*` requests plus `projectOpened/projectSaved` here. `ProjectManager::setTempo` / `setTimeSignature` just `markDirty()` and `ProjectManagerListener` has no property callback, so this bridge cannot see them either. So Play/Stop/Record, loop toggles, tempo and time-signature edits in MAGDA produce no event, while the contract says a delta is "what changed since the previous event" and the MCP resource list advertises `magda://transport` as play and record state. `osc_feedback_live.hpp` already names this gap and polls around it. Add a property-changed callback to `ProjectManagerListener` fired from those setters, and mark `Topic::Transport` from the transport state source (the AudioEngine listener fan-out or TE's `TransportControl` change broadcaster).

magda/daw/api/device_api_live.cpp
```diff
@@ -0,0 +1,246 @@
...
+        device.format = PluginFormat::Internal;
+        device.isInstrument = isInstrument;
+        device.deviceType = isInstrument ? DeviceType::Instrument : DeviceType::Effect;
+        return device;
+    };
+
+    if (const auto* spec = daw::audio::compiled::findCompiledPluginSpec(catalogId))
+        return fromSpec(spec->displayName, spec->isInstrument);
+
+    if (const auto* spec = daw::audio::findInternalPluginSpec(catalogId))
```
lucaromagnoli: `addDevice` accepts catalogue ids the catalogue deliberately hides. `getCatalog()` filters internal specs by `showInBrowser` ("the registry also holds internal devices that exist only as host wiring") and lists externals from `getExternalPlugins()` (preferred formats only), but `deviceFromCatalogId` resolves any id via `findInternalPluginSpec` with no `showInBrowser` check and scans `getAllExternalPlugins()`. `BaseDevicePack.cpp` registers Legacy Volume/Pan, Level Meter, External Insert, MIDI Receive, Sidechain Monitor, Audio Sidechain Monitor, Instrument Meter Tap, Track Measurement, Session Monitor, Follower Source Tap and Insert Capture without `showInBrowser`, so an `edit` client can insert host-wiring plugins (or a non-preferred plugin format) into a user chain even though `findCatalogEntry` returns nullopt for the same id and the header promises `INVALID_DEVICE_ID` when "the catalogue id does not resolve". Resolve through the same filtered set the catalogue is built from.

magda/daw/api/osc_feedback_live.cpp
```diff
@@ -0,0 +1,629 @@
...
+        projectMacros();
+    }
+    if (bindingsDirty_) {
+        bindingsDirty_ = false;
+        projectBindings();
+    }
+
+    // Always. Nothing marks the transport, and the playhead moves without
+    // telling anyone.
+    projectTransport();
```
lucaromagnoli, line 419: The fixed-namespace echo rule is defeated by the pump phase. `OscFeedback`'s design (its header) is "bound the suppression to the flush the arrival belongs to", and `flush()` clears `echoed_` unconditionally at its end. But the model change that makes this tick project the received value arrives via `ChangeSource`, whose delivery is a *separate* 33 ms `juce::Timer` (`Pump::timerCallback -> flush -> onChanges -> mixerDirty_`), so whenever this tick fires between the router's drain (which set the echo bit) and the pump's delivery, `flush()` finds nothing dirty, wipes the bit, and the next tick publishes the value with the bit gone and sends it straight back to the surface that produced it. For a surface sending slower than the tick (a slow finger drag in TouchOSC, or any motorised fader), that is every value, which is the chase the rule exists to stop. The tests never hit it because they call `markTracksChanged()` synchronously before `tick()`. Have the feedback tap also set the matching dirty flag (mixer/macros) so the tick after a drain always re-reads the model and sees the echo, or keep the slot's echo bit until that slot has actually been projected rather than clearing on every flush.

magda/daw/ui/dialogs/ConnectionsDialog.cpp
```diff
@@ -0,0 +1,1229 @@
...
+        message += " " + tr("connections.rotate_reconnect");
+
+        juce::AlertWindow::showAsync(
+            juce::MessageBoxOptions()
+                .withIconType(juce::MessageBoxIconType::QuestionIcon)
+                .withTitle(tr("connections.rotate_title").replace("{0}", transportName))
+                .withMessage(message)
+                .withButton(tr("connections.rotate_confirm"))
+                .withButton(tr("connections.rotate_cancel")),
+            [this](int result) {
```
lucaromagnoli: This lambda captures raw `this` and, on confirm, reads `transport_` and calls `refresh()`. The alert is an independent top-level modal, not a child of this window, and the window is deleted asynchronously by `SelfClosingDialogWindow::closeButtonPressed` (`callAsync([self = this]() { delete self; })`). JUCE only suppresses a native close under a modal on Windows (`WM_CLOSE` checks `isCurrentlyBlockedByAnotherModalComponent()`); macOS `windowShouldClose` calls `handleUserClosingWindow()` unconditionally, and so does the Linux `WM_DELETE_WINDOW` path. So on macOS/Linux: Rotate token, close the Connections window from the title bar, press Rotate in the still-open box, and the callback dereferences a destroyed `TransportPage`. The lambda already re-fetches the host defensively but not the page. Capture a `juce::Component::SafePointer<TransportPage>` and bail out when it is null.

lucaromagnoli (commented):
Scope: WebSocket / remote API / MCP / OSC only, the native engine is not covered here.

Overall this is solid and very well documented; the docs made it possible to check the code against what it promises. The comments below are what I could verify by reading the code and the pinned cpp-httplib. Four I consider blocking before release: the WebSocket `stop()`/disconnect hang against any client that answers pings, the bridge's `subscriptions/listen` dying on the first idle read, the legacy-era 404 that the bridge reads as "session gone" (with the session-exhaustion cascade), and the revision bumping on every automation tick. The rest are correctness gaps and doc/code mismatches, roughly in severity order.

magda/daw/audio/osc/OscFeedback.cpp
```diff
@@ -0,0 +1,182 @@
...
+        connected_ = false;
+    }
+    host_ = {};
+    port_ = 0;
+}
+
+bool OscSenderSink::send(const juce::String& address, float value) {
+    if (!connected_)
+        return false;
+    return sender_.send(juce::OSCMessage(juce::OSCAddressPattern(address), value));
```
lucaromagnoli: `juce::OSCAddressPattern(address)` throws `OSCFormatError` for a token containing `' '`, `'#'` (JUCE's `getDisallowedChars()` for patterns is `" #/"`), and for bound addresses this is the surface's literal string: `publishBinding` passes it straight in, `Binding::isValid()` only checks `startsWithChar('/')`, and the packet parser accepts any printable ASCII. A stored binding such as `/1/fader 3` (hand-edited bindings JSON today, a learn UI later) raises a C++ exception out of a `Timer` callback on the message thread the first time feedback is projected for it. Validate the address when the binding is built or published, or catch `OSCFormatError` here and return false.

manual/docs/reference/osc-addresses.md
```diff
@@ -0,0 +1,82 @@
...
+What gets sent:
+
+- **On change** - track and master volume, pan, mute, solo and sends, and the focused device's macros.
+- **Sampled at 30 Hz** - play, record, loop, tempo and the playhead.
+
+A value only goes out when it differs from what was last sent, and an address is sent at most once per flush however often it changed in between, so a moving fader does not flood the surface.
+
+A value the surface just sent is not echoed straight back, which would fight a motorised control. The one send you do get after a gesture ends is the confirmation of what MAGDA rounded your value to.
+
+A full snapshot is sent when the feedback destination changes, and when inbound traffic resumes after five seconds of silence. Large snapshots are spread over consecutive flushes rather than dropped, so a sixty-track project takes a few ticks to populate rather than arriving incomplete.
```
lucaromagnoli, line 78: "when inbound traffic resumes after five seconds of silence" has no implementation: `OscFeedback::requestSnapshot` is called only from `OscFeedbackProjector::syncSurfaces` for a surface that is new (first datagram from that host) or after a feedback-port change, `OscFeedbackProjector::requestSnapshot()` has no callers, and `OscPeers` keeps a peer for the session keyed by host with no inactivity tracking. A TouchOSC app restarted on the same host keeps its peer and gets no snapshot, so it shows template defaults until each value happens to change. Either implement it (use the peer's `lastSeenMs` to request a snapshot when a quiet peer speaks again) or drop the sentence.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,1055 @@
...
+                                  options.maxInFlightPerConnection);
+            refusal.isNotEmpty()) {
+            // Nothing was admitted, so this reply owns no slot. Releasing one
+            // when it is written would hand back capacity that belongs to a
+            // request still in flight.
+            connection->enqueue(errorReply({}, kTooManyRequests, refusal), false);
+            return;
+        }
+
+        if (message.size() > options.maxFrameBytes) {
```
lucaromagnoli, line 660: Unreachable at the defaults. `Options::maxFrameBytes` is `256 * 1024`, which equals the compile-time `CPPHTTPLIB_WEBSOCKET_MAX_PAYLOAD_LENGTH=262144` in `CMakeLists.txt`, and cpp-httplib's `read_websocket_frame` returns false as soon as `payload_len > max_len`, after which `WebSocket::read` sets `closed_` and returns `Fail`. So an oversized frame never reaches `handleMessage`: the connection drops with a bare TCP close (client sees 1006), no JSON error, no audit line, and not the 1009 `MessageTooBig` the header promises. A configured `maxFrameBytes` above the library cap is silently ignored for the same reason. Either set the default below the library cap so this is the path that runs, or document that the library cap is the real limit and that it closes abruptly.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,1055 @@
...
+    // to drop.
+    impl_->server.WebSocket(
+        kEndpoint, [this](const httplib::Request& request, httplib::ws::WebSocket& ws) {
+            impl_->runConnection(ws, juce::String(request.get_param_value("client")));
+        });
+
+    // Registered before the listener opens and cleared in stop(), so the
+    // settings UI can never route a disconnect to a server that has gone away.
+    if (impl_->options.clients != nullptr) {
+        impl_->options.clients->setDisconnectHandler(
```
lucaromagnoli: Registered before the bind. If `bind_to_port`/`bind_to_any_port` fails, `start()` returns false with `running` still false, `stop()` returns early on `running.exchange(false)`, and `RemoteApiHost::startTransport` then does `server_.reset()`, leaving a `std::function` with a dangling `this` in `RemoteClientRegistry` until the next successful `start()` overwrites it. Dormant today only because a disconnect needs a live `websocket` row to route to. Register it after the bind succeeds, or clear it on the failure path.

magda/daw/api/remote_websocket_server.hpp
```diff
@@ -0,0 +1,293 @@
...
+        /// Required. `start()` fails rather than open an unauthenticated
+        /// listener, so there is no configuration that accidentally allows
+        /// anonymous access.
+        juce::String bearerToken;
+
+        /// Browser origins permitted to upgrade. Empty means no browser may
+        /// connect; a native client sends no `Origin` at all and is unaffected.
+        /// Absent from a request is not the same as unrecognised — the first is
+        /// a non-browser client, the second is a page we did not authorise.
+        std::vector<juce::String> allowedOrigins;
```
lucaromagnoli, line 171: This cannot admit a browser. `authorise()` checks `Authorization` before `Origin`, and the browser `WebSocket` API cannot set request headers on the upgrade, so every page-originated upgrade fails 401 regardless of `allowedOrigins`; the "a listed browser origin is allowed" test passes only because the test client sends both headers, which no browser can. As written the option only affects non-browser clients that choose to send `Origin`. Either accept the token for browser upgrades through a channel a browser can use (query parameter or `Sec-WebSocket-Protocol`) or document that browsers are not a supported WebSocket client and keep origin validation for MCP.

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,998 @@
...
+        const std::lock_guard<std::mutex> lock(mutex_);
+        if (shutdown_)
+            return;
+
+        for (auto& client : clients_) {
+            client.refusedThisFlush = false;
+            client.acceptedThisFlush = false;
+        }
+
+        for (const auto& change : changes) {
```
lucaromagnoli: Revisions can go backwards across topics within one flush. `ChangeSource::flush` builds the list in `Topic` enum order, each entry carrying that topic's highest marked revision, and this loop publishes in that order. Two commits inside one 33 ms window, a clip edit at N then a track edit at N+1, deliver `tracks@N+1` before `clips@N` on the same connection, while the contract says "revision orders the stream" and the header says that is what makes a gap detectable. The "events arrive in revision order across topics" test commits one topic per flush so it does not see this. Stable-sort the changes by revision before projecting, or state that ordering is per topic.

magda/daw/api/remote_handlers.cpp
```diff
@@ -0,0 +1,768 @@
...
+    if (!type)
+        return HandlerResult::fail(ErrorCode::ValidationFailed,
+                                   "unsupported lane type: " + input["type"].toString());
+    const auto target = toAutomationTarget(input["target"]);
+    if (!target)
+        return HandlerResult::fail(ErrorCode::ValidationFailed,
+                                   "target does not resolve to an addressable parameter");
+    // Shape is valid; the thing it names must also still exist, or the lane
+    // would be created against a track that was deleted.
+    if (!targetResolves(api, *target))
```
lucaromagnoli, line 761: `targetResolves()` only checks the track exists; it never resolves `devicePath` or checks `parameterIndex` against the device, and `AutomationManager::createLane()` validates neither. So an `edit` client can create a lane for a nonexistent device step or `parameterIndex: 999999`, get `ok` and a fresh id, and leave an orphan lane on the track that automates nothing. Resolve the device through the track API for device-bound kinds and bounds-check the parameter index, returning `not_found` otherwise.

magda/daw/api/remote_mcp_server.cpp
```diff
@@ -0,0 +1,1496 @@
...
+        }
+    }
+
+    std::optional<juce::String> createSession(const juce::String& version,
+                                              const juce::String& clientName) {
+        std::vector<juce::String> collected;
+        juce::String id;
+        {
+            const std::lock_guard<std::mutex> lock(sessionMutex);
+            collectIdleSessionsLocked(collected);
```
lucaromagnoli, line 479: Only caller of `collectIdleSessionsLocked`. A legacy client that crashed or went quiet keeps its session (and its `noteSessionOpen` row, shown as connected in Settings > Connections) until some other client sends `initialize`, contrary to the Options comment that an idle session "is collected". Bounded by `maxSessions`, so not a leak, but the phantom rows persist and the slots count toward the 503 refusal. Run the collection on a tick or on every `touchSession`/`handlePost`.

magda/daw/api/remote_api_host.cpp
```diff
@@ -0,0 +1,556 @@
...
+ * connects to a port with no credential or presents a credential to nothing.
+ *
+ * `token`/`url`/`port` are the WebSocket's and keep their original names, so a
+ * script written before the split still finds what it was reading.
+ */
+bool writeTokenFile(const juce::File& file, const juce::String& wsToken, int port,
+                    const juce::String& mcpToken, int mcpPort) {
+    file.getParentDirectory().createDirectory();
+    file.deleteFile();
+    if (!file.create().wasOk())
```
lucaromagnoli, line 141: `File::create()` opens with mode 0644 (JUCE's POSIX `FileOutputStream` uses `O_CREAT, 00644`), and the `chmod` to 0600 happens after. The file is empty in that window, as the comment says, but a process that `open()`s it during the window keeps its descriptor and reads the token after the write, so "created empty and restricted before anything secret goes in" does not hold against another local user on a machine where the data dir is traversable. Create it with the final mode in one step (`open(path, O_WRONLY|O_CREAT|O_EXCL, 0600)`, or `umask(077)` around the create) rather than create-then-chmod.

lucaromagnoli (commented):
Round 2 on the remote API / MCP / OSC fixes. I re-read every change against the code and the pinned cpp-httplib and all 22 earlier comments are addressed: the `WebSocket::shutdown()` patch wakes the blocked reader without a second framed read (and the new test drives a ponging client), the bridge listen timeout, the legacy-era 404 and the bridge's narrowed re-handshake, automation writeback routed to `noteModelActivity`, the schema enum, explicit `meta.idempotencyKey`, the terminal listen result, MCP waiter cancellation on stop, correlated admission refusals, project property and transport events, the hidden catalogue ids, the OSC echo dirtying and resume snapshot, the SafePointer, the `OSCFormatError` catch, the frame cap guard, handler-after-bind on both servers, revision ordering, lane target resolution, idle session collection, and the 0600 create.

Two things left, both small: the transport detection landed as a poll where TE already broadcasts, and the new browser token compare is not constant-time.

magda/daw/api/remote_subscriptions.cpp
```diff
@@ -0,0 +1,1036 @@
...
+}
+
+void SubscriptionHub::sampleNow() {
+    bool transportChanged = false;
+    {
+        const std::lock_guard<std::mutex> lock(mutex_);
+        if (shutdown_)
+            return;
+
+        if (anySubscriberLocked(Topic::Transport)) {
```
lucaromagnoli: This detects local Play/Stop/Record/Loop by polling at 20 Hz, and the repo rule is listener callbacks over timers (timers are for animations). The source this reads already has the callbacks: `TransportApiLive::isPlaying()` reads `edit->getTransport()`, and `te::TransportControl` is a `juce::ChangeBroadcaster` that calls `sendChangeMessage()` from `startedOrStopped()` (play, stop, record) on the message thread, while `looping` is a `CachedValue` on the Edit state and can be watched with a `ValueTree::Listener`. A `ChangeListener` on the current Edit's transport in `ModelChangeBridge` (re-attached on `projectOpened`) marking `Topic::Transport` via `noteModelActivity` gives the same coalesced delta through the normal path with no sampler involvement, and keeps `sampleNow` to the two continuous topics it was written for. The 33 ms `ChangeSource` pump then bounds latency rather than a 50 ms poll on top of it.

magda/daw/api/remote_websocket_server.cpp
```diff
@@ -0,0 +1,1058 @@
...
+     */
+    httplib::Server::HandlerResponse authorise(const httplib::Request& request,
+                                               httplib::Response& response) {
+        const auto fromBrowser = request.has_header("Origin");
+        const auto browserToken = request.has_param("token")
+                                      ? juce::String(request.get_param_value("token"))
+                                      : juce::String();
+        const auto headerAuthorised = isAuthorised(
+            juce::String(request.get_header_value("Authorization")), options.bearerToken);
+        if (!headerAuthorised && (!fromBrowser || browserToken != options.bearerToken)) {
```
lucaromagnoli: `browserToken != options.bearerToken` is a plain `juce::String` compare; the header path two lines up goes through `secureEquals`. Same credential, so the query-parameter path should use `secureEquals(browserToken, options.bearerToken)` too (and only when `browserToken` is non-empty, so a browser request with no `token` does not reach the compare at all).

magda/daw/api/remote_subscriptions.cpp
lucaromagnoli: Fixed in 5018eda9e: transport polling and its sampler state cache are removed. TransportApiLive now observes TransportControl change broadcasts plus the loop ValueTree, ModelChangeBridge forwards those as coalesced transport activity, and the listener reattaches when the project/Edit changes. Added mock bridge coverage and a live JUCE listener/rebind test.

magda/daw/api/remote_websocket_server.cpp
lucaromagnoli: Fixed in 5018eda9e: browser query authentication now requires a non-empty token and compares it with secureEquals(), matching the Authorization-header path. Added coverage for an allowed Origin with an empty query token.

Issue #2172, opened 2026-08-23
Drag Midi File to DrumGrid Track


Axel-Erfurt:
LMDE7 / Debian 13
Magda Version 0.18

I can not drag midi files to a drum grid track. I can drag it to a new track and then move it to the drum grid track.
Is this a bug or intentional?

Click the image to show.

<img width="1287" height="727" alt="Image" src="https://github.com/user-attachments/assets/deeb5ea3-c8c3-4fec-8b96-8880e5dd7bc6" />

Axel-Erfurt:
Thanks, Works in  v0.19.0-rc0

PR #2177, opened 2026-08-23
Say what a track is, once, where the type is declared

lucaromagnoli:
`TrackType` named a distinction that stopped existing, and the app has been paying for it at every call site.

## The name

`Audio`, `Instrument` and `MIDI` were three types until they collapsed into one hybrid track. The survivor kept the name `Audio`, so the enum read as a statement about *content* for years after it had stopped being one — `TrackTypes.hpp` documented this in a comment (`// Regular hybrid track`) rather than in the name. It is `Media` now.

The ordinals Instrument and MIDI used (1 and 2) stay skipped. They are written into every `.mgd` and pinned by `tests/test_persisted_enum_pins.cpp`, whose header explains the failure mode: *"an Aux track becomes a Master. Nothing complains, because the integer is still in range."* Closing the gap would renumber five types in every saved project, and `trackTypeFromInt` could not tell an old `3` from a new one without threading the schema version through. That `static_assert` is also what caught the rename mid-refactor, which is the gate working.

## The larger half

Because the enum said what a track was *called* rather than what it could *do*, every call site needing a capability derived it by listing the types that lacked it. There are sixteen such lists in `magda/daw` and no two agree:

- `takesExternalInput()` excludes two types, `canHostInstrument()` excludes three — each documented as "single source of truth".
- The file drop excluded Group and Aux. The clip drag and the keyboard nudge excluded those **plus** Master, Chord and MultiOut.

#2172 is what falls out of that: a Drum Grid track refused the MIDI files it was likeliest to want, while the identical clip could be dragged onto it from a neighbour.

Each type now declares its facts once, in `TrackTypeTraits`. Asking is a lookup. The switch filling the table has no `default`, so a track type added without declaring itself fails `-Wswitch` rather than silently inheriting whatever the first exclusion list omitted.

## What a lane takes

Having a lane is not the same as taking anything aimed at it, and the first pass of this got it wrong by using a bool — a bool can only sort tracks into Media and not-Media, and two lanes fall in between:

| Lane | Takes |
| --- | --- |
| Media | audio or MIDI |
| MultiOut | MIDI only — those notes play the parent instrument, the same way `MidiInputRouter` already routes live input arriving there to its parent. Audio has nothing to do on one. |
| Chord | chord progressions only — a `.mid` carrying `CHORD:` markers |
| Group, Aux, Master | nothing |

So `UserClipAcceptance` replaces the bool, and a track is asked what it takes **of a given kind**. A progressions lane answers yes to MIDI, because a progression *is* a MIDI clip; whether the material really is one is a question about content, asked at the drop, which is the only place holding the file.

A mixed drop now imports the kinds the track takes and says what it left, rather than refusing all of it because one file was the wrong kind — refusing everything over one file is the shape of the bug this PR removes.

The nudge asks with the selection in hand for the same reason: it moves the selection as one block, so a destination has to accept every kind in it.

## Behaviour changes

- **The Drum Grid gate is gone** — the same deletion as #2176, which targets `main` for the rc0 re-tag. Whichever lands second resolves to the same end state.
- **The file drop now refuses Group, Aux and Master**, and refuses audio on a multi-out lane. Previously it accepted all of them, and an audio clip dropped on a multi-out lane was orphaned the moment that output was switched off.
- **The nudge now allows MIDI onto a multi-out lane**, which it previously refused outright.
- `getTrackTypeName` returns `"Media"` rather than `"Audio"`. Visible to the agent DSL (`dsl_interpreter.cpp`) and the CLI; not a user-facing UI string.

## Verified

Full headless suite: 2910 passed, 13 skipped, 0 failed across 2923 cases. New `test_track_type_traits.cpp` states each type's facts as sentences rather than by re-reading `traitsOf()`, so it fails if the table is changed to something else rather than agreeing with whatever it becomes.

### DIFF
--- a/docs/specs/track-model-architecture.md
+++ b/docs/specs/track-model-architecture.md
@@ -35,14 +35,24 @@ The track model is the single source of truth. Views are lenses that present and
 
 ```cpp
 enum class TrackType {
-    Audio,      // Audio clips, recording
-    Instrument, // MIDI → plugin → audio
-    MIDI,       // MIDI only, routes to other tracks
-    Group,      // Contains child tracks, routing hub
-    Aux,        // Receives from sends
-    Master      // Final output
+    Media = 0,     // Clips of any kind, instruments, effects, input
+    Group = 3,     // Contains child tracks, routing hub
+    Aux = 4,       // Receives from sends
+    Master = 5,    // Final output
+    MultiOut = 6,  // Output track for a multi-out instrument pair
+    Chord = 7      // Singleton chord-progression track, monitor-only
 };
 ```
+
+This began as `Audio`, `Instrument` and `MIDI` as three separate types. They
+collapsed into one hybrid track, which kept the name `Audio` and so read as a
+statement about content for years after it had stopped being one; it is `Media`
+now. The ordinals that Instrument and MIDI used are skipped rather than reused,
+because the numbers are written into every `.mgd` and are pinned
+(`tests/test_persisted_enum_pins.cpp`).
+
+What each type can do is declared once in `TrackTypeTraits`, not re-derived per
+call site. See `magda/daw/core/TrackTypes.hpp`.
 
 ### 1.2 Track Hierarchy
 
--- a/examples/lua_smoke.cpp
+++ b/examples/lua_smoke.cpp
@@ -41,12 +41,12 @@ int main(int argc, char** argv) {
         magda::TrackInfo a;
         a.id = 1;
         a.name = "Drums";
-        a.type = magda::TrackType::Audio;
+        a.type = magda::TrackType::Media;
         a.volume = 0.8f;
         magda::TrackInfo b;
         b.id = 2;
         b.name = "Bass";
-        b.type = magda::TrackType::Audio;
+        b.type = magda::TrackType::Media;
         b.volume = 0.6f;
         mock.tracks_.tracks.push_back(a);
         mock.tracks_.tracks.push_back(b);
--- a/magda/agents/dsl_interpreter.cpp
+++ b/magda/agents/dsl_interpreter.cpp
@@ -1822,14 +1822,14 @@ bool Interpreter::executeSelectClips(Tokenizer& tok) {
 
 TrackType Interpreter::parseTrackType(const Params& params) {
     if (!params.has("type"))
-        return TrackType::Audio;
+        return TrackType::Media;
 
     std::string typeStr = params.get("type");
     if (typeStr == "group")
         return TrackType::Group;
     if (typeStr == "aux")
         return TrackType::Aux;
-    return TrackType::Audio;
+    return TrackType::Media;
 }
 
 int Interpreter::findTrackByName(const juce::String& name) const {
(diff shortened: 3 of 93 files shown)


lucaromagnoli:
Review findings:

- [P1] Optional examples build is broken. examples/lua_smoke.cpp:44 and :49 still use the removed TrackType::Audio enumerator, so MAGDA_BUILD_EXAMPLES=ON will fail compilation. Replace both with TrackType::Media and consider covering this configuration in CI.

- [P1] Ordinary MIDI can now be nudged onto the chord track. TrackInfo::acceptsUserClip treats Progressions exactly like MidiOnly, while nudgeSelectedClipsToAdjacentTrack passes only ClipType. Shift+arrow can therefore move any MIDI clip onto the chord lane without chord extraction, since extraction only occurs during file import. The nudge decision needs the actual clip content rather than only its coarse type.

- [P1] Mouse dragging still bypasses the new acceptance policy. Single-clip and multi-clip drag choose any visible target and invoke MoveClipToTrackCommand without calling acceptsUserClip; the command itself also does not validate the target. Audio can therefore still be dragged onto MultiOut, Chord, Group, or Master tracks while file drops and nudges reject them, preserving the inconsistency this PR is intended to remove.

Validation: track-type tests passed 53 assertions; the chord-track file-drop test passed 9 assertions; git diff --check is clean. Linux, JUCE, quality, and security CI checks were passing when reviewed; macOS and Windows were pending.

lucaromagnoli:
All three fixed in 28c32cc8. Each was real; the third one in particular — the policy was three-quarters applied, which is worse than not applied at all, since the inconsistency had just moved rather than gone.

**[P1] `examples/lua_smoke.cpp`** — fixed. The rename sweep covered `magda` and `tests`; examples are off by default so nothing caught it. Verified by compiling the file, and `TrackType::Audio` now appears nowhere in the tree. On covering the config in CI: worth doing, but it configures the whole Faust toolchain from scratch, so I've left it out of this PR rather than adding a slow job on the way past.

**[P1] Ordinary MIDI nudged onto the chord track** — fixed, and the diagnosis was exactly right: `acceptsUserClip` treats `Progressions` like `MidiOnly` because a progression *is* a MIDI clip, so a kind-level check could never tell them apart. The nudge now asks with the clip in hand via a new `trackAcceptsClip(track, clip)`, which requires the clip to already carry chord annotations.

That draws the line this PR was missing: **an import can read a file and work the harmony out — a move can only take a clip as it already is.** Detection belongs to the drop; rewriting the contents of a clip somebody dragged would be a conversion wearing a move's undo step.

**[P1] Mouse drag bypassing the policy** — fixed at the choke point rather than at the gesture. `MoveClipToTrackCommand::canExecute` checked that the clip existed and the id was non-invalid, and *nothing about the destination*, so every caller was free to invent its own rule and the two that bothered disagreed. It now validates the target, which covers the API, the scripting layer and whatever comes next. The drag and nudge also ask first, since a refused command is still an undo step that does nothing.

**Verified:** full headless suite 2911 passed / 13 skipped / 0 failed, 600,079 assertions. New case in `test_track_type_traits.cpp` pins the kind-vs-clip distinction directly — a plain MIDI clip is accepted by kind and refused with the clip in hand, a progression is accepted both ways.

**Not addressed:** the drag still *shows* an invalid target as you drag over it and then declines to move the clip. Correct outcome, poor affordance — the preview should refuse to follow. That's in the drag preview logic and I'd rather it were its own change than buried here.

PR #2178, opened 2026-08-23
Feat/2139 rack parity cases

lucaromagnoli:


tests/NullDiffCorpus.cpp
```diff
@@ -1180,6 +1237,415 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
...
+            rackTrack(
+                kTrack, "Mono",
+                rackOf(710, "Mono",
+                       {gainChain(1, "Narrowing", {{961, 0.5f}, {962, 0.5f}, {963, 0.5f}})})));
+
+        magda::getRack(value.tracks.front().chain.fxChainElements.front())
+            .chains.front()
+            .elements[1] = monoGainDevice(962, 0.5f);
+
+        auto material = noise();
```
lucaromagnoli, line 1523: **P2 — Make the mono-rack fixture stereo**

`noise()` leaves `MaterialSpec::channels` at its default of 1, so both renderers receive the same mono signal duplicated to left and right. That means this case cannot detect choosing the wrong channel during narrowing, despite the asymmetric-input coverage described above, and such a regression can still null green. Please set `material.channels = 2`; the generator already uses channel-specific seeds, so that makes the two sides distinct.

PR #2180, opened 2026-08-23
Give an MCP host a bridge path that is still there next time (#1858)

lucaromagnoli:
An MCP host stores `magda-mcp`'s absolute path once and launches it later, on its own schedule, with MAGDA possibly closed. Two platforms could not meet that with the copy staged beside MAGDA, and each failed differently.

## Linux

Inside an AppImage the bridge lives under `/tmp/.mount_…` — torn down on exit, renamed every launch. The settings page published a placeholder and disabled Copy, and the release shipped `magda-mcp` as a **separate asset** the user had to download and put on `PATH` themselves.

Now MAGDA copies it out of the mount to `<data dir>/bin/magda-mcp` and publishes that. It re-stages when size or mtime stops matching the binary in the image, so upgrading the AppImage upgrades the bridge.

Staging is done once per session: `fieldText()` runs on a one-second timer while the page is open, and retrying a failed multi-megabyte copy every tick on the message thread is worse than showing the fallback text. Tradeoff: if staging fails (read-only or full data dir), it won't retry until the next launch.

The two Linux release assets (`magda-mcp-<version>-Linux-x86_64` + `.sha256`) go away with it.

## Windows

The installer now places the bridge in `%ProgramData%\MAGDA\bin` rather than `$INSTDIR`. Hosts disagree on whether a configured command is an argv array or a string they split on whitespace, and for the splitters the space in `C:\Program Files\MAGDA` breaks the command in two. `%ProgramData%` is a fixed name on every supported Windows and carries no user name, so it is space-free whoever is logged in, and its default ACL lets a non-elevated host execute it.

Both install and uninstall clear the old `$INSTDIR` copy, so an in-place upgrade leaves nothing pointing at the path with the space.

Signing is unaffected — `sign-release.ps1` signs the staged copy before NSIS packages it, so what lands in `%ProgramData%` is the signed binary.

macOS is unchanged; the bundle path was already permanent.

## Also

- Drops the hand-rolled `/proc/self/exe` lookup for `paths::executableDir()`, which already does exactly that, and corrects the doc comment that claimed otherwise.
- The `transport_check` packaging check asserted "magda-mcp is staged beside the MAGDA executable", now wrong on two platforms. It mirrors `McpPage::bridgeExecutable` instead.
- `mcp.bridge_missing` told users to download an asset we no longer publish (en.json only — Crowdin owns the rest).

## Verified

- Debug build: 1386/1386, links clean, no diagnostics from `ConnectionsDialog.cpp`
- `transport_check` self-test: 52 checks behaved as expected, renamed check registers as the intended expected-failure
- `release.yml` valid YAML, `lang/en.json` parses, Python compiles

## Not verified — needs CI

- The Linux staging path only runs when `$APPIMAGE`/`$APPDIR` are set, so a local build never enters it. Needs a real AppImage.
- The NSIS change needs an installer build; `makensis` isn't available locally, so the script isn't syntax-checked.

Both land on the same CI job, so a tag or `workflow_dispatch` run covers them.

## Note

`dev/0.20.0` branched off `main` before `f8aa3e4c` and won't have this after 0.19.0 ships. Worth carrying across before 0.20.0 touches `magda-installer.nsi` or the AppImage staging block.

### DIFF
--- a/magda/daw/ui/dialogs/ConnectionsDialog.cpp
+++ b/magda/daw/ui/dialogs/ConnectionsDialog.cpp
@@ -5,6 +5,11 @@
 #include <utility>
 #include <vector>
 
+// For the access(X_OK) check on the staged bridge in McpPage.
+#if !JUCE_WINDOWS && !JUCE_MAC
+    #include <unistd.h>
+#endif
+
 #include "../../api/remote_api_host.hpp"
 #include "../../api/remote_audit.hpp"
 #include "../../api/remote_clients.hpp"
@@ -462,6 +467,14 @@ class McpPage : public TransportPage {
         return std::getenv("APPIMAGE") != nullptr || std::getenv("APPDIR") != nullptr;
     }
 
+    /// Whether this process could actually exec the file. `juce::File` can set
+    /// the execute bit but not ask about it, and the question here is the
+    /// effective one — owner, group, and mode together — which is what
+    /// `access(2)` answers.
+    static bool isExecutable(const juce::File& file) {
+        return ::access(file.getFullPathName().toRawUTF8(), X_OK) == 0;
+    }
+
     /**
      * @brief The AppImage's bridge, copied to a path that outlives the mount.
      *
@@ -472,8 +485,10 @@ class McpPage : public TransportPage {
      *
      * Re-staged whenever the copy no longer matches the binary in the mount, so
      * upgrading the AppImage upgrades the bridge without the user having to know
-     * a copy was ever made. Size plus modification time is enough to tell: the
-     * squashfs carries the build's timestamps, so a rebuild moves both.
+     * a copy was ever made. Size plus modification time identifies the build —
+     * the squashfs carries the build's timestamps, so a rebuild moves both — and
+     * the execute bit is checked alongside them, because a copy that is the
+     * right build but no longer runnable is no more use than a stale one.
      *
      * Returns {} if the copy cannot be made, which the page renders as its
      * "point this somewhere yourself" fallback rather than a path that will not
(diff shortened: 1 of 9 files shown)


magda/daw/ui/dialogs/ConnectionsDialog.cpp
```diff
@@ -422,38 +417,103 @@ class McpPage : public TransportPage {
...
+            return staged;
+
+        if (!staged.getParentDirectory().createDirectory())
+            return {};
+
+        // JUCE's POSIX copy is a stream copy into a freshly created file: it
+        // carries neither the execute bit nor the timestamp, and this needs
+        // both — the first to run at all, the second so the next launch can tell
+        // this copy from the one a newer AppImage would bring.
+        if (!source.copyFileTo(staged) || !staged.setExecutePermission(true))
```
lucaromagnoli: **P2 — Replace the staged bridge atomically**\n\nJUCE’s `copyFileTo` deletes the existing destination before streaming the replacement. If an AppImage upgrade hits a full disk, is interrupted, or cannot set permissions, the previously working bridge is lost and already-configured hosts break at their durable path. Please copy to a sibling temporary file, set its mode and timestamp there, then atomically replace `staged` only after every step succeeds.

magda/daw/ui/dialogs/ConnectionsDialog.cpp
```diff
@@ -422,38 +417,103 @@ class McpPage : public TransportPage {
...
+        return staged;
+    }
+
+    static juce::File stageBridgeOutOfAppImage() {
+        const auto source = paths::executableDir().getChildFile("magda-mcp");
+        if (!source.existsAsFile())
+            return {};
+
+        const auto staged = paths::dataDir().getChildFile("bin").getChildFile("magda-mcp");
+        if (staged.existsAsFile() && staged.getSize() == source.getSize() &&
```
lucaromagnoli, line 522: **P2 — Validate the execute bit before reusing the staged bridge**\n\nThis cache hit checks only size and modification time. If a backup, sync, filesystem migration, or manual change preserves those fields but drops the execute bit, MAGDA publishes a bridge that the MCP host cannot launch. Please reapply/check execute permission before returning the existing file (and fall back if that fails).

PR #2184, opened 2026-08-23
Feat/2174 device layer

lucaromagnoli:


magda/daw/audio/plugins/engine/EngineMagdaDevice.cpp
```diff
@@ -0,0 +1,313 @@
...
+        midiScratch_.clear();
+
+        if (block.midiIn != nullptr)
+            for (const auto metadata : *block.midiIn) {
+                if (midiScratch_.size() >= midiScratch_.capacity()) {
+                    jassertfalse;  // a producer past the port's budget
+                    break;
+                }
+
+                auto message = metadata.getMessage();
```
lucaromagnoli, line 340: [P2] Avoid allocating while decoding long MIDI messages

`MidiMessageMetadata::getMessage()` constructs an owning `juce::MidiMessage`; for payloads larger than its inline storage (for example SysEx), JUCE calls `allocateSpace()` here on the audio thread. That violates `EngineDevice::process()`’s no-allocation contract. The Tracktion adapter wraps its existing `MidiMessageArray` directly, so it does not pay this per-block copy. Please use preallocated backing storage or a non-owning representation for long input messages.

magda/daw/audio/plugins/engine/EngineMagdaDevice.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    // Non-owning: the executor's buffer, seen through the container the SDK
+    // takes. Nothing is copied and nothing is allocated.
+    juce::AudioBuffer<float> audio(channels_.data(), static_cast<int>(numChannels), numSamples);
+
+    std::optional<EngineMidiBufferView> midi;
+    if (block.midiIn != nullptr || block.midiOut != nullptr) {
+        midiScratch_.clear();
+
+        if (block.midiIn != nullptr)
```
lucaromagnoli, line 333: [P1] Size this scratch for the merged input bound

This capacity is derived from one producer’s `kMaxMidiBytesPerPort` budget, but the buffer arriving here may be the output of `MergeMidi`. `PlanExecutor` explicitly sizes such ports from the sum of their upstream bounds. For example, two legal sources carrying 300 one-byte events each produce a legal 600-event merged input; this branch stops at roughly 585, asserts, and silently drops the tail. Please propagate the prepared input port’s fan-in bound (or an equivalent graph-derived bound) to the adapter so valid merged MIDI is not truncated.

Issue #2081, opened 2026-08-11
Validation slice 7: the real-project corpus


lucaromagnoli:
Slice 7 of #1896. Blocked on #1891 and #1892: a corpus of real projects needs racks and
modifiers to exist before those projects can be saved, let alone rendered.

#2040 decided cases are built in code rather than loaded from `.mgd`, and that decision was
right for what it covered: a load adds a step neither leg is testing and a binary to the repo,
and thirty lines of value initialisation is reviewable as a diff. It stays right for the parity
corpus.

This slice is the other thing #1896 asked for, and it answers a different question. Cases built
in code test the rules I knew to write down. A real project tests the combinations a person
actually makes, at the size they actually make them, and it is where the interactions live: a
rack inside a rack with a modifier on a send, a project with sixty tracks, an arrangement that
is forty minutes long.

- Projects saved from the app, covering stretch, warp, racks, modifiers, sends and automation as
  #1896 lists, each rendered through both legs under the tiers from #2075.
- Each records its environment, and the audio it references is generated rather than checked in,
  the way `NullDiffMaterial` already works, so the repo does not grow a sample library.
- A project whose tier is external-plugin gets the invariant treatment rather than a null, since
  a real project is exactly where an external plugin turns up.

These are also the projects #2082 measures, so the two slices share a corpus and the fixtures
are built once.

Whether these run in the ordinary test job or a longer one is decided when their cost is known.
The clip corpus is budgeted at thirty seconds total; a forty-minute arrangement rendered twice
is not going to fit that, and pretending otherwise at planning time just means the gate gets
disabled later.

---

**Sliced, and the blockers restated.** Written after reading the rig rather than at planning time,
which is where the third one came from.

- The `.mgd` fixture rig, which is a real project turned into a case: #2173. Unblocked.
- The device layer under both legs: #2174. This is the real blocker. Neither leg can run a device
  a real project contains -- the corpus knows one device, the gain in `tests/NullDiffGain.hpp`
  implemented twice -- so #1891 and #1892 being done does not make this issue startable.
- The projects that host an external plugin: #2175. Blocked on #1893, and split out so that this
  issue is finishable without it rather than waiting on a subsystem that has not started.

What stays here is the corpus itself: which projects, what each declares, and the cost question
this issue already asks. Where they run is one paragraph rather than a slice, and it is answered
by measuring rather than by deciding in advance -- a forty-minute arrangement rendered twice was
[...]

Issue #2175, opened 2026-08-23
Real projects that host an external plugin


lucaromagnoli:
Part of #1896. Blocked on #1893, and the reason #2081 can be finished without it.

Three of the twenty-two projects in the #2079 corpus carry external VST3 state, and a real-project
corpus that quietly dropped every project with a plugin in it would be dropping the projects most
worth having. It cannot have them yet: nothing hosts VST3 in the engine, so a `Device` op for one
resolves to whatever the host hands the store, which is nothing.

The gate for them is already built, and this is what lands in it. `AudioTier::Invariants` is
implemented and tested against known-bad pairs -- a NaN, a length difference, a step past the
bound, a tail that never decays -- and no corpus case is in that tier, so the runner's branch for
it has never executed on a real comparison. A project hosting a plugin is where it was always
going to: an external plugin owes nobody a sample, so there is no null to ask it for and those
four are what remains.

The block-size half is built too, and built so it cannot be widened by accident.
`Case::blockSizeEpsilonDb` defaults to bit identity written as a level, a case that raises it is
refused unless `externalDevicesIn` finds a plugin the difference can be attributed to, and that
function returns empty for the whole corpus today. It was written for this and should find the
gate rather than widen it.

What is missing is what a fixture says about a plugin it cannot assume is installed, and what the
corpus does on a machine where it is not. A project rendered without its plugin is not that
project, and reporting that as a null would be worse than not running it: it is a case that passes
by having tested less than it claims. The environment block on a `Case` records the rate, the
block size, the channel count and the seed, and deliberately records no plugin versions, because
until #1893 there was nothing to record. This is where that field goes in, together with the rule
that a case whose plugin is absent is unmeasurable rather than green -- the same treatment a proxy
that never arrived already gets, and for the same reason.

lucaromagnoli:
Landed in #2269 (the leg wiring and the master chain) and #2275 (the plugin field on the environment block, the absent-plugin gate, and the three projects that host one).

What the corpus found on its first run with a plugin in it: a hole in a device's contiguous parameter window was published as 0.0, and retrovid has no parameters at indices 0 and 1 because it predates the wet/dry pair being persisted, so the two fabricated zeros set the pair fully dry and fully attenuated. It rendered silent out of a plugin whose chunk had restored perfectly. An undeclared slot publishes no segments now.

[...]

PR #2190, opened 2026-08-24
The .mgd fixture rig: a real project as a null-diff case (#2173)

### LINKED ISSUE #2173: The .mgd fixture rig: a real project as a null-diff case
Part of #1896, and the first thing #2081 needs.

A real project reaches the corpus as a file, and every case in the corpus today is built in code.
#2040 settled that argument for the cases it covered and the reasoning still holds there: a load
adds a step neither leg is testing, and thirty lines of value initialisation is reviewable as a
diff. It stops holding for a project somebody actually made, because what those are worth testing
for is the combinations nobody would have thought to write down.

The load is already there and it already produces model values. `ProjectSerializer::loadAndStage`
fills a `StagedProjectData` -- tracks, the master, clips, pooled sources, automation lanes and
automation clips -- on any thread and without touching a singleton, which is close to the shape a
`Case` already holds. What is missing is everything a case declares on top of the project: the
range to render, the tier and the figures that tier needs, and the environment.

So a fixture is the file plus a manifest, and the manifest is where the declarations live. They
cannot live in the file. A `.mgd` is what the app writes, and there is nowhere in it to say what
two engines owe each other.

**The audio is generated rather than checked in.** A saved project references its sources by path
and those paths point at machines that are gone. Each fixture names, per source, the
`MaterialSpec` (NullDiffMaterial.hpp) that stands in for it; the rig writes it into the scratch
directory and repoints the source before either leg sees the case. Same rule as the rest of the
corpus and the same reason: a fixture recorded once is a fixture nobody can regenerate when a case
needs one bar more, and a corpus carrying a sample library grows a megabyte per project.

Material is chosen per source the way #2040 chooses it per case, and for the same reason. A
stretched clip fed impulses reports the distance between two interpolators rather than anything
about placement. So the manifest names a kind, and choosing it is a judgement about what stands
between the two engines on that source's path.

**A `SourceFact` says what a file is so no leg has to probe it**, which makes the manifest's
declared rate and duration something to check against what was actually written rather than
something to trust.

This slice renders nothing, and asserts four things about a fixture: it loads; the case it
produces matches what the manifest declares; every source it names exists and reads back at the
rate and length declared; and a source the project references that the manifest does not name
fails rather than passing quietly. That last one is the failure worth building the slice around --
a case that reached a leg with a missing source renders silence, and silence nulls against silence
perfectly well.

**Whether these are the #2079 fixtures is a decision this slice has to make.** Twenty-two real
projects are already checked in under `tests/corpus/legacy/projects`, saved by builds from the
oldest openable format through 0.18.0, and reusing them costs nothing to store. Two things push
the other way. Those bytes are migration fixtures and must never be rewritten, so a render corpus
can never ask one of them for one bar more or for a source that suits it better. And their sources
were whatever the person had on disk that day, which is not the same as sources chosen so that a
residual means something. Reuse where a project earns its place, save new ones where the material
has to be designed, and record which each one is.

`LegacyCorpus.hpp` points at `tests/corpus/legacy/README.md`, which does not exist. Whatever is
decided above belongs there, next to the rule about never rewriting the bytes.

lucaromagnoli:
Closes #2173. First of the three things #2081 is waiting on.

Every case in the corpus is built in code, and #2040 settled that argument for the cases it covered: a load adds a step neither leg is testing and a binary to the repository, and thirty lines of value initialisation is reviewable as a diff. That still holds and this does not replace it. It stops holding for a project somebody actually made, because what those are worth testing for is the combinations nobody would have thought to write down.

## A fixture is a file plus a manifest

`ProjectSerializer::loadAndStage` already produces model values on any thread without touching a singleton, and what it fills is close to the half of a `Case` that describes a project. The manifest carries the other half, the half a `.mgd` has nowhere to put: the range, the tier and its figures, the environment. The load fills the project half over the top, so a manifest can never quietly assert a project different from the one on disk.

The audio is generated rather than checked in. Every path in these files points at a machine that is gone, so the manifest names a `MaterialSpec` per source and the rig writes it, repoints the project, and reads the file back rather than trusting the spec. It drives the app's own install for that: `installStagedSources` clears the pool, preserves v2 ids, acquires the v1 sources under real ones and remaps the events that referenced the provisional ids, which is a paragraph of behaviour a second implementation would get subtly wrong and then agree with itself about.

## What it refuses

This slice renders nothing. It asserts the four things #2173 asked for, plus two the issue did not name:

1. The fixture loads.
2. The case is the project on disk plus the manifest's declarations, and nothing else.
3. Every source reads back at the rate and length declared.
4. **A source the project plays that the manifest does not name is refused.** This is what the rig is built around: such a source reaches a leg as a file that is not there, renders silence, and nulls against the other leg's silence at the ordinary floor. In a report that looks exactly like a case that passed.
5. A declaration no source claims is refused, for the reason the DAWproject loss table refuses one.
6. A project whose sources cannot be told apart by the only part of a path a manifest can name is refused. See below.

## The first fixture

`0.13.0-retrovid.mgd`, reused rather than saved: four tracks and nine clips at 92 bpm over eight sources from three dead volumes, two of them bounces of other clips in the same project. Impulses where a clip plays at its own rate, tones where the project stretched it, because impulses through two stretchers measure the distance between two interpolators rather than anything about the clip.

## The decision #2173 delegated

Reuse or save new is answered per fixture in `isMigrationFixture`: reuse where a project earns its place by being an arrangement, save new where the material has to be designed and the fixture has to be free to change, which a migration fixture never is. That is written into `tests/corpus/legacy/README.md`, which `LegacyCorpus.hpp` has pointed at since #2079 and which did not exist.

## Worth a close read: the collision guard

A manifest names a source by the last component of its path, because the rest names a volume that is gone. That leaves one thing a project can do the manifest cannot express: play two different files that share a name. They match one declaration, are written to one file, and collapse for real in the install, which dedups by canonical path. Two clips come out playing one sound and the case renders a project that never existed.

Four commits here are that guard, each fixing a defect in the previous one, and all four were found by review rather than by me:

- Names compared exactly missed `Loop.wav` against `loop.wav`, which is one file on macOS and Windows. Compared folded now, conservative on a case-sensitive filesystem on purpose: a corpus that refuses on Linux and accepts on macOS is worse than one that refuses everywhere.
- A follow-up check that compared the written paths claimed to catch a Unicode normalisation collapse and could not, for the same reason folding cannot: NFC and NFD spellings are two strings and one file. The load counts what its own directory holds after each write instead. There is a test that writes both spellings and checks what the machine actually does.
- Counting against sources processed then refused a project that names one file twice, which a v2 table entry beside a v1 clip produces and which `refuseIndistinguishableSources` allows on purpose. Counted against distinct paths now.

The reuse branch has no fixture exercising it, since no project here has that shape and synthesising one to break the rig is not worth checking in. What guards it instead is an invariant the load asserts on every fixture: one pooled source per source the manifest names. Too few means a collapse past every check above; too many means a repeat that did not collapse. Both are otherwise silent.

`FixtureLoad::written` is keyed by the manifest's name rather than ordered, because it was documented as manifest order and was in staged order, and the test reading files back was pairing the two by index.

## Testing

Full Catch2 suite green: 2927 passed, 13 skipped, 0 failed. The rig is 8 tests, 121 assertions.

The test's pool guard snapshots and restores rather than clearing. The rig's install empties the shared `SourcePool`, so a fixture loaded in a test binary would otherwise pull the null-diff corpus's sources out from under it and leave every `SourceFact` in it pointing at an id nobody holds. Ordering is not a defence: the corpus builds itself lazily and once. #2081 will need the same discipline when its runner starts loading fixtures.

### DIFF
--- a/tests/MgdFixture.cpp
+++ b/tests/MgdFixture.cpp
@@ -54,9 +54,19 @@ std::string quoted(const juce::String& text) {
 }
 
 /// The last component of a path that may name a volume this machine has never
-/// had. Built without checking, because checking is what it cannot do.
+/// had, and a machine that is not this one.
+///
+/// Split here rather than by juce::File, which cuts on the separator of the host
+/// doing the reading: getSeparatorChar() is a backslash on Windows, so a project
+/// saved on macOS hands it a path of forward slashes and it hands back the whole
+/// path as the name. Every manifest lookup then misses and the load refuses a
+/// fixture that is fine, which is what CI found. The mirror holds too, a project
+/// saved on Windows and read on macOS.
+///
+/// A path in a saved project belongs to the machine that saved it, so both
+/// separators are cut whatever this one uses.
 juce::String fileNameOf(const juce::String& path) {
-    return juce::File::createFileWithoutCheckingPath(path).getFileName();
+    return path.fromLastOccurrenceOf("/", false, false).fromLastOccurrenceOf("\\", false, false);
 }
 
 using Declarations = std::map<juce::String, const FixtureSource*>;
--- a/tests/MgdFixture.cpp
+++ b/tests/MgdFixture.cpp
@@ -110,12 +110,29 @@ std::string matchSources(const MgdFixture& fixture, const std::vector<Source*>& 
  * other than what it declared.
  */
 std::string writeAndRepoint(const std::vector<Source*>& sources, const Declarations& declared,
-                            const juce::File& directory, std::vector<juce::File>& written) {
+                            const juce::File& directory,
+                            std::map<juce::String, juce::File>& written) {
     auto& pool = SourcePool::getInstance();
 
+    // The stand-in already written for a path, so a project that names one file
+    // twice gets one file back. That is not a corner case: a v2 project can
+    // carry a pooled table entry beside a v1 clip that migrated to the same
+    // path, which is the pair installStagedSources exists to collapse. Two
+    // names for one sound is one sound.
+    std::map<juce::String, juce::File> byOriginalPath;
+
     for (auto* source : sources) {
-        const auto name = fileNameOf(source->filePath);
+        const auto original = source->filePath;
+        const auto name = fileNameOf(original);
         const auto* declaration = declared.at(name);
+
+        if (const auto seen = byOriginalPath.find(original); seen != byOriginalPath.end()) {
+            const auto facts = factsOf(seen->second);
+            source->filePath = seen->second.getFullPathName();
+            source->sampleRate = facts.sampleRate;
+            source->durationSeconds = facts.durationSeconds;
+            continue;
+        }
 
         const auto file = writeMaterial(directory, name.upToLastOccurrenceOf(".", false, false),
                                         declaration->material);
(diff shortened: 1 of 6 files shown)


tests/MgdFixture.cpp
```diff
@@ -0,0 +1,349 @@
...
+        result.failure = std::move(refusal);
+        return result;
+    }
+
+    // The app's own install, not a copy of it. It clears the pool, inserts the
+    // v2 table with ids preserved, acquires the v1 sources under real ids and
+    // remaps the events that referenced the provisional ones -- which is a
+    // paragraph of behaviour a second implementation would get subtly wrong and
+    // then agree with itself about.
+    ProjectSerializer::installStagedSources(staged.sources, staged.legacySources, staged.clips);
```
lucaromagnoli, line 434: **P1 — Keep previously loaded cases resolvable**

installStagedSources clears SourcePool and resets its ID allocator. Each returned Case retains only numeric source IDs, while the incumbent leg later resolves those IDs through the global pool. Loading fixture B can therefore make fixture A silently resolve its clips to B files, and loading any fixture also invalidates the existing code-built corpus mappings. Please install the Case SourceFacts immediately before rendering, or otherwise preserve globally unique mappings, and cover this with two simultaneously held fixture cases.

tests/MgdFixtures.cpp
```diff
@@ -0,0 +1,159 @@
...
+        // three different dead volumes -- an external SSD, a home Music folder,
+        // and a bounces directory under a project that no longer exists. Two of
+        // the eight are bounces of other clips in the same project, which is a
+        // shape a case built in code would never have thought to have.
+        //
+        // Reused rather than saved: it earns its place by being an arrangement
+        // rather than a demonstration, and being a migration fixture costs this
+        // case nothing it wanted.
+        MgdFixture fixture;
+        fixture.file = "legacy/projects/0.13.0-retrovid.mgd";
```
lucaromagnoli: **P1 — Use an internal-only project for this slice**

0.13.0-retrovid.mgd contains two Retrospect VST3 devices, but declarationFor currently marks the case AudioTier::Exact. Without Retrospect installed, both legs can pass by rendering passthroughs; with it installed, the incumbent may host it while the native leg cannot, so the result becomes machine-dependent. Issue #2175 explicitly reserves these external-plugin projects for the invariant tier plus an absent-plugin-is-unmeasurable gate. Please choose an internal-only project here, or move this fixture under that later handling.

PR #2191, opened 2026-08-24
Collapse the project media tree to three roots (#2170)

### LINKED ISSUE #2170: Revisit the project media folder layout: redundant and partially unmanaged subdirectories
A project's media tree is `<ProjectName>_Media/` with six declared subdirectories (`ProjectManager.cpp:22-27`): `recordings`, `renders`, `bounces`, `external-edits`, `imported`, `stems`. The set has grown one folder per feature and I no longer think it earns six roots.

### Redundancy

- **`renders` vs `bounces`** hold the same kind of thing: audio MAGDA rendered itself, living inside the project tree because a clip references it. `renders` = comp render (`CompService.cpp:234`) and clip/track render (`ClipCommands.cpp:1232`, `:1387`); `bounces` = the bounce commands (`ClipCommands.cpp:2067`, `:2277`). Nothing in the user-facing model draws that line.
- **`external-edits` / `imported` / `stems`** are three roots for "audio that arrived from outside this timeline", each with exactly one caller: `ClipManager.cpp:1002`, `MediaCollector.cpp:74`, `StemSeparationService.cpp:236`.

### Inconsistencies to fix whatever the layout ends up being

- `ensureMediaSubdirectories()` (`ProjectManager.cpp:730`) creates only 4 of the 6. `imported` and `stems` are created ad hoc by their callers.
- `migrateMediaFiles()` (`ProjectManager.cpp:745`) moves only the same 4. Save-As or a project move therefore leaves `imported/` and `stems/` behind in the old media tree while their clips keep pointing at the old path.
- The migration iterator is non-recursive (`RangedDirectoryIterator(srcDir, false)`), and stems nest one level deeper (`stems/<name> - <model>/`, `StemSeparationService.cpp:338`), so stems would still not move even after adding them to the list.

### Scope

1. Decide the final layout - which of the six survive, and what each one means to the user.
2. Make one list the single source of truth, consumed by creation, migration and `MediaCollector` alike.
3. Make migration recursive so nested content moves.
4. Migrate existing projects on load: fold retired folders into their replacement and relink the clips.

lucaromagnoli:
Closes #2170.

## Layout

Six roots become three, each with a meaning to the user:

```
MyProject_Media/
├── recordings/        captured from an input
├── renders/           anything MAGDA computed from the timeline
│   ├── comp_42.wav              comp render
│   ├── Kick_2026-08-24.wav      clip/track render, bounce
│   └── Vocals - htdemucs/       one folder per stem split
└── imported/          anything that arrived from outside
    ├── amen_break.wav           collected files
    └── Vox.wav                  copy-on-edit for an external editor
```

`renders` and `bounces` held the same kind of thing, and `external-edits` / `imported` / `stems` were three roots for "audio from outside this timeline" with exactly one caller each.

## Single source of truth

`kMediaSubdirs[]` is now the only list, read by both creation and migration. Previously `ensureMediaSubdirectories()` created four of the six and `migrateMediaFiles()` moved the same four — so `imported/` and `stems/` were created ad hoc by their callers, and a Save-As or project move stranded them in the old tree while their clips kept pointing at the old path. `kLegacyMediaSubdirs[]` maps each retired root to its replacement.

## Migration

- **Recursive**, so the folder a stem split nests (`stems/<source> - <model>/`) actually moves. The old iterator was `RangedDirectoryIterator(srcDir, false)`.
- Reports an old→new path map instead of rewriting path prefixes. A name already taken at the destination is uniquified rather than clobbered — folding two roots into one can collide, and overwriting would destroy a file another clip still references.
- Relinks **sampler and drum-pad samples** alongside clip sources and takes. Collected samples live in `imported/`, which never migrated at all before. Also walks every audio event per clip rather than only `primaryEvent()`.

## Existing projects

On load, `foldLegacyMediaDirectories()` moves the retired roots into their replacement and relinks in memory. That leaves the `.mgd` on disk naming folders that just went away, so the project is marked dirty and the user gets the save prompt on close. If they decline, the next load still resolves: a path under a retired root whose file is gone is followed to the same relative path under the replacement when that file exists. The one case that can't self-heal is a name that collided during the fold and was uniquified — that clip reads as missing until the user saves or relinks.

Save-As has no such window: `migrateMediaFiles()` folds on the way across and runs before serializing, so tree and file always agree.

If a move fails (read-only volume, permissions) no mapping is recorded, the legacy folder isn't deleted because files remain, and the clip keeps pointing at the file that's still there — it degrades to a no-op rather than to broken links.

## Call sites

Bounce commands → `getRendersDirectory()`; stem separation → `renders/<source> - <model>/`; external edits → `getImportedDirectory()`. `getBouncesDirectory()`, `getExternalEditsDirectory()` and `getStemsDirectory()` are gone. User-facing error strings updated to match.

## Testing

Three new sections in `test_project_serialization.cpp`: nested content survives Save-As, a saved project declares exactly the three roots and none of the retired ones, and loading a legacy project folds all three retired roots (including the nested stem folder), relinks the clips and marks dirty.

Verified negatively — making the iterator non-recursive fails the nested tests, and disabling the fold call fails the fold test.

Full Catch2 suite green (2920 passed, 13 skipped) plus the four ProjectManager-touching JUCE suites.

### DIFF
--- a/magda/daw/audio/ClipCommands.cpp
+++ b/magda/daw/audio/ClipCommands.cpp
@@ -1248,7 +1248,8 @@ void RenderClipCommand::execute() {
     juce::String trackName = trackInfo ? trackInfo->name : "Track";
     juce::String clipName =
         clip->name.isNotEmpty() ? clip->name : sourceFile.getFileNameWithoutExtension();
-    renderedFile_ = rendersDir.getChildFile(expandRenderPattern(clipName, trackName) + ".wav");
+    renderedFile_ =
+        rendersDir.getNonexistentChildFile(expandRenderPattern(clipName, trackName), ".wav", false);
 
     const double projectBPM = currentProjectBpm();
     const double renderStart = clip->getTimelineStart(projectBPM);
@@ -1403,8 +1404,8 @@ void RenderTimeSelectionCommand::execute() {
         juce::String trackName = trackInfo ? trackInfo->name : "Track";
         auto* firstClipInfo = clipManager.getClip(overlappingIds[0]);
         juce::String clipName = firstClipInfo ? firstClipInfo->name : trackName;
-        trackState.renderedFile =
-            rendersDir.getChildFile(expandRenderPattern(clipName, trackName) + ".wav");
+        trackState.renderedFile = rendersDir.getNonexistentChildFile(
+            expandRenderPattern(clipName, trackName), ".wav", false);
 
         const auto& project = ProjectManager::getInstance().getCurrentProjectInfo();
         OfflineRenderRequest request;
@@ -2093,7 +2094,12 @@ void BounceInPlaceCommand::execute() {
         auto* trackInfo = TrackManager::getInstance().getTrack(clip->trackId);
         juce::String trackName = trackInfo ? trackInfo->name : "Track";
         juce::String clipName = clip->name.isNotEmpty() ? clip->name : "clip";
-        renderedFile_ = rendersDir.getChildFile(expandBouncePattern(clipName, trackName) + ".wav");
+        // Renders and bounces share this folder and expand independently
+        // configurable patterns, so the two can name the same file. Claim a
+        // free name: overwriting would leave another clip playing this audio
+        // and let undo delete a file it does not own.
+        renderedFile_ = rendersDir.getNonexistentChildFile(expandBouncePattern(clipName, trackName),
+                                                           ".wav", false);
     }
 
     // Match the original bounce lifecycle: stop before capture or FX bypass,
@@ -2303,7 +2309,12 @@ void BounceToNewTrackCommand::execute() {
         auto* trackInfo = TrackManager::getInstance().getTrack(clip->trackId);
         juce::String trackName = trackInfo ? trackInfo->name : "Track";
         juce::String clipName = clip->name.isNotEmpty() ? clip->name : "clip";
-        renderedFile_ = rendersDir.getChildFile(expandBouncePattern(clipName, trackName) + ".wav");
+        // Renders and bounces share this folder and expand independently
+        // configurable patterns, so the two can name the same file. Claim a
+        // free name: overwriting would leave another clip playing this audio
+        // and let undo delete a file it does not own.
+        renderedFile_ = rendersDir.getNonexistentChildFile(expandBouncePattern(clipName, trackName),
+                                                           ".wav", false);
     }
 
     PlaybackResumeScope playbackScope(*engine_);
--- a/magda/daw/audio/ClipCommands.cpp
+++ b/magda/daw/audio/ClipCommands.cpp
@@ -2068,23 +2068,23 @@ void BounceInPlaceCommand::execute() {
         return;
     }
 
-    // Determine output file path — always the project's bounces directory.
+    // Determine output file path — always the project's renders directory.
     // A bounce file is referenced by a clip inside the project, so it must live
     // in the project media tree (so save/collect/move keeps it). It must NOT
     // honour Config::getRenderFolder(): that is the global Export-to-file
     // folder, and letting it hijack bounce output drops the file outside the
     // project (e.g. on the Desktop) where the project can't track it.
-    auto bouncesDir = ProjectManager::getInstance().getBouncesDirectory();
-    if (bouncesDir == juce::File())
-        bouncesDir = engine_->getEditFile().getParentDirectory().getChildFile("bounces");
+    auto rendersDir = ProjectManager::getInstance().getRendersDirectory();
+    if (rendersDir == juce::File())
+        rendersDir = engine_->getEditFile().getParentDirectory().getChildFile("renders");
     // Verify the destination is actually writable before handing the renderer a
     // dead destFile. On a read-only / unavailable project media tree the render
     // would otherwise fail with only a DBG line and no bounced clip, leaving the
     // user with nothing and no explanation. Bail with a visible error instead.
     // Safe to return here: the transport hasn't been touched yet.
-    if (bouncesDir.createDirectory().failed() || !bouncesDir.hasWriteAccess()) {
+    if (rendersDir.createDirectory().failed() || !rendersDir.hasWriteAccess()) {
         errorMessage_ =
-            "Bounce failed: can't write to the bounces folder:\n" + bouncesDir.getFullPathName();
+            "Bounce failed: can't write to the renders folder:\n" + rendersDir.getFullPathName();
         juce::Logger::writeToLog(errorMessage_);
         return;
     }
@@ -2093,7 +2093,7 @@ void BounceInPlaceCommand::execute() {
         auto* trackInfo = TrackManager::getInstance().getTrack(clip->trackId);
         juce::String trackName = trackInfo ? trackInfo->name : "Track";
         juce::String clipName = clip->name.isNotEmpty() ? clip->name : "clip";
-        renderedFile_ = bouncesDir.getChildFile(expandBouncePattern(clipName, trackName) + ".wav");
+        renderedFile_ = rendersDir.getChildFile(expandBouncePattern(clipName, trackName) + ".wav");
     }
 
     // Match the original bounce lifecycle: stop before capture or FX bypass,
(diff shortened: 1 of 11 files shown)


magda/daw/project/ProjectManager.cpp
```diff
@@ -66,6 +82,112 @@ void resetTransportForProjectBoundary() {
...
+        clipManager.forceNotifyMultipleClipPropertiesChanged(updatedClipIds);
+
+    // Collected samples live in imported/, so a media tree that moves takes the
+    // samplers and drum pads pointing into it along too.
+    bool relinkedSamplers = false;
+    if (auto* audioEngine = TrackManager::getInstance().getAudioEngine()) {
+        for (auto& reference : audioEngine->getSamplerMediaReferences()) {
+            const auto newPath = resolve(reference.source.getFullPathName());
+            if (newPath.isNotEmpty()) {
+                reference.replace(juce::File(newPath));
```
lucaromagnoli, line 231: [P1] Re-capture sampler state after Save-As migration

`saveProjectAs()` calls `onBeforeSave` before `migrateMediaFiles()`, so this replacement updates the live sampler only after `DeviceInfo::pluginState` has already been captured. The `.mgd` therefore retains the old sample path; after the media is moved, reopening the saved-as project leaves sampler and drum-pad samples missing. Please migrate before capture or capture plugin state again after relinking.

magda/daw/project/ProjectManager.cpp
```diff
@@ -811,6 +871,46 @@ void ProjectManager::migrateMediaFiles(const juce::File& oldDir, const juce::Fil
...
+        // them to where they live now — unless the fold had to rename, in which
+        // case the file is genuinely ambiguous and the clip stays unresolved
+        // until the user saves.
+        for (const auto& legacy : kLegacyMediaSubdirs) {
+            const auto prefix = rootPath + juce::File::getSeparatorString() + legacy.from +
+                                juce::File::getSeparatorString();
+            if (!path.startsWith(prefix))
+                continue;
+
+            const auto moved =
```
lucaromagnoli: [P2] Avoid filename-only stale recovery after a collision

If `renders/foo.wav` already exists and legacy `bounces/foo.wav` is folded, `moveMediaTree()` correctly moves the legacy file to a unique sibling. If the user closes without saving and reopens, the move map is empty and this fallback resolves stale `bounces/foo.wav` to the pre-existing `renders/foo.wav`, silently substituting unrelated audio. The unique mapping needs to survive reload, or collision cases must not use this fallback.

magda/daw/audio/ClipCommands.cpp
```diff
@@ -2093,7 +2093,7 @@ void BounceInPlaceCommand::execute() {
         auto* trackInfo = TrackManager::getInstance().getTrack(clip->trackId);
         juce::String trackName = trackInfo ? trackInfo->name : "Track";
         juce::String clipName = clip->name.isNotEmpty() ? clip->name : "clip";
-        renderedFile_ = bouncesDir.getChildFile(expandBouncePattern(clipName, trackName) + ".wav");
+        renderedFile_ = rendersDir.getChildFile(expandBouncePattern(clipName, trackName) + ".wav");
```
lucaromagnoli: [P2] Uniquify bounce output in the shared renders directory

Bounces now share `renders/` with clip and track renders, but both producers use deterministic `getChildFile()` paths with independently configurable patterns. If the names coincide, Tracktion truncates the existing file; another clip can then play the wrong audio, and undo can delete the shared file. Please choose a nonexistent sibling here and in `BounceToNewTrackCommand` (line 2306).

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Fixed in 382735b0 — `onBeforeSave()` now runs after `migrateMediaFiles()` rather than at the top of `saveProjectAs()`. Nothing in the capture depends on the migration, and the migration does not read captured state, so the reorder is the whole fix.

Guarded by `Plugin state is captured after the media migration`, which installs an `onBeforeSave` that records whether the migrated file had already landed at its destination. Verified negatively: moving the call back to the top fails that assertion.

magda/daw/project/ProjectManager.cpp
lucaromagnoli: You are right, and my code comment was wrong about it — I claimed the clip "stays unresolved", but the fallback resolves to the pre-existing file. Silent substitution, not a missing link.

Fixed in 382735b0 by making the collision unable to arise rather than by persisting the map. `moveMediaTree()` takes an `OnCollision` policy:

- The legacy fold uses `Skip`. A file whose destination name is taken is left where it is, so the path already in the `.mgd` stays valid and the folder survives to be retried next load.
- Save-As keeps `Uniquify` — it writes the new paths from the relinked model in the same operation, and must not strand anything in the directory it is leaving.

That makes the fallback sound: a file that would have had to share a name never moved, so a name that has left the retired root and is present in the replacement can only be the same file. I also added an `existsAsFile()` guard so a skipped file is never rewritten.

Covered by `Folding leaves a colliding legacy file where it is`, which asserts both files keep their original contents and the clip still points into `bounces/`. Verified negatively: switching the fold back to `Uniquify` fails it.

magda/daw/audio/ClipCommands.cpp
lucaromagnoli: Fixed in 382735b0, on all four output sites rather than the two bounce ones — `RenderClipCommand` and `RenderTrackCommand` write into the same folder, so uniquifying only the bounce side would still let a render truncate a bounce. All four now use `getNonexistentChildFile(pattern, ".wav", false)`, matching the idiom in `externalEditFileForClip`.

Undo deletes `renderedFile_` in each of these commands, so redo re-derives the same name from a now-free slot — no leak.

Not covered by a test: exercising it needs a real render through the engine. Flagging that rather than implying it is verified.

magda/daw/project/ProjectManager.cpp
```diff
@@ -66,6 +82,127 @@ void resetTransportForProjectBoundary() {
...
+        clipManager.forceNotifyMultipleClipPropertiesChanged(updatedClipIds);
+
+    // Collected samples live in imported/, so a media tree that moves takes the
+    // samplers and drum pads pointing into it along too.
+    bool relinkedSamplers = false;
+    if (auto* audioEngine = TrackManager::getInstance().getAudioEngine()) {
+        for (auto& reference : audioEngine->getSamplerMediaReferences()) {
+            const auto newPath = resolve(reference.source.getFullPathName());
+            if (newPath.isNotEmpty()) {
+                reference.replace(juce::File(newPath));
```
lucaromagnoli, line 231: [P1] Preserve sampler settings while relocating media

This `replace` callback is implemented by `TracktionEngineWrapper` as `MagdaSamplerPlugin::loadSample()`. That method treats the destination as a newly chosen sample: it overwrites the root note and resets sample start/end plus loop start/end. During Save-As, the reordered `onBeforeSave()` then captures those reset values into the `.mgd`; a legacy fold mutates the loaded project the same way. A sampler or drum pad with a custom root or trimmed/looped region therefore changes merely from moving its file. Please use a relocation path that preserves these interpretation settings, or snapshot and restore them around the reload.

magda/daw/project/ProjectManager.cpp
```diff
@@ -811,6 +896,50 @@ void ProjectManager::migrateMediaFiles(const juce::File& oldDir, const juce::Fil
...
+            const auto prefix = rootPath + juce::File::getSeparatorString() + legacy.from +
+                                juce::File::getSeparatorString();
+            if (!path.startsWith(prefix))
+                continue;
+            if (juce::File(path).existsAsFile())
+                return {};  // Skipped by the fold — it still lives where it says.
+
+            const auto moved =
+                mediaRoot.getChildFile(legacy.to).getChildFile(path.substring(prefix.length()));
+            if (moved.existsAsFile())
```
lucaromagnoli: [P2] Require evidence before following a missing legacy path

The collision policy makes reload safe when this version actually skipped or moved a file, but this fallback also runs on the first upgrade load. If the `.mgd` already references a missing `bounces/foo.wav` and an unrelated `renders/foo.wav` exists, there is no source file to collide or enter in `moves`; the `existsAsFile()` guard fails and this line silently relinks the clip to the unrelated render. Recovering a previously folded path needs durable evidence such as a stored migration map or marker, rather than destination-name existence alone.

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Confirmed and fixed in 40d5a93e. `loadSample()` re-derives the root note from file metadata and resets sample start/end and loop start/end, which is right for a sample the user picked and wrong for one that moved. Worth noting this was not only the new paths — `MediaCollector::apply()` drives the same callback, so "Collect files" has been doing it too.

Added `MagdaSamplerPlugin::relocateSample()`: reloads the audio, then restores root note and all four markers (clamped to the new length and the parameter range, via the file's existing `clampToParameterRange`). It bails without touching anything if `loadSample()` did not adopt the path, since in that case nothing was reset either. `getSamplerMediaReferences()` now calls it, which covers standalone samplers and drum pads and fixes the collect path as a side effect. `loadSample()` is unchanged.

New `tests/test_sampler_relocate_juce.cpp` covers both directions: relocation preserves root note and trim/loop markers, and choosing a genuinely different sample still resets them. Verified negatively — neutering `relocateSample()` fails 5 assertions (root note 48 -> 60, all four markers back to defaults).

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Right — the collision policy only covers files this version actually moved, and says nothing about a legacy path that was already dangling. Fixed in 40d5a93e by taking the durable-evidence option.

A fold that moves anything now writes `.magda-media-moves.json` at the media root: `{"bounces/foo.wav": "renders/foo.wav"}`, paths relative to the root with `/` separators so the record survives the project folder being moved, renamed, or opened on another platform. Resolution is an exact lookup in that record — the name-existence heuristic is gone entirely, so an unrecorded path is never rewritten no matter what shares its name.

`OnCollision::Skip` stays, but now on its own merit rather than as the justification for the heuristic: an automatic migration should not rename a user's files, and the retired folder simply gets retried on the next load.

Two tests: a folded-but-unsaved path resolves through the record, and your exact case — a `.mgd` naming a missing `bounces/ghost.wav` beside an unrelated `renders/ghost.wav` — leaves the clip pointing where it was and the render untouched. Verified negatively by restoring the name guess, which relinks across.

PR #2195, opened 2026-08-24
Run the compiled Faust devices under the native engine (#2192)

### LINKED ISSUE #2192: The devices MAGDA ships do not run under the native engine
The native engine (#1882) cannot run most of the devices MAGDA ships, and that is now the thing blocking validation rather than anything in the validation harness.

## The measurement

`tests/test_real_project_survey.cpp` loads all twenty-two projects in the legacy corpus and renders each through the native engine. One can be a render case today.

Two come back clean. `1.0.0-master-eq.mgd` has no tracks and no clips, so it renders silence and the runner's own silence guard would refuse it. That leaves `0.16.0-reverse.mgd`: one track, one clip, one source.

The other twenty stop at the same wall, and every diagnostic has the same form. Projects stopped, by device:

| device | projects | | device | projects |
| --- | --- | --- | --- | --- |
| `4osc` | 7 | | `magda_eq` | 2 |
| `magda_filter` | 4 | | `magda_grit` | 2 |
| `drumgrid` | 3 | | `magda_limiter` | 2 |
| `magda_polysynth` | 3 | | `magda_reverb` | 2 |
| `magda_saturator` | 3 | | `magda_ring_mod` | 2 |
| `faust` | 2 | | `sidechain` | 2 |
| `magda_chorus` | 2 | | eleven others | 1 each |
| `magda_compressor` | 2 | | | |

## Why

A device is in one of two shapes.

```cpp
// Bell: the native engine runs it
class MagdaBellCompiledPlugin : public MagdaCompiledPolyInstrument   // -> CompiledFaustDevice -> MagdaDevice
    spec: .createDevice = ...

// Clipper: it cannot
class MagdaClipperCompiledPlugin : public te::Plugin, public ICompiledFaustPlugin
    spec: .createPlugin = ...
```

A device whose DSP sits behind `MagdaDevice` runs under both engines: `TracktionMagdaDevicePlugin` wraps it for the fork, and `EngineDeviceFactory` hands it to the plan. A device that *is* a `te::Plugin` runs only under the fork, and the factory reports `no native device for <id>` rather than standing something in (#2174).

Twelve of sixty-two registered devices are in the first shape: nine compiled instruments through `MagdaCompiledPolyInstrument`, plus Faust Instrument, Oscilloscope and Spectrum Analyzer. The instruments crossed because they got a shared base that absorbs the plumbing; `MagdaBellCompiledPlugin` is 115 lines against `MagdaFilterCompiledPlugin`'s 752.

## What is actually left

The fifty are not fifty equal jobs.

**The compiled Faust effects (~22).** Bitcrusher, Chorus, Clipper, Compressor, Delay, Dimension, EQ, Filter, Flanger, Freq Shift, Gate/Expander, Grain Delay, Grit, Limiter, Mod, Multiband, Phaser, Pitch, Poly Synth, Reverb, Ring Mod, Saturator, Utility. These are the survey's blockers and they share one shape, so they want the effect equivalent of `MagdaCompiledPolyInstrument` and then a thin subclass each. This is the slice worth doing first.

**Fork plumbing (~9).** SidechainMonitor, AudioSidechainMonitor, FollowerSourceTap, InsertCapture, InstrumentMeterTap, SessionMonitor, Levels, TrackMeasurement, MidiReceive, and TE's own VolumeAndPan and LevelMeter. These should never cross: the plan does these structurally as ops (`Fader`, `Meter`, `ModSource`, `SendTap`), so they are Tracktion-only by design and the survey should stop counting them.

**Hand-written C++ devices (~10).** DrumGrid, Sampler, Arpeggiator, MIDI Chord Engine, Strum, the two step sequencers, Sidechain, Convolution, and the three Mutable ports. Individual jobs with no shared shape.

**4osc.** Its own decision. It is a lifted GPL port that #1902's v1 plan already has removed.

## What a port costs, per device

Not a rename. Measured on Clipper:

- the base class changes from `te::Plugin` to the SDK's;
- parameter storage moves from `te::AutomatableParameter` to the SDK's own handles, and the Tracktion adapter re-exposes them;
- `initialise` / `deinitialise` / `reset` / `applyToBuffer` become `prepare` / `release` / `reset` / `process`, over `DevicePrepareContext` and `DeviceProcessContext` instead of TE's;
- `properties()` replaces the TE capability overrides;
- the custom UI has to follow. `CompiledClipperCurveView::bindPlugin` takes a `te::Plugin*` and casts to the concrete class, which stops existing.

The registration is the last line, not the work.

## Done when

The survey reports the legacy corpus rendering, and #2081 can be written against real projects instead of one.

lucaromagnoli:
Closes #2192 for the compiled-Faust slice.

#2192 measured that the native engine could not run most of the devices MAGDA
ships, and named the compiled Faust effects as the slice worth doing first: they
are the survey's blockers and they share one shape. All twenty-two of them cross
here, plus Poly Synth.

## Why they were stuck

Not because any one of them was hard. Each carried its own copy of the same six
hundred lines: an `[idx:N]` zone harvester, a normalized parameter table, a
scratch-buffer dance around `dsp::compute()`, a sanitising pass. The device's
actual identity was about forty lines of that. Porting it is one job done
twenty-two times.

`MagdaCompiledEffect` is the effect half of what `MagdaCompiledPolyInstrument`
already was for the instruments. It owns the harvest, the slot table, the
engines, the block and project tempo; a device supplies its dsp factory, its
slots and its names. The twenty-one effects are now between forty and a hundred
and fifty lines each.

Poly Synth went onto the instrument base, which had assumed the layout a struck
drum voice wants. It composes that as a default now rather than imposing it: a
device may return its whole table and say where the control slots went.

## What the SDK had to grow

Each of these is a device asking for something the contract could not say, not a
convenience:

- **Project tempo.** `DeviceProcessContext::tempoMap` was never filled in on the
  Tracktion side, so the six tempo-synced effects would have lost sync. It reads
  the edit's sequence now, on the same thread those devices already read it from.
- **Sidechain.** The engine had `DeviceBlock::sidechain` and the SDK had nowhere
  to put it. The key arrives as further channels of the device's own buffer with
  `sidechainInputChannel` saying where they start. That is the layout the
  compiled dynamics DSPs already read and the one the fork already hands them,
  so neither host copies anything to supply it.
- **Channel widths, tail and latency**, for the mono-in/stereo-out and
  reverb-tail devices.
- **A dsp-index to slot mapping**, because the Filter's table interleaves
  controls no dsp owns.

## Two regressions this branch introduced and fixed

Worth reading as part of the review rather than buried:

- Delay and Grain Delay declare four-second tails and both told the host to keep
  pumping them after dry input stopped. The port carried the tail and dropped the
  pumping, so the echoes were cut. Restored.
- Converting a slot's value went through `ParameterInfo`, which carries a
  `std::vector<juce::String>` of choices, so every discrete parameter of every
  device allocated once per block inside `process()`. The devices this base
  replaced each avoided that by hand and the base reintroduced it for all of
  them. Both bases now cache a `ParameterUtils::ParameterDomain` per slot and the
  audio thread reads it by reference.

## What this does not do

The survey still reports `clean and renderable: 2 of 22`, and that is no longer a
device fact. The dominant remaining diagnostic is
`plan: track N: output routes to missing track -2, summed into the master`, which
is the plan compiler. #2192's "done when" assumed devices were the wall. They
were, and something else is now.

Still not native, and out of this slice: `4osc` (7 projects, and #1902's v1 plan
already removes it), `drumgrid` (3), `faust` (2), `sidechain` (2), and
Tracktion's own `toneGenerator` (1).

## Verification

- Catch2: 2943 cases, 601786 assertions, 0 failures.
- JUCE target: clean. One real failure caught here mid-branch and fixed: a
  declared output width was being read as "the device answers for both sides", so
  Reverb reported zero inputs and a nested rack stopped passing audio into the
  device after it.
- Survey: every compiled device gone from the blocked table.

The first two commits are the #2081 survey this is measured with. They have no PR
of their own and are small, so they ride along rather than paying separate CI.

### DIFF
--- a/magda/daw/audio/plugins/compiled/MagdaCompiledEffect.cpp
+++ b/magda/daw/audio/plugins/compiled/MagdaCompiledEffect.cpp
@@ -1,0 +1,491 @@
+#include "plugins/compiled/MagdaCompiledEffect.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <map>
+
+#include "core/ParameterUtils.hpp"
+#include "faust/dsp/dsp.h"
+#include "faust/gui/UI.h"
+#include "faust/gui/meta.h"
+#include "plugins/FaustMetadataParser.hpp"
+
+namespace magda::daw::audio::compiled {
+
+namespace {
+
+/// The one [idx:N] harvester the compiled effects used to each carry a copy of.
+///
+/// Records every numeric control by the slot index its label pins it to, plus
+/// its menu, its gate condition and its role tag: everything the base needs to
+/// bind a host slot to a zone without knowing which device it is looking at.
+class EffectZoneHarvester : public ::UI {
+  public:
+    struct Control {
+        int idx = -1;
+        FAUSTFLOAT* zone = nullptr;
+        std::vector<std::pair<float, juce::String>> menuChoices;
+        int gateSlotIndex = -1;
+        bool gateNegated = false;
+        bool isProjectTempo = false;
+    };
+
+    std::vector<Control> controls;
+
+    void openTabBox(const char*) override {}
+    void openHorizontalBox(const char*) override {}
+    void openVerticalBox(const char*) override {}
+    void closeBox() override {}
+
+    void addButton(const char* label, FAUSTFLOAT* zone) override {
+        emit(label, zone);
+    }
+    void addCheckButton(const char* label, FAUSTFLOAT* zone) override {
+        emit(label, zone);
+    }
+    void addVerticalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT, FAUSTFLOAT, FAUSTFLOAT,
+                           FAUSTFLOAT) override {
+        emit(label, zone);
+    }
+    void addHorizontalSlider(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT, FAUSTFLOAT,
+                             FAUSTFLOAT, FAUSTFLOAT) override {
+        emit(label, zone);
+    }
+    void addNumEntry(const char* label, FAUSTFLOAT* zone, FAUSTFLOAT, FAUSTFLOAT, FAUSTFLOAT,
+                     FAUSTFLOAT) override {
+        emit(label, zone);
+    }
+    void addHorizontalBargraph(const char*, FAUSTFLOAT*, FAUSTFLOAT, FAUSTFLOAT) override {}
+    void addVerticalBargraph(const char*, FAUSTFLOAT*, FAUSTFLOAT, FAUSTFLOAT) override {}
+    void addSoundfile(const char*, const char*, Soundfile**) override {}
+
+    void declare(FAUSTFLOAT* zone, const char* key, const char* value) override {
+        if (zone == nullptr)
+            return;
+        const auto k = juce::String::fromUTF8(key != nullptr ? key : "").toLowerCase();
+        const auto v = juce::String::fromUTF8(value != nullptr ? value : "");
+        applyFaustAnnotation(k, v, pendingByZone_[zone]);
+    }
+
+  private:
+    void emit(const char* rawLabel, FAUSTFLOAT* zone) {
+        if (zone == nullptr)
+            return;
+
+        const auto parsed =
+            parseFaustLabel(juce::String::fromUTF8(rawLabel != nullptr ? rawLabel : ""));
+        ControlMetadata merged = parsed.metadata;
+        if (auto it = pendingByZone_.find(zone); it != pendingByZone_.end()) {
+            mergeFaustMetadata(merged, it->second);
+            pendingByZone_.erase(it);
+        }
+
+        Control control;
+        control.idx = merged.slotIndex;
+        control.zone = zone;
+        control.menuChoices = merged.menuChoices;
+        control.gateSlotIndex = merged.gateSlotIndex;
+        control.gateNegated = merged.gateNegated;
+        control.isProjectTempo = merged.role == FaustControlRole::ProjectTempo;
+        controls.push_back(std::move(control));
+    }
+
+    std::map<FAUSTFLOAT*, ControlMetadata> pendingByZone_;
+};
+
+}  // namespace
+
+// ===========================================================================
+
+MagdaCompiledEffect::MagdaCompiledEffect() = default;
+MagdaCompiledEffect::~MagdaCompiledEffect() = default;
+
+::dsp* MagdaCompiledEffect::createEngineDsp(int) const {
+    return nullptr;
+}
+
+void MagdaCompiledEffect::initEffect() {
+    // The harvest first, because a device builds its slot table out of what the
+    // dsp declared: the tempo-synced effects take their Division choices
+    // straight from the menu rather than restating the list in C++.
+    //
+    // The real sample rate arrives with prepare(); this provisional one makes
+    // the device answerable from the moment it is constructed, which is what
+    // the app does with one before it ever reaches a graph.
+    constexpr int kProvisionalSampleRate = 44100;
+    createEngines(kProvisionalSampleRate);
+
+    hostSlotInfo_ = slotInfos();
+    applyHarvestedGates();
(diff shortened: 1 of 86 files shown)


magda/daw/audio/plugins/compiled/MagdaCompiledEffect.cpp
```diff
@@ -0,0 +1,518 @@
...
+        for (int i = 0; i < numSamples; ++i)
+            out[i] = sanitise(out[i]);
+    }
+
+    // A device that declares a fixed output width owns exactly that many
+    // channels, so anything above them is not its output -- it is the input it
+    // was handed, and passing that through would leak the dry signal around a
+    // widener that was asked to replace it.
+    if (outputChannelCount() > 0)
+        for (int channel = numOutputs; channel < hostChannels; ++channel)
```
lucaromagnoli: **P1 — Preserve the read-only sidechain block**

For the compressor, `hostChannels` includes the channels appended from `DeviceBlock::sidechain`, while `numOutputs` is 2. The native adapter obtains those pointers with `const_cast` under an explicit read-only contract, so this loop clears the sidechain source after compute. Because that source may fan out to other ops, this can corrupt their input or race a parallel reader. Please stop the clear range at `context.sidechainInputChannel` when it is non-negative; channels at and above that index are not device-owned outputs.

magda/daw/audio/plugins/compiled/MagdaPolySynthCompiledPlugin.hpp
```diff
@@ -82,139 +41,58 @@ class MagdaPolySynthCompiledPlugin : public te::Plugin, public ICompiledFaustPlu
...
     }
-    DeviceParameterHandle hostSlotParameter(int slotIndex) const override {
-        return tracktion_adapter::parameterHandle(getSlotParameter(slotIndex));
+    int numVoices() const override {
+        return kNumVoices;
     }
-    float displayToNormalized(int slotIndex, float displayValue) const override {
-        return displayValueToNativeValue(slotIndex, displayValue);
+    bool hasVoiceModes() const override {
+        return true;
```
lucaromagnoli, line 89: **P2 — Restore held-note preallocation**

This enables the shared mono/legato path for Poly Synth, but the port drops the old `heldNotes_.reserve(128)` from `initialise()`. The base `handleMonoNoteOn()` calls `heldNotes_.push_back()` from `process()`, and the base `prepare()` never reserves, so the first note and subsequent capacity growth allocate on the audio thread. Please reserve bounded capacity during prepare; the previous 128-note bound is sufficient.

magda/daw/audio/plugins/compiled/MagdaPolySynthCompiledPlugin.cpp
```diff
@@ -720,9 +211,8 @@ const CompiledPluginSpec& getMagdaPolySynthSpec() {
         .description = "Compiled Faust polyphonic synth: four detunable oscillators "
                        "(sine/saw/square/triangle) into a multimode filter with its own "
                        "envelope, plus an ADSR amp envelope. 16-voice, MIDI-driven.",
-        .createPlugin = [](const DevicePluginCreationContext& context) -> DevicePluginPtr {
-            return tracktion_adapter::pluginHandle(
-                new MagdaPolySynthCompiledPlugin(tracktion_adapter::creationInfo(context)));
+        .createDevice = [](const DevicePluginCreationContext&) -> std::unique_ptr<MagdaDevice> {
```
lucaromagnoli, line 214: **P2 — Update the remaining live-plugin cast**

Switching to `createDevice` means `getLivePlugin()` now returns a `TracktionMagdaDevicePlugin`. `DeviceCustomUIManager::refreshLivePluginBindings()` still casts `plugin.get()` directly to `MagdaPolySynthCompiledPlugin` at line 1883, so it always passes null into `PolySynthUI` and the 60 Hz live/effective filter curve never starts. Please unwrap with `deviceFromPlugin<MagdaPolySynthCompiledPlugin>()`, as the struck-instrument binding immediately below already does.

PR #2196, opened 2026-08-25
Sidechain, Faust, the test tone and a Drum Grid's pads under the native engine (#2192)

lucaromagnoli:
Follows #2195. Three more devices off the fork, the fixture bug that was hiding
behind them, and the model groundwork the fourth device needs.

Folds in what was #2197, which was stacked on this branch and is now part of it.

## The fixture, first

#2195 said the survey's remaining blocker was the plan compiler. That was wrong,
and this corrects it.

Fourteen of the twenty-two corpus projects predate the `masterTrack` section, and
the rig read their silence as "no master": `Case::master` stayed default
constructed, so its id was `INVALID_TRACK_ID`, and every track routing to
`"master"` resolved to `MASTER_TRACK_ID` and found no track with that id. That is
what `output routes to missing track -2` was. The compiler was right and it was
handed a master that was not one.

The app has never seen this. `TrackManager` builds a master with
`MASTER_TRACK_ID` before any project loads; an absent section means the master
saved no plugins. The rig does the same now.

Clean and renderable goes from two to five on that alone.

## The devices

- **Sidechain.** A plain four-parameter gain stage, so it implements the SDK
  hooks directly rather than borrowing `MagdaCompiledEffect`. Same duck DSP:
  stair reconstruction, attack and release poles, sides mode. Slot ids pinned to
  what the fork used.
- **Faust.** The interesting one. Sixty-four lifetime-stable pool slots stop
  being `te::AutomatableParameter`s and become normalized values the device owns,
  saved under the same `param_01..param_64` ids so a macro, mod or automation
  lane still finds its slot across a recompile. Project tempo comes from the
  context's tempo map. The bend-range constant left `FaustInstrumentPlugin` for
  `FaustParamInfo`, because the neutral pack was reaching into a host-engine
  header for one float.
- **Test tone.** A rewrite, not a port: MAGDA had a faceplate driving Tracktion's
  oscillator and no DSP of its own. The id, the four parameters and the six
  waveforms are unchanged, and regenerating the parameter-schema freeze produced
  no diff, which is the check that says so. Naive waveforms with PolyBLEP at the
  discontinuities when Band Limit is on.

Clean and renderable: **six of twenty-two**. What is left is `4osc`, which #1902
removes, and `drumgrid`, whose first step is below.

## Four host bugs the ports exposed

Each of these was already wrong for the devices #2195 landed:

1. `reset()` ran before parameters were ever synced, so a device seeding state
   from a parameter read its construction defaults.
2. Capture recorded a `parameters` blob the engine deletes on the next flush, so
   saving a project twice changed the file. Two corpus projects hit it.
3. The host baked each parameter's `ParameterInfo` at construction. Fine for a
   device whose parameters are fixed; the Faust pool rebinds on every compile, so
   the host formatted, parsed and scaled every slot against the passthrough patch
   it was built with.
4. Same for the name, which lives in a `const` member. A recompiled slot
   converted as `Cutoff` and was still called `(slot 1)`.

3 and 4 have a test written before the fix, which fails on the parent commit with
exactly `50.0%` against `4.8 kHz` and `(slot 1)` against `Cutoff`.

## Two more from review

Both are the same mistake in different places: a device was declared native
before everything that reaches it had followed.

- **The engine's factory built a device from its id and nothing else**, because
  for every device it had carried so far the parameters *are* the state and the
  plan's value layer writes those per block. Faust breaks that: its dsp source is
  state, so a native-engine Faust device ran the default passthrough and rendered
  a project nobody saved, which is the one thing that factory's own comment says
  never to do quietly. It restores the saved document before the adapter rather
  than after, because `EngineMagdaDevice` snapshots the device's parameter
  metadata when it is constructed and a later restore would leave that mapping
  pointing at the parameters the device used to have. Tested, confirmed failing
  without the fix.
- **`ToneGeneratorProcessor` resolved its device with a cast to the retired
  Tracktion plugin**, which is always null against the wrapper, so its defaults,
  its `DeviceInfo` parameter restore and every getter and setter had quietly
  become no-ops. It reaches the parameters by slot now, and the accessors keep
  their old units (Hz, linear level, the waveform enum) because their callers are
  written against those.

## Somewhere to put a Drum Grid's pads

Groundwork only, and the reason it is worth reviewing separately from the rest:
nothing writes `padRack` and nothing reads it, so there is no behaviour change
here. What follows it is a project format change; this part is not.

DrumGrid could not be ported like the others because it is not a DSP device.
Each pad hosts a full plugin chain (instrument plus FX) and `processChain()`
builds a `te::PluginRenderContext` and runs them; a pad can hold an external VST.
A `MagdaDevice` cannot host one, and the SDK has no device-hosts-devices
contract.

The plan already compiles racks structurally (`emitRack`, chains, `OpKey` carries
rack and chain), and a Drum Grid is that same shape: 64 pads, each a chain. So
the native answer is for the plan to expand it, not for the class to be ported.
Except the plan compiles from the model, and the pads are not in it. They live
inside the device's opaque `pluginState` as engine XML:

```xml
<PLUGIN type="drumgrid">
  <CHAIN index="0" lowNote="24" highNote="24" rootNote="24"
         padLevel="0.0" padPan="0.0" padMute="0" padSolo="0"
         padBypassed="0" busOutput="0">
    <PLUGIN type="magdasampler" source="..."/>
```

`PluginManagerSync` says it outright: "Pad plugins live inside the DrumGrid's
state, not in the chain."

So pads are stored as chains, because that is what they are. `ChainInfo` already
carries a pad's level, pan, mute, solo, bypass and output; the note range added
here is the only thing a pad has that a rack chain does not. A chain that answers
to every note is what a plain parallel chain already is, so nothing built before
this reads differently.

`DeviceInfo` holds them through `PadRack` rather than a bare pointer. `RackInfo`
is incomplete there, `DeviceInfo` is copied by value throughout the app, and two
copies of a device must not share one set of pads. `PadRack` owns the rack,
copies it deep, and keeps its rule of five out of line so `DeviceInfo` keeps the
compiler's own and nothing else has to change. No aggregate initialization of
`DeviceInfo` exists anywhere, which is what makes that safe.

Next, after this PR:

1. Read the pads out of the saved engine XML into `padRack`, and serialize them
2. Teach `PlanCompiler` to expand a device with pads like `emitRack`, gated by
   note range
3. Keep the fork rendering from that same model

## The pattern worth reviewing for

Most of the bugs above are the same shape: consolidating per-device plumbing into
a shared base made it assume a property only *most* devices have. Fixed tails,
fixed channel layout, fixed parameter metadata, parameters-are-the-state. Faust
breaks that assumption every time because it recompiles. Anything added to
`MagdaDevice` that a device could change at runtime is worth checking against it
specifically.

## Verification

Catch2 2945 cases / 601888 assertions, JUCE suite green. Full CI passed on this
exact head while it was #2197's; the fold was a fast-forward, so the tree under
review is byte-identical to what those runs covered.


magda/daw/audio/plugins/BaseDevicePack.cpp
```diff
@@ -412,13 +427,13 @@ void registerNativeDevices(InternalPluginRegistry& registry) {
...
                    .loadAliasCount = static_cast<int>(std::size(kFaustAliases)),
-                   .matchesPlugin = matches<FaustPlugin>,
+                   .matchesPlugin = matchesDevice<FaustPlugin>,
                    .createProcessor = makeProcessor<FaustProcessor>,
                    .showInBrowser = true,
                    .tags = kFaustTags,
                    .tagCount = static_cast<int>(std::size(kFaustTags)),
                    .createInSession = createValueTreePlugin,
-                   .createPlugin = createPlugin<FaustPlugin>});
+                   .createDevice = createDevice<FaustPlugin>});
```
lucaromagnoli, line 436: **P1 — Restore Faust state before declaring native-engine support**  This spec now makes `canCreateEngineDevice("faust-fx")` return true, but `createEngineDevice()` uses only `device.pluginId`: its creation context is empty and it never calls `restoreState()` with `device.pluginState`. Consequently every native-engine Faust instance runs the default passthrough rather than the serialized `dspSource`; because `EngineMagdaDevice` also snapshots parameter metadata in its constructor, restoring later would leave the mapping wrong. Please restore the device state before constructing the adapter, or continue reporting Faust as unsupported until the state slice exists.

magda/daw/audio/plugins/BaseDevicePack.cpp
```diff
@@ -238,20 +252,21 @@ void registerTracktionDevices(InternalPluginRegistry& registry) {
...
+         .matchesPlugin = matchesDevice<ToneGeneratorPlugin>,
          .createProcessor = makeProcessor<ToneGeneratorProcessor>,
          .showInBrowser = true,
-         .tags = kTracktionTags,
-         .tagCount = static_cast<int>(std::size(kTracktionTags)),
-         .createInSession = createTracktionPlugin});
+         .tags = kToneTags,
+         .tagCount = static_cast<int>(std::size(kToneTags)),
+         .createInSession = createValueTreePlugin,
+         .createDevice = createDevice<ToneGeneratorPlugin>});
```
lucaromagnoli, line 269: **P1 — Bridge the processor to the wrapped Test Tone**  This now creates a `TracktionMagdaDevicePlugin` containing the new `ToneGeneratorPlugin`, while the registered `ToneGeneratorProcessor` still resolves its device exclusively with `dynamic_cast<te::ToneGeneratorPlugin*>`. That cast is always null for this wrapper, so initialization, DeviceInfo parameter restore, and every UI getter/setter become defaults or no-ops. Please make the processor adapter-aware or replace it with the generic automatable processor used by native devices.

PR #2202, opened 2026-08-26
Compile a Drum Grid's pads instead of stopping at it (#2200)

### LINKED ISSUE #2200: The plan has no way to gate a Drum Grid's pads by note range (#2192)
#2199 put a Drum Grid's pads in the model. The plan still cannot render them,
because expanding a pad needs an op the plan does not have.

## What a pad is, as a plan op

A pad answers to a contiguous range of notes and transposes what it receives
onto its own root:

```
note in [lowNote, highNote]  ->  rootNote + (note - lowNote)
```

That is what makes a sampler mapped at C0 play from whichever pad triggered it,
and it is the one thing a pad has that a plain parallel rack chain does not.
`ChainInfo` already carries the three values (`lowNote`, `highNote`, `rootNote`,
with `lowNote > highNote` meaning "every note"), so the model side is done.

## Why it cannot be compiled today

`OpKind` has no note gate and no transpose. The whole list:

```
ClipAudio, ClipMidi, AudioInput, MidiInput, Device, MixAudio, MergeMidi,
Subtract, Delay, Crossfade, Gain, Fader, SendTap, Meter, ModSource, Output
```

`emitRack` is otherwise the right shape to copy: a Drum Grid is 64 pads, each a
chain, and `OpKey` already carries rack and chain so the ops key apart. What is
missing is the gate between the chain's MIDI input and its devices.

## What this touches

A new op kind is not a one-line addition:

- **Plan vocabulary.** `OpKind`, `OpRole`, and whatever the op needs bound at
  prepare time (three ints per chain).
- **Native executor.** The op's actual implementation.
- **The TE leg.** Parity is a hard requirement, so the Tracktion leg has to
  produce the same result for null-diff. Today `DrumGridPlugin::applyToBuffer`
  does the gating internally, so the two legs reach it by different routes and
  have to agree.
- **The differ.** A new op has to survive an edit round-trip like the rest.
- **`PlanCompiler`.** `emitRack`-shaped expansion for a device with a `padRack`,
  gated per chain.

## Worth deciding first

- **Whether the op gates, transposes, or both.** One op with three parameters is
  fewer ops per pad; two ops compose better if anything else ever needs a
  transpose. Sixty-four pads makes the count real either way.
- **A pad hosting an external plugin is NOT a new decision.** `1.0.0-drumgrid-rack.mgd`
  has one (Kick 2, `format="VST3"`), but `EngineDeviceFactory` already settles this
  case for every chain: `isRegisteredDevice()` is false for an external plugin, and
  "for those the engine running nothing is what every host does". The case worth
  reporting is the inverse -- `isRegisteredDevice()` true with
  `canCreateEngineDevice()` false, an internal device the app can build and the
  engine cannot. A pad follows the same rule a rack chain already follows, so
  nothing here needs deciding.

- **Pads with no saved DeviceId.** A Drum Grid allocates a pad plugin's id when
  it restores that plugin, so projects saved before instantiation carry none.
  #2199 refreshes the projection on capture for exactly this reason, but the
  compiler should report a pad it cannot key rather than emit a colliding op.

## Done when

The survey renders the three corpus projects that carry a Drum Grid, or reports
per pad why it cannot.

lucaromagnoli:
Closes #2200. The last of three steps: #2199 put a Drum Grid's pads in the model,
#2201 gave the plan a note gate, and this expands one into the other. A Drum Grid
renders under an engine that cannot host one.

## What it does

The device never becomes an op. Its pads are chains and each is an instrument
chain: it reads no audio, takes the notes its range claims through a
`MidiNoteGate`, and what it makes is summed with its siblings and injected into
the bus the way an instrument's output is.

A pad that answers to every note gets no gate at all, because a chain that takes
everything is what a plain parallel rack chain already is.

## A pad rack is a rack

It carries a `RackId` of its own, negative and derived from the device that owns
it. Rack ids the app allocates start at 1, so the negative space is free and
nothing has to reach an allocator that does not exist at projection time.

That is what keeps this from being a special case: the value layer's rack lookup
descends into a device's pads as well as a chain's racks, so a pad's mute, solo
and fader resolve through the same code every other chain uses.

## Two things only real projects found

**The rack mix and the inject collided.** They sat at one location and differed
only in role. The latency delays that land on a mix's inputs keep the id but not
the role, so one mix's delays keyed onto the other's and the plan did not
validate. The mix carries no device id now, which is what a rack's own mix
already does.

**Pads with no DeviceId.** A pad plugin's id arrives when the Drum Grid restores
that plugin, so a project not opened since carries none, and both corpus projects
are that old. Two such devices in one pad key the same op. The pad is now
reported and left out rather than compiled: a plan that does not validate costs
the whole project, where a diagnostic costs the pad.

Worth a decision, and the reason I am flagging it rather than burying it: those
pads only compile once a project has been opened and re-saved. If that is not
acceptable, the fix is resolving ids at load rather than at capture, which is its
own change.

## Survey

Clean and renderable: **six of twenty-two to seven**. `drumgrid` leaves the
blocked-device table entirely.

- `0.4.8-drumgrid.mgd`: clean.
- `1.0.0-drumgrid-rack.mgd`: from a refused plan to a render with a diagnostic
  naming the pad it could not key.

What blocks the rest is `4osc` on seven projects, which #1902 removes.

## Coverage

Eight cases: a well formed plan, a gate per pad with the right range and
transposition, the Drum Grid never appearing as a device op, a full-range pad
needing no gate, a bypassed device and a bypassed pad, a Drum Grid nested inside
a user rack (`1.0.0-drumgrid-rack.mgd`'s shape), and the unkeyable pad above.

## Not here

Parity. The Tracktion leg still gates internally in `DrumGridPlugin::applyToBuffer`,
so the two legs now reach the same result by different routes and null-diff over a
Drum Grid is the next thing to establish.

## Verification

Catch2 2980 cases / 602671 assertions green, re-run against the committed state
after clang-format.

### DIFF
--- a/magda/engine/exec/PlanValues.cpp
+++ b/magda/engine/exec/PlanValues.cpp
@@ -5,6 +5,7 @@
 #include <set>
 #include <unordered_map>
 
+#include "core/DrumGridPads.hpp"
 #include "core/RackInfo.hpp"
 #include "core/TrackInfo.hpp"
 #include "param/ParamTableCompiler.hpp"
@@ -108,12 +109,21 @@ const DeviceInfo* findDeviceIn(const std::vector<PostFxChainElement>& elements, 
 
 /// Whether a rack chain is in the mix. Solo is relative to its siblings, which
 /// is why chain activity cannot be answered from the chain alone.
+///
+/// A pad rack counts solo only on pads that have something to run. That is the
+/// Drum Grid's own rule, and it matters because removing a pad's last plugin
+/// leaves the chain behind: an empty soloed pad would otherwise silence every
+/// populated pad beside it, where the device plays them.
 bool isChainActive(const RackInfo& rack, const ChainInfo& chain) {
     if (chain.muted)
         return false;
-    const auto anySolo =
-        std::ranges::any_of(rack.chains, [](const ChainInfo& c) { return c.solo; });
-    return !anySolo || chain.solo;
+
+    const auto counts = [pads = isPadRackId(rack.id)](const ChainInfo& c) {
+        return c.solo && (!pads || !c.elements.empty());
+    };
+
+    const auto anySolo = std::ranges::any_of(rack.chains, counts);
+    return !anySolo || counts(chain);
 }
 
 class Resolver {
--- a/magda/engine/exec/PlanValues.cpp
+++ b/magda/engine/exec/PlanValues.cpp
@@ -71,10 +71,18 @@ const RackInfo* findRackIn(const RackInfo& rack, RackId rackId) {
 }
 
 const RackInfo* findRackIn(const std::vector<ChainElement>& elements, RackId rackId) {
-    for (const auto& element : elements)
-        if (isRack(element))
+    for (const auto& element : elements) {
+        if (isRack(element)) {
             if (const auto* found = findRackIn(getRack(element), rackId))
                 return found;
+            continue;
+        }
+        // A pad rack hangs off its device rather than sitting in the chain, and
+        // its chains carry mute, solo and a fader like any other's.
+        if (const auto& device = getDevice(element); device.padRack)
+            if (const auto* found = findRackIn(*device.padRack.get(), rackId))
+                return found;
+    }
     return nullptr;
 }
 
--- a/magda/engine/exec/PlanValues.cpp
+++ b/magda/engine/exec/PlanValues.cpp
@@ -178,8 +178,22 @@ class Resolver {
     /// outer chain disconnected.
     void collectSilencedRacks(const std::vector<ChainElement>& elements, bool insideInactiveChain) {
         for (const auto& element : elements) {
-            if (!isRack(element))
+            if (!isRack(element)) {
+                // A pad rack hangs off its device rather than sitting in the
+                // chain, and an inactive chain silences its pads the same way it
+                // silences a nested rack's chains. Without this a Drum Grid in a
+                // muted chain keeps its pad devices running behind a silent
+                // fader, advancing tails and publishing meters.
+                if (const auto& device = getDevice(element); device.padRack) {
+                    const auto& pads = *device.padRack.get();
+                    if (insideInactiveChain)
+                        silencedRacks_.insert(pads.id);
+                    for (const auto& pad : pads.chains)
+                        collectSilencedRacks(pad.elements,
+                                             insideInactiveChain || !isChainActive(pads, pad));
+                }
                 continue;
+            }
 
             const auto& rack = getRack(element);
             if (insideInactiveChain)
--- a/magda/engine/param/ParamTableCompiler.cpp
+++ b/magda/engine/param/ParamTableCompiler.cpp
@@ -540,6 +540,12 @@ void Builder::walkElements(const std::vector<magda::ChainElement>& elements, mag
             node.device = &device;
             node.sidechainSource = sidechainSourceOf(device.sidechain);
             nodes_.push_back(node);
+
+            // A pad rack's chains hold devices like any other rack's, and the
+            // plan compiles them, so their parameters, macros and mods need
+            // addresses or nothing consumes them.
+            if (device.padRack)
+                walkRack(*device.padRack.get(), trackId);
         } else if (magda::isRack(element)) {
             walkRack(magda::getRack(element), trackId);
         }
--- a/magda/engine/param/ParamTableCompiler.cpp
+++ b/magda/engine/param/ParamTableCompiler.cpp
@@ -544,8 +544,17 @@ void Builder::walkElements(const std::vector<magda::ChainElement>& elements, mag
             // A pad rack's chains hold devices like any other rack's, and the
             // plan compiles them, so their parameters, macros and mods need
             // addresses or nothing consumes them.
+            //
+            // Addressed under the owning device's id, not the pad rack's. A
+            // stored link to a pad device names the Drum Grid there -- see the
+            // chainDevice path syncDrumGridPadPlugins and the ADSR macro links
+            // are built with -- and ParamKey carries the rack id, so addressing
+            // these under the pad rack's own synthetic id would put them where
+            // no link can find them. The rack itself gets no node: it is
+            // synthesized, so it owns no macros or modifiers to address.
             if (device.padRack)
-                walkRack(*device.padRack.get(), trackId);
+                for (const auto& pad : device.padRack->chains)
+                    walkElements(pad.elements, trackId, device.id);
         } else if (magda::isRack(element)) {
             walkRack(magda::getRack(element), trackId);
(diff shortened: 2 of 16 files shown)


magda/engine/plan/PlanCompiler.cpp
```diff
@@ -952,6 +953,91 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
...
+        // No audio in: a pad generates rather than processes, so the bus flows
+        // past it the way it flows past an instrument.
+        const auto padOut = emitElements(pad.elements, padSite, ChainSignal{noInput(), padMidi});
+
+        const OpKey faderKey{site.trackId,           pads.id, pad.id,      INVALID_DEVICE_ID,
+                             OpRole::RackChainFader, 0,       site.segment};
+        const auto faderOp = addOp(OpKind::Fader, faderKey,
+                                   alignInputs(faderKey, OpKind::Fader, {padOut.audio, noInput()}),
+                                   {SignalKind::Audio});
+        padAudio.push_back(PortRef{faderOp, 0});
```
lucaromagnoli: **P1 — Route pads by `outputIndex` instead of always to the main mix.** `ChainInfo::outputIndex` is the projected Drum Grid `busOutput`, but every pad reaches this single `padAudio` vector and `multiOutTargets_` is never used. A pad assigned to Bus N will play on the parent track’s main output while the linked MultiOut track stays silent. Please group pad outputs by bus/pair and enqueue the non-main mixes for the corresponding multi-out targets (dropping unconnected pairs as the incumbent engine does).

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -952,6 +953,91 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
...
+            continue;
+        }
+
+        // No audio in: a pad generates rather than processes, so the bus flows
+        // past it the way it flows past an instrument.
+        const auto padOut = emitElements(pad.elements, padSite, ChainSignal{noInput(), padMidi});
+
+        const OpKey faderKey{site.trackId,           pads.id, pad.id,      INVALID_DEVICE_ID,
+                             OpRole::RackChainFader, 0,       site.segment};
+        const auto faderOp = addOp(OpKind::Fader, faderKey,
```
lucaromagnoli, line 1027: **P1 — Keep pad faders connected to the parameter table.** This `RackChainFader` resolves only from the projected `ChainInfo::volume/pan`. The Drum Grid actually renders pad gain/pan from its automatable `padLevelN`/`padPanN` parameters, and `ParamTableCompiler::walkElements()` does not descend into `device.padRack`, so automation, macros, and modulation for these controls and the emitted child devices have no consumer. Please add pad-rack traversal and bind this fader to the owning Drum Grid parameter slots.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -952,6 +953,91 @@ ChainSignal Compiler::emitRack(const RackInfo& rack, const ChainSite& site, Chai
...
+    if (padAudio.empty())
+        return signal;
+
+    // No device id, the way a rack's own mix carries none. The inject below is
+    // the same location and differs only in role, and the latency delays that
+    // land on a mix's inputs keep the id but not the role, so sharing it would
+    // key one mix's delays onto the other's.
+    const OpKey mixKey{site.trackId,    pads.id, INVALID_CHAIN_ID, INVALID_DEVICE_ID,
+                       OpRole::RackMix, 0,       site.segment};
+    const auto mixed = emitMix(mixKey, padAudio);
```
lucaromagnoli: **P1 — Preserve the Drum Grid slot gain before injection.** The expanded pads go straight from this mix into the bus, bypassing the `DeviceGain` and device-meter stages emitted for every ordinary instrument. Any non-unity gain set in the Drum Grid slot controls is therefore ignored. Please apply the owning device gain and metering to the wet pad mix before `DeviceInject`.

magda/engine/exec/PlanValues.cpp
```diff
@@ -71,10 +71,18 @@ const RackInfo* findRackIn(const RackInfo& rack, RackId rackId) {
...
-        if (isRack(element))
+    for (const auto& element : elements) {
+        if (isRack(element)) {
             if (const auto* found = findRackIn(getRack(element), rackId))
                 return found;
+            continue;
+        }
+        // A pad rack hangs off its device rather than sitting in the chain, and
+        // its chains carry mute, solo and a fader like any other's.
+        if (const auto& device = getDevice(element); device.padRack)
```
lucaromagnoli, line 83: **P2 — Mirror this traversal in `collectSilencedRacks()`.** `findRackIn` can now find a device-owned pad rack, but `collectSilencedRacks()` still skips every non-rack element. A Drum Grid inside a muted or non-soloed user-rack chain therefore leaves its pad devices running, advancing tails and publishing meters behind the silent outer fader. Please descend into `device.padRack` there and mark its synthetic rack ID when the containing chain is inactive.

magda/engine/plan/PlanCompiler.cpp
lucaromagnoli: **Still unresolved — the table walk does not bind this fader.** The new `walkRack()` call makes the child devices addressable, but this op remains a plain `RackChainFader`, and `PlanExecutor::prepare()` deliberately binds only `TrackFader` and `SendTap` mixer ops. The owning Drum Grid’s automatable `padLevelN`/`padPanN` slots therefore still have no consumer, so automation, macros, and modulation do not move the pad fader. Please give this op a binding to the owning device parameter pair and add a render-level test that changes those resolved parameters between blocks.

magda/engine/param/ParamTableCompiler.cpp
```diff
@@ -540,6 +540,12 @@ void Builder::walkElements(const std::vector<magda::ChainElement>& elements, mag
             node.device = &device;
             node.sidechainSource = sidechainSourceOf(device.sidechain);
             nodes_.push_back(node);
+
+            // A pad rack's chains hold devices like any other rack's, and the
+            // plan compiles them, so their parameters, macros and mods need
+            // addresses or nothing consumes them.
+            if (device.padRack)
```
lucaromagnoli, line 555: **P1 — Mirror this traversal in `collectRuntimeStateIds()`.** That model walk still records the Drum Grid itself and ordinary nested racks, but never descends into `device.padRack`. When bypass or chain power removes the pad `Device` ops from the live plan, neither the plan nor the model ID set names the child devices, so `RuntimeStateStore::releaseDeleted()` destroys their runtimes. Re-enabling then recreates the plugins and loses their tails and runtime state. Please collect the pad-rack devices there too and extend the existing bypass/chain-power retention test with a Drum Grid child.

Issue #2205, opened 2026-08-26
What a projected pad device carries is undocumented and mostly nothing


lucaromagnoli:
`DeviceInfo::padRack` (#2199, #2200) holds a `DeviceInfo` per pad plugin, built
field by field in `deviceFromNode()`. Nothing states which fields it fills, and it
fills a minority of them.

## What it carries

`pluginId`, `name`, `manufacturer`, `format`, `fileOrIdentifier`, `isInstrument`,
`deviceType`, `id`, `bypassed`, `pluginState`, and -- only after capture --
`parameters`.

## What it does not

`macros`, `mods`, `sidechain`, `multiOut`, `kitRows`, `deltaSolo`,
`browserCategoryOverride`, `audioInputChannels` / `audioOutputChannels`,
`uniqueId`, and the VST3 preset fields.

## Why this is not a bug yet

A pad plugin has no `DeviceInfo` in the chain model outside this projection.
`PluginManagerSync` says it: "Pad plugins live inside the DrumGrid's state, not in
the chain model." So a pad device cannot own a macro, a modifier or a sidechain
today, and a projection that drops them drops nothing.

The Drum Grid's own macros still work, because they live on the Drum Grid's
`DeviceInfo` and merely *target* pad device parameters -- which is what the ADSR
links in `ClipCommands` do.

## Why it is a trap

The moment a pad device can own one of those, three things have to change
together and nothing connects them:

1. `deviceFromNode()` has to carry the field.
2. Every walk that collects it has to descend into `padRack` -- `ModSources`,
   `SidechainTraversal`, the macro and modifier syncs, and whichever others (#2204).
3. Whatever writes it has to reach a pad device at all.

Miss any one and the feature silently does nothing, which is the exact failure
mode of every finding on #2202: not a crash, not a diagnostic, just a device
rendering without the thing you attached to it.

`audioOutputChannels` is the nearest one to mattering. `emitDevice()` clamps a
device's output width with it, and the projection leaves it at its default, so a
mono pad voice may be compiled as though it were stereo.

## What to do

Pick one, rather than leaving the set implicit:

- **State the contract.** Say in `DrumGridPads.hpp` which fields a projected pad
  device carries and why the rest are absent, and add a test that fails when
  `DeviceInfo` grows a field the projection has not been taught about. Cheap, and
  turns a silent gap into a build failure.
- **Or stop projecting.** If pads become first-class model chains (#2204's third
[...]

PR #2208, opened 2026-08-26
Give a Drum Grid's pads to the model (#2207)

### LINKED ISSUE #2207: The Drum Grid's pads are owned by the plugin, not the model
`DeviceInfo::padRack` (#2199, #2200, #2205) is not model state. It is decoded from
`device.pluginState` (`DrumGridPads.cpp:251`), and that string is only rewritten by
`capturePluginState()`. Pad edits never call it: `loadInternalPluginToPad`
(`DrumGridPlugin.cpp:648`) creates the plugin in the live grid and returns, and the
UI (`DeviceCustomUIManager.cpp:1245`) repaints the pad. The model is not told.

So the plan compiles a Drum Grid from the last snapshot of the plugin, not from
what is on screen.

## What that does

- **A new pad plugin is silent.** Drop a kick on a pad and press play. The live
  grid has it, the snapshot does not, `padRack` has no device there, the plan
  emits nothing for that pad.
- **A deleted pad plugin keeps playing.** The snapshot still names it. The plan
  compiles a `Device` op for it, `populatePadDeviceParameters()` finds no live
  plugin and skips it, and the op runs at defaults. A voice nobody can see or edit.
- **Old projects lose their pads.** State saved before pad ids existed projects
  `INVALID_DEVICE_ID`. The grid allocates ids on restore and writes them into the
  live plugin, where the model never reads them, so `emitPadRack()` refuses the pad.
- **Level, pan, mute, solo, note range** all follow the snapshot too.

Each of these heals on the next unrelated capture (save, duplicate track, remove
some other device), so it presents as "sometimes the pad does not play" and looks
like a race in the executor. Under Tracktion none of it shows because the plugin
plays itself. It lands the day the native path is live.

## Why

This is the Tracktion-era relationship, inverted from what the native engine is
built on. The model is supposed to own state; the plan is compiled from it; the
engine runs the plan. One direction, nothing to drift. Racks already work that
way: `RackInfo` and `ChainInfo` live in the model, edits go through
`TrackManager`, and `RackSyncManager` mirrors the model into TE. The plan reads
the same `ChainInfo`.

The Drum Grid predates that. It is a `te::Plugin` that owns its own 64 chains,
their sub-plugins, and per-pad level/pan/mute/note range in its own ValueTree
(`DrumGridPlugin.hpp:77`). The model never had pads. #2199 bridged that by
projecting the plugin's state into the model instead of moving the state there,
which makes the plugin the truth and the model a lagging mirror of it. #2205
already named the projection a trap on the field axis; this is the same trap on
the time axis.

## Fix

Flip ownership. Pads become model state and the plugin becomes a consumer of it,
the way a rack already is:

- The Drum Grid's `DeviceInfo` carries a real rack of 64 chains (`RackInfo` with
  a `ChainInfo` per pad, each holding its devices and `lowNote`/`highNote`/
  `rootNote`, level, pan, mute, solo, bypass, bus output). `padRack` and
  `readPadRack()` go away.
- Every pad edit goes through `TrackManager` as a chain edit: add device to
  chain, remove device from chain, set chain property. `DeviceCustomUIManager`
  stops calling `DrumGridPlugin` directly. `loadInternalPluginToPad`,
  `addInternalPluginToChain`, `loadSampleToPad`, `swapPadChains`, removal, and
  the pad property setters become model commands.
- `DrumGridPlugin` syncs from the model, the way `RackSyncManager::syncRack`
  fills a TE rack from `RackInfo`. Its internal chains are populated, not edited
  in place. Undo goes through the model like everything else.
- The plan compiler reads the Drum Grid's chains through `emitRack`-shaped
  expansion with the note gate from #2200. `populatePadDeviceParameters()` and
  the capture-time `refreshPadRack()` are no longer needed.
- Project load puts pads in the model directly and allocates their `DeviceId`s
  there, so the legacy-state case has nothing to heal.

Capturing on every pad edit is not the fix. It keeps the plugin as the source of
truth and every future pad feature has to remember the same call; that is how the
three failures above happened.

Blocks further work on `padRack`: each PR that builds on the projection has to be
undone by this one.

lucaromagnoli:
Closes #2207.

`DeviceInfo::padRack` was a projection: decoded from the device's `pluginState` and rebuilt on every capture, which made the plugin the truth and the model a lagging mirror of it. A pad added since the last capture was missing from the plan, a pad removed since was still in it, and neither showed under Tracktion because the plugin plays itself.

The pads are now `DeviceInfo::pads`, a rack of chains the device owns. The project file saves them, every pad edit writes them, the plan compiler expands them, and `DrumGridPlugin::syncFromModel()` fills the plugin from them the way `RackSyncManager` fills a `te::RackType` from a `RackInfo`. Nothing reads them back.

## Model

- `DeviceInfo::pads` is serialized by `TrackSerializer`, along with a chain's note range. `readLegacyPads()` is all that is left of the reader and runs once, at load, on a project saved before the pads moved.
- A pad rack answers to a `Rack` step in a `ChainNodePath` (its id is `padRackIdFor(deviceId)`, out of the negative space no allocated rack uses), so adding, removing, reordering, muting, soloing and fading a pad go through the chain commands `TrackManager` already has. `TrackManagerPads.cpp` holds only what is particular to a pad: found by note, made on demand, and two of them can trade places.
- `prepareNewDevice` re-keys a copied device's pads, `refreshIdCountersFromTracks` counts them, and `allocatePadDeviceIds()` hands an id to every pad device saved before pad ids existed, so the legacy-state case has nothing left to heal.

## Engine

- `DrumGridPlugin` has no edit API left. Its chains are built by `syncFromModel()`, which keeps a plugin the model still names so an unchanged pad is not rebuilt and does not glitch, and short-circuits on a structure fingerprint so a sync pass that changes nothing does nothing.
- Its chain `ValueTree`s are standalone, so the pads are no longer captured into the device's saved state or handed to Tracktion's graph builder as nested plugins.
- Pad plugins are registered at real model paths, so capture, macro and mod linking and the parameter refresh reach them the way they reach a rack device. `populatePadDeviceParameters()` and `DrumGridPadParameters.*` are gone with the projection.

## Two defects the flip exposed, both fixed here

Left alone, each would have been a regression:

- A compiled pad voice (every drum voice, every Faust effect) read out of legacy state fell into the external branch, so it came back as a VST3 with no instrument flag. The legacy reader now consults the compiled registry.
- `createPluginOnly` never seated an internal device's own saved *properties*, only its parameters, so a sampler inside a rack came back with no sample where one on a track kept it. Fixed at the shared creation point, which covers pads and racks.

## One behavioural change worth knowing

`DrumGridPlugin::valueTreePropertyChanged` is gone. It existed only to mirror pad CachedValues into the automation parameters, and it swallowed every property change without calling the base, so `te::Plugin`'s own handling was suppressed for Drum Grids. Removing it restores what every other MAGDA device gets.

## Tests

- `test_drum_grid_pads.cpp` rewritten: the model helpers every pad edit goes through, plus the load-time migration. The projection's field contract (#2205) goes with the projection.
- `test_drum_grid_serialization_juce.cpp` rewritten against the new ownership, including a pad the model drops leaving the grid on the next sync, and a Drum Grid's captured state carrying no second copy of its pads.
- `tests/PadSyncTestSupport.hpp` gives the three other JUCE suites that build a pad the two halves the real sync supplies.
- Corpus tests unchanged in substance: both released projects still reach the model, now through the migration.

Catch2: 2998 cases, 602738 assertions, 0 failures. JUCE: 0 failures.


magda/daw/core/TrackManager.cpp
```diff
@@ -3274,13 +3298,69 @@ void TrackManager::notifyDeviceAdded(const ChainNodePath& devicePath, const Devi
...
     newDevice.id = nextFxDeviceId_++;
+    rekeyPads(newDevice);
     applyCachedCapabilitiesToDevice(newDevice);
     stampDefaultKitIfMissing(newDevice);
     if (daw::audio::isInternalAnalysisPlugin(newDevice.pluginId))
         newDevice.deviceType = DeviceType::Analysis;
     return newDevice;
 }
 
+void TrackManager::rekeyPads(DeviceInfo& device) {
```
lucaromagnoli: `duplicateTrack` does not come through here. Its `reassignIds` walk re-stamps `device.id` and stops, so a duplicated Drum Grid keeps the source's pad device ids and a `pads->id` derived from the source's device id.

Two consequences I traced:
- `RuntimeStateStore` keys runtimes on (segment, DeviceId) with no track in the key, so the original's and the duplicate's pad devices share one runtime under the native engine.
- `padRackIdFor(newId)` no longer matches `pads->id`, so `padChainPath` on the duplicate resolves to nothing: `getChainByPath` fails, pad faders and mute/solo silently no-op, and `syncDrumGridPadPlugins` registers no processor for the pad plugins. `ensurePads` restamps on the next add, but a fader move never reaches it.

`reassignIds` needs to call `rekeyPads` per device, and `remap.devices` needs the pad ids too if a macro or mod on the grid targets one.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -1152,120 +1219,112 @@ bool DeviceCustomUIManager::createDrumGridUI(const magda::DeviceInfo& device,
...
-    // Level/pan/mute/solo callbacks - write directly to chain CachedValues
-    drumGridUI_->onPadLevelChanged = [getDrumGrid](int padIndex, float levelDb) {
-        if (auto* dg = getDrumGrid()) {
-            int midiNote = daw::audio::DrumGridPlugin::baseNote + padIndex;
-            if (auto* chain = dg->getChainForNote(midiNote))
-                const_cast<daw::audio::DrumGridPlugin::Chain*>(chain)->level = levelDb;
-        }
+    // A pad's fader, pan and switches are its chain's, so they are set the way
+    // any other chain's are.
+    drumGridUI_->onPadLevelChanged = [padPath](int padIndex, float levelDb) {
```
lucaromagnoli: This writes the model and nothing carries it to the plugin. `setChainVolume` and `setChainPan` raise `trackPropertyChanged`; `AudioBridge` answers that with `syncRackProperties`, and `syncRackProperties` (PluginManagerMacros.cpp) visits rack elements only. `syncDrumGridPads` runs from `syncTrackPlugins`, which only `trackDevicesChanged` triggers. So under Tracktion a pad level or pan move is inaudible until some device edit happens to run a full sync. Mute, solo and bypass are fine, they go through `trackDevicesChanged`.

Same drift the PR removes, in the other direction. `syncRackProperties` needs to reach every Drum Grid on the track (a `syncDrumGridPads` per grid would do; the fingerprint keeps it cheap).

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -751,8 +752,10 @@ void PluginManager::syncTrackPlugins(TrackId trackId) {
                     drumGrids.push_back({devicePath, dg});
             }
         }
-        for (auto& [devicePath, dg] : drumGrids)
+        for (auto& [devicePath, dg] : drumGrids) {
+            syncDrumGridPads(devicePath, *dg);
```
lucaromagnoli: The `drumGrids` collection above keys on `syncedDevices_` entries that carry a plugin, and for a rack-hosted Drum Grid the only thing that puts one there is the inner-plugin registration loop earlier in this function, which walks first-level rack chains only. A Drum Grid inside a nested rack is never handed to `syncDrumGridPads`, and the rehydration from `pluginState` that `createPluginOnly` did for it is gone in this PR, so it now comes up with no pads under Tracktion. Before this PR it played from its saved state.

Either recurse in that registration loop the way `loadRackContents` does, or collect the Drum Grids from `RackSyncManager`'s inner plugin map instead of from `syncedDevices_`.

lucaromagnoli (commented):
The flip itself is right and the model side reads clean. Three places where an existing path was not moved onto it, each verified against the branch.

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -701,6 +701,10 @@ void ProjectSerializer::commitStagedData(std::vector<TrackInfo>& stagedTracks,
             // After all tracks are restored, ensure TrackManager ID counters
             // (track/device/rack/chain) are updated to avoid ID collisions.
             trackManager.refreshIdCountersFromTracks();
+
+            // Then hand an id to every pad device saved before pads had them,
+            // now that the counters are past everything the project named.
+            trackManager.allocatePadDeviceIds();
```
lucaromagnoli: This is the only place pad ids get allocated, and the null-diff path never reaches it: `MgdFixture` and `test_real_project_survey` stop at `ProjectSerializer::loadAndStage` and render the staged tracks. So on this branch the survey still refuses `1.0.0-drumgrid-rack.mgd` with `pad 0: a plugin has no device id, so the pad is not compiled` (7 of 22 clean; that one would be 8). The app heals it here and the harness does not, which is exactly the app/harness split the rig was built to avoid.

The allocation belongs on the load: run it at the end of `loadAndStage` after the counters are refreshed from the staged tracks, or give the rig the same post-load step the app runs. Either way the survey should show the drum-grid-in-a-rack project clean before this merges, since that project is the one this PR is about.

lucaromagnoli (commented):
One more from running the survey on this head.

lucaromagnoli:
Audited what the pad ownership flip leaves behind, in one pass rather than one finding at a time: #2211.

Nothing in it blocks this PR. It covers the track- and rack-scoped link that does not survive duplication, `findDevicePath` not seeing pad devices, pad edits having no undo command, three pad controls wired to nothing (bus output, note range, slot meters), and multi-out being inert at both ends.

PR #2213, opened 2026-08-27
Finish the Drum Grid pad ownership flip (#2211)

### LINKED ISSUE #2211: Finish the Drum Grid pad ownership flip: links, undo, and the controls it left unwired
#2207 moved a Drum Grid's pads into the model. This is the audit of what that
left behind, done in one pass over every place a pad can be written, read,
addressed, captured or restored, rather than one finding at a time.

Nothing here is a reason to hold #2208: the flip itself is correct and every
regression it introduced is fixed on that branch. What follows is the work the
flip exposed and did not finish.

## 1. A track- or rack-scoped link into a pad does not survive duplication

`remapDuplicatedPath` rewrites a Rack step through `remap.racks` and a Device
step through `remap.devices`. A pad device's address is
`Rack(gridDeviceId) > Chain(pad) > Device(padDevice)`, and the grid's DeviceId
deliberately does NOT go into `remap.racks`: rack ids and device ids come out of
counters that both start at 1, so putting it there would rewrite the paths of any
rack that shares the number.

The grid's own links are handled (`TrackManager::retargetPadLinks`), but a track
macro or a rack macro pointing at a pad device parameter is not. Those get their
Device step remapped and their Rack step left alone, so the copy's link addresses
`Rack(oldGridId) > Chain(pad) > Device(newPadId)`: a path to nothing.

`DeviceSlotDrumGridBridge` hands `trackMacros_` and `trackMods_` to the pad slot
link context, so a track-scoped link into a pad is reachable from the UI.

The fix is the same whole-path retarget `retargetPadLinks` does, applied to the
track's and each rack's links as well, keyed on the grid ids the duplication
re-stamped.

## 2. A pad device cannot be found by id

`TrackManager::findDevicePath(DeviceId)` walks the track's own list and one level
of rack chains. It does not descend into `DeviceInfo::pads`, so no pad device
resolves. Anything that starts from a DeviceId and needs a path -- alias
generation, automation targets, link repair -- cannot see a pad.

This was true before #2207 as well, because a pad device was not in the model at
all. It is worth fixing now precisely because it IS: a pad device is an ordinary
`DeviceInfo` in every other respect, and the asymmetry is the kind that produces
a silent miss rather than an error.

## 3. Pad edits are not undoable

Every other chain edit has a command: `AddDeviceByPathCommand`,
`RemoveDeviceByPathCommand` and friends in `TrackCommands.cpp`. The pad commands
on `TrackManager` (`setPadDevice`, `addDeviceToPad`, `removeDeviceFromPad`,
`moveDeviceInPad`, `clearPad`, `swapPads`, the property setters) write the model
directly, so Ctrl+Z does not reach a pad.

#2207's own text asks for this: "Undo goes through the model like everything
else." It does not yet.

## 4. Controls that are wired to nothing

Each of these is a real control in the pad UI whose callback is never assigned.
All predate #2207; they are listed because the flip is what makes them fixable,
and because two of them now display model state they cannot change.

- `DrumGridUI::onPadOutputChanged` (DrumGridUI.hpp:120). The per-pad bus
  selector calls it; `DeviceCustomUIManager` never sets it. `ChainInfo::outputIndex`
  is model state now, so this is a one-line wire-up.
- `DrumGridUI::getNoteRange` / `onPadRangeChanged` (DrumGridUI.hpp:126, 130).
  Nothing assigns either, and `PadChainRangeRowComponent`, the component that
  edits a pad's note range, is not constructed anywhere. So `lowNote`,
  `highNote` and `rootNote` are model state with no way to edit them, on a device
  whose whole point is being keyed by pitch.
- `PadChainPanel::PluginSlotInfo::getMeterLevels` (PadChainPanel.hpp:52). Pad
  slot meters never move.

## 5. Multi-out is inert

`DrumGridPlugin::setMultiOutEnabled`, `assignBusOutputs` and `setChainBusOutput`
were removed by #2207 because nothing called them. Nothing did: `multiOutEnabled_`
is persisted state with no UI, so it was always false and `assignBusOutputs()` only
ever reset every pad to bus 0.

`PluginManager::syncDrumGridMultiOutTracks` still reads a pad's bus and would
create multi-out child tracks for it, and `PlanCompiler::emitPadRack` compiles a
per-bus mix and reports a bus that reaches no track. So the engine and the plan
both support per-pad outputs and nothing can turn them on. Either wire the toggle
and the bus selector (item 4) or delete the machinery.

## 6. Slice-generated pad state reaches the model only at save

`buildDrumGridFromSlices` puts a sampler on each pad through the model, then
writes `sampleStart` and `sampleEnd` onto the live plugin. Those values reach
`DeviceInfo` at the next capture, which is the same contract every other
parameter edit in the app has, so this is consistent rather than wrong. Noted
because it is the one remaining place a pad's state is written engine-first, and
because if parameter capture is ever made stricter this is where it will bite.

## Not in scope

The Drum Grid's own `mixerExpanded` and `multiOutEnabled` stay in its
`pluginState`. They are the device's own properties, not pads, and belong there.

lucaromagnoli:
Closes #2211.

#2207 gave a Drum Grid's pads to the model. This is the work that audit found unfinished: items 2 through 7 (item 1 shipped on #2208, item 6 was noted rather than a task).

## Found by a pad device's id (item 2)

`findDevicePath()` descends into `DeviceInfo::pads`, so a pad device resolves to the address every stored link to one already carries: `Rack(gridDeviceId) > Chain(pad) > Device(padDevice)`.

Rewriting it as one recursion also closed two gaps it already had: it went only one rack level deep, and it never looked in the master track's racks at all.

## Collected as a modulation source (item 7)

`ModSources` walks a pad's elements the way it walks a rack chain's. A pad device can own macros and mods since #2207 and nothing read them, so the plan carried no edge for a modifier the parameter table has an address for.

Under the pad's own track, not the grid's sidechain source: a device inside a rack hears its own sidechain, and the rack's is the rack's.

## Undoable (item 3)

`EditPadsCommand` snapshots the grid's whole pad rack. That is one field on one `DeviceInfo`, and it is the only snapshot that stays correct for the edits that touch more than one pad at a time: a swap, a replace, a clear. The per-op commands in `TrackCommands.cpp` cannot serve a pad, because a pad path's Rack step spells a DeviceId and would resolve onto whichever rack shares the number.

Redo re-runs the edit rather than restoring an after-state, so an edit that allocates ids hands out the ones a fresh run would. A `mergeKey` coalesces a fader drag into one step.

### A use-after-free the flip introduced

Every pad edit in the UI is now posted rather than called inline. Since #2207 a pad edit writes the model and notifies `trackDevicesChanged`; `TrackChainContent::trackDevicesChanged` calls `rebuildNodeComponents()`, which clears `nodeComponents_` synchronously and owns the slots by `unique_ptr`. That freed the device slot, the pad UI inside it and the row whose button was still on the stack. Before #2207 these callbacks wrote the plugin's `CachedValue` directly and notified nothing, so there was nothing to tear down.

Posting lets the gesture finish first. The faders stay inline: they notify `trackPropertyChanged`, which by design does not rebuild.

The general shape is untouched. Any UI callback that writes a device property firing `trackDevicesChanged` has the same exposure, because that handler rebuilds every node component. Only the pad path is closed here.

## Controls wired to nothing (item 4)

- `onPadOutputChanged` writes `ChainInfo::outputIndex` through a new `TrackManager::setPadOutput`.
- `getNoteRange` / `onPadRangeChanged` read and write a pad's note range through a new `TrackManager::setPadNoteRange`, and `PadChainRangeRowComponent` is built for every chain row and laid out under the selected one. Built with the rows rather than on selection, because selection changes from a row's own `mouseUp`.
- A pad slot's meters drain the grid's own per-plugin peaks. A pad plugin runs inside the grid, so it has no `DeviceMeteringManager` tap of its own.

## Multi-out (item 5)

Wiring the bus selector makes the machinery live: the engine already routes a pad's bus, `PluginManager` already creates the child track for it, and the device slot's multi-out button already opens the pairs.

`DrumGridPlugin::multiOutEnabled` is deleted. It gated `assignBusOutputs()`, which #2207 removed, and nothing ever wrote or read either. This is the "delete the machinery" half of item 5 applied to the one piece nothing reads; the issue's out-of-scope note was about where `mixerExpanded` and `multiOutEnabled` live, not about keeping a flag with no readers.

## Seeing it in the app

Open a track with a Drum Grid and expand the device slot. The chains panel on the right lists one row per loaded pad; selecting a row now shows a second row under it with Low, High and Root note fields. The bus button on a pad row (Main / B1 / B2 ...) now sticks and creates the multi-out child track. Pad slot meters in the detail panel move while a pad plays. Ctrl+Z steps back through pad edits.

## Tests

Six new cases: pad lookup by id, output bus, key range clamping and low-first ordering, a pad retuned off its own note, undo/redo of a pad edit, fader coalescing, and the modulation-source descent. The `ModSources` one was checked against the unpatched walk and fails there.

3014 Catch2 cases and the full JUCE suite pass locally.

### DIFF
--- a/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
+++ b/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
@@ -2971,7 +2971,11 @@ void PluginManager::syncDrumGridMultiOutTracks(const ChainNodePath& drumGridPath
     const auto trackId = drumGridPath.trackId;
     const auto deviceId = drumGridPath.getDeviceId();
     auto& tm = TrackManager::getInstance();
-    auto* devInfo = tm.getDevice(trackId, deviceId);
+
+    // By path, not by (track, device): the flat lookup misses a grid nested in
+    // a rack, which is a placement the model supports and the pad bus selector
+    // can now reach (#2211).
+    auto* devInfo = tm.getDeviceInChainByPath(drumGridPath);
     if (!devInfo || !devInfo->multiOut.isMultiOut)
         return;
 
--- a/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
+++ b/magda/daw/audio/plugin_manager/PluginManagerSync.cpp
@@ -2979,6 +2979,18 @@ void PluginManager::syncDrumGridMultiOutTracks(const ChainNodePath& drumGridPath
     if (!devInfo || !devInfo->multiOut.isMultiOut)
         return;
 
+    // A grid can arrive here in a placement where buses do not work: wrapping a
+    // top-level one in a rack moves it and keeps its pads, and a project can be
+    // loaded already like that. Nothing carries a bus off a nested grid, so a
+    // pad left on one would simply go silent, and any child track it had made
+    // would linger with nothing feeding it. Put the pads back on the grid's own
+    // mix and take the children down.
+    if (!tm.padBusesAvailable(drumGridPath)) {
+        tm.resetPadBuses(drumGridPath);
+        tm.deactivateAllMultiOutPairs(trackId, deviceId);
+        return;
+    }
+
     auto& pairs = devInfo->multiOut.outputPairs;
     const auto& chains = drumGrid->getChains();
 
--- a/magda/daw/core/PadCommands.cpp
+++ b/magda/daw/core/PadCommands.cpp
@@ -35,11 +35,8 @@ PadRack snapshotPads(const ChainNodePath& gridPath) {
 }  // namespace
 
 EditPadsCommand::EditPadsCommand(const ChainNodePath& gridPath, juce::String description,
-                                 std::function<void()> edit, juce::String mergeKey)
-    : gridPath_(gridPath),
-      description_(std::move(description)),
-      edit_(std::move(edit)),
-      mergeKey_(std::move(mergeKey)) {}
+                                 std::function<void()> edit)
+    : gridPath_(gridPath), description_(std::move(description)), edit_(std::move(edit)) {}
 
 void EditPadsCommand::execute() {
     auto& tm = TrackManager::getInstance();
@@ -74,27 +71,69 @@ void EditPadsCommand::undo() {
     TrackManager::getInstance().setPads(gridPath_, padsBefore_);
 }
 
-bool EditPadsCommand::canMergeWith(const UndoableCommand* other) const {
-    if (mergeKey_.isEmpty())
-        return false;
-
-    const auto* otherEdit = dynamic_cast<const EditPadsCommand*>(other);
-    return otherEdit != nullptr && otherEdit->gridPath_ == gridPath_ &&
-           otherEdit->mergeKey_ == mergeKey_;
-}
-
-void EditPadsCommand::mergeWith(const UndoableCommand* other) {
-    // The later state is where redo lands; the state to come back to stays the
-    // one this command was made against, which is the point of merging. The
-    // other command has already run by the time this is called.
-    if (const auto* otherEdit = dynamic_cast<const EditPadsCommand*>(other))
-        padsAfter_ = otherEdit->padsAfter_;
-}
-
 void editPads(const ChainNodePath& gridPath, const juce::String& description,
-              std::function<void()> edit, const juce::String& mergeKey) {
+              std::function<void()> edit) {
     UndoManager::getInstance().executeCommand(
-        std::make_unique<EditPadsCommand>(gridPath, description, std::move(edit), mergeKey));
+        std::make_unique<EditPadsCommand>(gridPath, description, std::move(edit)));
+}
+
+// ============================================================================
+// SetPadFaderCommand
+// ============================================================================
+
+SetPadFaderCommand::SetPadFaderCommand(const ChainNodePath& gridPath, int padIndex, Target target,
+                                       float value, int gesture)
+    : gridPath_(gridPath),
+      padIndex_(padIndex),
+      target_(target),
+      newValue_(value),
+      gesture_(gesture) {
+    if (const auto* pad = TrackManager::getInstance().getPad(gridPath_, padIndex_)) {
+        oldValue_ = target_ == Target::Volume ? pad->volume : pad->pan;
+        valid_ = true;
+    }
+}
+
+void SetPadFaderCommand::apply(float value) const {
+    auto& tm = TrackManager::getInstance();
+    if (target_ == Target::Volume)
+        tm.setPadVolume(gridPath_, padIndex_, value);
+    else
+        tm.setPadPan(gridPath_, padIndex_, value);
+}
+
+void SetPadFaderCommand::execute() {
+    if (valid_)
+        apply(newValue_);
+}
+
+void SetPadFaderCommand::undo() {
+    if (valid_)
+        apply(oldValue_);
+}
+
+juce::String SetPadFaderCommand::getDescription() const {
+    return target_ == Target::Volume ? "Set Pad Level" : "Set Pad Pan";
+}
+
+bool SetPadFaderCommand::canMergeWith(const UndoableCommand* other) const {
+    const auto* otherFader = dynamic_cast<const SetPadFaderCommand*>(other);
+    return otherFader != nullptr && otherFader->gridPath_ == gridPath_ &&
+           otherFader->padIndex_ == padIndex_ && otherFader->target_ == target_ &&
+           otherFader->gesture_ == gesture_;
+}
+
+void SetPadFaderCommand::mergeWith(const UndoableCommand* other) {
+    // The value to redo to moves on; the one to come back to stays where the
+    // gesture started, which is the point of merging.
+    if (const auto* otherFader = dynamic_cast<const SetPadFaderCommand*>(other))
+        newValue_ = otherFader->newValue_;
+}
+
+void setPadFader(const ChainNodePath& gridPath, int padIndex, SetPadFaderCommand::Target target,
+                 float value, int gesture) {
+    UndoManager::getInstance().executeCommand(
+        std::make_unique<SetPadFaderCommand>(gridPath, padIndex, target, value, gesture));
 }
 
 }  // namespace magda
--- a/magda/daw/core/PadCommands.cpp
+++ b/magda/daw/core/PadCommands.cpp
@@ -1,0 +1,63 @@
+#include "PadCommands.hpp"
+
+#include <memory>
+#include <utility>
+
+#include "RackInfo.hpp"
+
+namespace magda {
+
+EditPadsCommand::EditPadsCommand(const ChainNodePath& gridPath, juce::String description,
+                                 std::function<void()> edit, juce::String mergeKey)
+    : gridPath_(gridPath),
+      description_(std::move(description)),
+      edit_(std::move(edit)),
+      mergeKey_(std::move(mergeKey)) {}
+
+void EditPadsCommand::execute() {
+    if (!edit_)
+        return;
+
+    auto& tm = TrackManager::getInstance();
+
+    // Taken before the edit and only on the first run: redo re-runs the edit,
+    // so the state to come back to is the one this command was made against.
+    if (!executed_) {
+        if (const auto* pads = tm.getPads(gridPath_))
+            padsBefore_.reset(std::make_unique<RackInfo>(*pads));
+    }
+
+    edit_();
+    executed_ = true;
+}
+
+void EditPadsCommand::undo() {
+    if (!executed_)
+        return;
+
+    TrackManager::getInstance().setPads(gridPath_, padsBefore_);
+}
+
+bool EditPadsCommand::canMergeWith(const UndoableCommand* other) const {
+    if (mergeKey_.isEmpty())
+        return false;
+
+    const auto* otherEdit = dynamic_cast<const EditPadsCommand*>(other);
+    return otherEdit != nullptr && otherEdit->gridPath_ == gridPath_ &&
+           otherEdit->mergeKey_ == mergeKey_;
+}
+
+void EditPadsCommand::mergeWith(const UndoableCommand* other) {
+    // The later edit is what redo replays; the state to come back to stays the
+    // one this command was made against, which is the point of merging.
+    if (const auto* otherEdit = dynamic_cast<const EditPadsCommand*>(other))
+        edit_ = otherEdit->edit_;
+}
+
+void editPads(const ChainNodePath& gridPath, const juce::String& description,
+              std::function<void()> edit, const juce::String& mergeKey) {
+    UndoManager::getInstance().executeCommand(
+        std::make_unique<EditPadsCommand>(gridPath, description, std::move(edit), mergeKey));
+}
+
+}  // namespace magda
--- a/magda/daw/core/PadCommands.cpp
+++ b/magda/daw/core/PadCommands.cpp
@@ -3,9 +3,36 @@
 #include <memory>
 #include <utility>
 
+#include "../audio/AudioBridge.hpp"
+#include "../engine/AudioEngine.hpp"
 #include "RackInfo.hpp"
 
 namespace magda {
+
+namespace {
+
+/// Flush the grid's live plugins into the model before it is snapshotted.
+///
+/// A pad plugin's patch reaches `DeviceInfo` only when the Drum Grid's state is
+/// captured (`PluginManager::captureDrumGridPads`), and the ordinary sync does
+/// not do it. Without this, undoing the removal of a pad whose sampler had been
+/// edited since it was added restores the sampler as it was added, not as it
+/// sounded. `RemoveDeviceByPathCommand` captures for the same reason.
+void capturePadPluginStates(const ChainNodePath& gridPath) {
+    auto& tm = TrackManager::getInstance();
+    if (auto* engine = tm.getAudioEngine())
+        if (auto* bridge = engine->getAudioBridge())
+            bridge->getPluginManager().capturePluginState(gridPath);
+}
+
+PadRack snapshotPads(const ChainNodePath& gridPath) {
+    PadRack pads;
+    if (const auto* live = TrackManager::getInstance().getPads(gridPath))
+        pads.reset(std::make_unique<RackInfo>(*live));
+    return pads;
+}
+
+}  // namespace
 
 EditPadsCommand::EditPadsCommand(const ChainNodePath& gridPath, juce::String description,
                                  std::function<void()> edit, juce::String mergeKey)
@@ -15,19 +42,28 @@ EditPadsCommand::EditPadsCommand(const ChainNodePath& gridPath, juce::String des
       mergeKey_(std::move(mergeKey)) {}
 
 void EditPadsCommand::execute() {
+    auto& tm = TrackManager::getInstance();
+
+    // Redo restores the after-state rather than replaying the edit. Replaying
+    // would run `addDeviceToPad()` again, and that allocates a fresh DeviceId
+    // every time: the pad device would come back under a new id while every
+    // later command in the redo chain still named the old one, and each of them
+    // would silently do nothing. Restoring the snapshot brings the ids back
+    // with it.
+    if (executed_) {
+        tm.setPads(gridPath_, padsAfter_);
+        return;
+    }
+
     if (!edit_)
         return;
 
-    auto& tm = TrackManager::getInstance();
-
-    // Taken before the edit and only on the first run: redo re-runs the edit,
-    // so the state to come back to is the one this command was made against.
-    if (!executed_) {
-        if (const auto* pads = tm.getPads(gridPath_))
-            padsBefore_.reset(std::make_unique<RackInfo>(*pads));
-    }
+    capturePadPluginStates(gridPath_);
+    padsBefore_ = snapshotPads(gridPath_);
 
     edit_();
+
+    padsAfter_ = snapshotPads(gridPath_);
     executed_ = true;
 }
 
@@ -48,10 +84,11 @@ bool EditPadsCommand::canMergeWith(const UndoableCommand* other) const {
 }
 
 void EditPadsCommand::mergeWith(const UndoableCommand* other) {
-    // The later edit is what redo replays; the state to come back to stays the
-    // one this command was made against, which is the point of merging.
+    // The later state is where redo lands; the state to come back to stays the
+    // one this command was made against, which is the point of merging. The
+    // other command has already run by the time this is called.
     if (const auto* otherEdit = dynamic_cast<const EditPadsCommand*>(other))
-        edit_ = otherEdit->edit_;
+        padsAfter_ = otherEdit->padsAfter_;
 }
 
 void editPads(const ChainNodePath& gridPath, const juce::String& description,
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -803,6 +803,25 @@ void TrackManager::deleteTrack(TrackId trackId) {
 // Multi-Output Management
 // =============================================================================
 
+namespace {
+
+/// The multi-out device @p deviceId names, wherever it sits on @p parentTrackId.
+///
+/// `getDevice(trackId, deviceId)` walks the track's flat lists only, so a
+/// multi-out instrument inside a rack chain answered nothing and its pairs
+/// could never be activated. A Drum Grid reaches this on any pad given a bus,
+/// and nesting one in a rack is explicitly supported (#2211).
+DeviceInfo* multiOutDevice(TrackManager& tm, TrackId parentTrackId, DeviceId deviceId) {
+    if (auto* device = tm.getDevice(parentTrackId, deviceId))
+        return device;
+
+    const auto path = tm.findDevicePath(deviceId);
+    return path.isValid() && path.trackId == parentTrackId ? tm.getDeviceInChainByPath(path)
+                                                           : nullptr;
+}
+
+}  // namespace
+
 TrackId TrackManager::activateMultiOutPair(TrackId parentTrackId, DeviceId deviceId,
                                            int pairIndex) {
     auto* parentTrack = getTrack(parentTrackId);
@@ -810,7 +829,7 @@ TrackId TrackManager::activateMultiOutPair(TrackId parentTrackId, DeviceId devic
         return INVALID_TRACK_ID;
 
     // Find the device
-    DeviceInfo* device = getDevice(parentTrackId, deviceId);
+    DeviceInfo* device = multiOutDevice(*this, parentTrackId, deviceId);
     if (!device || !device->multiOut.isMultiOut)
         return INVALID_TRACK_ID;
 
@@ -874,7 +893,7 @@ void TrackManager::deactivateMultiOutPair(TrackId parentTrackId, DeviceId device
     if (!parentTrack)
         return;
 
-    DeviceInfo* device = getDevice(parentTrackId, deviceId);
+    DeviceInfo* device = multiOutDevice(*this, parentTrackId, deviceId);
     if (!device || !device->multiOut.isMultiOut)
         return;
 
@@ -907,7 +926,7 @@ void TrackManager::deactivateAllMultiOutPairs(TrackId parentTrackId, DeviceId de
     // Re-fetch device pointer each iteration since deactivateMultiOutPair
     // calls tracks_.erase() which can invalidate pointers
     for (int i = 0;; ++i) {
-        DeviceInfo* device = getDevice(parentTrackId, deviceId);
+        DeviceInfo* device = multiOutDevice(*this, parentTrackId, deviceId);
         if (!device || !device->multiOut.isMultiOut)
             break;
         if (i >= static_cast<int>(device->multiOut.outputPairs.size()))
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -872,9 +872,14 @@ TrackId TrackManager::activateMultiOutPair(TrackId parentTrackId, DeviceId devic
         tracks_.push_back(std::move(newTrack));
     }
 
-    // Re-fetch pointers after insert (vector reallocation invalidates them)
+    // Re-fetch pointers after insert (vector reallocation invalidates them).
+    // Through the same lookup as above: the flat one misses a device inside a
+    // rack, and this dereferences what it answers (#2211).
     parentTrack = getTrack(parentTrackId);
-    device = getDevice(parentTrackId, deviceId);
+    device = multiOutDevice(*this, parentTrackId, deviceId);
+    if (device == nullptr || pairIndex >= static_cast<int>(device->multiOut.outputPairs.size()))
+        return INVALID_TRACK_ID;
+
     auto& pairRef = device->multiOut.outputPairs[static_cast<size_t>(pairIndex)];
 
     // Update the output pair state
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <map>
+#include <set>
 #include <unordered_set>
 
 #include "../audio/AudioBridge.hpp"
@@ -28,6 +29,9 @@ namespace {
 struct DuplicatePadRemap {
     DeviceId newOwnerId = INVALID_DEVICE_ID;
     std::map<DeviceId, DeviceId> devices;  ///< old pad DeviceId -> new
+    /// The owner's pad chain ids. Unchanged by duplication, and what tells a
+    /// pad prefix from a rack that merely shares the owner's number.
+    std::set<ChainId> chains;
 };
 
 struct DuplicateIdRemap {
@@ -39,6 +43,14 @@ struct DuplicateIdRemap {
     /// Keyed by the OLD DeviceId of the device that owns the pads.
     std::map<DeviceId, DuplicatePadRemap> pads;
 };
+
+template <typename Id> bool remapDuplicateId(const std::map<Id, Id>& ids, int& value) {
+    auto it = ids.find(value);
+    if (it == ids.end())
+        return false;
+    value = it->second;
+    return true;
+}
 
 /// Rewrite @p path if it addresses a pad device, and say whether it did.
 ///
@@ -55,29 +67,51 @@ struct DuplicateIdRemap {
 /// passes through a rack numbered like a grid is left to the ordinary remap
 /// (#2207).
 bool remapDuplicatedPadPath(ChainNodePath& path, const DuplicateIdRemap& remap) {
-    if (path.steps.size() != 3 || path.steps[0].type != ChainStepType::Rack ||
-        path.steps[1].type != ChainStepType::Chain || path.steps[2].type != ChainStepType::Device)
+    if (path.steps.size() < 3 || path.steps[0].type != ChainStepType::Rack ||
+        path.steps[1].type != ChainStepType::Chain)
         return false;
 
     const auto owner = remap.pads.find(path.steps[0].id);
     if (owner == remap.pads.end())
         return false;
 
-    const auto device = owner->second.devices.find(path.steps[2].id);
-    if (device == owner->second.devices.end())
+    // A pad device, named directly by the pad's chain.
+    if (path.steps.size() == 3 && path.steps[2].type == ChainStepType::Device) {
+        const auto device = owner->second.devices.find(path.steps[2].id);
+        if (device == owner->second.devices.end())
+            return false;
+
+        path.trackId = remap.newTrackId;
+        path.steps[0].id = owner->second.newOwnerId;
+        path.steps[2].id = device->second;
+        return true;
+    }
+
+    // Or something further down, inside a rack the pad's chain holds. The
+    // prefix is still the pad's; what follows is an ordinary route and moves
+    // with the ordinary maps.
+    if (!owner->second.chains.contains(path.steps[1].id))
         return false;
 
     path.trackId = remap.newTrackId;
     path.steps[0].id = owner->second.newOwnerId;
-    path.steps[2].id = device->second;
-    return true;
-}
 
-template <typename Id> bool remapDuplicateId(const std::map<Id, Id>& ids, int& value) {
-    auto it = ids.find(value);
-    if (it == ids.end())
-        return false;
-    value = it->second;
+    for (std::size_t index = 2; index < path.steps.size(); ++index) {
+        auto& step = path.steps[index];
+        switch (step.type) {
+            case ChainStepType::Rack:
+                remapDuplicateId(remap.racks, step.id);
+                break;
+            case ChainStepType::Chain:
+                remapDuplicateId(remap.chains, step.id);
+                break;
+            case ChainStepType::Device:
+                remapDuplicateId(remap.devices, step.id);
+                break;
+            case ChainStepType::Segment:
+                break;
+        }
+    }
     return true;
 }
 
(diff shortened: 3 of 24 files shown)


magda/daw/core/PadCommands.cpp
```diff
@@ -0,0 +1,63 @@
...
+void EditPadsCommand::execute() {
+    if (!edit_)
+        return;
+
+    auto& tm = TrackManager::getInstance();
+
+    // Taken before the edit and only on the first run: redo re-runs the edit,
+    // so the state to come back to is the one this command was made against.
+    if (!executed_) {
+        if (const auto* pads = tm.getPads(gridPath_))
```
lucaromagnoli: **P1 — Capture live pad plugin state before snapshotting**

This copies only the current `DeviceInfo::pads`. A pad plugins live patch is written back to that model only by `PluginManager::captureDrumGridPads()`, reached through explicit Drum Grid state capture; ordinary sync does not capture it. Therefore clearing, replacing, or removing a pad after editing its plugin and then undoing can restore stale `pluginState` and parameter values. The existing remove-device commands explicitly capture live state before their undo snapshot for this reason. Please capture the Drum Grid (which also captures its pads) before copying `padsBefore_`.

magda/daw/core/PadCommands.cpp
```diff
@@ -0,0 +1,63 @@
...
+            padsBefore_.reset(std::make_unique<RackInfo>(*pads));
+    }
+
+    edit_();
```
lucaromagnoli, line 61: **P1 — Preserve device identity across redo**

Re-running an add is not equivalent to restoring its after-state: `addDeviceToPad()` calls `prepareNewDevice()`, which allocates a fresh global DeviceId on every execution. A sequence such as add pad FX → disable or change its gain → undo twice → redo twice recreates the FX under a new ID, while the later command still targets the original captured ID and silently becomes a no-op. Automation and links created after the add can likewise retain the old identity. Redo needs to restore an after-snapshot with the original IDs, or otherwise guarantee stable identity for the whole redo chain.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1284,30 +1284,77 @@ void TrackManager::setDeviceKitRows(TrackId trackId, DeviceId deviceId,
...
+            if (!device.pads)
+                continue;
+
+            // A pad's address spells its rack step with the grid's own
+            // DeviceId: what padChainPath() builds, and what every link a
+            // project has saved to a pad device carries.
+            const auto gridPath = devicePathUnder(parentChain, device.id);
+            for (const auto& pad : device.pads->chains)
+                if (auto found = findDeviceUnder(
+                        pad.elements, TrackManager::padChainPath(gridPath, pad.id), deviceId);
```
lucaromagnoli, line 1564: **P1 — Return a path the model resolver can follow**

This builds the saved/synthetic pad address `Rack(gridDeviceId) > Chain > Device`, but `getDeviceInChainByPath()` resolves the chain through `getRackByPath()`, which deliberately traverses allocated racks only and does not resolve `DeviceInfo::pads`. Consequently `findDevicePath()` now returns a valid-looking path for a pad device that generic callers cannot dereference. For example, `AutoAliasGenerator::regenerateForDevice(DeviceId)` calls `findDevicePath()` and then `getDeviceInChainByPath()`, so it still exits without generating aliases; parameter writeback by bare DeviceId fails the same way. Please either make the generic resolver understand this pad path or route these callers through a pad-aware lookup.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1076,6 +1097,45 @@ DeviceInfo* TrackManager::getDeviceInChainByPath(const ChainNodePath& devicePath
...
+    // rather than a chain element, so `getRackByPath()` cannot follow it and
+    // was never meant to (#2207).
+    //
+    // Tried only after the ordinary route has failed, so an allocated rack that
+    // shares the number resolves exactly as it did before. That ordering is
+    // what makes this unambiguous rather than a guess: every device in the
+    // Rack > Chain > Device space comes out of one counter, so the leaf id
+    // names one device in the whole project, and whichever route reaches it
+    // reaches the same one (#2211).
+    if (devicePath.steps.size() != 3)
```
lucaromagnoli: **P2 — Resolve the recursive paths that `findDevicePath` returns**

This fallback handles only a device directly in a pad. `findDeviceUnder()` continues recursively through rack elements inside `pad.elements`, so for a device in such a rack it returns `Rack(grid) > Chain(pad) > Rack > Chain > Device`. The ordinary resolver fails at the synthetic first Rack step, then this exact-size check rejects the path, leaving the same valid-looking-but-undereferenceable result for that supported model shape. The plan, parameter-table, modulation, and runtime-state walks all recurse through pad elements, so the fallback should consume the synthetic owner/pad prefix and then traverse the remaining rack path rather than require exactly three steps.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -1180,15 +1182,43 @@ bool DeviceCustomUIManager::createDrumGridUI(const magda::DeviceInfo& device,
...
+    // resolver reads this manager and the rebuild may have taken it by then.
+    auto postPadEdit = [this, gridPath](const juce::String& description,
+                                        std::function<void(const magda::ChainNodePath&)> edit,
+                                        std::function<void()> then = {},
+                                        const juce::String& mergeKey = {}) {
+        const auto grid = gridPath();
+        if (!grid.isValid())
+            return;
+
+        juce::MessageManager::callAsync(
```
lucaromagnoli, line 1218: **P1 — Do not rebuild continuous controls before mouse-up**

`callAsync` waits only until the current `onValueChanged` callback returns; it does not wait for the full mouse gesture. Both the key-range `TextSlider`s and the pad-device gain `TextSlider` call `postPadEdit` on every drag update. The first queued edit calls `setPadNoteRange` or `setPadDeviceGainDb`, which synchronously emits `trackDevicesChanged` and rebuilds this UI, destroying the slider before its `mouseUp` and truncating the drag after the first delta. Buttons are safe with this helper, but continuous controls need to commit on `onDragEnd`, or update through a notification path that does not rebuild their owner during the gesture.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1284,30 +1344,77 @@ void TrackManager::setDeviceKitRows(TrackId trackId, DeviceId deviceId,
...
+}
+
+}  // namespace
+
+ChainNodePath TrackManager::findDevicePath(DeviceId deviceId) const {
+    for (const auto& track : tracks_)
+        if (auto found = findDeviceUnder(track.chain.fxChainElements,
+                                         ChainNodePath::trackLevel(track.id), deviceId);
+            found.isValid())
+            return found;
```
lucaromagnoli, line 1592: **P1 — Use this canonical path for pad parameter controls**

The pad-chain UI still bypasses this lookup: `PadChainPanel::applyLinkContextToSlot()` constructs `ChainNodePath::topLevelDevice(trackId, info.deviceId)` for every pad plugin. That path has no Rack component, while `ParamTableCompiler` keys a pad device under the owning grid DeviceId, so links, automation targets, MIDI learn, and parameter selection created from those controls do not address the compiled pad parameter. Now that this method returns the canonical pad path, that consumer should use `findDevicePath(info.deviceId)` instead of manufacturing a top-level path.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,38 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+void TrackManager::setPadOutput(const ChainNodePath& gridPath, int padIndex, int outputIndex) {
+    auto* pad = mutablePad(gridPath, padIndex);
+    if (pad == nullptr || pad->outputIndex == outputIndex)
+        return;
+
+    pad->outputIndex = outputIndex;
+
+    // trackDevicesChanged, not trackPropertyChanged: a pad's bus decides
+    // whether the grid needs a multi-out child track for it, and that is
+    // settled by the device sync (PluginManager::syncDrumGridMultiOutTracks).
```
lucaromagnoli, line 413: **P2 — Resolve the Drum Grid by its full path for multi-out sync**

`setPadOutput()` accepts a `gridPath` for a Drum Grid nested in a rack (that placement is explicitly supported), but the downstream `syncDrumGridMultiOutTracks()` looks the grid up with `tm.getDevice(trackId, deviceId)`. That overload only scans the flat top-level/post-FX/analysis lists, so a nested grid returns null here and the selected bus never activates its multi-out child track. Please carry/use `drumGridPath` with `getDeviceInChainByPath()` through the multi-out helpers, rather than falling back to the bare track/device lookup.

magda/daw/core/PadCommands.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    TrackManager::getInstance().setPads(gridPath_, padsBefore_);
+}
+
+bool EditPadsCommand::canMergeWith(const UndoableCommand* other) const {
+    if (mergeKey_.isEmpty())
+        return false;
+
+    const auto* otherEdit = dynamic_cast<const EditPadsCommand*>(other);
+    return otherEdit != nullptr && otherEdit->gridPath_ == gridPath_ &&
```
lucaromagnoli: **P2 — Bound merging to one drag gesture**

This compares only the grid and stable merge key. `UndoManager` merges adjacent commands after they execute and has no timeout or mouse-up boundary, so dragging a pad fader, releasing it, and later dragging the same fader again collapses both gestures into one undo entry (until some unrelated command intervenes). One Undo then reverts multiple user gestures, despite the UI contract here being one step per drag. Please include a per-gesture token or otherwise close the merge group on drag end.

magda/daw/core/PadCommands.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // with it.
+    if (executed_) {
+        tm.setPads(gridPath_, padsAfter_);
+        return;
+    }
+
+    if (!edit_)
+        return;
+
+    capturePadPluginStates(gridPath_);
```
lucaromagnoli, line 58: **P2 — Avoid capturing every pad plugin on every fader tick**

Pad level and pan call `editPads()` for every `TextSlider` value change, and every one of those commands reaches this capture before merging. `capturePluginState(gridPath_)` walks all Drum Grid chains, flushes each external plugin state, repopulates parameters, and the command then deep-copies the entire pad rack twice. Because merging happens only after `execute()`, it does not remove any of that per-pixel work; a populated grid can stall badly while a fader is dragged. Please capture once at gesture start (or reserve live-plugin capture for destructive edits) and commit/merge the lightweight model changes separately.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,38 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+    notifyTrackDevicesChanged(gridPath.trackId);
+}
+
+void TrackManager::setPadNoteRange(const ChainNodePath& gridPath, int padIndex, int lowNote,
+                                   int highNote, int rootNote) {
+    auto* pad = mutablePad(gridPath, padIndex);
```
lucaromagnoli: **P2 — Identify the chain directly when ranges overlap**

`padIndex` stops being an unambiguous chain key as soon as this editor allows overlapping ranges. During UI refresh, the later chain overwrites `padInfos_` for an overlapping note, but `mutablePad()`/`findPadChain()` returns the first model chain covering that note. A range row that appears to represent the later chain can therefore change the earlier chain instead; with complete overlap, one chain also disappears from the rows entirely. Please have the row carry the actual `ChainId` into this setter (and the other row actions), or reject/normalize overlapping ranges so one note always has one owner.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -1334,9 +1435,15 @@ bool DeviceCustomUIManager::createDrumGridUI(const magda::DeviceInfo& device,
...
 
     // Delete from chain row — same as clear
-    drumGridUI_->onPadDeleteRequested = [this, gridPath](int padIndex) {
-        magda::TrackManager::getInstance().clearPad(gridPath(), padIndex);
-        drumGridUI_->updatePadInfo(padIndex, "", false, false, 0.0f, 0.0f, -1);
+    drumGridUI_->onPadDeleteRequested = [this, postPadEdit](int padIndex) {
+        postPadEdit(
+            "Clear Pad",
+            [padIndex](const magda::ChainNodePath& grid) {
+                magda::TrackManager::getInstance().clearPad(grid, padIndex);
```
lucaromagnoli: **P2 — Allow a ranged chain to be deleted**

The new range editor can make a chain answer to multiple notes, but this chain-row Delete action still delegates to `clearPad()`. That method intentionally returns without doing anything unless `lowNote == highNote == padNoteFor(padIndex)`, so after widening a range, both this Delete action and the detail Clear button silently stop working and the chain cannot be removed from this UI. The chain-row action should remove the represented `ChainId` directly (with an explicit whole-range confirmation if desired), rather than using the single-pad clear guard.

magda/daw/core/TrackManager.cpp
```diff
@@ -803,14 +803,33 @@ void TrackManager::deleteTrack(TrackId trackId) {
...
+
 TrackId TrackManager::activateMultiOutPair(TrackId parentTrackId, DeviceId deviceId,
                                            int pairIndex) {
     auto* parentTrack = getTrack(parentTrackId);
     if (!parentTrack)
         return INVALID_TRACK_ID;
 
     // Find the device
-    DeviceInfo* device = getDevice(parentTrackId, deviceId);
+    DeviceInfo* device = multiOutDevice(*this, parentTrackId, deviceId);
```
lucaromagnoli, line 866: **P1 — Re-fetch the nested device with this lookup too**

This initial lookup is correct, but after inserting the child track (lines 875–878), the function must re-fetch pointers because `tracks_` may have reallocated and it still does so with the old flat-only `getDevice(parentTrackId, deviceId)`. For a Drum Grid inside a rack that returns null, and the following `device->multiOut` dereference crashes during bus activation. Please use `multiOutDevice(*this, parentTrackId, deviceId)` for the post-insert re-fetch as well, with a null guard.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,75 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+    // One note, one owner. Every chain whose range covers an incoming note
+    // plays it, so an overlap layers two sounds the UI has no way to tell
+    // apart: the rows show the last chain to claim a note and the setters
+    // reach the first, so a row would edit a chain other than the one it
+    // shows. Refused rather than silently normalised, so the slider snaps
+    // back to what the model kept.
+    for (const auto& other : pads->chains)
+        if (other.id != pad->id && !other.answersToEveryNote() && low <= other.highNote &&
+            high >= other.lowNote)
+            return false;
```
lucaromagnoli, line 450: **P2 — Propagate a rejected range back to the UI**

This returns before `notifyTrackDevicesChanged()`, while `onPadRangeChanged` discards the boolean and the range row has already kept the dragged/typed slider value. Consequently the rejected overlap does not “snap back”; the UI continues showing a range the model did not accept. The surrounding `EditPadsCommand` also snapshots both sides and is pushed as a no-op undo entry, marking the project dirty. Please return the result through the callback and refresh the row on failure (and avoid committing a command when validation fails), or validate before creating the undo command.

magda/daw/ui/components/chain/drum_grid/DrumGridUI.hpp
```diff
@@ -237,6 +245,7 @@ class DrumGridUI : public juce::Component,
     };
 
     std::array<PadInfo, kTotalPads> padInfos_;
+    int faderGesture_ = 0;
```
lucaromagnoli: **P2 — Do not reuse gesture IDs when this UI is recreated**

This counter is local to one `DrumGridUI` and always restarts at 0. Closing/reopening the device UI, switching tracks, or otherwise rebuilding the component is not an undoable action, so the last fader command can remain at the top of the undo stack; the first drag in the new UI then reuses gesture 0 and `SetPadFaderCommand::canMergeWith()` folds the two separate sessions together. Use a process-wide monotonic gesture token (or explicitly terminate merging independently of component lifetime) so IDs cannot collide after UI reconstruction.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,75 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+                                   int highNote, int rootNote) {
+    auto* pads = getPads(gridPath);
+    if (pads == nullptr)
+        return false;
+
+    auto* pad = findPadChain(*pads, padIndex);
+    if (pad == nullptr)
+        return false;
+
+    const auto low = juce::jlimit(0, 127, std::min(lowNote, highNote));
```
lucaromagnoli, line 428: **P2 — Keep edited ranges reachable from the 64-pad UI**

The sliders and this setter accept the full MIDI range, but `DrumGridUI` creates rows only from notes `DrumGridPlugin::baseNote .. baseNote + maxPads - 1` (currently 24–87). If a user moves a chain wholly to, for example, 0–10, the model and engine keep the chain and it still answers to those notes, but no `padInfo` or chain row is created for it, so it disappears from the UI and can no longer be edited or deleted. Either require every range to intersect the visible grid, or build the chain list directly from all chains so out-of-window ranges remain manageable.

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -2971,7 +2971,11 @@ void PluginManager::syncDrumGridMultiOutTracks(const ChainNodePath& drumGridPath
     const auto trackId = drumGridPath.trackId;
     const auto deviceId = drumGridPath.getDeviceId();
     auto& tm = TrackManager::getInstance();
-    auto* devInfo = tm.getDevice(trackId, deviceId);
+
+    // By path, not by (track, device): the flat lookup misses a grid nested in
+    // a rack, which is a placement the model supports and the pad bus selector
+    // can now reach (#2211).
+    auto* devInfo = tm.getDeviceInChainByPath(drumGridPath);
```
lucaromagnoli, line 2978: **P1 — Route nested multi-out sources through the manager that owns them**

This lookup fixes discovery here, but `syncMultiOutTrack()` still gates child-track routing on the flat-only `TrackManager::getDevice(link.sourceTrackId, link.sourceDeviceId)` around line 1724, so a racked Drum Grid currently creates no output RackInstance. There is a deeper ownership issue too: devices inside MAGDA racks are loaded by `RackSyncManager`, whereas that function unconditionally calls `InstrumentRackManager::createOutputInstance()`; the nested grid has no entry in `InstrumentRackManager::wrapped_`, so merely replacing the model lookup would still return no routing instance. Please implement the output route through the owning rack machinery (and preserve enough path in the link to find it), or disable bus assignment for nested grids until that route exists.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -1214,95 +1247,166 @@ bool DeviceCustomUIManager::createDrumGridUI(const magda::DeviceInfo& device,
...
+                        magda::TrackManager::getInstance().setPadBypassed(grid, padIndex, bypassed);
+                    });
     };
 
-    drumGridUI_->onPadSoloChanged = [gridPath](int padIndex, bool soloed) {
-        magda::TrackManager::getInstance().setPadSolo(gridPath(), padIndex, soloed);
+    // The pad's output bus. `ChainInfo::outputIndex` is model state, and the
+    // device sync turns a pad on a bus into a multi-out child track, so the row
+    // selector only ever had to write the model (#2211).
+    drumGridUI_->onPadOutputChanged = [postPadEdit](int padIndex, int busIndex) {
```
lucaromagnoli: **P2 — Preflight or disable bus selection for nested grids**

`PadChainRowComponent` changes `currentBusOutput_` and its button text before invoking this callback. On a nested grid this still posts an `EditPadsCommand`; `setPadOutput()` then returns false without notifying, and its result is discarded here. The row therefore continues to show “B1” although the model stayed on Main, while the no-op snapshot command adds an undo entry and marks the project dirty. Mirror the range path by checking `padBusesAvailable()` before creating the command and refreshing on refusal, or disable/hide non-Main menu items when buses are unavailable.

magda/daw/ui/components/chain/drum_grid/DrumGridUI.cpp
```diff
@@ -261,6 +263,7 @@ DrumGridUI::DrumGridUI() {
         if (onPadLevelChanged)
             onPadLevelChanged(selectedPad_, static_cast<float>(value));
     };
+    levelSlider_.onDragEnd = [this]() { faderGesture_ = nextFaderGesture_.fetch_add(1); };
```
lucaromagnoli: **P2 — Advance the token for non-drag fader edits too**

The token changes only in `onDragEnd`. `TextSlider::commitText()` and `resetToDefaultValue()` call `onValueChanged` but never `onDragEnd`, so two distinct typed values (or a reset followed by a typed value) on the same pad/target keep the same token and merge into one undo step indefinitely. In each fader `onValueChanged`, advance the gesture after dispatch when `!slider.isBeingDragged()` (and do the same for chain-row sliders), or expose a completion callback that all direct-edit/reset paths invoke.

magda/daw/ui/components/chain/drum_grid/PadChainRangeRowComponent.cpp
```diff
@@ -36,9 +42,26 @@ PadChainRangeRowComponent::PadChainRangeRowComponent(int padIndex) : padIndex_(p
     setupNoteSlider(highNoteSlider_);
     setupNoteSlider(rootNoteSlider_);
```
lucaromagnoli: **P2 — Keep the root-note control at the full MIDI range**

The visibility problem applies only to `lowNote`/`highNote`, but this shared setup now restricts `rootNoteSlider_` to the grid window as well (currently 24–87). `rootNote` does not decide whether a row is reachable; `DrumGridPlugin::processChain()` uses it as the MIDI transposition target, and both model and engine still support 0–127. This prevents valid mappings such as triggering a sample whose natural/root pitch is above or below the displayed pads. Please bound only the two range endpoints to the grid and leave the root slider at 0–127.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,115 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
     }
 }
 
+bool TrackManager::padBusesAvailable(const ChainNodePath& gridPath) const {
+    // A multi-out child track is fed by the output instance
+    // InstrumentRackManager makes when it wraps a top-level instrument. A grid
+    // inside a MAGDA rack is loaded by RackSyncManager instead and has no entry
+    // there, so nothing would carry a bus off it: the pads on that bus would
+    // simply go silent. Refused until that route exists (#2211).
+    return gridPath.getType() == ChainNodeType::TopLevelDevice;
```
lucaromagnoli, line 370: **P1 — Normalize existing bus assignments when a grid becomes nested**

This predicate prevents a new non-Main assignment only after the grid is already nested. A user can assign Bus 1 while the grid is top-level and then call `wrapDeviceInRack()`; that move preserves `DeviceInfo::pads` and does not deactivate/reset its multi-out pairs, so the now-nested grid still has `outputIndex > 0` and its pads go silent through the unsupported route (with stale child tracks potentially left behind). Loaded projects can carry the same state. Please reject wrapping while buses are active, or atomically deactivate/remove the children and reset every pad to Main when the grid crosses into a placement where buses are unavailable; sync/load should sanitize it too.

magda/daw/ui/components/chain/drum_grid/DrumGridUI.cpp
```diff
@@ -1011,14 +1034,42 @@ void DrumGridUI::rebuildChainRows() {
 
         row->setSelected(i == selectedPad_);
```
lucaromagnoli: **P2 — Preserve chain selection when rebuilding ranged rows**

A row uses the lowest visible pad in its chain as `i`, but the selected pad may be any other note covered by that range. `setSelectedPad()` correctly selects by `chainIndex`; this rebuild resets it using index equality instead. After selecting (say) the second note of a ranged chain, the next `DeviceCustomUIManager::update()` rebuild leaves its only row unselected, so `resized()` no longer lays out the range editor. Initialize selection with the same chain-identity comparison used in `setSelectedPad()`, not `i == selectedPad_`.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,134 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+    if (pad == nullptr)
+        return false;
+
+    const auto low = juce::jlimit(0, 127, std::min(lowNote, highNote));
+    const auto high = juce::jlimit(0, 127, std::max(lowNote, highNote));
+
+    // Reachable. The grid shows kPadCount pads from kPadBaseNote, and its rows
+    // are built from those notes: a chain moved clear of them keeps playing and
+    // disappears from the UI, with no way left to edit or delete it.
+    if (high < kPadBaseNote || low > kPadBaseNote + kPadCount - 1)
```
lucaromagnoli: **P2 — Reject partially out-of-grid endpoints too**

This only rejects a range that is wholly outside the grid. For example, `0..kPadBaseNote` is accepted, but the endpoint sliders are restricted to the grid and therefore display it as `kPadBaseNote..kPadBaseNote`; additionally `padParameterSlot()` derives the slot from `lowNote` and returns -1, so the plan no longer binds this chain fader to the grid level/pan parameters. Either require both endpoints to be inside the displayed range, or support external endpoints consistently in the UI and parameter-slot mapping.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,134 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+    if (pads == nullptr)
+        return;
+
+    const auto found = std::ranges::find_if(
+        pads->chains, [padChainId](const ChainInfo& chain) { return chain.id == padChainId; });
+    if (found == pads->chains.end())
+        return;
+
+    const auto chainPath = padChainPath(gridPath, padChainId);
+    for (const auto& element : found->elements)
```
lucaromagnoli: **P2 — Clear selections below nested pad racks before erasing the chain**

This visits only direct devices. A pad chain may contain a rack (the recursive path support and tests explicitly allow that), and a selected rack or device below it is not equal to `chainPath`; `clearSelectionForDeletedChainNode` does not treat that path as an ancestor. Deleting the ranged chain can therefore leave the inspector holding a path to an erased nested node. Please recurse through rack elements, or make the selection cleanup clear descendants of the deleted chain.

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -2971,10 +2971,26 @@ void PluginManager::syncDrumGridMultiOutTracks(const ChainNodePath& drumGridPath
...
         return;
 
+    // A grid can arrive here in a placement where buses do not work: wrapping a
+    // top-level one in a rack moves it and keeps its pads, and a project can be
+    // loaded already like that. Nothing carries a bus off a nested grid, so a
+    // pad left on one would simply go silent, and any child track it had made
+    // would linger with nothing feeding it. Put the pads back on the grid's own
+    // mix and take the children down.
+    if (!tm.padBusesAvailable(drumGridPath)) {
+        tm.resetPadBuses(drumGridPath);
```
lucaromagnoli, line 2989: **P1 — Keep bus normalization inside the undoable edit**

This reset runs later from engine synchronization, outside `WrapChainElementsInRackCommand`. That command only moves the same device object back during undo, so wrapping a top-level grid with active buses resets every `outputIndex` and removes the child tracks, then Undo returns the grid to the top level with all of its routing permanently lost. Please make the normalization part of the wrap command state (or otherwise preserve and restore the assignments/pairs when undoing) so an undoable structural edit does not destroy routing configuration.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +334,134 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+    }
+
+    if (moved)
+        notifyTrackDevicesChanged(gridPath.trackId);
+
+    return moved;
+}
+
+bool TrackManager::setPadOutput(const ChainNodePath& gridPath, int padIndex, int outputIndex) {
+    if (outputIndex != 0 && !padBusesAvailable(gridPath))
```
lucaromagnoli, line 400: **P2 — Validate the bus index before storing it**

This accepts negative and arbitrarily large outputs. The live `DrumGridPlugin` clamps the value to `0..maxBusOutputs-1`, but `PlanCompiler` groups and routes pads using the raw model `outputIndex`; the two engines can therefore disagree, and the plan reports an unclaimed bus and drops the pad audio. Please reject or normalize invalid indices here, and sanitize deserialized pad outputs through the same rule.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -839,6 +898,12 @@ RackId TrackManager::wrapChainElementsInRack(const std::vector<ChainNodePath>& p
         if (!path.isValid() || getParentChainPathForElementPath(path) != sourceChainPath)
             return INVALID_RACK_ID;
 
+        // Same refusal as the single-device wrap: a pad on a bus would lose its
+        // routing to a clean-up the undo of this move does not cover (#2211).
+        if (const auto* device = getDeviceInChainByPath(path);
```
lucaromagnoli, line 1042: **P1 — Apply the same refusal to moves into existing racks**

These guards cover the two wrap APIs, but the normal drag/drop path into an existing rack uses `MoveChainElementsCommand -> moveChainElement()` and never reaches either one. A top-level grid with a pad on B1 can therefore still be dragged into a rack; the subsequent device sync resets its buses outside the move command, and Undo moves the already-reset grid back to the top level, reproducing the routing loss this change is meant to prevent. Please enforce the placement transition centrally in `moveChainElement` as well (while still allowing top-level reordering).

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1284,30 +1386,77 @@ void TrackManager::setDeviceKitRows(TrackId trackId, DeviceId deviceId,
...
+
+            if (!device.pads)
+                continue;
+
+            // A pad's address spells its rack step with the grid's own
+            // DeviceId: what padChainPath() builds, and what every link a
+            // project has saved to a pad device carries.
+            const auto gridPath = devicePathUnder(parentChain, device.id);
+            for (const auto& pad : device.pads->chains)
+                if (auto found = findDeviceUnder(
```
lucaromagnoli, line 1563: **P1 — Rekey pad-owned nodes when copying a Drum Grid**

The Ctrl-drag/copy path goes through `reassignCopiedElementIds()`, which allocates a new id for the outer grid but never descends into `device.pads` or restamps its synthetic rack id. The clone therefore keeps every pad-device id (and any nested pad rack/chain ids) from the source. With this new recursive lookup, `findDevicePath(padDeviceId)` returns whichever copy is visited first, and copied macro/mod targets still name the original pad path. Please include pad-owned nodes in copied-element id reassignment and remap their stored paths; this needs to recurse through racks inside pads too.

magda/daw/core/TrackManagerPads.cpp
```diff
@@ -312,6 +361,142 @@ void TrackManager::setPadBypassed(const ChainNodePath& gridPath, int padIndex, b
...
+
+    return moved;
+}
+
+bool TrackManager::setPadOutput(const ChainNodePath& gridPath, int padIndex, int outputIndex) {
+    // The bus has to be one that exists. The live plugin clamps what it is
+    // given and the plan compiler takes the model's value as it finds it, so an
+    // out-of-range index makes the two engines disagree: the plan reports a bus
+    // that reaches no track and the pads on it go silent (#2211).
+    if (outputIndex < 0 || outputIndex >= kPadBusCount)
```
lucaromagnoli, line 397: **P2 — Apply bus validation when loading model state too**

This protects only calls through the setter. `deserializeChainInfo()` still assigns `outputIndex` verbatim, so a top-level project containing -1 or `kPadBusCount` bypasses this check; sync does not call `resetPadBuses` for a top-level grid, leaving the live plugin clamped and the plan routing the invalid raw bus exactly as before. Please validate or normalize pad outputs during deserialization/load as well.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -613,6 +719,34 @@ static void reassignCopiedElementIds(TrackManager& tm, std::vector<ChainElement>
                 const auto oldId = device.id;
                 device.id = tm.allocateDeviceId();
```
lucaromagnoli, line 754: **P1 — Clear inherited multi-out pair ownership on copied grids**

A copied grid with a pad on B1 keeps both the pad `outputIndex` and the source device’s `multiOut.outputPairs[*].active/trackId`. After insertion, sync sees the cloned pair as already active and does not create a child track for the clone, while the inherited child track link still names the original device; the copied pad is therefore silent. `duplicateTrack()` explicitly clears pair activity after rekeying. Please do the equivalent for copied elements so sync can activate a fresh child for the new device.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -15,13 +16,49 @@ namespace magda {
...
+/// The same shape `remapDuplicatedPadPath()` handles for a duplicated track,
+/// and for the same reason: a pad device's address is
+/// `Rack(gridDeviceId) > Chain(pad) > Device(padDevice)`, and none of its steps
+/// may go through the ordinary maps. The rack step is a DeviceId, and rack ids
+/// come out of a counter that also starts at 1; the chain step is a pad chain
+/// id, rack-local and untouched by copying. Both ends are matched before either
+/// is rewritten, so a link merely passing through a rack numbered like a grid
+/// is left to the ordinary remap (#2211).
+bool remapPresetPadPath(ChainNodePath& path, const PresetIdRemap& remap) {
+    if (path.steps.size() != 3 || path.steps[0].type != ChainStepType::Rack ||
```
lucaromagnoli: **P1 — Remap paths to devices nested inside pad racks**

The recursive rekey now handles racks inside a pad, but this accepts only the three-step direct-device path. A target such as `Rack(grid) > Chain(pad) > Rack(inner) > Chain(inner) > Device(x)` falls through to the ordinary remapper: its owner prefix is not changed (or is changed through a colliding ordinary rack/chain id), so the copied link does not resolve to the cloned nested device. Please recognize the pad owner/pad-chain prefix for longer paths, rewrite that prefix with the pad remap, then apply the ordinary rack/chain/device maps only to the suffix. The duplicated-track helper has the same exact-size limitation.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -688,6 +822,18 @@ bool TrackManager::moveChainElement(const ChainNodePath& sourceElementPath,
         return false;
     }
 
+    // Refused for the same reason a wrap is: dropping a grid into a rack takes
+    // it somewhere no bus is carried, so the move would have to put every pad
+    // back on the main mix and take its child tracks down, and that clean-up is
+    // not part of the undoable step the move is. Undo would return the grid to
+    // the top level with its routing gone. Reordering within the track's own
+    // list is not a placement change and stays allowed (#2211).
+    if (destinationChainPath.getType() != ChainNodeType::Track) {
```
lucaromagnoli: **P1 — Treat cross-track moves as a routing placement change**

A destination can be track-level but belong to a different track. Dragging a top-level grid with B1 from track A to track B passes this guard and moves its active pair state, but the existing multi-out child still links to source track A; sync on B sees the pair already active and creates nothing, so the pad becomes silent. Allow same-track top-level reordering, but reject or migrate active-bus grids when `destinationChainPath.trackId != sourceElementPath.trackId` as well.

Issue #2221, opened 2026-08-27
Centralize chain and pad structural edits behind one mutation boundary


lucaromagnoli:
## Context

Move, copy, paste, wrap, duplicate-track, and pad edits currently perform their own mixture of traversal, placement validation, id allocation, link retargeting, mutation, and notification. PR #2213 showed that these implementations drift:

- a top-level cross-track move can bypass the nested-placement guard;
- copied subtrees can retain instance routing ownership;
- different recursive rekey/remap walkers cover different container shapes;
- invalid nested multi-out state is repaired later by engine synchronization instead of being rejected or normalized at the edit boundary.

Adding another guard to every command will keep reproducing this class of bug.

## Goal

Introduce one model-level structural mutation service used by all chain and pad transformations. An operation should preflight the complete source subtree and destination, construct the transformed value, and commit it atomically with one notification/undo boundary.

## Acceptance criteria

- Move, copy, paste, wrap, unwrap, device/rack duplication, track duplication, and pad structural edits use the same placement validation.
- Validation examines the complete subtree and is independent of whether the immediate destination is a track, rack chain, or pad chain.
- One canonical recursive clone operation allocates fresh ids and remaps every internal link; instance-owned state is reset by an explicit policy.
- A rejected operation leaves the serialized model unchanged and emits no partial notification.
- Undo/redo restores byte-equivalent serialized model state for structural operations.
- Engine synchronization reconciles runtime objects from valid model state; it does not silently mutate the project model to repair invalid placement.
- Transition-matrix tests cover active multi-out grids moved or copied among top-level tracks, rack chains, pad chains, and nested racks.

## Depends on

The typed pad paths (#2219) to validate against, and the single ownership record (#2220) to reset on clone. Do those first.

Shares its subject with #2204, split by direction: #2204 is the one shared way to READ the chain tree, this is the one shared way to WRITE it.

The pad ownership work this came out of is merged, so the command-specific placement and clone fixes it accumulated are in the tree now; this replaces them.

Issue #2220, opened 2026-08-27
Make Drum Grid multi-out ownership canonical and instance-local


lucaromagnoli:
## Context

Drum Grid multi-out ownership is represented twice:

- `DeviceInfo::multiOut.outputPairs` persists `active` and `trackId`;
- each generated child `TrackInfo` persists a `multiOutLink` back to the source device, track, and bus.

Those copies can disagree. More importantly, `active` and `trackId` are copied with `DeviceInfo`, even though they describe routing resources belonging to one placed device instance. PR #2213 demonstrated the result: copying a grid with an assigned B1 bus can leave the clone claiming the source grid child track, so the clone remains silent. Nested or moved grids can likewise retain ownership that is invalid at the new location, and engine sync currently repairs some of it by mutating the model.

The pad bus selection is declarative device content and may be copyable. Ownership of generated child tracks is not.

## Goal

Choose one authoritative representation for `(source device, output bus) -> child track` ownership and derive all secondary state from it. Prefer child `TrackInfo::multiOutLink` or a dedicated routing table; do not persist a second independently mutable owner in the device.

## Acceptance criteria

- Exactly one model record is authoritative for each multi-out child-track assignment.
- `active` and displayed assignment state are derived rather than independently synchronized.
- Copying or duplicating a Drum Grid preserves declarative pad bus choices but never inherits child-track ownership from the source instance.
- Moving a grid across tracks either transfers/rebuilds ownership atomically or is rejected before mutation according to one documented rule.
- Placing a multi-out grid where child-track routing is unsupported is rejected or normalized at the structural mutation boundary, not during engine sync.
- Deleting the source grid, deleting a child track, changing a bus, undo/redo, save/load, and duplicate-track operations cannot leave dangling or multiply-owned links.
- Loading legacy projects with both representations is deterministic and reports or repairs conflicts once.
- Tests exercise every lifecycle transition above against both the serialized model and the live engine plan.

## Depends on

The typed pad paths (#2219), which is what the ownership record is addressed and remapped through.

The pad ownership work this came out of is merged, so the local ownership repairs it carried are in the tree now; this removes them.

lucaromagnoli:
Done in PR #2224, merged into dev/0.20.0. Closing by hand: the PR merged into the release branch rather than the default one, so the closing keyword did not fire.

PR #2223, opened 2026-08-27
Give pad-owned chains explicit path step types

### LINKED ISSUE #2219: Give pad-owned chains explicit path step types
## Context

PR #2213 exposed an identity ambiguity in `ChainNodePath`. A pad device is addressed today as:

`Rack(gridDeviceId) > Chain(padChainId) > Device(padDeviceId)`

The `Rack` step normally contains a `RackId`, but here it contains the owning Drum Grid `DeviceId`. Generic resolvers and remappers cannot distinguish those meanings. They consequently depend on exact path shapes or contextual guesses. This has already produced broken duplication when a real rack and a grid share a numeric id, and broken remapping for pad-owned paths that contain nested racks.

This is the addressing half of #2204; that issue remains responsible for consolidating traversal.

## Goal

Represent pad ownership explicitly in paths. One compatible option is to append `PadRack` and `PadChain` step types to `ChainStepType`; an equivalent typed representation is fine. Existing serialized enum values must not be renumbered.

## Acceptance criteria

- A path identifies whether each owner id is a `RackId`, `DeviceId`, or pad-local `ChainId` without inspecting the project model or assuming a fixed number of steps.
- Resolve, find, serialization, link retargeting, copy, duplication, and preset import use the typed representation.
- Paths through racks nested inside pad chains resolve and remap recursively.
- Legacy serialized paths continue to load and are normalized or migrated deliberately.
- Tests cover colliding rack/device numeric ids, nested pad racks, copy/duplicate, master-track lookup, and serialization round trips.

## Depends on

Nothing. Implemented in PR #2223, and closes when that merges.

lucaromagnoli:
Closes #2219.

Based on `fix/2211-pad-ownership-followups` (#2213), which merges first. `padChainPath()` only exists on that branch, so this cannot sit on main until it lands.

## The problem

A pad device's address names its owning Drum Grid by that device's own id. Spelled as an ordinary `Rack > Chain` pair it was character for character how an allocated rack of the same number is spelled, and rack ids and device ids come out of counters that both start at 1. Every resolver and remapper therefore recognised pads by the shape of the path and by trying the routes in a fixed order.

## The change

`ChainStepType` gains `PadRack` and `PadChain`, appended so the persisted integer values of the existing steps stay stable. `padChainPath()` builds the typed form and the types decide from there:

- `getDeviceInChainByPath()` dispatches a pad-owned address straight to the pad route rather than walking a rack tree that cannot contain it.
- `getChainByPath()` gains the same dispatch through a new `getChainInPadByPath()`, so a pad chain is looked up through the generic call like any other chain.
- The shape-matching pad remappers in the track-duplication and preset-import paths are **removed rather than extended**. A `PadRack` step carries a DeviceId, so it moves with the ordinary devices map like any other device id; a `PadChain` step is rack-local and stays put. The pad-local id maps they needed (`DuplicatePadRemap`, `PresetPadRemap`) go with them.
- `isRackStep()` / `isChainStep()` cover the walkers that ask which rack encloses a node, so a pad path still answers them. `getRackIdAt()` stays strict, because a pad owner id is not a RackId, which turns the old silent collisions in the rack macro and modifier lookups into a clean miss.
- The remote API projects the two steps as `pad_rack` / `pad_chain`, so an external client is not left inferring ownership from position.

## Legacy projects

`pad_paths::migrateLegacyPadPaths()` retypes saved addresses on load, in its own unit beside `LegacyDeviceAliases` and `DeviceParamMigrations`. It keeps the old try-rack-then-pad order as its tie-break: a prefix a top-level rack of the track can account for is left alone, and only one a pad-owning device can account for is retyped. A project resolves exactly as it did before and leaves the ambiguity behind the next time it is saved. Until then, untyped addresses still resolve, tried after the ordinary rack route as they always were.

`fromVar` validates the pair structurally: `PadRack` only leads, `PadChain` only follows it, so the ambiguity cannot be readmitted through a hand-written file.

## Also fixed

`PasteChainElementsCommand` looked its destination chain up with `getChain(trackId, path.getRackId(), path.getChainId())` and dereferenced the result unchecked. That is a null dereference for any destination those two accessors cannot name, which now includes pad chains, and it was already wrong for a destination nested deeper than one rack. It goes through `getChainByPath()` with a guard.

## Tests

New `tests/test_pad_path_types.cpp`:

- total id collision: the rack's id equals the grid's DeviceId **and** the rack chain's id equals the pad chain's, so the two addresses are the same sequence of numbers and only the step types tell them apart. Both resolve to their own device.
- a rack nested inside a pad chain resolves, with the typed pair leading and an ordinary route as the tail.
- a pad address is answered only by the track it names, checked against the master track, which `getTrack()` returns a live track for.
- duplicating a track re-points a pad link at the copy's own pad device and leaves the original naming its own.
- serialization round trip, and rejection of misplaced pad steps.
- both migration directions: a legacy address gets retyped, and one an allocated rack can account for alone is left as it is.

Full Catch2 suite (3037 cases) and the JUCE target both pass locally.

## Follow-ups

#2220 (canonical multi-out ownership) and #2221 (one structural mutation boundary) land after this, in that order.

### DIFF
--- a/magda/daw/api/remote_api_projection.cpp
+++ b/magda/daw/api/remote_api_projection.cpp
@@ -450,6 +450,15 @@ DevicePathDto makeDevicePathDto(const ChainNodePath& path) {
             case ChainStepType::Device:
                 dto.steps.push_back({"device", step.id});
                 break;
+            case ChainStepType::PadRack:
+                // The id is the owning Drum Grid's DeviceId, not a RackId; the
+                // public step type says so rather than leaving a client to
+                // guess from position (#2219).
+                dto.steps.push_back({"pad_rack", step.id});
+                break;
+            case ChainStepType::PadChain:
+                dto.steps.push_back({"pad_chain", step.id});
+                break;
             case ChainStepType::Segment:
                 // Non-leading segments are not produced by any factory; drop
                 // rather than emit a step type the public contract lacks.
@@ -482,6 +491,10 @@ std::optional<ChainNodePath> toChainNodePath(const DevicePathDto& dto) {
             path.steps.push_back({ChainStepType::Chain, step.id});
         else if (step.type == "device")
             path.steps.push_back({ChainStepType::Device, step.id});
+        else if (step.type == "pad_rack")
+            path.steps.push_back({ChainStepType::PadRack, step.id});
+        else if (step.type == "pad_chain")
+            path.steps.push_back({ChainStepType::PadChain, step.id});
         else
             return std::nullopt;
     }
--- a/magda/daw/core/ChainNodePath.cpp
+++ b/magda/daw/core/ChainNodePath.cpp
@@ -88,7 +88,7 @@ bool fromVar(const juce::var& v, ChainNodePath& out) {
             if (!readRequiredInt(*stepObj, "type", rawType))
                 return false;
             if (rawType < static_cast<int>(ChainStepType::Rack) ||
-                rawType > static_cast<int>(ChainStepType::Segment))
+                rawType > static_cast<int>(ChainStepType::PadChain))
                 return false;
 
             ChainPathStep step;
@@ -105,6 +105,19 @@ bool fromVar(const juce::var& v, ChainNodePath& out) {
                     step.id > static_cast<int>(ChainSegment::MixerAnalysis))
                     return false;
             }
+
+            // The pad steps come as a leading pair. A pad path names its owning
+            // grid by DeviceId rather than by a route to it, so PadRack is always
+            // step 0 and PadChain always step 1; accepting either anywhere else
+            // would readmit exactly the ambiguity the types exist to remove.
+            if (step.type == ChainStepType::PadRack && i != 0)
+                return false;
+            if (step.type == ChainStepType::PadChain &&
+                (i != 1 || path.steps.front().type != ChainStepType::PadRack))
+                return false;
+            if (i == 1 && path.steps.front().type == ChainStepType::PadRack &&
+                step.type != ChainStepType::PadChain)
+                return false;
 
             path.steps.push_back(step);
         }
--- a/magda/daw/core/ChainNodePath.cpp
+++ b/magda/daw/core/ChainNodePath.cpp
@@ -123,6 +123,14 @@ bool fromVar(const juce::var& v, ChainNodePath& out) {
         }
     }
 
+    // A PadRack with nothing after it is not an address. The pair is the whole
+    // point: alone, the step names the pad rack, which nothing resolves, and
+    // getType() would report the path as an ordinary Rack whose id is a
+    // DeviceId, which is the ambiguity the types exist to remove. The in-loop
+    // checks cannot catch it, because there is no second step to check.
+    if (!path.steps.empty() && path.steps.back().type == ChainStepType::PadRack)
+        return false;
+
     if (obj->hasProperty("isTrackLevel")) {
         const auto flag = obj->getProperty("isTrackLevel");
         if (!flag.isBool())
--- a/magda/daw/core/PadPathMigration.cpp
+++ b/magda/daw/core/PadPathMigration.cpp
@@ -1,8 +1,7 @@
 #include "PadPathMigration.hpp"
 
 #include <functional>
-#include <map>
-#include <set>
+#include <vector>
 
 #include "AutomationInfo.hpp"
 #include "ChainNodePath.hpp"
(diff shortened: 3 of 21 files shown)


magda/daw/core/PadPathMigration.cpp
```diff
@@ -0,0 +1,103 @@
...
+            collectLinks(element.device.macros, element.device.mods, addressing.paths);
+    };
+
+    const auto retype = [](const TrackAddressing& addressing, ChainNodePath& path) {
+        if (path.steps.size() < 2 || path.steps[0].type != ChainStepType::Rack ||
+            path.steps[1].type != ChainStepType::Chain)
+            return;
+
+        const auto rack = addressing.topLevelRacks.find(path.steps[0].id);
+        if (rack != addressing.topLevelRacks.end() && rack->second.contains(path.steps[1].id))
```
lucaromagnoli: **Migration tie-break needs to resolve the complete ordinary route**

This currently preserves a legacy path whenever an allocated rack accounts for only the leading `Rack > Chain` pair. The old resolver gave the rack route priority only when the full route, including the leaf device or nested tail, resolved; otherwise it fell back to the pad route. If a rack and grid share the prefix but the leaf exists only on the pad, this leaves the path untyped. Track duplication or preset import then remaps its owner and chain through the ordinary rack maps and breaks the link. Please preserve the legacy spelling only when the complete ordinary path resolves, and add the colliding-prefix/pad-only-leaf case.

— Codex

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1320,6 +1257,29 @@ DeviceInfo* TrackManager::getDeviceInPadByPath(const ChainNodePath& devicePath)
     return nullptr;
 }
 
+ChainInfo* TrackManager::getChainInPadByPath(const ChainNodePath& chainPath) {
+    // `PadRack(gridDeviceId) > PadChain(pad)` and nothing after it: a pad chain
+    // is the whole of a pad-owned address, and anything deeper names something
+    // inside the chain rather than the chain (#2219).
+    if (chainPath.steps.size() != 2 || !chainPath.isPadOwned() ||
```
lucaromagnoli: **Continue through nested rack/chain steps after the pad prefix**

Restricting this resolver to exactly two steps means `getChainByPath()` rejects a valid nested chain such as `PadRack > PadChain > Rack > Chain`. `getElementContainerForChainPath()` delegates here, so copy, move, remove, wrap, insert, and paste operations cannot act on elements inside a rack nested beneath a pad even though device lookup can resolve the same typed route. Please resolve the root pad chain first and then walk the remaining ordinary `Rack > Chain` pairs.

— Codex

magda/daw/api/remote_api_projection.cpp
```diff
@@ -450,6 +450,15 @@ DevicePathDto makeDevicePathDto(const ChainNodePath& path) {
             case ChainStepType::Device:
                 dto.steps.push_back({"device", step.id});
                 break;
+            case ChainStepType::PadRack:
+                // The id is the owning Drum Grid's DeviceId, not a RackId; the
+                // public step type says so rather than leaving a client to
+                // guess from position (#2219).
+                dto.steps.push_back({"pad_rack", step.id});
```
lucaromagnoli, line 457: **Add the new step names to the published device-path schema**

`devicePathSchema()` still constrains `steps[].type` to `rack`, `chain`, or `device`. This projection therefore emits output that violates the API/MCP output schema, and input validation rejects the same `pad_rack`/`pad_chain` path when a client sends it back. Please extend the schema and the public DTO contract tests with both new names.

— Codex

magda/daw/core/PadPathMigration.cpp
lucaromagnoli: Fixed in the follow-up commit.

You are right that the resolver only gave the rack route priority when the whole route came out somewhere: `getDeviceInChainByPath()` walked the rack tree, searched the chain it landed in for the leaf, and fell through to `getDeviceInPadByPath()` when that search failed. Matching the leading pair alone reproduced the wrong half of that.

`resolvesOrdinary()` now walks the complete route recursively, so nested racks inside the tail are followed, and it handles a bare-Rack or Chain-terminated address as well as a Device leaf. An address is retyped only when the rack tree cannot answer it and a pad can.

Added the case you named: a rack and a grid sharing both leading numbers, the rack chain empty, the leaf on the pad. I checked it fails against the previous logic before keeping it.

lucaromagnoli (commented):
Follow-up review at `64cca27d`: the complete-route migration tie-break is fixed, and the new colliding-prefix/pad-only-leaf regression test passes. I found no new issue in that follow-up commit.

The nested pad-chain traversal and remote device-path schema threads remain open on the current head.

— Codex

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed.

`getChainInPadByPath()` now resolves the root pad chain and then walks the remaining ordinary `Rack > Chain` pairs, so `PadRack > PadChain > Rack > Chain` resolves and `getElementContainerForChainPath()` can name it. A trailing half pair is rejected rather than guessed at, and a rack id that names nothing under the pad still resolves to nothing.

Covered by a new case building a rack inside a pad chain and looking its chain up through the generic `getChainByPath()`. It fails against the two-step version.

magda/daw/api/remote_api_projection.cpp
lucaromagnoli: Fixed.

`devicePathSchema()` now admits `pad_rack` and `pad_chain`, so the projection no longer emits output that violates the published schema and a client can send the same address back.

The public DTO contract test covers both: a pad address and one continuing through a rack nested inside the pad chain, each round-tripped and then validated against the `devices.list` output schema, which embeds the device path schema. It fails against the previous enum.

magda/daw/core/ChainNodePath.cpp
```diff
@@ -106,6 +106,19 @@ bool fromVar(const juce::var& v, ChainNodePath& out) {
                     return false;
             }
 
+            // The pad steps come as a leading pair. A pad path names its owning
+            // grid by DeviceId rather than by a route to it, so PadRack is always
+            // step 0 and PadChain always step 1; accepting either anywhere else
+            // would readmit exactly the ambiguity the types exist to remove.
+            if (step.type == ChainStepType::PadRack && i != 0)
```
lucaromagnoli, line 113: **[P2] Reject a `PadRack` when the input ends before its `PadChain`**

The checks inside this loop reject a misplaced `PadRack` and a wrong second step, but they never run a second-step check when the array contains only `{PadRack, id}`. `fromVar()` therefore returns true for that input, and `getType()` classifies it as a valid Rack path even though the new invariant and the test comment require pad steps to be a leading pair. Please reject an unterminated `PadRack` after parsing the steps and cover the one-step input explicitly.

— Codex

lucaromagnoli (commented):
Re-reviewed at `34019a09c`. The nested pad-chain traversal and remote schema findings are fixed and their focused regression tests pass. I found one additional serialization validation gap and posted it inline.

— Codex

magda/daw/core/ChainNodePath.cpp
lucaromagnoli: Fixed.

You are right that the in-loop checks cannot catch it: with only one step there is no second step for them to reject, so `{PadRack, id}` parsed and `getType()` reported it as an ordinary Rack path whose id is really a DeviceId.

Rejected after the steps are parsed, where the absent second step is visible. Nothing resolves a bare pad rack anyway, so there is no reading under which the input was meaningful.

The one-step input is now an explicit case in the misplaced-steps test. I checked it fails without the new check.

lucaromagnoli (commented):
Re-reviewed at `3a8f479cb`. The lone-`PadRack` serialization gap is fixed and covered by the new one-step regression test. I found no additional actionable defects in this follow-up or the current PR diff.

Focused pad-path and remote-contract tests pass.

— Codex

PR #2224, opened 2026-08-28
Make Drum Grid multi-out ownership canonical and instance-local

### LINKED ISSUE #2220: Make Drum Grid multi-out ownership canonical and instance-local

lucaromagnoli:
Closes #2220.

## The problem

Which child track a multi-out pair drives was recorded twice:

- `active` and `trackId` on the device's `MultiOutOutputPair`;
- `multiOutLink` on the generated child track.

The two could disagree, and the device's copy travelled with `DeviceInfo`. A grid copied by duplication, preset import or paste arrived claiming the source's child track and stayed silent while the link still named the original. Engine sync repaired part of it by writing back onto the project model.

Ownership of a generated child track belongs to one placed instance of a device. The pair's bus selection is declarative device content and is copyable; the child track it drives is not.

## The change

The child track's `multiOutLink` is the only record.

- `MultiOutOutputPair` keeps `outputIndex`, `name`, `firstPin` and `numChannels`, and says nothing about where the pair goes. A copy has no ownership to inherit, so the resets duplication and preset import performed are gone with it.
- `multiOutChildTrack()` answers which track a pair drives; `multiOutPairIsActive()` whether it drives one. Both derive from the links, keyed by source track, source device and pair index, so two devices opening the same numbered bus cannot be confused.
- The UI menu, the Drum Grid pad bus reconciliation and the activate/deactivate mutators all ask rather than read a flag. The plan compiler already built its targets from the links, which is what confirmed the direction.

## Engine sync no longer edits the model

The sync used to write `active` and `trackId` back onto the device on the grounds that a project load rebuilt pairs before the tracks were restored. A derived answer cannot get that wrong, so the repair is deleted rather than kept.

## Legacy projects

`active` and `trackId` are no longer serialized, and both are dropped when reading a project that has them. The child tracks in the same project carry the links, so the assignments come back from those; a pair the device claimed with no child track to match was already an orphan and is now correctly closed. Deterministic in one pass, with no conflict to report.

## Moving a device that owns child tracks

Refused before any mutation, across the whole moved subtree, because the links name the track the device stands on and a device off the top level has no output instance to carry a bus. Closing the pairs first is the one documented way to move it.

This generalises the existing guard, which covered only a Drum Grid with a pad on a bus and so let a plain multi-out instrument move across tracks and strand its child.

## Tests

Six new cases in `tests/test_multi_out_tracks.cpp`, plus the copy case in the pad command tests rewritten to assert the property rather than the old reset:

- duplicating a track carries the declarative pair description and no ownership, with the original still driving its own child;
- deleting the child track directly ends the assignment and the pair can be reopened onto a fresh track. This is the case the old model got wrong: nothing cleared the source device's pair, so it kept `active` with a `trackId` naming a deleted track;
- two pairs of one device own two different tracks, and closing one leaves the other;
- two devices opening the same pair index own different tracks, and neither answers for the other;
- an unopened pair owns nothing, and out-of-range or invalid addresses are answered rather than guessed at;
- a device driving child tracks cannot change placement, is unchanged by the refusal, and moves once its pairs are closed.

3045 Catch2 cases and the JUCE target pass locally.

## Next

#2221, the structural mutation boundary, is the last of the three. It wants the single ownership record this adds, to reset on clone.

### DIFF
--- a/magda/daw/core/TrackManagerDevices.cpp
+++ b/magda/daw/core/TrackManagerDevices.cpp
@@ -32,6 +32,41 @@ template <typename Id> bool remapId(std::map<Id, Id> const& ids, int& value) {
     return true;
 }
 
+/// Whether anything in @p element drives a multi-out child track of @p sourceTrackId.
+///
+/// The whole subtree, because a rack being moved carries its devices with it.
+/// Ownership of a generated child track belongs to the device where it stands,
+/// and the child track's link names the track it stands on, so carrying the
+/// device somewhere else would leave the link naming a track that no longer
+/// hosts it (#2220).
+bool ownsMultiOutChildTracks(const TrackManager& tm, const ChainElement& element,
+                             TrackId sourceTrackId) {
+    if (magda::isDevice(element)) {
+        const auto& device = magda::getDevice(element);
+        if (device.multiOut.isMultiOut) {
+            for (std::size_t pair = 0; pair < device.multiOut.outputPairs.size(); ++pair)
+                if (tm.multiOutPairIsActive(sourceTrackId, device.id, static_cast<int>(pair)))
+                    return true;
+        }
+
+        if (device.pads)
+            for (const auto& pad : device.pads->chains)
+                for (const auto& padElement : pad.elements)
+                    if (ownsMultiOutChildTracks(tm, padElement, sourceTrackId))
+                        return true;
+        return false;
+    }
+
+    if (!magda::isRack(element))
+        return false;
+
+    for (const auto& chain : magda::getRack(element).chains)
+        for (const auto& nested : chain.elements)
+            if (ownsMultiOutChildTracks(tm, nested, sourceTrackId))
+                return true;
+    return false;
+}
+
 bool targetPointsAtDevice(const ControlTarget& target, DeviceId deviceId) {
     return deviceId != INVALID_DEVICE_ID && target.devicePath.getDeviceId() == deviceId;
 }
@@ -689,16 +724,6 @@ static void reassignCopiedElementIds(TrackManager& tm, std::vector<ChainElement>
                 device.id = tm.allocateDeviceId();
                 remap.devices[oldId] = device.id;
 
-                // The copy owns no child tracks yet. Inheriting the source's
-                // pair state would leave the sync thinking they were already
-                // made, so it would build none and the copy's pads on a bus
-                // would be silent while the link still named the original.
-                // `duplicateTrack()` clears them for the same reason.
-                for (auto& pair : device.multiOut.outputPairs) {
-                    pair.active = false;
-                    pair.trackId = INVALID_TRACK_ID;
-                }
-
                 // A pad-per-chain device's pads are its own: their DeviceIds and
                 // the rack id derived from the owner's would otherwise key the
                 // ops of the device this was copied from, and the copy's links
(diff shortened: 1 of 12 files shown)


magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -835,6 +858,21 @@ bool TrackManager::moveChainElement(const ChainNodePath& sourceElementPath,
...
+    // track's link names the track the device stands on, so moving it would
+    // leave that link pointing at a track no longer hosting it, and a device
+    // taken off the top level has no output instance to carry a bus at all.
+    // Refused here, before anything is mutated, rather than repaired later by
+    // engine sync: closing the pairs first is the one documented way to move a
+    // device that owns them (#2220).
+    //
+    // The whole subtree, because a rack carries its devices with it. Reordering
+    // within the track's own list is not a placement change and stays allowed.
+    const bool placementChanges = destinationChainPath.getType() != ChainNodeType::Track ||
```
lucaromagnoli: **[P2] Preserve in-container reordering for nested multi-out devices**

This classifies every non-track destination as a placement change. Consequently, if a multi-out device already lives inside a rack or pad chain and owns a child track, dragging it to another index in that same container returns `false`, even though neither `sourceTrackId` nor `deviceId`—the ownership key—changes. Nested multi-out devices are explicitly supported by `multiOutDevice()`, and ordinary nested elements can be reordered through this API, so this is a new UI/command regression. Please distinguish a true container/track change from `sourceElements == destinationElements` after resolving the containers, and cover reordering an active multi-out device within its existing nested chain.

— Codex

lucaromagnoli (commented):
Found one actionable P2 regression in the new move guard. The canonical child-track ownership change otherwise looks coherent across model queries, serialization, engine sync, UI callers, duplication, and preset/paste paths.

Focused verification passed: 93 assertions in `[multi_out]` and 272 assertions in `[drumgrid][pads][commands]`. The unfiltered local binary reached 3045 cases but is not clean in this workspace because fixture/environment-dependent tests cannot read their audio material, write the user cache, or find `stdfaust.lib`; GitHub CI is green.

— Codex

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed.

You are right that the ownership key does not change when a nested device moves to another index in the chain it already lives in, and `sameContainer` just below is the exact test I should have used: leaving the container is the placement change, reordering inside it is not, whatever the container is. The guard now runs off that instead of off the destination path shape.

Covered by reordering an active multi-out device within its own rack chain, checking the child track and its link are untouched, with the leaving case still refused in the same test. I checked it fails against the previous condition.

lucaromagnoli (commented):
Re-reviewed at `f5e095dec`. The nested in-container reorder regression is fixed correctly: the guard now keys off the resolved container identity, the active device can reorder without changing its ownership key, and leaving the container remains refused. The new regression test covers both sides. I found no additional actionable defects in the follow-up or current PR diff.

Focused verification passes: 109 assertions in `[multi_out]` and 272 assertions in `[drumgrid][pads][commands]`.

— Codex

PR #2225, opened 2026-08-28
Collapse four chain re-key walkers into one, and fix the two that skipped pads

lucaromagnoli:
First step of #2221, which stays open: one canonical clone before one preflight and one commit. Two live bugs fall out of it.

## What was there

Duplication, preset import, rack presets and chain presets each carried their own recursive re-key of a chain subtree. Four walkers, same job, allocating from the same three counters, and they had drifted:

| walker | descends into pads | records the synthetic pad rack id |
| --- | --- | --- |
| `duplicateTrack` | yes | yes |
| preset import (copy/paste) | yes | no |
| `applyRackPreset` | **no** | no |
| `applyChainPreset` | **no** | no |

The rack-preset one is commented "Mirrors the recursive walk in duplicateTrack". It does not.

## The two bugs

**A preset carrying a Drum Grid brought the preset's pad DeviceIds into the live project.** They collide with whatever already holds them, and the pad rack id stayed derived from the id the grid had in the preset rather than the one it has now, which is what every pad path is built from. A Drum Grid in a rack chain is ordinary, so both preset paths reach it.

**A link naming a pad rack survived a track duplication but not a preset import.** Only duplication recorded `padRackIdFor(old) -> padRackIdFor(new)`; the synthetic negative id a pad rack holds is what a stored link to one carries, and nothing else follows it.

## The change

`TrackManager::reassignChainElementIds()` is the one walk, and all four call it. It descends into a device's pads, records the synthetic rack id, and leaves pad chain ids alone because they are rack-local, which is what keeps a macro, a mod or an automation lane naming a pad still naming it.

`ChainIdRemap` carries what moved where. The two preset remap structs and the duplication one take their maps from it rather than filling their own: link retargeting stays per-caller, because who owns the links to follow afterwards genuinely differs (a track's macros, a rack's, a loose fragment's).

Nothing else moved. This is the id half; placement validation and the notification boundary are the next two steps on the issue.

## Tests

Two new cases in `tests/test_rack_audio.cpp`: a rack preset and a chain preset, each carrying a Drum Grid whose pad device claims the id of a device already live on the track. Both assert the grid, its pad device and the pad rack id all come back re-keyed.

I ran both against the previous walkers first and confirmed they fail there.

3048 Catch2 cases and the JUCE target pass locally.

## Still to come on #2221

- one preflight: placement validation is spread over five sites with four predicates (`canHostInstrument` at five call sites, `anyPadOnABus` at three, plus `chainPathContainsRack`, `elementContainsInstrument` and the multi-out ownership guard), and each command picks its own subset;
- one commit: preflight the whole subtree and destination, build the transformed value, apply and notify once with a single undo boundary.

### DIFF
--- a/magda/daw/core/ChainIdRemap.hpp
+++ b/magda/daw/core/ChainIdRemap.hpp
@@ -1,0 +1,28 @@
+#pragma once
+
+#include <map>
+
+#include "TypeIds.hpp"
+
+namespace magda {
+
+/**
+ * @brief What one re-keyed chain subtree's ids were moved to.
+ *
+ * Filled by `TrackManager::reassignChainElementIds()`, and read by everything
+ * that has to follow the subtree's internal references afterwards: macro and
+ * mod link targets, automation lane targets, and the addresses stored inside
+ * the subtree itself.
+ *
+ * `racks` carries two kinds of key. An allocated rack's id is one. The other is
+ * the synthetic negative id a pad rack holds (`padRackIdFor()`), which moves
+ * with the DeviceId it is derived from; a link naming a pad rack is remapped
+ * through this map like any other.
+ */
+struct ChainIdRemap {
+    std::map<DeviceId, DeviceId> devices;
+    std::map<RackId, RackId> racks;
+    std::map<ChainId, ChainId> chains;
+};
+
+}  // namespace magda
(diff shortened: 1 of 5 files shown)


lucaromagnoli (commented):
Reviewed at `8b5c277d`. The canonical re-key walk preserves the prior allocation semantics for duplication and copy/paste, fixes rack and chain presets by descending into pad-owned subtrees, stamps the synthetic pad-rack ID from the new owner, and leaves rack-local pad-chain IDs stable. The four callers still perform their context-specific link retargeting after the shared ID pass. I found no actionable defects.

Focused verification passes: 124 assertions in `[rack_audio]` and 272 assertions in `[drumgrid][pads][commands]`. The full PR diff also passes `git diff --check`.

— Codex

PR #2226, opened 2026-08-28
Ask one placement question instead of four subsets of it

lucaromagnoli:
Second step of #2221, which stays open for the third. Two more live bugs fall out of it.

## What was there

Move, paste and wrap each decided for themselves whether a subtree could go where it was being put, and each picked a different subset of the rules:

| operation | destination can host | pad on a bus | drives child tracks | rack inside itself |
| --- | --- | --- | --- | --- |
| `moveChainElement` | yes | the element only | yes | yes |
| `wrapDeviceInRack` | no | the element only | **no** | n/a |
| `wrapChainElementsInRack` | no | each element | **no** | n/a |
| `insertChainElementsByPath` | **no** | **no** | **no** | n/a |

## The two bugs

**Wrapping a device that drives multi-out child tracks was allowed.** Wrapping takes it off the top level, where the output instance that carries a bus is made, so it strands the child track exactly as a move would, and a move of the same device is refused. Wrap only ever asked about pads, which a plain multi-out instrument never trips.

**Pasting an instrument onto a track that cannot host one was allowed**, because `insertChainElementsByPath()` asked nothing.

## The change

`TrackManager::checkPlacement()` is the question. It takes what is being placed, where it is now, where it is going, and whether it leaves its container at all, and returns a typed `PlacementRefusal` rather than a bool so a caller can say why. Every rule examines the complete subtree, and none of them care whether the destination is a track's own list, a rack chain or a pad chain.

Move, paste, and both wrap paths call it. The rules themselves are unchanged apart from one widening: pad-on-a-bus is now subtree-wide, where the callers it replaces looked only at the element itself. That is not reachable today, because a grid inside a rack cannot put a pad on a bus in the first place, but the rule now says what it means rather than what happened to be enough.

Reordering inside the container a subtree already lives in is not a placement change and skips all of it, which is the distinction the multi-out guard learned in #2224.

## Tests

Three new cases: wrapping a device with an open multi-out pair, wrapping a selection containing one, and pasting an instrument onto an aux track. Each also checks the refusal left the model unchanged, and the first checks it wraps once the pair is closed.

I ran all three against the previous code first and confirmed they fail there.

3051 Catch2 cases and the JUCE target pass locally.

## Still to come on #2221

One commit boundary: preflight the whole subtree and destination, build the transformed value, then apply and notify once with a single undo boundary. That is the last of the three.

### DIFF
--- a/magda/daw/core/TrackManagerDevices.cpp
+++ b/magda/daw/core/TrackManagerDevices.cpp
@@ -830,11 +830,22 @@ PlacementRefusal TrackManager::checkPlacement(const PlacementRequest& request) c
     }
 
     // A pad on a bus is carried by the output instance made for a top-level
-    // instrument. Nowhere else carries one, so the move would have to put every
-    // pad back on the main mix and take its child tracks down, and that clean-up
-    // is not part of the undoable step (#2211).
-    if (anyPadOnABusInSubtree(*request.subtree))
-        return PlacementRefusal::PadOnABus;
+    // instrument. Nowhere else carries one, so a destination that is not a
+    // track's own list would silently drop the routing, and leaving the track
+    // the buses were realised on would leave its child tracks behind. Neither
+    // clean-up is part of the undoable step (#2211).
+    //
+    // A source-less reconstruction is neither: a paste, or an undo putting a
+    // deleted device back, owns no child tracks and leaves no track. Refusing it
+    // would have made undoing the deletion of a routed grid restore nothing at
+    // all, silently, because the callers discard the result (#2221).
+    if (anyPadOnABusInSubtree(*request.subtree)) {
+        const bool hasSource = request.sourcePath.trackId != INVALID_TRACK_ID;
+        const bool leavesItsTrack =
+            hasSource && request.destination.trackId != request.sourcePath.trackId;
+        if (!request.destinationIsTrackTopLevel || leavesItsTrack)
+            return PlacementRefusal::PadOnABus;
+    }
 
     // And the same for a multi-out pair: the child track's link names the track
     // the device stands on, so moving it strands the link, and a device off the
@@ -897,8 +908,8 @@ bool TrackManager::moveChainElement(const ChainNodePath& sourceElementPath,
     const bool sameContainer = sourceElements == destinationElements;
 
     // One question, asked before anything is mutated (#2221).
-    if (checkPlacement({&*sourceIt, sourceElementPath, destinationChainPath, !sameContainer}) !=
-        PlacementRefusal::Allowed)
+    if (checkPlacement({&*sourceIt, sourceElementPath, destinationChainPath, !sameContainer,
+                        destinationChainPath.steps.empty()}) != PlacementRefusal::Allowed)
         return false;
 
     const int sourceIndex = static_cast<int>(std::distance(sourceElements->begin(), sourceIt));
@@ -1004,7 +1015,9 @@ bool TrackManager::insertChainElementsByPath(const ChainNodePath& destinationCha
     // the rules keyed on where they came from pass and the ones about what is
     // being placed still apply (#2221).
     for (const auto& element : elements) {
-        if (checkPlacement({&element, {}, destinationChainPath, true}) != PlacementRefusal::Allowed)
+        if (checkPlacement(
+                {&element, {}, destinationChainPath, true, destinationChainPath.steps.empty()}) !=
+            PlacementRefusal::Allowed)
             return false;
     }
 
@@ -1039,8 +1052,8 @@ RackId TrackManager::wrapChainElementsInRack(const std::vector<ChainNodePath>& p
         // whether a pad was on a bus, so wrapping a multi-out instrument
         // stranded the child tracks a move of it refuses to strand (#2221).
         if (const auto* element = findChainElement(*this, path);
-            element != nullptr &&
-            checkPlacement({element, path, sourceChainPath, true}) != PlacementRefusal::Allowed)
+            element != nullptr && checkPlacement({element, path, sourceChainPath, true, false}) !=
+                                      PlacementRefusal::Allowed)
             return INVALID_RACK_ID;
 
         const int index = getChainElementIndex(path);
@@ -2068,7 +2081,8 @@ RackId TrackManager::wrapDeviceInRack(TrackId trackId, DeviceId deviceId,
     // on a bus, so wrapping a multi-out instrument stranded the child tracks a
     // move of it refuses to strand (#2211, #2221).
     if (checkPlacement({&*it, ChainNodePath::topLevelDevice(trackId, deviceId),
-                        ChainNodePath::trackLevel(trackId), true}) != PlacementRefusal::Allowed)
+                        ChainNodePath::trackLevel(trackId), true, false}) !=
+        PlacementRefusal::Allowed)
         return INVALID_RACK_ID;
 
     int insertIndex = static_cast<int>(std::distance(elements.begin(), it));
@@ -2121,7 +2135,7 @@ RackId TrackManager::wrapDeviceInRackByPath(const ChainNodePath& devicePath,
     // A nested device can drive multi-out child tracks too, and wrapping moves
     // it into a container it was not in, so this is a placement change like any
     // other and was going round the boundary (#2221).
-    if (checkPlacement({&*it, devicePath, chainPath, true}) != PlacementRefusal::Allowed)
+    if (checkPlacement({&*it, devicePath, chainPath, true, false}) != PlacementRefusal::Allowed)
         return INVALID_RACK_ID;
 
     int insertIndex = static_cast<int>(std::distance(elements.begin(), it));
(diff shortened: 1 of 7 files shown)


magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1992,12 +2062,13 @@ RackId TrackManager::wrapDeviceInRack(TrackId trackId, DeviceId deviceId,
...
-    // take its child tracks down -- and that clean-up would not be part of the
-    // undoable step the move is, so undoing it would return the device to the
-    // top level with its routing gone for good (#2211).
-    if (anyPadOnABus(magda::getDevice(*it)))
+    // Wrapping takes the device off the top level, where the output instance
+    // that carries a bus is made, so it is a placement change and asks the same
+    // question every other one does. This used to check only whether a pad was
+    // on a bus, so wrapping a multi-out instrument stranded the child tracks a
+    // move of it refuses to strand (#2211, #2221).
+    if (checkPlacement({&*it, ChainNodePath::topLevelDevice(trackId, deviceId),
```
lucaromagnoli, line 2117: **[P2] Apply the same preflight in the nested path-based wrap**

`wrapDeviceInRackByPath()` delegates its top-level case here, but the nested branch below erases and wraps the device without calling `checkPlacement()`. A nested multi-out device can own a child track (the existing multi-out tests construct exactly that state), so this public API still permits a container-changing wrap that `moveChainElement()` refuses and leaves the new centralized boundary incomplete. Please preflight `&*it` with `devicePath`, `chainPath`, and `leavesItsContainer = true` before extracting it, and cover the active nested multi-out case.

— Codex

lucaromagnoli (commented):
Found one actionable P2: the nested path-based wrap still bypasses the new placement boundary. The focused multi-out, drum-grid placement, and path-move suites pass locally.

— Codex

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed.

You are right that it was going round the boundary: the top-level branch delegates to `wrapDeviceInRack()`, which asks, but the nested branch extracted and wrapped the device itself. So a public API still allowed a container-changing wrap that `moveChainElement()` refuses, on a state the multi-out tests already build.

`checkPlacement({&*it, devicePath, chainPath, true})` now runs before the device is extracted, so a refusal leaves the chain as it was rather than half-wrapped.

Covered by wrapping a nested multi-out device with an open pair: refused, chain unchanged, ownership intact, and allowed once the pair is closed. I checked it fails without the call.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -752,6 +780,71 @@ static bool elementContainsInstrument(const ChainElement& element) {
...
+            return PlacementRefusal::DestinationCannotHostInstrument;
+    } else {
+        return PlacementRefusal::DestinationCannotHostInstrument;
+    }
+
+    // A pad on a bus is carried by the output instance made for a top-level
+    // instrument. Nowhere else carries one, so the move would have to put every
+    // pad back on the main mix and take its child tracks down, and that clean-up
+    // is not part of the undoable step (#2211).
+    if (anyPadOnABusInSubtree(*request.subtree))
```
lucaromagnoli: **[P1] Allow source-less routed grids to be reconstructed at track level**

This unconditional rule also runs for every `insertChainElementsByPath()` call, whose request has an empty source path. A copied Drum Grid retains each pad's `outputIndex`, so pasting one with a routed pad at a valid track-level destination now returns `false`. More critically, both device-removal commands use this same insert API with IDs preserved during `undo()`; deleting a top-level routed grid and undoing therefore silently fails to restore the device at all. A source-less reconstruction owns no old child tracks, and a top-level destination can carry its buses, so this case should be allowed while retaining the refusal for nested destinations/wraps and live ownership-changing moves. Please cover both routed-grid paste and delete/undo.

— Codex

lucaromagnoli (commented):
The previous nested-wrap P2 is fixed and covered. A second full pass found one new P1 in the generalized insert path: the unconditional pad-bus rule rejects valid top-level paste and undo reconstruction, so undoing deletion of a routed Drum Grid leaves it deleted. Focused placement suites pass, but they do not cover this source-less routed-grid case.

— Codex

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed, and thank you for catching the undo case: that one failed silently, because both removal commands discard the result of the insert.

The rule was about the act of placing when it should have been about whether the routing survives. A pad's bus is carried by the output instance made for a top-level instrument, so what matters is whether the destination is a track's own list, and whether the subtree is leaving the track its buses were realised on. A source-less reconstruction leaves no track and owns no child tracks, so only the first half applies to it.

`PlacementRequest` now says which case it is. `sourcePath` is empty for a reconstruction. `destinationIsTrackTopLevel` is the part a path cannot express: a wrap lands in a rack that does not exist yet, so it passes false while naming the chain that rack will live in, and stays refused.

Covered by deleting a routed grid and undoing it (restored, with `outputIndex` intact), and by pasting one into a rack chain (refused, nothing lands) and at track level (allowed, routing preserved on the copy). Both fail against the unconditional rule.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -752,6 +780,82 @@ static bool elementContainsInstrument(const ChainElement& element) {
...
+    //
+    // A source-less reconstruction is neither: a paste, or an undo putting a
+    // deleted device back, owns no child tracks and leaves no track. Refusing it
+    // would have made undoing the deletion of a routed grid restore nothing at
+    // all, silently, because the callers discard the result (#2221).
+    if (anyPadOnABusInSubtree(*request.subtree)) {
+        const bool hasSource = request.sourcePath.trackId != INVALID_TRACK_ID;
+        const bool leavesItsTrack =
+            hasSource && request.destination.trackId != request.sourcePath.trackId;
+        if (!request.destinationIsTrackTopLevel || leavesItsTrack)
```
lucaromagnoli, line 880: **[P2] Do not apply the root's top-level status to routed descendants**

`destinationIsTrackTopLevel` says where the subtree root lands, but `anyPadOnABusInSubtree()` deliberately searches every descendant. For a source-less rack containing a routed Drum Grid, inserting the rack in the track's own list passes this branch because the rack is top-level; the grid remains nested inside it, however, so it still has no output instance to carry that bus and sync will drop the routing. The same mismatch applies to deeper device/pad subtrees. Please make the exemption conditional on the routed grid itself being the root element (or evaluate each routed grid against its effective destination), and cover a pasted rack containing such a grid.

— Codex

lucaromagnoli (commented):
The routed-grid paste and deletion-undo regression is fixed and covered. One P2 remains: the new top-level flag describes only the subtree root, while the pad-bus query is recursive, so a routed grid nested inside a pasted rack is incorrectly treated as top-level. Focused multi-out, Drum Grid, path-move, and placement suites pass.

— Codex

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed.

You are right that the two did not line up: the flag describes the root, the query was recursive, so a rack carrying a routed grid passed on the rack being top-level while the grid inside it stayed nested.

Split in two rather than made conditional. Anything routed strictly below the root is refused outright, because it stays nested inside that root wherever the root goes and no destination can make it the top-level instrument its bus needs. The destination test then applies only to a routed root, which is the one element the caller's flag actually describes.

Covered by pasting a rack carrying a routed grid into the track's own list: refused, and nothing lands. I checked it fails against the recursive form.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -752,6 +807,89 @@ static bool elementContainsInstrument(const ChainElement& element) {
...
+    //
+    // A source-less reconstruction is neither: a paste, or an undo putting a
+    // deleted device back, owns no child tracks and leaves no track. Refusing it
+    // would have made undoing the deletion of a routed grid restore nothing at
+    // all, silently, because the callers discard the result (#2221).
+    //
+    // Only the root lands where the caller says it lands. A routed grid deeper
+    // in the subtree stays nested inside that root wherever it goes, so it can
+    // never be the top-level instrument its bus needs, whatever the destination.
+    if (anyPadOnABusBelowRoot(*request.subtree))
```
lucaromagnoli, line 873: Verified: separating routed descendants from the routed root aligns the recursive subtree check with the caller's root-level destination flag. The new rack-carrier regression test exercises the previously accepted shape, while source-less restoration of a routed root remains allowed. No further issue here.

— Codex

lucaromagnoli (commented):
No further findings. The routed-root/routed-descendant split closes the previous P2, all placement callers remain consistent with the request contract, the focused suites pass locally, and every CI job is green.

— Codex

PR #2228, opened 2026-08-28
Make redo of a paste and a wrap reproduce the model they made

lucaromagnoli:
Third step of #2221. The harness is as much the deliverable as the two fixes it found.

## The harness

`tests/test_structural_undo_roundtrip.cpp` snapshots every track through `serializeTrackInfo()`, runs one structural command through the undo stack, and requires that **undo restores the bytes it started from** and **redo reproduces the bytes it made**.

That is a sharper question than the tests we had. "The device is back" cannot tell a restored model from a merely plausible one: an id allocated on the way out, a link retargeted and not put back, a notification with no matching mutation are all invisible to it. Comparing the serialized track is not.

Seven operations covered: move within a list, move into a rack chain, cross-track move, paste, wrap, device removal, and removal of a Drum Grid with a routed pad.

## What it found

Five passed. Paste and wrap both failed, on the redo.

**Paste** re-copied `templateElements_` on every execute and inserted it with ids reassigned, so a redo gave each pasted device a fresh id.

**Wrap** called `wrapChainElementsInRack()` again, which allocated a fresh rack id and chain id.

In both cases undo followed by redo left the model holding elements that no earlier link, automation lane or alias could name. That is exactly the failure the removal commands already guard against, and say so in their comments: "undo would restore it under a different id and orphan every automation lane, macro link, and alias that targeted it." Paste and wrap had the same problem and no test that could see it.

## The fixes

Paste records what its first run actually produced, ids and all, and replays that with re-keying off. The template is still what a first execute copies.

Wrap passes the ids its first run allocated back into `wrapChainElementsInRack()`, which uses them when given them and allocates otherwise. Same shape as the removal commands preserving a DeviceId.

`serializeTrackInfo()` is now public, for the same reason `serializeClipInfo()` already is: it is a pure function of the value it is given, and the serialized form is the only way to state this property.

## Not changed, and why

Two of the issue's criteria were already met and I am not touching them:

- **Engine sync does not mutate the project model.** The only `TrackManager` write left in `PluginManagerSync` is `setEnabled` on a plugin, which is runtime state. The structural repair went with #2224.
- **`moveChainElement` emits two notifications**, but for two different tracks on a cross-track move. That is correct rather than partial; "notify once" means once per affected track.

3062 Catch2 cases and the JUCE target pass locally.

## What is left on #2221

The remaining criterion is the transition matrix: active multi-out grids moved or copied among top-level tracks, rack chains, pad chains and nested racks, as a systematic sweep rather than the individual cases the last three PRs added. Worth doing once this is in, and it may be worth closing the issue and filing that separately, since the three structural halves it asked for are now done.

### DIFF
--- a/magda/daw/core/TrackCommands.cpp
+++ b/magda/daw/core/TrackCommands.cpp
@@ -570,10 +570,18 @@ PasteChainElementsCommand::PasteChainElementsCommand(const ChainNodePath& destin
 
 void PasteChainElementsCommand::execute() {
     auto& tm = TrackManager::getInstance();
-    auto elements = deepCopyChainElements(templateElements_);
+    const bool replaying = !materialised_.empty();
+
+    // A redo replays what the first run produced, ids and all. Re-copying the
+    // template would re-key it, giving every pasted device a fresh id and
+    // orphaning the links, automation lanes and aliases made against the first
+    // paste -- the same reason the removal commands restore under the id the
+    // device had (#2221).
+    auto elements =
+        replaying ? deepCopyChainElements(materialised_) : deepCopyChainElements(templateElements_);
     const int requestedIndex = insertIndex_;
     executed_ = tm.insertChainElementsByPath(destinationChainPath_, std::move(elements),
-                                             requestedIndex, true);
+                                             requestedIndex, !replaying);
     insertedPaths_.clear();
     if (!executed_)
         return;
@@ -607,6 +615,9 @@ void PasteChainElementsCommand::execute() {
         } else if (isRack(element)) {
             insertedPaths_.push_back(destinationChainPath_.withRack(getRack(element).id));
         }
+
+        if (!replaying)
+            materialised_.push_back(deepCopyElement(element));
     }
 }
 
@@ -653,7 +664,11 @@ void WrapChainElementsInRackCommand::execute() {
         return;
     }
 
-    rackId_ = tm.wrapChainElementsInRack(sourceElementPaths_, rackName_);
+    // A redo reuses the ids the first run allocated, so undo followed by redo
+    // leaves the rack with the identity it had rather than a fresh one (#2221).
+    const auto newRackId =
+        tm.wrapChainElementsInRack(sourceElementPaths_, rackName_, rackId_, chainId_);
+    rackId_ = newRackId;
     executed_ = rackId_ != INVALID_RACK_ID;
 
     if (executed_) {
--- a/magda/daw/core/TrackCommands.cpp
+++ b/magda/daw/core/TrackCommands.cpp
@@ -1,5 +1,6 @@
 #include "TrackCommands.hpp"
 
+#include <algorithm>
 #include <limits>
 
 #include "../audio/AudioBridge.hpp"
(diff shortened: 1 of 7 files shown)


magda/daw/core/TrackCommands.cpp
```diff
@@ -607,6 +615,9 @@ void PasteChainElementsCommand::execute() {
         } else if (isRack(element)) {
             insertedPaths_.push_back(destinationChainPath_.withRack(getRack(element).id));
         }
+
+        if (!replaying)
```
lucaromagnoli, line 639: **[P1] Track the insertion position using the pre-insertion size**

`insertChainElementsByPath()` clamps `requestedIndex` against the destination size *before* inserting, but this code computes `start` against the larger post-insertion size. If a two-element paste targets index 2 in a one-element list, the model inserts at 1 while bookkeeping starts at 2, so it records only the second element; a larger stale index records none. Undo then leaves unrecorded pasted elements behind, and redo can paste another fresh copy because `materialised_` is empty/incomplete. I reproduced this with the new round-trip helper using two elements and index 99: undo left both pasted devices and redo duplicated them. Please capture the effective index against the pre-mutation destination size (or have the insert API return the actual inserted paths) and add the out-of-range round trip.

— Codex

tests/test_structural_undo_roundtrip.cpp
```diff
@@ -0,0 +1,185 @@
...
+        ChainNodePath::trackLevel(trackId), std::move(pasted), 1));
+}
+
+TEST_CASE("Undoing a wrap restores the serialized model", "[structural][undo][roundtrip]") {
+    resetState();
+    auto& tm = TrackManager::getInstance();
+
+    const auto trackId = tm.createTrack("Track");
+    const auto first = tm.addDeviceToTrack(trackId, effect("First"));
+    const auto second = tm.addDeviceToTrack(trackId, effect("Second"));
```
lucaromagnoli, line 163: **[P2] Cover the noncontiguous selection that wrap undo cannot restore**

This adjacent pair masks a current round-trip failure. The context menu accepts an arbitrary multi-selection, but `WrapChainElementsInRackCommand` remembers only the minimum `sourceIndex_`; undo reinserts every child contiguously from there. For `[First, Gap, Third]`, wrapping First + Third and undoing produces `[First, Third, Gap]`, changing the effect order and serialized model. I reproduced that directly with this helper. Please retain each selected element's original index (or otherwise restore the gaps) and make this case noncontiguous so the harness covers the public operation's full shape.

— Codex

lucaromagnoli (commented):
Found two reproducible undo round-trip gaps: one P1 in paste bookkeeping when the requested insertion index is clamped, and one P2 when wrapping a noncontiguous selection. Both fail through the new serialized round-trip helper; the committed focused suites pass.

— Codex

magda/daw/core/TrackCommands.cpp
lucaromagnoli: Fixed.

You are right about the mismatch: the insert clamps against the destination as it was, and the bookkeeping read the index back against the larger list, so it started past what had been inserted. The effective index is now taken before the insert, from the pre-mutation size, so the recorded run is exactly the inserted one.

Your out-of-range case is in the round-trip suite now, with two elements and index 99. It fails without the fix, both on the undo leaving them behind and on the redo duplicating them.

tests/test_structural_undo_roundtrip.cpp
lucaromagnoli: Fixed, and you are right that the adjacent pair was masking it.

The command now remembers each wrapped element's own index rather than only the lowest, and each goes back to it. The rack is still standing while the children move, occupying one slot, so a child whose home is past it aims one higher and lands right when the rack goes.

The wrap case in the harness is noncontiguous now: four devices, wrapping the first and third. It fails without the fix, with exactly the reordering you describe.

lucaromagnoli (commented):
No further findings. Both reported round-trip failures are fixed and covered: paste bookkeeping uses the pre-mutation clamp, and wrap undo restores each noncontiguous element to its own index. Focused suites pass locally; the long-running CI jobs are still in progress.

— Codex

magda/daw/core/TrackCommands.cpp
```diff
@@ -570,10 +571,36 @@ PasteChainElementsCommand::PasteChainElementsCommand(const ChainNodePath& destin
...
+        replaying ? deepCopyChainElements(materialised_) : deepCopyChainElements(templateElements_);
+    const int elementCount = static_cast<int>(elements.size());
+
+    // The index the insert will actually use, taken against the destination as
+    // it is NOW. `insertChainElementsByPath()` clamps against the pre-insertion
+    // size, so reading it back afterwards against the larger list would start
+    // the bookkeeping past what was inserted: a stale or out-of-range index
+    // recorded some of the pasted elements, or none, and undo then left them
+    // behind while redo pasted another copy (#2221).
+    const auto* destinationBefore =
```
lucaromagnoli, line 591: Verified: computing the effective index from the destination before mutation now matches `insertChainElementsByPath()` exactly. The two-element/index-99 regression covers both the prior incomplete undo and duplicate-on-redo behavior.

— Codex

magda/daw/core/TrackCommands.cpp
```diff
@@ -682,9 +723,23 @@ void WrapChainElementsInRackCommand::undo() {
...
-        tm.moveChainElement(childPath, sourceChainPath_, insertIndex++);
+    // Each child goes back to the index it came from, not to a run starting at
+    // the lowest one: a selection can have gaps, and closing them reorders the
+    // chain (#2221).
+    //
+    // The rack is still standing while they move, occupying one slot at
+    // `rackIndex`, so a child whose home is past it aims one higher and lands
+    // right when the rack goes. `rackIndex` rises as children are put in front
+    // of it.
+    int rackIndex = sourceIndex_;
```
lucaromagnoli, line 734: Verified: the sorted original indices align with the rack's sorted children, and adjusting targets around the rack's moving slot restores both contiguous and gapped selections. The noncontiguous round-trip now reproduces the original chain order.

— Codex

PR #2230, opened 2026-08-29
Sweep the structural matrix, and fix the five things it found (#2229)

### LINKED ISSUE #2229: Sweep the structural transition matrix with the serialized round-trip harness
## Context

The last of the acceptance criteria from #2221. That issue's three structural halves are done:

- one recursive clone (#2225), which found two preset paths that skipped a device's pads;
- one placement preflight (#2226), which found wrap stranding multi-out child tracks and paste accepting an instrument onto a track that cannot host one;
- one reversible commit (#2228), which found paste and wrap both re-allocating ids on redo, a paste at a clamped index losing its bookkeeping, and wrap reordering a noncontiguous selection on undo.

Each of those landed with the individual cases that exposed it. What is missing is the systematic version: the same operations run across every combination of container, rather than the combinations that happened to come up.

## The tool exists

`tests/test_structural_undo_roundtrip.cpp` (#2228) snapshots every track through `serializeTrackInfo()`, runs one structural command through the undo stack, and requires that undo restores the bytes it started from and redo reproduces the bytes it made.

That is the right oracle for this. "The device is back" cannot tell a restored model from a plausible one; an id allocated on the way out, a link retargeted and not put back, or an element reinserted at the wrong index are all invisible to it and all caught by comparing the serialized track. Every bug found in #2228 was found this way.

## Goal

Drive the harness from a matrix rather than from hand-written cases, so a container shape that nobody thought of is covered by construction.

The two axes:

- **Operation**: move, copy/paste, wrap, unwrap, device duplication, rack duplication, track duplication, remove, and the pad structural edits.
- **Container**: a track's own list, a rack chain, a pad chain, a rack nested in a rack chain, a rack nested in a pad chain, the post-fader list, and the mixer-analysis section.

Run each operation from each container to each other container, with the subject being in turn an ordinary effect, an instrument, a rack, a Drum Grid with a routed pad, and a multi-out device driving a child track. Refusals count as outcomes: a rejected operation must leave the serialized model byte-identical and emit no notification, which the harness can assert the same way.

## Acceptance criteria

- The matrix is generated rather than enumerated by hand, so adding a container shape or an operation extends the coverage without editing every case.
- Each cell asserts one of two outcomes: the operation is refused and the serialized model is unchanged, or it succeeds and undo/redo round-trips byte-identically.
- Combinations that are legitimately impossible are excluded explicitly, with the reason, rather than silently absent.
- The suite reports which cells it skipped, so a shape that stops being generated is visible rather than quietly dropping out.
- Anything it finds is fixed, or filed with the failing cell named.

## Notes

This is coverage rather than architecture, which is why it is separate from #2221 rather than holding it open.

Worth coordinating with #2204: that issue is the read side of the same tree, and a matrix that knows every container shape is close to being the list of shapes a shared traversal has to handle.

lucaromagnoli:
Closes #2229. The last acceptance criterion from #2221.

The hand-written round-trip cases each record a bug that came up. This adds the systematic other half: 996 generated cells over operation x container x subject, against the same oracle, so a container shape nobody thought of is covered by construction.

**Operations**: move, paste, wrap, device and rack duplication, removal, track duplication, and the seven pad structural edits.
**Containers**: a track's own list, a rack chain, a pad chain, a rack nested in either, the post-fader list and the mixer-analysis section, on the subject's own track, another media track and an aux track.
**Subjects**: an effect, an instrument, a rack, a Drum Grid with a routed pad, and a device driving a multi-out child track.

Each cell asserts one of two outcomes: refused, and then the serialized model is byte-identical and nothing was announced; or applied, and then undo restores the bytes it started from and redo reproduces the bytes it made. The 169 impossible combinations are excluded by name with the reason, grouped and reported, and every axis value has to survive into a cell that ran, so a shape that stops being generated fails rather than quietly dropping out.

## What it found

- **Undoing a move towards the front of a chain put the element back one slot early.** The index a move records is where the element stood, counted with itself in the list; `moveChainElement()` takes a drop position and drops one when the element travels up a list it is already in. The multi-element undo made that conversion, the single-element one did not. Now there is one conversion and both call it.
- **A rack pasted into a track's own list could not be undone.** The paste recorded it under `trackLevel(track).withRack(id)`, a path claiming to be both the track and a rack in it, which reads back as the track. Extending a path now clears `isTrackLevel`, so every caller gets a well-formed path.
- **Wrapping a device inside a pad chain could not be undone.** The wrap read its new rack's chain id back through the rack walk, which no pad address answers to. `getRackByPath()` now dispatches pad-owned addresses to the pad route, the way `getChainByPath()` already dispatched chains, so every path-based rack call works inside a pad rather than failing silently. This is the widest of the five: bypass, volume, macros and `addChainToRack` were all silently failing in there.
- **Redoing a track duplication made a different track.** It duplicated again, allocating a fresh TrackId and fresh device, rack and chain ids, orphaning every link made against the first duplicate. It now restores what the first run made, as a paste replays what it materialised.
- **Undoing a track deletion moved the track to the end of the project.** The restore appended; it now takes the index the track stood at.

Each has its own named case in `test_structural_undo_roundtrip.cpp`, because a sweep says a cell failed and a case says what broke.

## Left standing

No undoable command removes a rack: the chain view calls `removeRackFromChainByPath()` straight off the model, so deleting a rack cannot be undone at all. Named in the sweep's own exclusions rather than fixed here, because it needs a new command plus UI wiring and engine-teardown verification.

## Testing

Full Catch2 suite green (713,712 assertions, 3,063 cases) and `magda_juce_tests` green.

### DIFF
--- a/magda/daw/core/TrackCommands.cpp
+++ b/magda/daw/core/TrackCommands.cpp
@@ -202,7 +202,7 @@ void DeleteTrackCommand::execute() {
     // Store full track info, position and clips for undo (only on first execute)
     if (!executed_) {
         storedTrack_ = *track;
-        storedIndex_ = trackManager.getTrackIndex(trackId_);
+        storedPosition_ = trackManager.restorePositionOf(trackId_);
     }
 
     // Store and remove all clips on this track
@@ -228,7 +228,7 @@ void DeleteTrackCommand::undo() {
         return;
     }
 
-    TrackManager::getInstance().restoreTrack(storedTrack_, storedIndex_);
+    TrackManager::getInstance().restoreTrack(storedTrack_, storedPosition_);
 
     // Restore clips that were on this track
     auto& clipManager = ClipManager::getInstance();
@@ -260,7 +260,7 @@ void DuplicateTrackCommand::execute() {
     if (executed_) {
         if (duplicatedTrackId_ == INVALID_TRACK_ID)
             return;
-        trackManager.restoreTrack(storedTrack_, storedIndex_);
+        trackManager.restoreTrack(storedTrack_, storedPosition_);
         for (const auto& clip : storedClips_)
             clipManager.restoreClip(clip);
         return;
@@ -294,7 +294,7 @@ void DuplicateTrackCommand::execute() {
     // What it made, so a redo can make exactly that again.
     if (const auto* made = trackManager.getTrack(duplicatedTrackId_)) {
         storedTrack_ = *made;
-        storedIndex_ = trackManager.getTrackIndex(duplicatedTrackId_);
+        storedPosition_ = trackManager.restorePositionOf(duplicatedTrackId_);
         storedClips_.clear();
         for (auto clipId : clipManager.getClipsOnTrack(duplicatedTrackId_))
             if (const auto* clip = clipManager.getClip(clipId))
--- a/magda/daw/core/TrackCommands.cpp
+++ b/magda/daw/core/TrackCommands.cpp
@@ -92,6 +92,31 @@ bool describeChainElementPath(const ChainNodePath& path, ChainStepType& type, in
 
     return false;
 }
+
+/// The index to hand `moveChainElement()` so @p elementPath comes to rest at
+/// @p homeIndex of @p homeChain.
+///
+/// The two are not the same number. `homeIndex` is where the element stood,
+/// counted with itself still in the list. `moveChainElement()` takes a drop
+/// position -- insert before whatever stands there now -- and drops one when
+/// the element is travelling up a list it is already in, because removing it
+/// first shifts everything past it down. So an element on its way back up its
+/// own container has to aim one slot past its home to land on it.
+///
+/// One conversion, because it was written twice and only one of them was right:
+/// the multi-element undo made the correction, the single-element one passed the
+/// home index straight through, and so undoing a drag that moved a device
+/// towards the front of its chain put it back one slot too early
+/// (#2229). `WrapChainElementsInRackCommand::undo()` makes the same correction
+/// against the standing rack.
+int dropIndexForHome(TrackManager& tm, const ChainNodePath& elementPath,
+                     const ChainNodePath& homeChain, int homeIndex) {
+    if (parentChainOf(elementPath) != homeChain)
+        return homeIndex;
+
+    const int currentIndex = tm.getChainElementIndex(elementPath);
+    return currentIndex >= 0 && currentIndex < homeIndex ? homeIndex + 1 : homeIndex;
+}
 }  // namespace
 
 // ============================================================================
@@ -174,9 +199,10 @@ void DeleteTrackCommand::execute() {
         return;
     }
 
-    // Store full track info and clips for undo (only on first execute)
+    // Store full track info, position and clips for undo (only on first execute)
     if (!executed_) {
         storedTrack_ = *track;
+        storedIndex_ = trackManager.getTrackIndex(trackId_);
     }
 
     // Store and remove all clips on this track
@@ -202,7 +228,7 @@ void DeleteTrackCommand::undo() {
         return;
     }
 
-    TrackManager::getInstance().restoreTrack(storedTrack_);
+    TrackManager::getInstance().restoreTrack(storedTrack_, storedIndex_);
 
     // Restore clips that were on this track
     auto& clipManager = ClipManager::getInstance();
@@ -225,6 +251,20 @@ DuplicateTrackCommand::DuplicateTrackCommand(TrackId sourceTrackId, bool duplica
 
 void DuplicateTrackCommand::execute() {
     auto& trackManager = TrackManager::getInstance();
+    auto& clipManager = ClipManager::getInstance();
+
+    // A redo puts back what the first run made rather than duplicating again:
+    // duplicating again allocates a fresh TrackId and fresh device, rack and
+    // chain ids, so undo followed by redo would leave a different project than
+    // the one the undo took away (#2229).
+    if (executed_) {
+        if (duplicatedTrackId_ == INVALID_TRACK_ID)
+            return;
+        trackManager.restoreTrack(storedTrack_, storedIndex_);
+        for (const auto& clip : storedClips_)
+            clipManager.restoreClip(clip);
+        return;
+    }
 
     // Capture current plugin state so the duplicate gets the source's live settings.
     // Skipped when we're stripping the FX chain anyway — nothing to carry over.
@@ -239,7 +279,6 @@ void DuplicateTrackCommand::execute() {
     duplicatedTrackId_ = trackManager.duplicateTrack(sourceTrackId_, duplicateDevices_);
 
     if (duplicateContent_ && duplicatedTrackId_ != INVALID_TRACK_ID) {
-        auto& clipManager = ClipManager::getInstance();
         auto clipIds = clipManager.getClipsOnTrack(sourceTrackId_);
         const double projectBpm = ProjectManager::getInstance().getCurrentProjectInfo().tempo;
         const double bpm = isValidBpm(projectBpm) ? projectBpm : DEFAULT_BPM;
@@ -250,6 +289,16 @@ void DuplicateTrackCommand::execute() {
                                             bpm);
             }
         }
+    }
+
+    // What it made, so a redo can make exactly that again.
+    if (const auto* made = trackManager.getTrack(duplicatedTrackId_)) {
+        storedTrack_ = *made;
+        storedIndex_ = trackManager.getTrackIndex(duplicatedTrackId_);
+        storedClips_.clear();
+        for (auto clipId : clipManager.getClipsOnTrack(duplicatedTrackId_))
+            if (const auto* clip = clipManager.getClip(clipId))
+                storedClips_.push_back(*clip);
     }
 
     executed_ = true;
@@ -263,7 +312,7 @@ void DuplicateTrackCommand::undo() {
 
     // Delete all clips on the duplicated track before deleting the track
     auto& clipManager = ClipManager::getInstance();
-    auto clipIds = clipManager.getClipsOnTrack(duplicatedTrackId_);
+    const auto clipIds = clipManager.getClipsOnTrack(duplicatedTrackId_);
     for (auto clipId : clipIds) {
         clipManager.deleteClip(clipId);
     }
@@ -415,7 +464,9 @@ void MoveChainElementCommand::undo() {
     if (!executed_)
         return;
 
-    TrackManager::getInstance().moveChainElement(movedElementPath_, undoChainPath_, undoIndex_);
+    auto& tm = TrackManager::getInstance();
+    tm.moveChainElement(movedElementPath_, undoChainPath_,
+                        dropIndexForHome(tm, movedElementPath_, undoChainPath_, undoIndex_));
 }
 
 // ============================================================================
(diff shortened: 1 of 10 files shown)


tests/test_structural_matrix.cpp
```diff
@@ -0,0 +1,904 @@
...
+    INFO(cell);
+    auto& undo = UndoManager::getInstance();
+    undo.clearHistory();
+
+    NotificationCounter notifications;
+    const auto before = snapshot();
+    undo.executeCommand(std::move(command));
+    const auto after = snapshot();
+
+    if (after == before) {
```
lucaromagnoli: **P1 — Encode the expected disposition for each matrix cell**

This branch treats any byte-identical result as an acceptable refusal. If move, paste, wrap, or remove starts rejecting every legal input, all of those cells still pass because undo/redo remain no-ops; the ledger also counts attempted/refused cells toward axis coverage. Since this matrix is the main acceptance mechanism, please pass an expected outcome into the cell (or at least require application for the known-legal combinations) and reserve refusal for explicitly enumerated policy cases.

— Codex

magda/daw/core/TrackCommands.cpp
```diff
@@ -202,7 +228,7 @@ void DeleteTrackCommand::undo() {
         return;
     }
 
-    TrackManager::getInstance().restoreTrack(storedTrack_);
+    TrackManager::getInstance().restoreTrack(storedTrack_, storedIndex_);
```
lucaromagnoli: **P2 — Restore the child sibling position too**

`storedIndex_` only captures the flat `tracks_` position. For a grouped child B in `childIds = [A, B, C]`, deletion removes B from the parent and this restore appends it, producing `[A, C, B]` even though the flat insertion is now correct. Grouped ordering comes from `childIds`, so this still fails the structural round trip. Please store the parent sibling index and reinsert there, and cover a middle-child deletion in the new test.

— Codex

lucaromagnoli (commented):
Review outcome: changes needed before merge.

The five targeted production fixes are generally narrow and coherent, but two structural gaps remain:

- P1: the matrix accepts unexpected no-ops as valid refusals, so even a blanket regression that rejects every legal operation can leave the sweep green.
- P2: deleting and undoing a middle child track restores its flat index but appends it in the parent `childIds`, changing group order.

I posted both findings inline on the current head (`607bd1d`). CI is still running; `git diff --check` is clean. A local rebuild was started but not completed because this checkout required a broad 393-object rebuild, so I am not claiming a local test pass.

— Codex

magda/daw/core/TrackCommands.cpp
```diff
@@ -202,7 +228,7 @@ void DeleteTrackCommand::undo() {
         return;
     }
 
-    TrackManager::getInstance().restoreTrack(storedTrack_);
+    TrackManager::getInstance().restoreTrack(storedTrack_, storedPosition_);
```
lucaromagnoli: **P1 — Preserve descendants when undoing a group deletion**

`deleteTrack()` recursively deletes every child of a group, but `execute()` captures only the group’s `TrackInfo` and clips on the group itself. This undo therefore restores just the group: its `childIds` still name tracks that no longer exist, while every child track and its clips remain deleted. Deleting a non-empty group and pressing undo causes permanent data loss and leaves a dangling hierarchy. Please store and restore the complete subtree (including positions and clips), or explicitly reparent children if that is the intended deletion behavior, and add a group-deletion round-trip case.

— Codex

tests/test_structural_matrix.cpp
```diff
@@ -0,0 +1,1021 @@
...
+            if (const auto reason = cannotPlace(subject, source); reason.isNotEmpty()) {
+                ledger.exclude(cellName("move", subject, source), reason);
+                continue;
+            }
+
+            for (auto destination : kContainers) {
+                for (auto role : kRoles) {
+                    const auto cell = cellName("move", subject, source, destination, role);
+
+                    auto world = buildWorld();
```
lucaromagnoli: **P1 — Keep the matrix inside the Windows CI budget**

The Windows build succeeds, but its `Run tests` step now hits the hard 15-minute timeout. The matrix output alone spans roughly 06:10:37–06:16:34 in that log, and the job is terminated at 06:20:50; rebuilding this full three-track world for every cell is a material part of the overrun. This leaves a required platform check red. Please reduce/reuse the per-cell fixture, shard or tag the matrix into a separately budgeted job, or extend the Windows timeout before merge.

— Codex

lucaromagnoli (commented):
Second-pass outcome: changes still needed before merge.

The new commit correctly addresses both earlier findings: each matrix cell now declares its expected disposition independently, and track restoration preserves the grouped sibling index. The Linux JUCE/full test jobs and macOS job pass, and `git diff --check` remains clean.

Two new blockers remain and are posted inline:

- P1: undoing deletion of a non-empty group restores only the group, not its recursively deleted child tracks or clips, leaving data loss and dangling `childIds`.
- P1: the new matrix pushes the Windows `Run tests` step beyond its hard 15-minute timeout; the build succeeds, but the required Windows check is red.

Reviewed current head `d1354d67b`.

— Codex

magda/daw/core/TrackCommands.cpp
```diff
@@ -174,43 +234,38 @@ void DeleteTrackCommand::execute() {
...
-    auto& clipManager = ClipManager::getInstance();
-    auto clipIds = clipManager.getClipsOnTrack(trackId_);
-    storedClips_.clear();
-    for (auto clipId : clipIds) {
-        const auto* clip = clipManager.getClip(clipId);
-        if (clip) {
-            storedClips_.push_back(*clip);
-        }
-        clipManager.deleteClip(clipId);
+        storedTracks_ = collectSubtree(trackId_);
```
lucaromagnoli, line 240: **P1 — Restore routing cleaned up outside the deleted subtree**

`collectSubtree()` captures only the root and descendants, but `deleteTrack()` also mutates every surviving track: it erases sends targeting each deleted track, clears audio/MIDI inputs listening to it, and clears device/rack sidechains referencing it (including on master). This undo restores none of those external changes. For example, if track A sends to child B, deleting B’s group and undoing leaves A’s send permanently gone even though the subtree returns. Please capture and restore the affected external routing state (or have deletion produce a cleanup record), and cover at least an external send/input/sidechain round trip.

— Codex

lucaromagnoli (commented):
Third-pass outcome: one blocker remains.

The latest commit correctly restores deleted group descendants (including their clips and positions), adds the missing group round-trip, and materially reduces the matrix fixture cost. The previous Windows-timeout fix is structurally sensible; the new platform jobs are still running, while code-quality and the release/security gates are green. `git diff --check` is clean.

I posted one signed P1 inline: track deletion also removes sends, track-input routes, and sidechains from surviving tracks outside the captured subtree, and undo does not restore those external mutations. The project therefore still does not round-trip when anything routes to the deleted root or one of its descendants.

Reviewed current head `9d0d889c5`.

— Codex

lucaromagnoli (commented):
Follow-up status: the PR head has not changed since the prior review (`9d0d889c5`), so the outstanding P1 remains unresolved.

All CI checks are now green, including Windows, confirming that the matrix-fixture optimization fixed the timeout. The remaining issue is still the signed inline finding already posted: undoing deletion restores the deleted subtree but not sends, track-input routes, or sidechains that `deleteTrack()` removed from surviving tracks outside that subtree.

I did not duplicate the inline comment because it still applies to the same code and commit.

— Codex

tests/test_structural_matrix.cpp
lucaromagnoli: Fixed in d1354d6. Each cell now declares its expected outcome through `expectedOutcome()`, with the rules restated on the test side rather than asked of `checkPlacement()`, which would agree by construction. Applied cells assert `before != after` before the round trip.

Verified by making `moveChainElement()` refuse everything: the move matrix goes red where it used to pass.

magda/daw/core/TrackCommands.cpp
lucaromagnoli: Fixed in d1354d6. `restoreTrack()` takes a `TrackRestorePosition` carrying both the project index and the sibling index, and both the deletion and the duplicate redo record where the track stood in each. Covered by a middle-child case.

Verified by making the sibling insert append again: the case goes red.

magda/daw/core/TrackCommands.cpp
lucaromagnoli: Fixed in 9d0d889. `DeleteTrackCommand` records the whole subtree, each track with both its indices and its clips, and restores them by ascending track index so the list refills left to right. A child restored before its parent needs no parent to rejoin: the parent own stored `childIds` still names it.

Verified by making the collection stop at the named track: the new group case goes red.

tests/test_structural_matrix.cpp
lucaromagnoli: Addressed in 9d0d889 and 2b59304.

The waste came out first: a cell was standing up all seven container shapes on three tracks while addressing one or two, and the oracle serializes every track four times per cell, so that was multiplied by four thousand. A cell now builds only the shapes it addresses on only the tracks it reaches. The aux-destination cells for a subject carrying no instrument are excluded by name as well, since an aux track differs from another media track in exactly one thing this sweep tests. 61s to 11.8s locally, and the Windows step went from timing out to 13m38s.

That left 1m22s of margin, which is the wrong place to leave it, so the guard is raised to thirty in 2b59304. The step comment already says a timeout is here to catch a hang and not to enforce a performance budget, and records that five minutes had stopped doing that job. Fifteen went the same way: the suite reached 11:32 on dev before this branch, and Debug MSVC runs it about ten times slower than a Debug clang build, so this is the step every addition meets first. Trimming further would have meant deleting real cells to fit a mis-calibrated guard.

magda/daw/core/TrackCommands.cpp
lucaromagnoli: Fixed in 9bfdc97. `TrackManager` gains `externalRoutingInto()` and `restoreExternalRouting()`. The first is asked before the deletion and records the whole of each surviving track sends and inputs rather than a difference, because the deletion only ever removes from them; the second replays it after the subtree is back, so a send, an input or a sidechain naming a restored track names something that exists again. Sidechains go back through their own setters so the engine follows.

Covered by a case carrying a send, both input routes and a device sidechain through one round trip. Verified by dropping the restore call: it goes red.

magda/daw/core/TrackCommands.cpp
```diff
@@ -160,6 +186,40 @@ juce::String CreateTrackCommand::getDescription() const {
...
+        record.position = tm.restorePositionOf(id);
+
+        auto& clipManager = ClipManager::getInstance();
+        for (auto clipId : clipManager.getClipsOnTrack(id))
+            if (const auto* clip = clipManager.getClip(clipId))
+                record.clips.push_back(*clip);
+
+        records.push_back(std::move(record));
+
+        for (auto childId : track->childIds)
```
lucaromagnoli, line 209: **P1 — Include owned multi-out tracks in the deletion snapshot**

This walk follows only hierarchy `childIds`, but active multi-out tracks deliberately are not hierarchy children: `activateMultiOutPair()` gives them a `multiOutLink` and leaves `parentId`/`childIds` alone. `deleteTrack()` nevertheless removes them via `deactivateAllMultiOutPairs()` before recursing through `childIds`. Undo therefore restores a deleted multi-out instrument (or routed Drum Grid) without the output tracks that carried its active pairs, changing the model and routing permanently. Please collect/restore tracks whose `multiOutLink.sourceTrackId` belongs to the doomed subtree as well, and add a deletion round trip with an activated output pair.

— Codex

lucaromagnoli (commented):
Fourth-pass outcome: one blocker remains.

The latest changes correctly restore external sends, track-input routes, and sidechains after undoing a deletion. The sidechain traversal preserves full rack paths, device IDs are restored through the engine-aware setters, and the targeted round-trip covers the central external-routing case. The Windows timeout is now a 30-minute hang guard; the new CI run is still in progress. `git diff --check` is clean.

I posted one signed P1 inline: `collectSubtree()` still omits active multi-out tracks because those tracks are linked through `multiOutLink`, not `childIds`, while `deleteTrack()` erases them. Undo restores the source instrument/grid without its active output-pair tracks.

Reviewed current head `2b5930420`.

— Codex

PR #2235, opened 2026-08-30
Render real projects through both engines, and the six things it found (#2081)

### LINKED ISSUE #2081: Validation slice 7: the real-project corpus

lucaromagnoli:
Closes #2081.

Seven projects saved by released builds between 0.4.8 and 0.17.0 are loaded from the bytes they were written as, rendered through both engines, and compared by the same runner the code-built corpus uses. The code-built corpus (#2040) tests the rules whoever wrote it knew about. This is where the combinations live: a v1 project whose clips carry their sources inline, a five-device chain under an FM synth, two tracks coupled through a sidechain rather than through routing, sixty-four session clips beside an arrangement, a warp map made by a real transient pass with a repeated marker in it.

The runner moves out of the corpus test into `NullDiffRunner` so both corpora judge a case the same way. A second copy of the tiers, the two-silences guard and the diagnostics contract is the thing this harness exists to refuse.

`MgdFixtures.cpp` says which seven and why the other fifteen are out: half host a plugin (#2175), seven contain 4OSC, two render silence, one has no instrument on its only track, one stops on a parameter gap.

**Cost**, which #2081 left open until it was known: twenty seconds for the seven, half of it the demo project alone. It stays in the ordinary job. The figure is printed on every run rather than recorded in a comment, so the day it stops being true the run says so.

## What it found

Six defects, each fixed here:

- **`PrefetchStream` read past the end of a chunk.** It computed what was left as `length - position`. A looping reader reports int64's maximum for its length and a position can be negative -- an event anchored before its own loop start, a reversed clip trimmed longer than its source -- so the subtraction wrapped to a large negative that truncated back to a count larger than a chunk: 8414 samples read into a buffer of 4096. Saturating now, with the regression test asserting the bound rather than the number.
- **A `te::Plugin` outlived its Edit and terminated the process.** `PluginManager::drumGridChainsChanged` posted an async holding a `Plugin::Ptr`. A plugin's teardown locks the Edit's parameter change handler inside a `noexcept` function, so a message delivered after the project closed called `std::terminate` rather than crashing somewhere a stack trace explains. It resolves the device path up front and carries nothing that can keep the plugin alive.
- **A Faust device whose saved source fails to compile deleted the project's sidechain routing.** `activeDspMatchesSource` was left true after a failed restore, so the guard that clears a stale audio sidechain fired on a device that had never loaded. Two tests already asserted this and could not see it, because the fallback patch did not compile in the test binary either.
- **The incumbent leg told the Edit it was rendering offline and never told the plugin layer**, while the native leg builds every device with `offlineRender` set.
- **`prepareForRendering` un-bypasses every tone generator and nothing put them back**, so they stayed audible under a stopped transport until the next play or stop. A tone playing out of a stopped project after a bounce.
- **The test binaries opened the machine's real audio output.** Inaudible while nothing put a sound source in a live Edit; the demo project's four tone generators ended that. The device is closed after initialise. Not `MAGDA_HEADLESS`, which would also take down the plugin window manager a test correctly asserts is there.

## What the corpus needed

- The incumbent leg baked automation onto the corpus's own gain device and converted with the identity. It resolves any target through the app's own `ControlTargetResolver` now and converts through `makeParameterValueConverter`, lifted out of `AutomationPlaybackEngine` so the playback path and the corpus share one conversion.
- The native leg's arrangement lane takes arrangement clips, the same rule `ClipSynchronizer` already applies at the same point. The demo project has sixty-four session clips and reported one diagnostic for each.
- The Faust standard library is staged beside both test binaries, and `getFaustLibrariesPath` falls back to the executable's own directory on macOS when there is no bundle. Without it a Faust device silently passed audio through and the case measured two chains that were not running it.

## The finding worth its own issue

Four real projects go onto the block-size gate's list (#2078). They are the first projects in that gate containing a device MAGDA ships -- the code-built corpus has only the four devices written for it -- and each differs between 512 and the other sizes by 2 to 9 dB. No mechanism is claimed for any of them.

The gate now asks first whether a project renders the same twice at one block size, so a dependence on process history can never be read as a dependence on block size. Every project does, so this is the block size.

## Verification

Both suites green: `magda_tests` 3089 passed, `magda_juce_tests` all passed. The code-built corpus holds its same nine calibrating cases and nothing else moved.


magda/engine/io/PrefetchStream.cpp
```diff
@@ -246,8 +247,25 @@ bool PrefetchStream::fill() {
...
+        // length is int64's maximum (SourceReaders.cpp), and a position can be
+        // negative: an event anchored before its own loop start is anchored at
+        // a phase within it, and a reversed clip trimmed longer than its source
+        // begins ahead of the first sample. Subtracting a negative position
+        // from that maximum wraps to a large negative, whose low thirty-two
+        // bits truncate back to a positive count larger than a chunk -- 8414
+        // samples into a chunk of 4096, which is a read past the end of the
+        // buffer rather than a wrong number. A real project found it: a loop
+        // whose region starts after the clip does.
+        const auto headroom = std::numeric_limits<std::int64_t>::max() + fillPosition_;
```
lucaromagnoli: **P2 — Avoid overflowing `headroom` for ordinary positive positions**

This initializer is evaluated before the condition below. After a normal stream fills its first chunk, `fillPosition_` is positive, so the next iteration evaluates `INT64_MAX + fillPosition_` and invokes signed-overflow undefined behavior. The `fillPosition_ < 0` check does not guard the addition because it happens afterward. Please compute `headroom` only inside the negative-position branch (and extend the regression test across zero; its four chunks currently remain negative).

lucaromagnoli: Fixed in 1f97e3c9d. You are right, and it is the same overflow moved one line up: the sum was computed unconditionally and the sign test sat in the ternary's condition, which runs afterwards and guards nothing.

Branched now, so the sign is established before the arithmetic that depends on it. A non-negative position takes the plain subtraction, where neither operand can make it wrap; the saturating form is reached only where INT64_MAX + fillPosition_ is representable.

The regression test plays across zero rather than up to it, and asserts the bound every round rather than once at the end. Four chunks of read-ahead from -8415 never reach a non-negative position, so the ordinary path was the one branch it did not cover, which is where this lived. Re-checked against the unfixed subtraction: still 8414 into a chunk of 256.

tests/test_prefetch_stream.cpp
```diff
@@ -41,9 +42,20 @@ class CountingReader final : public AudioFileReader {
...
 
-        const auto available =
-            static_cast<int>(std::clamp<std::int64_t>(length_ - startSample, 0, numSamples));
+        // What the file has from here, saturating rather than wrapping. A
+        // reader with no end reports int64's maximum for its length, and it is
+        // asked for positions ahead of its first sample -- LoopingAudioFileReader
+        // answers those by phase and pads what is in front of them -- so the
+        // plain subtraction overflows for exactly the pair this file is here to
+        // stand in for.
+        const auto headroom = std::numeric_limits<std::int64_t>::max() + startSample;
```
lucaromagnoli: **P2 — Avoid repeating the signed overflow in the test double**

The expanded regression now deliberately crosses zero, so `CountingReader::read()` is eventually called with a positive `startSample`. This initializer then evaluates `INT64_MAX + startSample` before the condition below can check the sign, invoking the same signed-overflow undefined behavior that the production fix just removed. Please use the same guarded branching here; otherwise the regression itself has undefined behavior on the newly covered path.

lucaromagnoli: Fixed in 4d9526c4c. Correct, and it is the same mistake in the code written to catch it: the double carries the saturating subtraction because it stands in for a reader with no end, and it formed int64's maximum plus the position before establishing the sign.

It did not matter while the case only played negative positions. Widening it across zero is exactly what made the double reach that path, so the regression would have had undefined behaviour on the ground it was widened to cover.

Branched the same way. Both remaining uses of that sum, here and in PrefetchStream::fill, now sit inside a branch that has already established the position is negative; grep for the addition finds nothing outside one.

Issue #2240, opened 2026-08-30
The native engine has no denormal handling: no flush-to-zero, no output sanitising


lucaromagnoli:
Part of #1882. Found writing the external plugin adapter (#1893), where the fork's own denormal
handling is the one line of `ExternalPlugin::processPluginBlock` the adapter does not reproduce.

The native engine has no denormal handling anywhere. There is no `juce::ScopedNoDenormals` around
a render, and nothing sanitises what an op wrote. The fork has both halves:

- Every plugin's output is passed through `zeroDenormalisedValuesIfNeeded`, which on Intel checks
  whether the CPU raised its denormal flag during the call and, if it did, runs `sanitiseValues`
  over the buffer: NaNs and infinities cleared, values clamped to a magnitude of 100, denormals
  flushed. Two of its own devices assert the flag was not raised at all.
- The audio callback itself runs under flush-to-zero, so a device that feeds its own output back
  through a filter decays to zero rather than into the denormal range where every multiply costs a
  microcode trap.

What this is worth is bounded and worth stating rather than assuming. The check is `#if JUCE_INTEL`
in the fork, so on Apple silicon it is a no-op and there is nothing to diverge from today. It is
Intel and Windows where it matters, and it matters twice: as a CPU figure (#2082 measures dropout
rate, and a reverb tail running in denormals is where that shows up) and as a divergence, since a
plugin that leaves a NaN in the buffer is clamped by the fork and passed on by the engine.

The two halves are separable and the second is not obviously right to copy. Flush-to-zero around
the callback is what every host does and costs nothing. Clamping a plugin's output to a magnitude
of 100 is a policy, and the argument for it is that a plugin which produced an infinity has already
failed and the alternative is a burst of full-scale noise into the speakers -- but it is the fork's
policy, and adopting it should be a decision rather than a transcription.

Whichever way that goes, the engine says what it does: the parity envelope covers what the incumbent
delivers by default (#1896), so a divergence here needs its mechanism written down the way the
render divergences do.

lucaromagnoli:
Fixed by #2443.

Issue #1895, opened 2026-07-27
Engine: live input, monitoring, recording


lucaromagnoli:
Part of #1882. Depends on the engine core.

Live input, monitoring, and recording.

- Input device instances with monitor modes (in, auto, off) and per-track arming.
- Audio and MIDI recording with loop-record takes; session-slot recording with quantized start.
- Latency-aligned capture, count-in interaction, live MIDI preview path.
- This subsystem plus the launcher are where verification is interactive rather than null-diff; budget hands-on QA here.

lucaromagnoli:
Scope note from breaking down #1896: this issue carries its own final parity slice, the way
#2040 was slice 7 of #1890.

Cases for what is ported here are written here, not in #1896. A parity case is worth what its
author's knowledge of the feature is worth, and that knowledge is at its peak on the day the
port lands. Written later by someone reading the code back, a case asserts the code rather than
the behaviour, which is the one thing a parity corpus must not do.

#1896 keeps the cross-cutting infrastructure the slice builds on: the whole-project case and the
tiered oracle (#2075), the goldens (#2076), the block-size gate (#2078), and the corpora that
span features (#2081, #2082).

lucaromagnoli:
Broken into eight slices, in dependency order. 1 and 2 are independent of each other; 3 needs both; 4 sits beside 3; 5 needs 3 and 4; 6 needs the launcher, which is done; 7 needs 3; 8 is the parity slice this issue carries per the scope note above.

- #2459 slice 1: a live input reaching an input op
- #2460 slice 2: the capture queue and the writer thread
- #2461 slice 3: an arrangement audio take, from armed to clip
- #2462 slice 4: MIDI recording, takes, and the notes a loop leaves open
- #2463 slice 5: the recording pass reported without a poll
- #2464 slice 6: session slot recording and the arrangement capture
- #2465 slice 7: a take that survives a plan swap
- #2466 slice 8: parity cases for recording

Three things found while cutting them up, which is why the slices are shaped this way:

**Arming is already structural and already compiled.** `PlanCompiler` emits `AudioInput` and `MidiInput` ops for a track whose input route is External, tags them `LivenessDomain::Live`, and leaves them out for a track that is neither armed nor monitoring; `PlanBindings` has `audioInputs` and `midiInputs` waiting. So none of this is plan shape. What is missing is that `EngineSession::process` takes an output buffer and nothing else, so the callback's input has no way in.

[...]

Issue #1894, opened 2026-07-27
Engine: clip launcher and session playback


lucaromagnoli:
Part of #1882. Depends on the engine core. Deliberately late in the sequence: it is the most from-scratch subsystem and the least null-testable.

Clip launcher and session playback.

- Launch-handle state machine: play/stop/queued states, quantized launch and stop against monotonic beats, played-range reporting for playheads.
- Follow actions (including engine-initiated launches being reflected back into app state without an adoption dance).
- Per-track session/arranger switching with de-click.
- Handle lifetime via the plan-swap epoch reclamation; no hand-rolled deferred-release protocol.
- Session slot recording integration point (capture handled in the recording issue).

lucaromagnoli:
Scope note from breaking down #1896: this issue carries its own final parity slice, the way
#2040 was slice 7 of #1890.

Cases for what is ported here are written here, not in #1896. A parity case is worth what its
author's knowledge of the feature is worth, and that knowledge is at its peak on the day the
port lands. Written later by someone reading the code back, a case asserts the code rather than
the behaviour, which is the one thing a parity corpus must not do.

#1896 keeps the cross-cutting infrastructure the slice builds on: the whole-project case and the
tiered oracle (#2075), the goldens (#2076), the block-size gate (#2078), and the corpora that
span features (#2081, #2082).

lucaromagnoli:
Sliced.

- [x] #2300 the launch handle, and what a block straddling a launch point means
- [x] #2301 session sources in the plan, and a slot's voice across a swap
- [x] #2302 per-track session and arranger switching with de-click
- [x] #2303 played ranges reported without a poll
- [x] #2304 follow actions, and an engine-initiated launch reaching the app
- [x] #2305 handle lifetime on the plan-swap epoch
- [x] #2306 behaviour cases for the launcher

Order is dependency order rather than priority order, with one exception: #2305 is numbered sixth and wants doing early, because moving handles onto the epoch after every call site holds a raw pointer to one is a retrofit rather than a decision.

Slice 1 landed in #2307. Two things came out of it that the later slices inherit: the transport now carries a monotonic beat on `BlockInfo`, which is the domain a queued position is named in, and the rule that anything quantized lands on its beat while a loop wrap gives way.

PR #2249, opened 2026-08-30
Host an external plugin under a Device op, and find the one a project meant (#2241, #2243)

### LINKED ISSUE #2241: External plugins slice 1: a plugin under a Device op
Slice 1 of #1893.

The adapter, and only the adapter: one `juce::AudioPluginInstance` behind one Device op, doing
with a block what `te::ExternalPlugin` does with the same block.

A plugin is the one thing in a project neither engine wrote and neither can ask anything of, so
what the two hosts do around it is the whole of the difference between them. That makes this slice
a transcription rather than a design, and each piece of it is worth naming because each is a
divergence in every project that hosts a plugin if it is got wrong:

- Channel adaptation at `max(max(1, inputs), outputs)`, with the fork's bridging rules on either
  side of it: a mono block into a stereo input duplicates, a stereo block into a mono input averages,
  a mono output is spread back across a stereo slot.
- The wet/dry pair the fork gives every external plugin and the plugin never declared. MAGDA
  persists both (`DeviceInfo::wrapperParameters`), so a project can carry a plugin at 40% wet.
- Parameter identity: the fork's automatable list is dry, wet, then the plugin's automatable
  parameters in plugin order, and those positions are what a project saved. The mapping is rebuilt
  from the live instance rather than from the model, because that is where the fork's comes from.
- Latency reported into the executor's PDC.
- A playhead, because a tempo-synced delay that is told nothing renders a different project.
- Sidechain audio as the channels after the plugin's own bus.
- `setNonRealtime` for a bounce, which the fork sets and clears around a render.

**Not in this slice, and each for its own reason.** Instantiation, which is a message-thread job with
a format manager behind it and an async path the plan has to have a state for. State restore, which
is the three-step contract (#1893's own list) and wants a first-class API rather than the fork's
internal headers. Hardware inserts, which are a graph-level op pair rather than a device. The parity
cases, which are this issue's final slice and are written the day the port lands rather than later.

Tested against a plugin written in the test rather than one that happens to be installed. What is
being asserted is what the host does -- which channels a plugin is handed, which slot a saved value
lands on, what the wrapper pair means -- and a stub that reports exactly what it was given answers
those better than a real plugin does. What a real plugin is for is the corpus (#2175).

### LINKED ISSUE #2243: External plugins slice 2: instantiation and one rule for which plugin a project meant
Slice 2 of #1893. Depends on #2241, which is the adapter the instance is handed to.

Where the instance comes from. A Device op naming a VST3 or an AU has to resolve to a
`juce::AudioPluginInstance` before the adapter has anything to bind, and today nothing on the
engine side can make one: `createEngineDevice` asks the two internal catalogs and returns null for
everything else.

**Resolution is the part with a rule in it, and the rule already exists twice.** A saved device
carries a name, a manufacturer, a file path and a unique id, and matching those against what the
machine actually has installed is four passes in `PluginManagerSync`: file path plus instrument
flag, name plus manufacturer plus instrument flag, file path alone, then name plus format alone for
a project that came from another host and carries no path of ours. Those passes are not
incidental; each was added for a project that would otherwise have loaded silence or loaded the
wrong plugin. The engine cannot have a second, simpler opinion about which plugin a project meant,
so this lifts them out of the sync path into something both engines call, the way #2235 lifted the
parameter conversion.

**Creation, then, is short**: `AudioPluginFormatManager::createPluginInstance` for a render that is
waiting for it and `createPluginInstanceAsync` for a session that is not, `enableAllBuses` the way
the fork does at the same point, and a failure that is reported rather than swallowed.

The plan already behaves correctly while an instance is missing: an unbound Device op passes its
audio through and says so in the diagnostics, which is what the fork does with a plugin that has not
finished loading. What this slice owes is the other end of it -- a binding that appears after the
plan was prepared is a re-prepare, and the host has to be told that is what completion means.

Not in this slice: the saved state chunk, which is #1893's three-step contract and is its own slice.
A plugin created here runs at its own defaults, which is exactly what a project with no saved state
gets.

lucaromagnoli:
Closes #2241. Closes #2243.

The first two slices of external plugin hosting (#1893): a plugin runs under a Device op, and the
engine can find and load one.

## The adapter (#2241)

`EngineExternalDevice` is the third device adapter and the one with no MAGDA device behind it.
EngineMagdaDevice hosts something this repository wrote and can ask anything of; this hosts a binary
somebody else shipped whose only contract is `juce::AudioPluginInstance`. So it is a transcription
of `te::ExternalPlugin` rather than a design, and each piece is written out beside the fork's own
reason for it, because a difference here is a difference in every project that hosts a plugin:

- **Channel adaptation** at `max(max(1, inputs), outputs)`, whatever the chain around it is, with
  the fork's bridging rules on either side: a mono block into a stereo input duplicates, a stereo
  block into a mono input averages at half gain, a mono output is spread back over a stereo slot so
  the next device has two channels to read.
- **The wet/dry pair** the fork gives every external plugin and the plugin never declared. MAGDA
  persists both (`DeviceInfo::wrapperParameters`), so a project can carry a plugin at 40% wet and
  the two engines have to agree what that means. Fully dry still runs the plugin, which is what
  makes a wet-level automation lane usable at all.
- **Parameter identity.** The fork's automatable list is dry, wet, then the plugin's automatable
  parameters in plugin order, and those positions are what a project saved as
  `ParameterInfo::paramIndex`. The mapping is rebuilt from the live instance rather than from the
  model, because that is where the fork's comes from: a project saved against another build of the
  plugin would otherwise write its values onto whatever the parameters are called now. A parameter
  is written only when the value differs from what the plugin reports, which is also the fork's
  rule -- a plugin is entitled to treat every write as a gesture.
- **Latency** read once after prepareToPlay and reported into the executor's PDC.
- **A playhead**, because a tempo-synced delay that is told nothing renders a different project.
  Position, bar start, signature and bpm come off the block's own tempo map.
- **Sidechain audio** as the channels after the plugin's own bus, and `setNonRealtime` for a bounce,
  which the fork sets and clears around a render.

Three differences are named in the header rather than left to be found: the all-notes-off flag
`juce::MidiBuffer` has nowhere to carry (the gap EngineMagdaDevice already declares), the transport
state a device under the plan is not handed (#1894, #1895), and denormal handling (#2240, below).

## Instantiation, and one answer to which plugin a project meant (#2243)

The search for the installed plugin a saved device names was written twice in `PluginManagerSync`,
and the two copies had drifted. The track path tried the file path on its own after the vendor pass;
the rack path did not. **A plugin renamed in place by its vendor therefore resolved on a track and
not inside a rack, in the same project, on the same machine.** They are one function now
(`ExternalPluginLookup.hpp`) with the four passes in one place and their order tested, because the
order is the whole content of the search: every pass matches things the later ones also match, so a
test that only asked whether something was found would pass with them in any order and let the wrong
plugin load.

The lifted version also declines to match on an empty file path, which both copies would have done
had their callers not guarded the whole search. An empty path equalling an empty path is a match on
the absence of evidence; the guard belongs to the two passes that need it rather than to the call
site, so the other two still answer for a device that saved no file. A dead debug loop that walked
every scanned plugin on every load, with an empty body, goes with them.

Loading is the engine's half: a blocking create for a render that is waiting for the answer, an
async one for a session that is not, `enableAllBuses` at the point the fork does it, and a failure
that comes back as a sentence rather than a null. **Refusing a plugin the scan does not know is
deliberate and differs from the app**, which tries the saved description anyway and lets the format's
own lookup have a go. That is right where a user is watching and can be told, and wrong for a
render: a plugin resolved by a route nothing recorded is a difference no corpus can attribute
afterwards.

## The finding worth its own issue

**The native engine has no denormal handling anywhere** (#2240). No `ScopedNoDenormals` around a
render, and nothing sanitising what an op wrote, where the fork has both halves: flush-to-zero
around the callback, and every plugin's output passed through `zeroDenormalisedValuesIfNeeded`. It
is `#if JUCE_INTEL` in the fork, so there is nothing to diverge from on Apple silicon today, and it
is an executor-wide decision rather than this adapter's -- the second half in particular (clamping a
plugin's output to a magnitude of 100) is a policy worth deciding rather than transcribing.

## Testing

Against a plugin written in the test rather than one that happens to be installed. What is asserted
is what the host does with a plugin -- which channels it hands it, which slot a saved value lands
on, what the wrapper pair means -- and a stub that reports exactly what it was given answers those
better than a real plugin does. The stub is deliberately awkward in the ways real plugins are: a
non-automatable parameter in the middle of its list, mono and stereo buses, a sidechain, latency it
reports. A real plugin under both engines is the corpus's job (#2175).

The lookup's tests show each pass resolving the case it was added for and not resolving the case the
pass before it owns.

## Verification

`magda_tests` 3114 passed, 13 skipped. `magda_juce_tests` all passed. `make debug` clean.

### DIFF
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
@@ -41,31 +41,39 @@ ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
                                          const juce::KnownPluginList& knownPlugins) {
     const auto saved = describeSavedPlugin(device);
 
-    // The four passes, as predicates over one list. Written this way rather
-    // than as four loops because what matters about them is the order, and the
-    // order is invisible when it is four hundred lines of the same loop.
+    // The passes, as predicates over one list. Written this way rather than as
+    // five loops because what matters about them is the order, and the order is
+    // invisible when it is four hundred lines of the same loop.
     //
-    // The two file passes ask nothing when the device saved no file, because an
-    // empty path matches every description that also has none and that is a
-    // match on the absence of evidence. The app guarded this at its call sites
-    // by declining to search at all; the guard belongs to the pass that needs
-    // it, so the other two still answer.
+    // A pass asks nothing when the device saved nothing for it to ask with. An
+    // empty path matches every description that also has none, and an empty
+    // name matches every plugin with no name: both are matches on the absence
+    // of evidence, and they would resolve a device that saved neither to
+    // whichever scanned plugin happens to be missing the same field.
     const auto hasFile = device.fileOrIdentifier.isNotEmpty();
+    const auto hasName = device.name.isNotEmpty();
+    const auto hasIdentity = device.uniqueId.isNotEmpty();
 
-    const std::array<std::function<bool(const juce::PluginDescription&)>, 4> passes{
+    const std::array<std::function<bool(const juce::PluginDescription&)>, 5> passes{
+        [&](const juce::PluginDescription& known) {
+            return hasIdentity && known.createIdentifierString() == device.uniqueId;
+        },
         [&](const juce::PluginDescription& known) {
             return hasFile && known.fileOrIdentifier == device.fileOrIdentifier &&
                    known.isInstrument == device.isInstrument;
         },
         [&](const juce::PluginDescription& known) {
-            return known.name == device.name && known.manufacturerName == device.manufacturer &&
+            return hasName && known.name == device.name &&
+                   known.manufacturerName == device.manufacturer &&
+                   known.pluginFormatName == saved.pluginFormatName &&
                    known.isInstrument == device.isInstrument;
         },
         [&](const juce::PluginDescription& known) {
             return hasFile && known.fileOrIdentifier == device.fileOrIdentifier;
         },
         [&](const juce::PluginDescription& known) {
-            return known.name == device.name && known.pluginFormatName == saved.pluginFormatName;
+            return hasName && known.name == device.name &&
+                   known.pluginFormatName == saved.pluginFormatName;
         },
     };
 
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
@@ -1,0 +1,80 @@
+#include "ExternalPluginLookup.hpp"
+
+#include <array>
+#include <functional>
+
+namespace magda {
+
+namespace {
+
+/// JUCE's own name for a format, which is what a PluginDescription carries and
+/// what the scan wrote. Empty for an internal device, which has no format and
+/// never reaches this file.
+juce::String formatNameOf(PluginFormat format) {
+    switch (format) {
+        case PluginFormat::VST3:
+            return "VST3";
+        case PluginFormat::AU:
+            return "AudioUnit";
+        case PluginFormat::LV2:
+            return "LV2";
+        case PluginFormat::Internal:
+            break;
+    }
+
+    return {};
+}
+
+}  // namespace
+
+juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
+    juce::PluginDescription description;
+    description.name = device.name;
+    description.manufacturerName = device.manufacturer;
+    description.fileOrIdentifier = device.fileOrIdentifier;
+    description.isInstrument = device.isInstrument;
+    description.pluginFormatName = formatNameOf(device.format);
+    return description;
+}
+
+ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
+                                         const juce::KnownPluginList& knownPlugins) {
+    const auto saved = describeSavedPlugin(device);
+
+    // The four passes, as predicates over one list. Written this way rather
+    // than as four loops because what matters about them is the order, and the
+    // order is invisible when it is four hundred lines of the same loop.
+    //
+    // The two file passes ask nothing when the device saved no file, because an
+    // empty path matches every description that also has none and that is a
+    // match on the absence of evidence. The app guarded this at its call sites
+    // by declining to search at all; the guard belongs to the pass that needs
+    // it, so the other two still answer.
+    const auto hasFile = device.fileOrIdentifier.isNotEmpty();
+
+    const std::array<std::function<bool(const juce::PluginDescription&)>, 4> passes{
+        [&](const juce::PluginDescription& known) {
+            return hasFile && known.fileOrIdentifier == device.fileOrIdentifier &&
+                   known.isInstrument == device.isInstrument;
+        },
+        [&](const juce::PluginDescription& known) {
+            return known.name == device.name && known.manufacturerName == device.manufacturer &&
+                   known.isInstrument == device.isInstrument;
+        },
+        [&](const juce::PluginDescription& known) {
+            return hasFile && known.fileOrIdentifier == device.fileOrIdentifier;
+        },
+        [&](const juce::PluginDescription& known) {
+            return known.name == device.name && known.pluginFormatName == saved.pluginFormatName;
+        },
+    };
+
+    for (const auto& matches : passes)
+        for (const auto& known : knownPlugins.getTypes())
+            if (matches(known))
+                return {.description = known, .found = true};
+
+    return {.description = saved, .found = false};
+}
+
+}  // namespace magda
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
@@ -37,6 +37,11 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
     return description;
 }
 
+SavedPluginIdentity savedPluginIdentity(const DeviceInfo& device) {
+    return {.identifier = device.uniqueId,
+            .fields = describeSavedPlugin(device).createIdentifierString()};
+}
+
 ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
                                          const juce::KnownPluginList& knownPlugins) {
     const auto saved = describeSavedPlugin(device);
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
@@ -39,7 +39,11 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
 
 SavedPluginIdentity savedPluginIdentity(const DeviceInfo& device) {
     return {.identifier = device.uniqueId,
-            .fields = describeSavedPlugin(device).createIdentifierString()};
+            .name = device.name,
+            .manufacturer = device.manufacturer,
+            .fileOrIdentifier = device.fileOrIdentifier,
+            .format = device.format,
+            .isInstrument = device.isInstrument};
 }
 
 ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.hpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.hpp
@@ -91,18 +91,26 @@ ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
  *
  * Two forms, because a project may carry either. @ref identifier is JUCE's own
  * identifier string, which is what DeviceInfo::uniqueId holds for a device
- * added from the browser; @ref fields is the saved name, format and file, which
- * is all an imported project has.
+ * added from the browser. The fields beside it are what an imported project has
+ * instead, and they are compared as themselves rather than through
+ * PluginDescription::createIdentifierString().
  *
- * Both are needed and neither replaces the other. The identifier contains the
- * plugin's numeric uid, which a description reconstructed from saved fields
- * does not have: comparing one against the other is comparing a real uid
- * against a zero, and it never matches. So a comparison has to be told which
- * form it is looking at, which is what this type is for.
+ * Not through that string for two reasons. It contains the plugin's numeric
+ * uid, which a description rebuilt from saved fields does not have, so
+ * comparing a rebuilt one against a scanned one never matches. And it encodes
+ * only the format, the name and a hash of the file: not the vendor, and not
+ * whether the plugin is an instrument. A bundle shipping an effect and an
+ * instrument of the same name out of one file -- the case the lookup's own
+ * first pass exists for -- produces one string for both.
  */
 struct SavedPluginIdentity {
     juce::String identifier;
-    juce::String fields;
+
+    juce::String name;
+    juce::String manufacturer;
+    juce::String fileOrIdentifier;
+    PluginFormat format = PluginFormat::VST3;
+    bool isInstrument = false;
 
     /**
      * @brief Whether @p other names the same plugin.
@@ -117,7 +125,9 @@ struct SavedPluginIdentity {
         if (identifier.isNotEmpty() && other.identifier.isNotEmpty())
             return identifier == other.identifier;
 
-        return fields == other.fields;
+        return name == other.name && manufacturer == other.manufacturer &&
+               fileOrIdentifier == other.fileOrIdentifier && format == other.format &&
+               isInstrument == other.isInstrument;
     }
 };
 
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.hpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.hpp
@@ -52,20 +52,36 @@ struct ExternalPluginMatch {
 juce::PluginDescription describeSavedPlugin(const DeviceInfo& device);
 
 /**
- * @brief The installed plugin @p device names, searched for in four passes.
+ * @brief The installed plugin @p device names, searched for in five passes.
  *
  * In order, each stricter than the one after it:
  *
- * 1. The file it was loaded from, and the instrument flag agreeing. The flag is
+ * 1. JUCE's own identifier, which is what DeviceInfo::uniqueId holds
+ *    (PluginDescription::createIdentifierString). Exact identity, and it is
+ *    first because nothing below it can tell two plugins apart that a bundle
+ *    ships out of one file with one role: a shell format's dozen effects share
+ *    a path, and the pass below would return whichever of them the scan listed
+ *    first.
+ * 2. The file it was loaded from, and the instrument flag agreeing. The flag is
  *    checked here because a plugin often ships an effect and an instrument out
  *    of one file, and loading the wrong one is silent.
- * 2. Name, vendor and the instrument flag. What answers when the plugin moved.
- * 3. The file alone. What answers when a plugin was renamed in place.
- * 4. Name and format alone. What answers for a project that came from another
+ * 3. Name, vendor, format and the instrument flag. What answers when the plugin
+ *    moved. The format is required because a machine commonly has the same
+ *    plugin twice, as an AU and as a VST3, with the same name, vendor and role:
+ *    the two have different parameter layouts and different state, so
+ *    substituting one for the other loads a plugin and renders a different
+ *    project.
+ * 4. The file alone. What answers when a plugin was renamed in place.
+ * 5. Name and format alone. What answers for a project that came from another
  *    host: a DAWproject carries the plugin's name and its class id but neither
  *    our file path nor, reliably, its role -- Bitwig exports Serum, an
  *    instrument, as an audio effect -- so neither the vendor nor the flag can
  *    be required by the pass that has to resolve it.
+ *
+ * A pass that has nothing to ask with asks nothing: a device that saved no file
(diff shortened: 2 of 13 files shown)


magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // The two file passes ask nothing when the device saved no file, because an
+    // empty path matches every description that also has none and that is a
+    // match on the absence of evidence. The app guarded this at its call sites
+    // by declining to search at all; the guard belongs to the pass that needs
+    // it, so the other two still answer.
+    const auto hasFile = device.fileOrIdentifier.isNotEmpty();
+
+    const std::array<std::function<bool(const juce::PluginDescription&)>, 4> passes{
+        [&](const juce::PluginDescription& known) {
+            return hasFile && known.fileOrIdentifier == device.fileOrIdentifier &&
```
lucaromagnoli, line 71: [P1] Match the saved JUCE identity before falling back to the file. DeviceInfo::uniqueId is populated from PluginDescription::createIdentifierString(), but this matcher never reads it. A bundle can expose several effects or several instruments from the same file, so this first pass returns whichever same-role component happens to be first even when its name and UID differ from the saved device. Please add an exact identifier pass before the file-based fallbacks and cover a same-file, same-role bundle. — Codex

magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
```diff
@@ -0,0 +1,422 @@
...
+        else
+            audio.clear(destination, 0, numSamples);
+    }
+
+    processPluginBlock(audio);
+
+    for (int channel = 0; channel < destChannels; ++channel) {
+        auto* destination = block.audio.getChannelPointer(static_cast<std::size_t>(channel));
+
+        if (channel < processChannels_)
```
lucaromagnoli: [P1] Base copy-back on the plugin output count, not the scratch width. processChannels_ is max(inputs, outputs), so it includes input-only channels. For example, a mono-output plugin with a mono sidechain has processChannels_ == 2; in a stereo slot this branch copies the sidechain/input-only channel into the right output instead of duplicating output channel 0. The same distinction is needed by the in-place fast-path test below for stereo-input/mono-output plugins. Track the actual output width and use it for both decisions. — Codex

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -0,0 +1,80 @@
...
+                   known.isInstrument == device.isInstrument;
+        },
+        [&](const juce::PluginDescription& known) {
```
lucaromagnoli, line 74: [P1] Keep the saved format in this moved-plugin pass. On macOS the known list commonly contains AU and VST3 descriptions with the same name, vendor, and role. When the saved VST3 path has moved, pass 1 fails and this pass can return the AU solely because it appears first, even though the formats have different state and parameter layouts. Require known.pluginFormatName to match saved.pluginFormatName here and add a case where the competing format satisfies this earlier pass. — Codex

magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
```diff
@@ -0,0 +1,422 @@
...
+    // either is destroyed, because JUCE leaves the pointer where it was and a
+    // plugin asking after the fact would read freed memory. The fork does the
+    // same thing in the same order (deinitialise: "must be done first").
+    instance_->setPlayHead(nullptr);
+
+    if (prepared_)
+        instance_->releaseResources();
+}
+
+void EngineExternalDevice::prepare(const magda::engine::RenderContext& context) {
```
lucaromagnoli, line 176: [P2] Pair a repeated prepare with releaseResources(). RuntimeStateStore prepares every retained device again when the RenderContext changes, and EngineMagdaDevice explicitly releases before its second prepare for that lifecycle. Here the same AudioPluginInstance receives prepareToPlay again without the matching stop/release call; plugins that allocate or register resources in prepare can leak, retain the old-rate resources, or fail after an audio-device change. — Codex

magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -2142,66 +2143,12 @@ te::Plugin::Ptr PluginManager::createPluginOnly(TrackId trackId, const DeviceInf
             }
         }
     } else {
-        // External plugin — same lookup logic as loadDeviceAsPlugin but without track insertion
+        // External plugin. Which installed plugin the saved device meant is one
+        // question with one answer (ExternalPluginLookup.hpp), asked here, by
+        // the track path below, and by the native engine's device factory.
         if (device.uniqueId.isNotEmpty() || device.fileOrIdentifier.isNotEmpty()) {
```
lucaromagnoli: [P2] Remove the remaining whole-search guard. The shared matcher deliberately supports name/vendor and name/format lookup when no file was saved, and the imported-project test constructs exactly such a device with both uniqueId and fileOrIdentifier empty. This guard, and the identical one in loadDeviceAsPlugin, prevent the real app paths from ever exercising that behavior, so they return no plugin while the helper test passes. — Codex

magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+        if (channel < sidechainChannels)
+            juce::FloatVectorOperations::copy(
+                audio.getWritePointer(destination),
+                block.sidechain.getChannelPointer(static_cast<std::size_t>(channel)), numSamples);
+        else
+            audio.clear(destination, 0, numSamples);
+    }
+
+    processPluginBlock(audio);
```
lucaromagnoli, line 452: **P1 — Publish the plugin additional output pairs**

`processChannels_` includes every plugin output channel, and the executor supplies one cleared `DeviceBlock::extraOutputs` block for each pair declared by `DeviceInfo::multiOut`, but this copy-back only fills `block.audio`. A multi-output instrument therefore renders its main pair while every MultiOut track receives cleared buffers. Please copy scratch channels 2/3, 4/5, and so on into `block.extraOutputs` (respecting each block width), and add an adapter/executor regression test.

— Codex

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -84,6 +86,57 @@ void restoreSavedState(MagdaDevice& device, const juce::String& savedState) {
...
+/// The adapter over an instance the format manager just handed back.
+///
+/// enableAllBuses first, at the same point the fork does it
+/// (completePluginInstanceCreation) and for the same reason: a plugin whose
+/// sidechain or second output bus is disabled reports channels it does not
+/// have, and everything downstream -- the adapter's own channel adaptation, the
+/// plan's declared widths -- is read off those numbers.
+ExternalDeviceResult adaptInstance(std::unique_ptr<juce::AudioPluginInstance> instance,
+                                   const magda::DeviceInfo& device, bool offlineRender) {
+    instance->enableAllBuses();
```
lucaromagnoli, line 100: **P1 — Restore the external plugin native state**

`adaptInstance` enables the buses and immediately wraps a newly created default instance, but it never applies `device.pluginState` (the base64 native chunk) or `device.vst3Preset`. The parameter array is not an equivalent substitute: native state also carries programs, samples/mappings, non-automatable values and other private data, and the incumbent restore path deliberately makes the chunk authoritative when it disagrees with stale saved parameters. As a result, a saved synth patch can render the initialized/default voice under the native engine, despite `EngineExternalDevice` documenting that the instance arrives with saved state applied. Please restore the applicable state before publishing the adapter, preserve chunk-over-parameter precedence, and cover both sync and async creation with a chunk whose parameter value disagrees with the model.

— Codex

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -84,6 +86,105 @@ void restoreSavedState(MagdaDevice& device, const juce::String& savedState) {
...
+ExternalDeviceResult adaptExternalPluginInstance(
+    std::unique_ptr<juce::AudioPluginInstance> instance, const magda::DeviceInfo& device,
+    bool offlineRender) {
+    instance->enableAllBuses();
+
+    // Before the adapter, not after, and for the same reason the internal path
+    // restores before wrapping: the adapter reads the plugin's parameter list
+    // when it is constructed, and a plugin that changes which parameters it has
+    // when its state is applied would be mapped against the ones it had first.
+    restoreSavedChunk(*instance, device.pluginState);
```
lucaromagnoli: **P1 — Preserve the chunk when the saved parameter array is stale**

This restore is undone for every automatable value on the first call to `EngineExternalDevice::process`, because `writeParameters` writes the plan values back over the instance. That is not an undecided precedence in this repository: `test_external_plugin_state_restore_juce.cpp` covers projects produced by the pre-fix build where `DeviceInfo::parameters` is stale, and the incumbent path deliberately applies the array as a baseline, overlays the chunk, then refreshes its cache so the chunk wins. The new test instead restores Tone = 0.9 and explicitly expects the block to replace it with 0.4, codifying a null-diff divergence and recreating the saved-patch bug from the first block. Please seed the effective static parameter values from the restored instance (while still allowing real automation to override them), and add the existing stale-array/chunk scenario to this adapter path.

— Codex

magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
```diff
@@ -0,0 +1,544 @@
...
+                // block and render the initialised voice.
+                //
+                // Anything that moves the value off that position -- an
+                // automation lane, a macro, a modifier, a host edit -- is a
+                // write, and it stays a write from then on. Latched rather than
+                // compared each block, because a lane that returns to the
+                // saved position later has still taken the parameter away from
+                // the plugin's own state, and stopping there would leave it
+                // wherever the last write put it.
+                if (!mapping.moved) {
```
lucaromagnoli: **P1 — Keep the saved parameter baseline when no chunk restores**

This suppression is unconditional, so it also skips the model values when `pluginState` is empty, malformed, or a valid chunk the plugin declines. In those cases the incumbent sequence leaves the saved parameter array in place: it applies that baseline first and only then overlays the chunk. Here, for example, a model saving Tone = 0.7 beside an empty state resolves 0.7, sees that it equals `mapping.base`, and breaks forever while the instance stays at its 0.25 factory default. The new no-state test does not call `process`, so it cannot catch this. Please mirror the baseline-then-overlay sequence before latching these writes off, and add empty/invalid-state cases with a saved value different from the plugin default.

— Codex

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -115,4 +180,59 @@ bool isRegisteredDevice(const juce::String& pluginId) {
...
+        return;
+    }
+
+    // The device is copied into the callback rather than captured by reference.
+    // What it is read for is its parameter metadata, and the model it came from
+    // is free to be edited, moved or deleted while a plugin loads: that is what
+    // taking seconds means.
+    services.formats->createPluginInstanceAsync(
+        resolved.description, services.context.sampleRate, services.context.maxBlockSize,
+        [saved = device, offlineRender, completed = std::move(completed)](
```
lucaromagnoli: **P1 — Do not overwrite edits made while the async load is in flight**

The callback restores from this by-value `saved` snapshot and returns parameter values that the API contract tells the caller to write into the live model. If loading takes a few seconds and the user changes a parameter or applies a preset meanwhile, completion restores the old array/chunk and `applyRestoredParameters` overwrites that newer edit. The comment explicitly allows the model to be edited while loading, but copying it only makes the memory safe; it does not make the stale values authoritative. Please carry an identity/revision and merge the snapshot only when the corresponding model value is unchanged, or re-read the current `DeviceInfo` on the message thread before running the restore transaction. Add a delayed-success callback test with a model edit between request and completion; the current async test returns an immediate missing-plugin failure and cannot exercise this race.

— Codex

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -115,4 +180,85 @@ bool isRegisteredDevice(const juce::String& pluginId) {
...
+
+    if (instance == nullptr)
+        return {.device = {},
+                .failure = "external plugin \"" + device->name + "\" could not be loaded: " +
+                           (error.isNotEmpty() ? error : "no reason given")};
+
+    // And it has to still be the same plugin. A device whose plugin was
+    // replaced while this one loaded is not waiting for this answer, and
+    // restoring onto it would put one plugin's saved patch on another.
+    if (magda::describeSavedPlugin(*device).createIdentifierString() !=
```
lucaromagnoli: **P1 — Compare against the saved JUCE identity, not a reconstructed zero UID**

`describeSavedPlugin` does not populate `PluginDescription::uniqueId` or `deprecatedUid`, while `requested` is the scanned description and normally has a nonzero UID. JUCE includes that integer in `createIdentifierString()`, so an unchanged real plugin compares unequal here and every successful async load is rejected as “changed plugin.” The tests miss this because they construct `requested` with `describeSavedPlugin` too, leaving both UIDs at zero. Please use the model’s saved `device.uniqueId` with `requested.matchesIdentifierString`/the exact identifier (and the existing lookup fallback when an imported device has no saved identity), then add a completion test whose requested description has a nonzero UID and whose model stores `requested.createIdentifierString()`.

— Codex

magda/daw/audio/plugin_manager/ExternalPluginState.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    // DeviceInfo::vst3Preset is deliberately not applied here. It is the
+    // portable .vstpreset a DAWproject carries, applied once on import and
+    // cleared, and its own field comment says what this relies on: interchange
+    // only, native state is pluginState. When the native engine grows an import
+    // path of its own (#2244) it belongs in this function beside the chunk,
+    // rather than in whichever engine happens to be loading the project.
+    try {
+        instance.setStateInformation(chunk.getData(), static_cast<int>(chunk.getSize()));
+    } catch (...) {
```
lucaromagnoli, line 92: **P2 — Do not publish a partially restored instance after an exception**

Catching the exception does not mean the baseline is still standing. A third-party `setStateInformation` implementation can change parameters or private state and then throw; this returns `false`, but the factory ignores that result, snapshots the partially mutated instance, and publishes it back to the model and engine. The current test throws before touching any state, so it cannot exercise this case. Please treat a thrown restore as a load failure and discard the instance (or recreate it and reapply the baseline); simply replaying parameters cannot roll back partially changed programs/samples/private state. Add a stub that changes Tone and then throws to verify no partial state is published.

— Codex

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -0,0 +1,93 @@
...
+    description.manufacturerName = device.manufacturer;
+    description.fileOrIdentifier = device.fileOrIdentifier;
+    description.isInstrument = device.isInstrument;
+    description.pluginFormatName = formatNameOf(device.format);
+    return description;
+}
+
+SavedPluginIdentity savedPluginIdentity(const DeviceInfo& device) {
+    return {.identifier = device.uniqueId,
+            .fields = describeSavedPlugin(device).createIdentifierString()};
```
lucaromagnoli: **P2 — Keep role/vendor in the fallback identity**

`PluginDescription::createIdentifierString()` only encodes format, name, the file hash, and numeric UID; it drops both `manufacturerName` and `isInstrument`. Here the reconstructed description has UID 0, so an imported device’s fallback identity treats an effect and instrument with the same name/file/format as identical. If the user swaps between those variants while the async load is pending, `completeExternalPluginLoad()` accepts the stale instance and restores the new device’s state onto the wrong plugin—the same shared-bundle ambiguity that pass 2 above explicitly avoids. Please compare an explicit saved-field tuple that includes at least the instrument role (and preferably manufacturer), rather than JUCE’s identifier string.

— Codex

magda/daw/audio/plugin_manager/ExternalPluginLookup.hpp
```diff
@@ -0,0 +1,137 @@
...
+     * otherwise, which is what an imported device has and what a device that
+     * has only just been resolved has: a device that gains a scanned identifier
+     * has not changed plugin, and a comparison that read that as a change would
+     * refuse an answer it should have taken.
+     */
+    bool namesSamePluginAs(const SavedPluginIdentity& other) const {
+        if (identifier.isNotEmpty() && other.identifier.isNotEmpty())
+            return identifier == other.identifier;
+
+        return name == other.name && manufacturer == other.manufacturer &&
```
lucaromagnoli, line 128: **P2 — Do not treat scan metadata correction as a plugin swap**

This equality is stricter than the lookup that started the load. The fallback passes deliberately resolve the same plugin after a moved path, missing vendor, rename, or unreliable DAWproject role; the app also corrects `isInstrument` from the resolved description so routing is accurate. If that correction happens while an imported synth is loading (e.g. Bitwig saved Serum as `audioFX`), the current device differs only in `isInstrument`, yet this guard discards the correct instance as a replacement. The native plan needs that correction because `PlanCompiler::consumesMidi()` reads `isInstrument`; leaving it stale gives the synth no MIDI, while applying it trips this check. Please base the completion guard on the resolved installed identity, while allowing saved-field enrichment/correction that the lookup itself considers the same plugin.

— Codex

PR #2250, opened 2026-08-30
MPE pitch glide: bend a segment between two points (#2198)

### LINKED ISSUE #2198: [FR]: MPE Pitch glide curve between 2 points
### What would you like?

The ability to drag the line between 2 points of the MPE linear curve and reshape to non-linear

In Bitwig this is achieved by holding ALT & LMB DRAG when the mouse is hovered slightly off center of the line ( see video )

https://github.com/user-attachments/assets/54d17eaa-ce6f-42f5-8ea7-17969229f1bb

### Use case

Great for having more expressive MPE curves

lucaromagnoli:
Closes #2198.

Drag a glide segment with Alt held and it bends, the way Bitwig does it.

## The short version

The spline machinery was already here. `core/CurveMath.hpp`'s `evalSegment` backs the automation lanes, the LFO editor and the CC/pitch-bend lanes, and both `MidiCCData` and `MidiPitchBendData` carry a `curveType` and a `tension`. The per-note MPE glide was the one curve in the app that never got connected to it — `MidiPitchExpressionPoint` was `{beat, semitones}` and interpolated linearly, with no way to say anything else. Which is why this reads as an oversight rather than a missing feature.

So this is mostly wiring, not new maths.

## What changed

**The model.** `MidiPitchExpressionPoint` gains a `tension`, in the same −3..+3 scale as `CurvePoint` and `MidiCCData`, owned by the segment's left point as segments are everywhere else here. Zero is the straight line it always was, so a glide authored before this reads back unchanged. The serializer writes the property only when a segment is actually bent, so a project of straight glides is written exactly as it was before the field existed.

**One evaluator, three callers.** The piano roll drew one curve, `MidiClipCompiler` compiled a second and `ClipSynchronizer` handed a third to Tracktion, each with its own copy of the same interpolation. They agreed only because none of them had a shape to disagree about; a tension ends that, so the answer moved to `core/PitchExpressionCurve.hpp` and all three ask it. Neither compiler needed anything else — both already densified each segment finely enough for a bend to survive.

**The gesture.** Alt is what Bitwig uses and the one modifier still free in expression mode (Shift already means "do not snap to semitones" on both of the other gestures there). `pitchExpressionTensionThrough` inverts `evalSegment` at the parameter the mouse is at, so the curve passes under the pointer wherever along the segment it was grabbed — not a nudge-until-it-looks-right knob.

**The cursor**, in its own commit. Alt already showed the note-draw pencil in the piano roll, including in expression mode where it cannot draw, so the bend gesture was announcing itself with a pencil pointing at the curve it was about to reshape. There is a bend cursor now — an S-curve for what changes, a small double-headed arrow for the drag being vertical — shown only while Alt is held and only over a segment the gesture would take, so it doubles as the discoverability hint.

## One limitation, deliberate

A segment between two points at the same pitch stays flat, and the gesture declines it. The shape is a warp of the travel between the endpoints and there is none — the same property the automation lanes and the LFO editor have. Bitwig can bend a flat segment because it bends in absolute terms; matching that means a second kind of maths for glides only, which is worth its own issue rather than a quiet divergence here.

## Tests

- `test_pitch_expression_curve.cpp` — the shape headless: endpoints held, monotonic, mirrored about the straight line, confined to its own segment, and the drag's inverse round-tripped back through the evaluator.
- A compiler case asserting the wheel stream follows the curve rather than the line, to the wheel's own resolution.
- A serialization case pinning both the round-trip and the absent property on a straight segment.

Verified negatively: forcing the evaluator back to linear fails four cases, one of them the compiled output.

Full headless suite green (3113 passed, 13 skipped). Gesture confirmed by hand in the running app.

## Worth a reviewer's eye

The tension→shape mapping is the existing one, reused as-is. The new maths is only its inverse, and it is covered by round-tripping rather than by asserting particular numbers.

### DIFF
--- a/magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp
+++ b/magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp
@@ -902,6 +902,7 @@ void PianoRollGridComponent::mouseMove(const juce::MouseEvent& e) {
             hoveredExpressionPoint_ = hit;
             repaint();
         }
+        updateExpressionCursor(e.mods, e.getEventRelativeTo(this).getPosition());
     } else if (hoveredExpressionPoint_) {
         hoveredExpressionPoint_.reset();
         repaint();
@@ -2034,7 +2035,25 @@ double PianoRollGridComponent::absolutePlayheadBeatForDisplayX(int mouseX) const
     return juce::jlimit(0.0, timelineLengthBeats_, beat);
 }
 
+void PianoRollGridComponent::updateExpressionCursor(const juce::ModifierKeys& mods,
+                                                    juce::Point<int> pos) {
+    // Only while the modifier is actually held: the cursor is how the gesture
+    // announces itself, and showing it unprompted over every glide would make
+    // Alt look like it was already down.
+    if (isBendExpressionGesture(mods) && hitTestExpressionSegment(pos).has_value())
+        setMouseCursor(CursorManager::getInstance().getCurveBendCursor());
+    else
+        setMouseCursor(juce::MouseCursor::NormalCursor);
+}
+
 void PianoRollGridComponent::updateEmptyGridCursor(const juce::ModifierKeys& mods, int /*mouseX*/) {
+    // Alt means bend while glides are being edited, not draw, and a pencil
+    // pointing at a curve the click will reshape is the wrong promise.
+    if (pitchExpressionMode_) {
+        updateExpressionCursor(mods, getMouseXYRelative());
+        return;
+    }
+
     if (mods.isAltDown()) {
         setMouseCursor(CursorManager::getInstance().getNoteDrawCursor());
     } else {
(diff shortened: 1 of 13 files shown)


magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp
```diff
@@ -2972,6 +2977,55 @@ std::optional<PianoRollGridComponent::ExpressionHit> PianoRollGridComponent::hit
...
+                continue;
+
+            auto visibleNote = clip->midiNotes[i];
+            if (!ClipOperations::clipMidiNoteToVisibleRange(*clip, visibleNote))
+                continue;
+
+            const double noteStartDisplayBeat = displayBeatForClipBeat(clipId, note.startBeat);
+            const double centerY = noteNumberToY(note.noteNumber) + noteHeight_ * 0.5;
+
+            for (size_t p = 0; p + 1 < points.size(); ++p) {
```
lucaromagnoli, line 3005: Flat segments still pass this hit test, so they advertise the bend cursor and begin an expression drag even though `bendExpressionSegment()` later refuses them. Mouse-up then commits unchanged points, adding a no-op undo entry. Please exclude segments whose endpoint pitches are equal here (ideally via a predicate shared with the drag handler), so the cursor and gesture are offered only where bending can take effect.

— Codex

lucaromagnoli: Fixed in 55ae145f.

Agreed that a bend between two points at the same pitch is not a thing — nothing here tries to make it one, and adding a point stays the way to shape that region. What was wrong was that the editor still offered the gesture: the hit test found the segment, so the cursor promised a bend, the drag started, `bendExpressionSegment` refused it for the whole gesture, and mouse-up committed regardless.

`pitchExpressionSegmentCanBend` now answers that in core beside the maths that imposes it, and the hit test and the drag handler both ask it — one epsilon, one rule. The drag handler keeps its own check rather than trusting the hit test, because points move under a live drag and a segment that was bendable at mouse-down need not still be.

Five test cases on the predicate, including that what it permits is exactly what the evaluator actually changes.

PR #2254, opened 2026-08-30
Add project metadata fields (#2253)

### LINKED ISSUE #2253: Add project metadata fields
A project has no author-facing metadata. `ProjectInfo` carries tempo, time signature, sample rate, bit depths, key, timeline length, loop, markers and view state — all technical. There is nowhere to put who wrote the thing.

DAWproject already defines exactly the set we want, in `third_party/dawproject/MetaData.xsd`:

`Title`, `Artist`, `Album`, `OriginalArtist`, `Composer`, `Songwriter`, `Producer`, `Arranger`, `Year`, `Genre`, `Copyright`, `Website`, `Comment` — all optional strings.

### Where we are

- `DawProjectArchive::toMetadataXml()` writes a single `<Title>`, taken from `info.name`. Every other element is absent from every archive we export.
- The import side (`DawProjectArchive::readFromFile()`) reads `<Title>` back into `info.name` and drops the other twelve on the floor, so a project exported by Bitwig or Studio One loses its credits the moment we open it.
- Nothing in the `.mgd` format holds any of this either, so there is nothing to round-trip even if the adapter wanted it.

### Scope

- A `ProjectMetadata` block on `ProjectInfo` holding the thirteen fields, all defaulting to empty.
- Serialize it in `.mgd`; absent keys load as empty so existing projects keep opening.
- Map it both ways in `DawProjectArchive`, writing only the elements that have text (the schema is a `xs:sequence` of `minOccurs="0"`, so element order matters). `Title` stays backwards compatible: export falls back to `info.name` when the title is empty, and import still seeds `info.name` from `<Title>` when the name is empty.
- A Metadata section in `ProjectSettingsDialog` (File > Project Settings) with the thirteen text fields, above the existing technical rows. New `project_settings.*` source strings in `lang/en.json`.
- Round-trip coverage in `tests/test_project_serialization.cpp` and `tests/test_dawproject_round_trip.cpp`, including the empty-fields case (no stray elements in `metadata.xml`) and schema validation of a fully-populated one.

Related: #2169 covers per-project *defaults* seeded from `Config`, which is separate work.

lucaromagnoli:
Closes #2253.

A project had nowhere to record who made it. `ProjectInfo` carried tempo, sample rate, bit depths, key and timeline length — all technical. On the DAWproject side we wrote a single `<Title>` taken from the project name and read only `<Title>` back, so the other twelve elements were dropped on import: a session exported from Bitwig or Studio One lost its credits the moment we opened it.

## The field set

DAWproject already defines exactly what we want, in `third_party/dawproject/MetaData.xsd`, so `ProjectMetadata` is that element verbatim — Title, Artist, Album, OriginalArtist, Composer, Songwriter, Producer, Arranger, Year, Genre, Copyright, Website, Comment, all optional strings.

`kProjectMetadataFields` pairs each member with its `.mgd` key and its XML element name, in schema order. The four sites that map metadata — native save, native load, DAWproject write, DAWproject read — iterate that table rather than spelling out thirteen fields apiece and drifting apart. The order is load-bearing: `metaData` is an `xs:sequence`, so an export that writes the elements in any other order fails schema validation.

`year` is a string because the schema says `xs:string`, and because the things people actually type there ("2003-2005", "MMXIV") do not all survive a parse to int.

## Serialization

`ProjectSerializer` parses the project object twice — once in `loadAndStage`, once in `deserializeProject` — so both now read through one shared `readProjectMetadata()` helper. A field only one of them knew about would go missing down whichever path the caller happened to take.

Only filled-in fields are written, so an uncredited project carries no `metadata` object at all, and every project saved before today loads with the block empty.

## Title, both ways

`<Title>` stays backwards compatible. An empty title exports the project name, because every DAW that reads this expects to find something there and because archives written before this change should look the same as the ones written after. An incoming `<Title>` still names the imported project, as it did before.

The cost: a round trip copies the title into the name, since DAWproject has nowhere to put a project name. `A song title outranks the project name in metadata.xml` asserts that explicitly rather than leaving it to be discovered.

The title is a field of its own rather than a reuse of `info.name`, because a song called *Blue* can live in `blue_v7.mgd`. But it opens empty and shows the project name as its placeholder, so nobody has to retype a name they already gave: leaving the box alone is what writes the name, and keeping it empty keeps the inheritance live, so renaming the project carries through instead of freezing whatever it was called the first time somebody opened the dialog.

## UI

**File → Project Settings** becomes two tabs: **General** (the existing technical controls) and **Metadata** (two columns of short fields plus a full-width comment box). Naming a song and setting a sample rate are unrelated jobs and nobody opens the dialog to do both. The dialog sizes to the taller tab, at 408px.

The widgets moved into a `GeneralPage` and a `MetadataPage` that each load from and apply to a `ProjectInfo`. The pages are forms and nothing else; the dialog owns the side effect on the live timeline.

Two footer changes came out of the split:

- **The *Save as default for new projects* tick is gone.** It only ever governed the four General values, so on a tabbed dialog it read as belonging to whichever tab you were looking at.
- **OK is replaced by Apply.** OK was only Apply that also closed. Apply commits; Cancel closes and discards whatever has not been applied.

## Preferences gets the defaults

Dropping that tick left nothing in the UI writing Config's new-project sample rate, bit depths and timeline length — Preferences had punted all four to Project Settings, with comments in both pages saying so, and they would have become unreachable.

They belong in Preferences, which is where per-install settings live:

A new **Defaults** tab holds everything a project is seeded with. Called Defaults rather than Project because a Project tab in Preferences reads like it edits the project you have open, which is what Project Settings is for; the one thing every control there has in common is that it changes nothing until the next project is created.

- **Format** — Default Total Length, Sample Rate, Render Bit Depth, Bounce Bit Depth.
- **Credits** — Artist, Composer, Songwriter, Producer, Arranger, Genre, Copyright, Website. A solo producer types the same eight strings into every project they make; now they type them once.

Title, Album, Original Artist, Year and Comment get no default: they describe the work rather than the person, and a stored value would be wrong in every project after the first. Which fields take a default is a `seededFromDefaults` flag on `kProjectMetadataFields`, so Config, the dialog and `ProjectManager` read one list instead of repeating three.

Config keys the credits by the same string the project format uses, so it can hold them without knowing what a `ProjectMetadata` is — the field list lives in a layer Config doesn't include. The Preferences fields borrow their labels from the Project Settings keys rather than duplicating eight strings that must stay identical.

*General → Timeline* keeps **Default View**, which is about how much of the timeline is on screen rather than how long it is. *Rendering* goes back to being only about where rendered files go and what they are called.

The rates and depths themselves move to `ProjectFormatChoices.hpp`. Two dialogs now offer the same choices for the same fields, and two copies of the lists would drift — the first symptom being a default that cannot be expressed, offered on one dialog and missing from the other.

The manual already documented all four as Preferences settings; it had never been updated when they moved. It is true again, and now says which dialog owns which.

Thirteen new `en.json` source strings for Crowdin. Manual pages for Project Settings and DAWproject updated.

## Verification

Full Catch2 suite green — 3122 passed, 13 skipped, 0 failures.

Both legs verified negatively: reverting the export loop to title-only fails `DawProjectArchive roundtrips the whole MetaData block`, and short-circuiting the native read fails `every DAWproject metadata field survives a save and load`.

New coverage:

- `tests/test_project_serialization.cpp` — all thirteen fields round-trip; an uncredited project writes no `metadata` object; only filled-in fields are written; loading a project with no credits clears credits already in the struct, so the previous project's artist can't show under this one's name.
- `tests/test_project_document_adapters.cpp` — a fully populated block round-trips through the archive and validates against the schema (which is also what asserts element order); an uncredited project exports a `<Title>` and nothing else; a song title outranks the project name.
- A new project is seeded with the stored credit defaults — asserting both halves: the eight carry over, and the other five stay empty even when the config file holds a value for them. Verified negatively by disabling the seeding loop and watching it fail.

The dialog layout was checked by rendering it offscreen to a PNG through a throwaway JUCE snapshot test, which was then removed — `tests/CMakeLists.txt` is unmodified.


magda/daw/ui/dialogs/PreferencesDialog.cpp
```diff
@@ -1352,6 +1355,161 @@ class AppearancePage : public juce::Component {
...
+            const auto entry = credits.find(row->field->key);
+            row->editor.setText(entry == credits.end() ? juce::String()
+                                                       : juce::String(entry->second),
+                                juce::dontSendNotification);
+        }
+    }
+
+    void applySettings(Config& config) {
+        config.setDefaultTimelineLengthBars(static_cast<int>(lengthSlider.getValue()));
+        config.setRenderSampleRate(
```
lucaromagnoli, line 1482: **[P2] Give new-project defaults independent storage**

These controls still write `Config::renderSampleRate` and `renderBitDepth`, but `ExportAudioDialog` also overwrites those same fields on every export (lines 117–126 there). As a result, choosing a one-off export rate/format silently changes what every future project starts with, and this Defaults tab later shows export history as if it were an explicit default. Please decouple the new-project seeds from the export-dialog preferences, or stop the export dialog from writing these fields.

— Codex

magda/daw/ui/dialogs/PreferencesDialog.cpp
```diff
@@ -3171,6 +3341,7 @@ void PreferencesDialog::applySettings() {
     renderingPage->applySettings(config);
     shortcutsPage->applySettings(config);
     pathsPage->applySettings(config);  // no-op for path values
+    defaultsPage->applySettings(config);
```
lucaromagnoli, line 3348: **[P2] Keep future-project length from resizing the open project**

After this applies the new default, the unchanged block below dispatches `getDefaultTimelineLengthBars()` into the live `TimelineController`. Applying Preferences > Defaults therefore resizes the currently open arrangement even though this tab and the manual promise that nothing here affects the open project; it also leaves `currentProject_.timelineLengthBars` unchanged, so the model and timeline disagree. Please remove or gate that live dispatch so this value is consumed only when `newProject()` seeds a project.

— Codex

PR #2255, opened 2026-08-30
Use stable assignment tokens for async external-plugin loading (#2252)

### LINKED ISSUE #2252: Use stable assignment tokens for async external-plugin loading
Follow-up to #2249.

## Context

PR #2249 introduces the native external-plugin factory and its asynchronous completion path. That path is currently scaffolding: `createEngineExternalDeviceAsync()` and `SavedPluginIdentity` have no production callers outside their implementation and tests.

The current completion guard decides whether a device changed plugin by comparing mutable saved metadata. This is not a stable identity boundary:

- External-plugin lookup deliberately tolerates a moved file, renamed plugin, missing vendor, and an unreliable DAWproject instrument role.
- The model may legitimately gain corrected scan metadata while the plugin is loading.
- Strict field equality therefore rejects the correct instance after metadata correction.
- Loosening the comparison risks accepting a stale instance after a genuine plugin replacement.

Review context: https://github.com/Conceptual-Machines/magda-core/pull/2249#discussion_r3890330507

## Required design

Use a stable plugin-assignment generation/request token rather than inferring replacement from mutable plugin metadata.

A suitable flow is:

1. Resolve the saved device to an installed `PluginDescription` once.
2. Surface and apply resolved metadata/capabilities before compiling the native plan, so corrected instruments receive MIDI and channel topology is accurate.
3. Associate the device's plugin assignment with a monotonically changing generation or equivalent stable token.
4. Capture the device ID, assignment generation, and exact resolved installed identity when an async load starts.
5. At completion, accept the instance only if the device still exists and has the same assignment generation.
6. Restore the instance from the current model state, then publish it and its corrected parameter snapshot.
7. Discard results for deleted devices or replaced assignments without mutating the model.

## Acceptance criteria

- Metadata enrichment or correction while loading does not discard the correct instance.
- Changing only an imported DAWproject's incorrect `isInstrument` value to the scanned value is accepted.
- A moved path, filled-in vendor, or scanned identifier gained during loading is accepted when the assignment is unchanged.
- Replacing the plugin with another variant from the same bundle is rejected.
- Replacing it with a same-name plugin from another vendor is rejected.
- Deleting the device while loading is rejected.
- Parameter edits and preset/state changes made while loading are restored from the current model.
- Resolved role and capabilities are applied before plan compilation.
- Production wiring of `createEngineExternalDeviceAsync()` is blocked on this issue.

PR #2249 may merge independently because this async path is not yet used by production code.

lucaromagnoli:
## Summary

- give each live plugin assignment a transient, monotonically generated token that survives model copies but changes for replacement devices
- apply the exact resolved plugin metadata, role, and channel topology before native plan compilation
- complete async loads only when the captured device ID and assignment token still match, while restoring parameters and plugin state from the current model
- cover metadata enrichment, incorrect imported instrument roles, same-bundle variants, same-name plugins from different vendors, deletion, and in-flight parameter/state edits

## Tests

- cmake --build cmake-build-debug --target magda_tests -j8
- magda_tests [engine][external] — 41 cases, 137 assertions
- magda_tests [plugins][lookup] — 13 cases, 32 assertions
- magda_tests [engine][plan][compiler] — 55 cases, 431 assertions
- magda_tests [project][serialization][device] — 6 cases, 56 assertions

Closes #2252

### DIFF
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
     return description;
 }
 
-SavedPluginIdentity savedPluginIdentity(const DeviceInfo& device) {
-    return {.identifier = device.uniqueId,
-            .name = device.name,
-            .manufacturer = device.manufacturer,
-            .fileOrIdentifier = device.fileOrIdentifier,
-            .format = device.format,
-            .isInstrument = device.isInstrument};
+void applyResolvedPluginDescription(DeviceInfo& device,
+                                    const juce::PluginDescription& description) {
+    device.name = description.name;
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
+    device.isInstrument = description.isInstrument;
+    if (device.deviceType != DeviceType::MIDI)
+        device.deviceType = description.isInstrument ? DeviceType::Instrument : DeviceType::Effect;
+    device.uniqueId = description.createIdentifierString();
+    device.fileOrIdentifier = description.fileOrIdentifier;
+    device.audioInputChannels = description.numInputChannels;
+    device.audioOutputChannels = description.numOutputChannels;
 }
 
 ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
--- a/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
@@ -37,21 +37,6 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
     return description;
 }
 
-void applyResolvedPluginDescription(DeviceInfo& device,
-                                    const juce::PluginDescription& description) {
-    device.name = description.name;
-    device.pluginId = description.createIdentifierString();
-    device.manufacturer = description.manufacturerName;
-    device.format = pluginFormatFromName(description.pluginFormatName);
-    device.isInstrument = description.isInstrument;
-    if (device.deviceType != DeviceType::MIDI)
-        device.deviceType = description.isInstrument ? DeviceType::Instrument : DeviceType::Effect;
-    device.uniqueId = description.createIdentifierString();
-    device.fileOrIdentifier = description.fileOrIdentifier;
-    device.audioInputChannels = description.numInputChannels;
-    device.audioOutputChannels = description.numOutputChannels;
-}
-
 ExternalPluginMatch matchInstalledPlugin(const DeviceInfo& device,
                                          const juce::KnownPluginList& knownPlugins) {
     const auto saved = describeSavedPlugin(device);
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
@@ -100,15 +100,6 @@ ExternalDeviceResult adaptExternalPluginInstance(
     bool offlineRender) {
     instance->enableAllBuses();
 
-    // Scan descriptions report the format's default buses, not the instance
-    // after the host has enabled its sidechains and extra outputs. These are
-    // the only channel counts safe to compile a live plan from.
-    auto resolvedDevice = device;
-    resolvedDevice.audioInputChannels = instance->getTotalNumInputChannels();
-    resolvedDevice.audioOutputChannels = instance->getTotalNumOutputChannels();
-    resolvedDevice.canReceiveMidi = !resolvedDevice.isInstrument && instance->acceptsMidi();
-    resolvedDevice.producesMidi = instance->producesMidi() || instance->isMidiEffect();
-
     // The saved array, then the chunk over it, then what that left behind. The
     // order and the reasons are ExternalPluginState.hpp's; what matters here is
     // that all three happen before the adapter exists, so the adapter never has
@@ -117,6 +108,33 @@ ExternalDeviceResult adaptExternalPluginInstance(
         return {.device = {},
                 .failure = "external plugin \"" + device.name +
                            "\" failed while restoring its own saved state"};
+
+    // Buses again, and the widths only now. setStateInformation is free to
+    // change a plugin's bus layout -- an instrument the patch switches to
+    // multi-out, a compressor whose sidechain the patch turns on -- so widths
+    // read before the chunk was applied describe a layout the instance no
+    // longer has, while EngineExternalDevice::prepare() goes on to process the
+    // one it does. Re-enabling keeps the all-buses policy the first call
+    // establishes: a chunk that disabled one would otherwise leave the device
+    // reporting channels the rest of the chain is wired for.
+    instance->enableAllBuses();
+
+    // Scan descriptions report the format's default buses, not the instance
+    // after the host has enabled its sidechains and extra outputs. These are
+    // the only channel counts safe to compile a live plan from.
+    auto resolvedDevice = device;
+    resolvedDevice.audioInputChannels = instance->getTotalNumInputChannels();
+    resolvedDevice.audioOutputChannels = instance->getTotalNumOutputChannels();
+
+    // Promote only, the rule the model's other two writers follow
+    // (PluginManagerSync::updateDeviceCapabilityFlags,
+    // applyCachedCapabilitiesToDevice). A raw AudioProcessor::acceptsMidi() is
+    // narrower than what the incumbent engine asks -- it takes MIDI input for
+    // plugins whose processor does not advertise it -- so assigning it here
+    // clears a saved true and PlanCompiler stops routing MIDI to the device.
+    if (!resolvedDevice.isInstrument && (instance->acceptsMidi() || instance->isMidiEffect()))
+        resolvedDevice.canReceiveMidi = true;
+    resolvedDevice.producesMidi = instance->producesMidi() || instance->isMidiEffect();
 
     auto restored = magda::snapshotHostParameters(*instance);
 
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
@@ -216,34 +216,33 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
     // The model first, before anything is done with the instance. Everything
     // below restores from it, and it is a different object from the one the
     // load was requested with: seconds have passed.
-    const auto* device = currentDevice ? currentDevice() : nullptr;
+    const auto* device = currentDevice ? currentDevice(requested.deviceId) : nullptr;
+
+    const auto& requestedName = requested.resolvedIdentity.name;
 
     if (device == nullptr)
         return {.device = {},
-                .failure = "the device was removed while \"" + requested.name + "\" was loading"};
+                .failure = "the device was removed while \"" + requestedName + "\" was loading"};
+
+    // This is the identity boundary. Scan metadata may have changed in every
+    // field while the plugin loaded, including the role the plan compiles from;
+    // those are facts learned about this assignment, not another assignment.
+    // A replacement DeviceInfo carries a newly authored generation even when
+    // it is another variant in the same bundle or has the same display name.
+    if (device->id != requested.deviceId ||
+        device->pluginAssignmentGeneration != requested.assignmentGeneration)
+        return {.device = {},
+                .failure = "the device changed plugin while \"" + requestedName + "\" was loading"};
 
     if (instance == nullptr)
         return {.device = {},
                 .failure = "external plugin \"" + device->name + "\" could not be loaded: " +
                            (error.isNotEmpty() ? error : "no reason given")};
 
-    // And it has to still be the same plugin. A device whose plugin was
-    // replaced while this one loaded is not waiting for this answer, and
-    // restoring onto it would put one plugin's saved patch on another.
-    //
-    // The model's identity on both sides. The scanned description's identifier
-    // carries the plugin's numeric uid and a device's saved fields do not, so
-    // comparing this against that would find every unchanged plugin changed and
-    // refuse every load there is.
-    if (!magda::savedPluginIdentity(*device).namesSamePluginAs(requested.identity))
-        return {.device = {},
-                .failure =
-                    "the device changed plugin while \"" + requested.name + "\" was loading"};
-
     return adaptExternalPluginInstance(std::move(instance), *device, offlineRender);
 }
 
-void createEngineExternalDeviceAsync(const magda::DeviceInfo& device,
+void createEngineExternalDeviceAsync(magda::DeviceInfo& device,
                                      const ExternalPluginServices& services, bool offlineRender,
                                      CurrentDeviceLookup currentDevice,
                                      std::function<void(ExternalDeviceResult)> completed) {
@@ -256,13 +255,18 @@ void createEngineExternalDeviceAsync(const magda::DeviceInfo& device,
         return;
     }
 
-    // Nothing of the model is captured here. What the callback needs from it is
-    // read when the callback runs, because by then the project has had seconds
-    // to move on: see CurrentDeviceLookup.
+    // This is synchronous on purpose. A caller compiles the native plan after
+    // starting the loads, while their instances are still in flight; it needs
+    // the resolved instrument role and channel topology now, not at completion.
+    // These are corrected facts about the existing assignment, so its stable
+    // generation is preserved.
+    magda::applyResolvedPluginDescription(device, resolved.description);
+
     services.formats->createPluginInstanceAsync(
         resolved.description, services.context.sampleRate, services.context.maxBlockSize,
-        [requested = RequestedPlugin{.identity = magda::savedPluginIdentity(device),
-                                     .name = resolved.description.name},
+        [requested = RequestedPlugin{.deviceId = device.id,
+                                     .assignmentGeneration = device.pluginAssignmentGeneration,
+                                     .resolvedIdentity = resolved.description},
          currentDevice = std::move(currentDevice), offlineRender, completed = std::move(completed)](
             std::unique_ptr<juce::AudioPluginInstance> instance, const juce::String& error) {
             completed(completeExternalPluginLoad(std::move(instance), error, requested,
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
@@ -3,6 +3,7 @@
 #include <utility>
 
 #include "core/DeviceState.hpp"
+#include "core/PluginCapabilities.hpp"
 #include "plugin_manager/ExternalPluginLookup.hpp"
 #include "plugin_manager/ExternalPluginState.hpp"
 #include "plugins/InternalPluginRegistry.hpp"
@@ -99,6 +100,15 @@ ExternalDeviceResult adaptExternalPluginInstance(
     bool offlineRender) {
     instance->enableAllBuses();
 
+    // Scan descriptions report the format's default buses, not the instance
+    // after the host has enabled its sidechains and extra outputs. These are
+    // the only channel counts safe to compile a live plan from.
+    auto resolvedDevice = device;
+    resolvedDevice.audioInputChannels = instance->getTotalNumInputChannels();
+    resolvedDevice.audioOutputChannels = instance->getTotalNumOutputChannels();
+    resolvedDevice.canReceiveMidi = !resolvedDevice.isInstrument && instance->acceptsMidi();
+    resolvedDevice.producesMidi = instance->producesMidi() || instance->isMidiEffect();
+
     // The saved array, then the chunk over it, then what that left behind. The
     // order and the reasons are ExternalPluginState.hpp's; what matters here is
     // that all three happen before the adapter exists, so the adapter never has
@@ -110,10 +120,11 @@ ExternalDeviceResult adaptExternalPluginInstance(
 
     auto restored = magda::snapshotHostParameters(*instance);
 
-    return {.device =
-                std::make_unique<EngineExternalDevice>(std::move(instance), device, offlineRender),
+    return {.device = std::make_unique<EngineExternalDevice>(std::move(instance), resolvedDevice,
+                                                             offlineRender),
             .failure = {},
-            .restoredParameters = std::move(restored)};
+            .restoredParameters = std::move(restored),
+            .resolvedDevice = std::move(resolvedDevice)};
 }
 
 namespace {
@@ -125,32 +136,16 @@ juce::String describeMissingPlugin(const magda::DeviceInfo& device) {
            ") is not installed on this machine";
 }
 
-/// What the two entry points check before either asks a format manager for
-/// anything: the services are complete, and the scan knows the plugin.
-///
-/// Returns the description to load, or the reason there is not one.
-struct ResolvedPlugin {
-    juce::PluginDescription description;
-    juce::String failure;
-};
+/// Apply only a role JUCE's installed description can author. MIDI and Analysis
+/// are explicit MAGDA roles, and replacing either with Effect/Instrument would
+/// change the plan rather than enrich it.
+void applyInstalledRole(magda::DeviceInfo& device, bool isInstrument) {
+    if (device.deviceType == magda::DeviceType::MIDI ||
+        device.deviceType == magda::DeviceType::Analysis)
+        return;
 
-ResolvedPlugin resolvePlugin(const magda::DeviceInfo& device,
-                             const ExternalPluginServices& services) {
-    if (services.formats == nullptr || services.knownPlugins == nullptr)
-        return {.description = {},
-                .failure = "no plugin formats or scan results were given to the engine"};
-
-    const auto match = magda::matchInstalledPlugin(device, *services.knownPlugins);
-
-    // Unfound is refused rather than attempted. The app tries the saved
-    // description anyway and lets the format's own lookup have a go, which is
-    // right for a session where a user is watching and can be told; a render is
-    // not that, and a plugin resolved by a route nothing recorded is the kind
-    // of difference a null-diff corpus cannot attribute afterwards.
-    if (!match.found)
-        return {.description = {}, .failure = describeMissingPlugin(device)};
-
-    return {.description = match.description, .failure = {}};
+    device.isInstrument = isInstrument;
+    device.deviceType = isInstrument ? magda::DeviceType::Instrument : magda::DeviceType::Effect;
 }
 
 }  // namespace
@@ -184,6 +179,37 @@ bool isRegisteredDevice(const juce::String& pluginId) {
            compiled::findCompiledPluginSpec(pluginId) != nullptr;
 }
 
+ExternalPluginResolution resolveEngineExternalPlugin(const magda::DeviceInfo& device,
+                                                     const ExternalPluginServices& services) {
+    ExternalPluginResolution resolved{.planDevice = device};
+
+    if (services.formats == nullptr || services.knownPlugins == nullptr) {
+        resolved.failure = "no plugin formats or scan results were given to the engine";
+        return resolved;
+    }
+
+    const auto match = magda::matchInstalledPlugin(device, *services.knownPlugins);
+
+    // Unfound is refused rather than attempted. The app tries the saved
+    // description anyway and lets the format's own lookup have a go, which is
+    // right for a session where a user is watching and can be told; a render is
+    // not that, and a plugin resolved by a route nothing recorded is the kind
+    // of difference a null-diff corpus cannot attribute afterwards.
+    if (!match.found) {
+        resolved.failure = describeMissingPlugin(device);
+        return resolved;
+    }
+
+    resolved.description = match.description;
+    applyInstalledRole(resolved.planDevice, match.description.isInstrument);
+
+    // Keep the project's identity and preference key. The capability cache is
+    // nevertheless queried with the exact installed identity resolution found.
+    const auto resolvedIdentifier = match.description.createIdentifierString();
+    magda::applyCachedCapabilitiesToDevice(resolved.planDevice, resolvedIdentifier);
+    return resolved;
+}
+
 bool isInstalledExternalPlugin(const magda::DeviceInfo& device,
                                const juce::KnownPluginList& knownPlugins) {
     return magda::matchInstalledPlugin(device, knownPlugins).found;
@@ -192,7 +218,7 @@ bool isInstalledExternalPlugin(const magda::DeviceInfo& device,
 ExternalDeviceResult createEngineExternalDevice(const magda::DeviceInfo& device,
                                                 const ExternalPluginServices& services,
                                                 bool offlineRender) {
-    const auto resolved = resolvePlugin(device, services);
+    const auto resolved = resolveEngineExternalPlugin(device, services);
     if (resolved.failure.isNotEmpty())
         return {.device = {}, .failure = resolved.failure};
 
@@ -205,7 +231,7 @@ ExternalDeviceResult createEngineExternalDevice(const magda::DeviceInfo& device,
                 .failure = "external plugin \"" + device.name + "\" could not be loaded: " +
                            (error.isNotEmpty() ? error : "no reason given")};
 
-    return adaptExternalPluginInstance(std::move(instance), device, offlineRender);
+    return adaptExternalPluginInstance(std::move(instance), resolved.planDevice, offlineRender);
 }
 
 ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPluginInstance> instance,
@@ -218,7 +244,7 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
     // load was requested with: seconds have passed.
     const auto* device = currentDevice ? currentDevice(requested.deviceId) : nullptr;
 
-    const auto& requestedName = requested.resolvedIdentity.name;
+    const auto& requestedName = requested.displayName;
 
     if (device == nullptr)
         return {.device = {},
@@ -229,8 +255,7 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
     // those are facts learned about this assignment, not another assignment.
     // A replacement DeviceInfo carries a newly authored generation even when
     // it is another variant in the same bundle or has the same display name.
-    if (device->id != requested.deviceId ||
-        device->pluginAssignmentGeneration != requested.assignmentGeneration)
+    if (device->pluginAssignmentGeneration != requested.assignmentGeneration)
         return {.device = {},
                 .failure = "the device changed plugin while \"" + requestedName + "\" was loading"};
 
@@ -239,39 +264,36 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
                 .failure = "external plugin \"" + device->name + "\" could not be loaded: " +
                            (error.isNotEmpty() ? error : "no reason given")};
 
-    return adaptExternalPluginInstance(std::move(instance), *device, offlineRender);
+    auto resolvedDevice = *device;
+    applyInstalledRole(resolvedDevice, requested.resolvedIsInstrument);
+    return adaptExternalPluginInstance(std::move(instance), resolvedDevice, offlineRender);
 }
 
-void createEngineExternalDeviceAsync(magda::DeviceInfo& device,
-                                     const ExternalPluginServices& services, bool offlineRender,
-                                     CurrentDeviceLookup currentDevice,
-                                     std::function<void(ExternalDeviceResult)> completed) {
+ExternalPluginResolution createEngineExternalDeviceAsync(
+    const magda::DeviceInfo& device, const ExternalPluginServices& services, bool offlineRender,
+    CurrentDeviceLookup currentDevice, std::function<void(ExternalDeviceResult)> completed) {
     jassert(completed != nullptr);
     jassert(currentDevice != nullptr);
 
-    const auto resolved = resolvePlugin(device, services);
+    auto resolved = resolveEngineExternalPlugin(device, services);
     if (resolved.failure.isNotEmpty()) {
         completed({.device = {}, .failure = resolved.failure});
-        return;
+        return resolved;
     }
-
-    // This is synchronous on purpose. A caller compiles the native plan after
-    // starting the loads, while their instances are still in flight; it needs
-    // the resolved instrument role and channel topology now, not at completion.
-    // These are corrected facts about the existing assignment, so its stable
-    // generation is preserved.
-    magda::applyResolvedPluginDescription(device, resolved.description);
 
     services.formats->createPluginInstanceAsync(
         resolved.description, services.context.sampleRate, services.context.maxBlockSize,
         [requested = RequestedPlugin{.deviceId = device.id,
                                      .assignmentGeneration = device.pluginAssignmentGeneration,
-                                     .resolvedIdentity = resolved.description},
+                                     .displayName = device.name,
+                                     .resolvedIsInstrument = resolved.description.isInstrument},
          currentDevice = std::move(currentDevice), offlineRender, completed = std::move(completed)](
             std::unique_ptr<juce::AudioPluginInstance> instance, const juce::String& error) {
             completed(completeExternalPluginLoad(std::move(instance), error, requested,
                                                  currentDevice, offlineRender));
         });
+
+    return resolved;
 }
 
 }  // namespace magda::daw::audio::engine_adapter
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.hpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.hpp
@@ -186,22 +186,24 @@ ExternalDeviceResult createEngineExternalDevice(const magda::DeviceInfo& device,
  * means the device is gone, and the load is abandoned rather than completed
  * against a project that no longer has it.
  */
-using CurrentDeviceLookup = std::function<const magda::DeviceInfo*()>;
+using CurrentDeviceLookup = std::function<const magda::DeviceInfo*(magda::DeviceId deviceId)>;
 
 /**
  * @brief What an asynchronous load remembers about what it asked for.
  *
- * The identity is the model's, taken when the load was requested, so that
- * comparing it with the model's identity at completion compares like with like.
- * Not the scanned description's identifier: that carries the plugin's numeric
- * uid, a device's saved fields do not, and comparing the two would call every
- * unchanged plugin a change.
+ * The device id and assignment generation are the stable question the loader
+ * is answering. Saved plugin metadata is not: resolving a moved plugin or an
+ * imported DAWproject legitimately corrects it before this request completes.
+ *
+ * @ref resolvedIdentity is the exact installed description handed to JUCE. It
+ * is retained so the request records the precise variant that is in flight;
+ * completion does not try to re-resolve mutable model fields and infer whether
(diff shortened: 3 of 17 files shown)


magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
+    device.isInstrument = description.isInstrument;
+    if (device.deviceType != DeviceType::MIDI)
+        device.deviceType = description.isInstrument ? DeviceType::Instrument : DeviceType::Effect;
+    device.uniqueId = description.createIdentifierString();
+    device.fileOrIdentifier = description.fileOrIdentifier;
+    device.audioInputChannels = description.numInputChannels;
+    device.audioOutputChannels = description.numOutputChannels;
```
lucaromagnoli: `audioOutputChannels` is written straight from the scan description with no `> 0` guard. `TrackSerializer.cpp:756-761` refuses exactly this value on the way in ("Checked on the way in too, so a hand-edited or older project cannot silence a chain either. Zero comes from the live plugin or not at all."), and `PlanCompiler.cpp:711` turns a zero into `outputWidth = 0`, i.e. an op that emits nothing.

A `KnownPluginList` entry whose scanned default layout reports no output bus (MIDI-only VST3s, or a plugin whose main out is only enabled by `enableAllBuses()`) will silently mute the chain slot the moment the device resolves, with no way back short of a rescan. The guard the serializer already has belongs here too.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
+    device.name = description.name;
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
+    device.isInstrument = description.isInstrument;
+    if (device.deviceType != DeviceType::MIDI)
+        device.deviceType = description.isInstrument ? DeviceType::Instrument : DeviceType::Effect;
+    device.uniqueId = description.createIdentifierString();
+    device.fileOrIdentifier = description.fileOrIdentifier;
+    device.audioInputChannels = description.numInputChannels;
```
lucaromagnoli: These two come from the scan's *default* bus layout, but `adaptExternalPluginInstance()` (EngineDeviceFactory.cpp:96) calls `instance->enableAllBuses()` first, and says why: "a plugin whose sidechain or second output bus is disabled reports channels it does not have, and everything downstream -- the adapter's own channel adaptation, the plan's declared widths -- is read off those numbers." The field's own doc in DeviceInfo.hpp agrees: "read from the live plugin via getChannelNames".

After this change the plan's declared widths come from the pre-`enableAllBuses` scan while the adapter adapts against the post-`enableAllBuses` instance. Any plugin with a bus disabled by default (sidechain compressors, multi-out instruments) now gets a plan width that disagrees with the live channel count the adapter sees.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
+void applyResolvedPluginDescription(DeviceInfo& device,
+                                    const juce::PluginDescription& description) {
+    device.name = description.name;
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
+    device.isInstrument = description.isInstrument;
+    if (device.deviceType != DeviceType::MIDI)
+        device.deviceType = description.isInstrument ? DeviceType::Instrument : DeviceType::Effect;
+    device.uniqueId = description.createIdentifierString();
```
lucaromagnoli: Rewriting `uniqueId` here (and `pluginId` on line 43) changes the key that `PluginPreferences::identifierForDevice()` (PluginPreferences.cpp:209) and `PluginCapabilityCache::identifierForDevice()` (PluginCapabilities.cpp:143) derive from the device: both prefer `uniqueId`, falling back to `pluginId` then `fileOrIdentifier`.

Nothing here re-runs `applyCachedCapabilitiesToDevice()` or re-reads the user-global preferences under the new key, so an imported device that had no `uniqueId` silently loses its stamped drum-kit rows and AI prompt, and keeps whatever stale `canReceiveMidi` / `canSidechain` / `producesMidi` it was constructed with. That last one undercuts the header's own claim: `PlanCompiler.cpp:47` routes MIDI on `device.isInstrument || device.canReceiveMidi || deviceType == MIDI`, and this function only fixes the first term. A MIDI-receiving effect stays unrouted.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
-            .isInstrument = device.isInstrument};
+void applyResolvedPluginDescription(DeviceInfo& device,
+                                    const juce::PluginDescription& description) {
+    device.name = description.name;
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
+    device.isInstrument = description.isInstrument;
+    if (device.deviceType != DeviceType::MIDI)
+        device.deviceType = description.isInstrument ? DeviceType::Instrument : DeviceType::Effect;
```
lucaromagnoli: Only `DeviceType::MIDI` is guarded, but the enum has four values (DeviceInfo.hpp:114) and `Analysis` is load-bearing for the plan: `PlanCompiler.cpp:43` reads `deviceType == DeviceType::Analysis` as `isTransparentTap()`, which forces full-stereo pass-through widths, suppresses the delta button and skips the device's own width clamping.

A device typed `Analysis` that goes through resolution is rewritten to `Effect` and stops being a transparent tap. Guard `Analysis` alongside `MIDI`, or key the rewrite on the device's role actually being unknown rather than on "not MIDI".

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
-            .fileOrIdentifier = device.fileOrIdentifier,
-            .format = device.format,
-            .isInstrument = device.isInstrument};
+void applyResolvedPluginDescription(DeviceInfo& device,
+                                    const juce::PluginDescription& description) {
+    device.name = description.name;
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
+    device.isInstrument = description.isInstrument;
```
lucaromagnoli: Flipping `isInstrument` on an already-placed device bypasses every guard that decided the placement. `TrackManager::addDeviceToChainByPath()` refuses a device when `!track->canHostInstrument() && device.isInstrument` (TrackManagerDevices.cpp:378 and :431), and `PlanCompiler.cpp:98` / `:1127` end the primary-instrument search at the first non-bypassed `isInstrument` device.

The header names the case directly: an imported DAWproject effect that resolves to an instrument, because "Bitwig exports Serum, an instrument, as an audio effect". Correcting the role in place puts an instrument on a track that refuses instruments, or inserts a second instrument ahead of the real one in the chain, with no re-validation of either invariant.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
-SavedPluginIdentity savedPluginIdentity(const DeviceInfo& device) {
-    return {.identifier = device.uniqueId,
-            .name = device.name,
-            .manufacturer = device.manufacturer,
-            .fileOrIdentifier = device.fileOrIdentifier,
-            .format = device.format,
-            .isInstrument = device.isInstrument};
+void applyResolvedPluginDescription(DeviceInfo& device,
+                                    const juce::PluginDescription& description) {
+    device.name = description.name;
```
lucaromagnoli: `device.name` is the display name, and it is not always the scanned name. `LegacyDeviceAliases.cpp:574-577` deliberately writes `retired->targetDisplayName` over it, and an imported project carries the exporting host's own device label. Overwriting it unconditionally on resolution throws both away.

It is also the one field in this function where "the installed plugin's fact wins" is not self-evidently right: nothing downstream reads `name` for routing, only for chrome and error strings.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
+    device.pluginId = description.createIdentifierString();
+    device.manufacturer = description.manufacturerName;
+    device.format = pluginFormatFromName(description.pluginFormatName);
```
lucaromagnoli: `pluginFormatFromName()` (DeviceInfo.hpp:100) swallows an unrecognised format name and returns `PluginFormat::VST3`. `maybePluginFormatFromName()` sits directly above it and exists precisely to express "unknown", which is what this call site wants.

A wrong format is not inert here: `describeSavedPlugin()` feeds `device.format` back through `formatNameOf()` into passes 3 and 5 of `matchInstalledPlugin()`, both of which require `known.pluginFormatName` to match. A mis-stamped format makes the device fail to re-resolve by name on the next open, which is the failure mode this file exists to prevent.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
```diff
@@ -37,13 +37,19 @@ juce::PluginDescription describeSavedPlugin(const DeviceInfo& device) {
...
 }
 
-SavedPluginIdentity savedPluginIdentity(const DeviceInfo& device) {
-    return {.identifier = device.uniqueId,
-            .name = device.name,
-            .manufacturer = device.manufacturer,
-            .fileOrIdentifier = device.fileOrIdentifier,
-            .format = device.format,
-            .isInstrument = device.isInstrument};
+void applyResolvedPluginDescription(DeviceInfo& device,
```
lucaromagnoli: This description-to-DeviceInfo mapping already exists four times, and none of them was updated to call the new canonical helper:

- `plugin_api_live.cpp:24-40` (`toDeviceInfo`)
- `DeviceCustomUIManager.cpp:451-462` (`externalPadDevice`)
- `TrackChainContent.cpp:2970-2985`
- `ChainPanel.cpp:808` onwards

The last two even hand-roll the VST3/AU/LV2 switch instead of calling `pluginFormatFromName()`. Adding a fifth copy in the file whose own header opens by complaining that "the app itself had two [opinions]" about plugin identity leaves the divergence in place rather than closing it.

Minor, same function: `description.createIdentifierString()` is built twice (lines 43 and 49) for the same string.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -216,34 +216,33 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
...
-    if (!magda::savedPluginIdentity(*device).namesSamePluginAs(requested.identity))
-        return {.device = {},
-                .failure =
-                    "the device changed plugin while \"" + requested.name + "\" was loading"};
-
     return adaptExternalPluginInstance(std::move(instance), *device, offlineRender);
 }
 
-void createEngineExternalDeviceAsync(const magda::DeviceInfo& device,
+void createEngineExternalDeviceAsync(magda::DeviceInfo& device,
```
lucaromagnoli: The enrichment is only on the asynchronous entry point. `createEngineExternalDevice()` (line 192) resolves the same description and then hands the *unenriched* `const DeviceInfo&` straight to `adaptExternalPluginInstance()`.

That is the path the header reserves for "a render waiting for the answer -- an offline bounce, a corpus case", which is exactly where the plan is compiled from `device.isInstrument`, `audioInputChannels` and `audioOutputChannels`. So #2252's acceptance criterion "Resolved role and capabilities are applied before plan compilation" holds for one of the two entry points: an offline bounce still compiles a corrected instrument as an effect that receives no MIDI, and at whatever width the project happened to save.

Both entry points already funnel through `resolvePlugin()`; that, or a shared step both call, is where this belongs.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -256,13 +255,18 @@ void createEngineExternalDeviceAsync(const magda::DeviceInfo& device,
...
 
-    // Nothing of the model is captured here. What the callback needs from it is
-    // read when the callback runs, because by then the project has had seconds
-    // to move on: see CurrentDeviceLookup.
+    // This is synchronous on purpose. A caller compiles the native plan after
+    // starting the loads, while their instances are still in flight; it needs
+    // the resolved instrument role and channel topology now, not at completion.
+    // These are corrected facts about the existing assignment, so its stable
+    // generation is preserved.
+    magda::applyResolvedPluginDescription(device, resolved.description);
```
lucaromagnoli: This mutates the caller's model `DeviceInfo` before it is known whether the plugin instantiates at all. If `createPluginInstanceAsync` then fails, the model has already had `name`, `pluginId`, `uniqueId`, `fileOrIdentifier`, `manufacturer`, `format`, `isInstrument`, `deviceType` and both channel counts rewritten to describe a plugin that never loaded, and nothing rolls it back. The device is left permanently re-pointed at a scan entry it could not instantiate, and the next save persists it.

The synchronous-failure branch above (lines 253-256) has the mirror problem now that the parameter is a mutable model reference rather than a copy: `completed()` runs re-entrantly while the caller still holds the `DeviceInfo&` it passed in, so a handler that removes the device (the natural response to "not installed on this machine") leaves that reference dangling for the rest of this function and for the caller.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -216,34 +216,33 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
...
         return {.device = {},
-                .failure = "the device was removed while \"" + requested.name + "\" was loading"};
+                .failure = "the device was removed while \"" + requestedName + "\" was loading"};
+
+    // This is the identity boundary. Scan metadata may have changed in every
+    // field while the plugin loaded, including the role the plan compiles from;
+    // those are facts learned about this assignment, not another assignment.
+    // A replacement DeviceInfo carries a newly authored generation even when
+    // it is another variant in the same bundle or has the same display name.
+    if (device->id != requested.deviceId ||
```
lucaromagnoli: `device->id != requested.deviceId` cannot fire through the intended contract: `currentDevice` is invoked fourteen lines above with `requested.deviceId`, so any lookup that honours its argument returns either a device with that id or null.

No test exercises it either, because every `CurrentDeviceLookup` in test_engine_external_device.cpp discards the new parameter (`[&](magda::DeviceId) { return &model; }`). So the id half of the guard is unreachable in production and untested in the suite. Either drop it and let the generation carry the check, or state the lookup's contract in `CurrentDeviceLookup`'s doc and cover a lookup that returns a mismatched device.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -256,13 +255,18 @@ void createEngineExternalDeviceAsync(const magda::DeviceInfo& device,
...
+    // generation is preserved.
+    magda::applyResolvedPluginDescription(device, resolved.description);
+
     services.formats->createPluginInstanceAsync(
         resolved.description, services.context.sampleRate, services.context.maxBlockSize,
-        [requested = RequestedPlugin{.identity = magda::savedPluginIdentity(device),
-                                     .name = resolved.description.name},
+        [requested = RequestedPlugin{.deviceId = device.id,
+                                     .assignmentGeneration = device.pluginAssignmentGeneration,
+                                     .resolvedIdentity = resolved.description},
```
lucaromagnoli: `resolvedIdentity` is a whole `juce::PluginDescription` copied into every in-flight completion lambda, and the only thing anything reads from it is `.name`, for the two error strings on lines 221/226/234.

The header justifies it as "the request records the precise variant that is in flight", but nothing ever compares it, so it is bookkeeping that cannot be checked, paid for on every async load. The `juce::String name` it replaced carried the whole of the current behaviour at a fraction of the copy.

magda/daw/core/DeviceInfo.hpp
```diff
@@ -194,6 +207,13 @@ struct MeterInfo {
  */
 struct DeviceInfo {
     DeviceId id = INVALID_DEVICE_ID;
+
+    // Runtime identity of the plugin assignment in this slot. Copies preserve
+    // it; a newly constructed replacement gets another monotonically increasing
+    // value. Scan enrichment, parameter edits and preset restores must not
+    // change it. Transient by design: serializers neither read nor write it.
+    std::uint64_t pluginAssignmentGeneration = nextPluginAssignmentGeneration();
```
lucaromagnoli, line 215: The invariant this PR rests on ("a newly constructed replacement gets another monotonically increasing value") is enforced by nothing but this default member initializer, and the codebase already reassigns a slot's plugin *in place*, which keeps the old token:

- `LegacyDeviceAliases.cpp:580-581` writes a different `pluginId`, `uniqueId`, `parameters` and `pluginState` onto the existing device.
- `DrumGridPads.cpp:243` swaps `pluginId` to a successor device.

And the converse: `TrackManager::prepareNewDevice()` (TrackManager.cpp:3505-3506) copies rather than authors, so a device added from a browser template inherits the template's generation instead of getting a fresh one.

Neither direction is reachable from a live load today because nothing outside tests calls `createEngineExternalDeviceAsync()`. But the moment it is wired up (which #2252 says is blocked on this change), an in-place reassignment will pass the generation check and restore one plugin's saved patch onto another, which is the exact failure the removed field comparison was there to catch. This wants a single choke point on `DeviceInfo` that bumps the token, not a comment asking callers to remember.

tests/test_engine_external_device.cpp
```diff
@@ -1296,91 +1335,104 @@ TEST_CASE("A scanned plugin's own identifier is not read as a change", "[engine]
...
     model.isInstrument = false;
     model.uniqueId = {};  // imported: nothing scanned it
 
-    const adapter::RequestedPlugin requested{.identity = magda::savedPluginIdentity(model),
-                                             .name = model.name};
+    const auto requested = requestFor(model);
 
-    // The swap: the instrument out of the same bundle.
-    model.isInstrument = true;
+    auto replacement = externalDeviceSaving(1.0f, 0.5f);
```
lucaromagnoli: These replacement tests no longer test production behaviour. The old versions mutated the model's identity fields, which is what a real replacement path might plausibly do; the new ones construct a fresh `DeviceInfo` in the test body and move it in, so what is asserted is that *the test* built a new object.

Nothing in the codebase guarantees a slot replacement does the same (see `LegacyDeviceAliases.cpp:580` and `DrumGridPads.cpp:243`, which reassign in place). A version that starts from `model` and reassigns its plugin the way production does would be the one with teeth; as written, both "is a change" cases pass no matter what `completeExternalPluginLoad` compares.

tests/test_engine_external_device.cpp
```diff
@@ -1188,7 +1203,7 @@ TEST_CASE("The asynchronous entry point reports a missing plugin the same way",
     bool called = false;
     adapter::ExternalDeviceResult delivered;
     adapter::createEngineExternalDeviceAsync(
-        missing, services, false, [&]() { return &missing; },
+        missing, services, false, [&](magda::DeviceId) { return &missing; },
```
lucaromagnoli, line 1423: This is the only test of `createEngineExternalDeviceAsync()`, and it returns from `resolvePlugin()` before ever reaching the new synchronous enrichment (EngineDeviceFactory.cpp:263). So the PR's headline behaviour, "apply the exact resolved plugin metadata, role, and channel topology before native plan compilation", has no coverage at the entry point that performs it.

`test_external_plugin_lookup.cpp` covers `applyResolvedPluginDescription` in isolation, but nothing asserts that the async entry point calls it, that the generation survives the call, or that the enriched values are what ends up captured in `RequestedPlugin`. A test with an installed stub in `KnownPluginList` that checks `device.isInstrument` / `audioOutputChannels` immediately after the call returns would close it.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
lucaromagnoli: Addressed in d78996709: async start never mutates the model. It copies the request token/name/role before dispatch, does not touch the DeviceInfo reference after the loader call, and failures return no publishable resolvedDevice.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: resolution no longer flips the role on an already placed persistent model. It returns a transient planDevice for placement validation/compilation, and only a successful load returns a resolvedDevice that the caller may validate and publish.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: applyInstalledRole() preserves both DeviceType::MIDI and DeviceType::Analysis. Coverage now asserts that a scanned instrument cannot replace an Analysis role.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: scan output counts are no longer copied into DeviceInfo at all. The result now reports output channels only from the live instance after enableAllBuses(), so a zero/default scan count cannot mute the plan.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: resolution leaves saved/default channel widths unchanged. adaptExternalPluginInstance() enables every bus first and then publishes getTotalNumInputChannels()/getTotalNumOutputChannels() in the successful result.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: applyResolvedPluginDescription() and its duplicate identity/format mapping were removed entirely. Resolution now applies only the narrow role and capability facts needed by the plan.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: the resolved path no longer rewrites DeviceInfo::format, so there is no unknown-format-to-VST3 fallback in this flow.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: device.name is never replaced by the scan name. The planning and successful runtime snapshots retain the project/user display label.

tests/test_engine_external_device.cpp
lucaromagnoli: Addressed in d78996709: stale-load tests now perform production-style in-place reassignment by calling beginNewPluginAssignment() before changing bundle role/vendor/path, instead of constructing a fresh DeviceInfo and moving it over the model.

magda/daw/audio/plugin_manager/ExternalPluginLookup.cpp
lucaromagnoli: Addressed in d78996709: resolved loading no longer rewrites pluginId, uniqueId, or the preference key. The transient planning copy queries PluginCapabilityCache with the exact resolved identifier, cached false values now clear stale MIDI flags, and successful creation replaces them with live processor capabilities.

tests/test_engine_external_device.cpp
lucaromagnoli: Addressed in d78996709: a StubFormat registered with AudioPluginFormatManager now drives the real async entry point through successful completion. The test checks immediate transient role enrichment, no persistent mutation, the exact description requested, and live channel facts returned on completion.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
lucaromagnoli: Addressed in d78996709: both blocking and async creation now share resolveEngineExternalPlugin(). The blocking path adapts from its resolved planning snapshot and returns live topology/capabilities after successful creation.

magda/daw/core/DeviceInfo.hpp
lucaromagnoli: Addressed in d78996709: DeviceInfo::beginNewPluginAssignment() is now the explicit choke point. prepareNewDevice(), copied pad devices, LegacyDeviceAliases migration, and DrumGridPads successor replacement all call it. A TrackManager test proves two additions from one reusable browser template get distinct tokens.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
lucaromagnoli: Addressed in d78996709: the redundant device->id comparison was removed. CurrentDeviceLookup is documented and tested as the id boundary; assignmentGeneration is the stale-assignment boundary.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
lucaromagnoli: Addressed in d78996709: RequestedPlugin no longer retains a PluginDescription. It stores only deviceId, assignmentGeneration, displayName, and the resolved instrument bit that completion actually needs.

magda/daw/core/PluginCapabilities.cpp
```diff
@@ -269,14 +269,17 @@ bool supportsSidechainRoutingMenu(const DeviceInfo& device) {
...
+void applyCachedCapabilitiesToDevice(DeviceInfo& device, const juce::String& pluginIdentifier) {
+    auto snapshot = PluginCapabilityCache::getInstance().find(pluginIdentifier);
     if (!snapshot)
         return;
 
     device.producesMidi = snapshot->hasMidiOutput;
-    if (!device.isInstrument && snapshot->hasMidiInput)
-        device.canReceiveMidi = true;
+    if (!device.isInstrument)
+        device.canReceiveMidi = snapshot->hasMidiInput;
```
lucaromagnoli: **This flips the flag from promote-only to assign, and it can erase saved MIDI routing.**

The old body was `if (!device.isInstrument && snapshot->hasMidiInput) device.canReceiveMidi = true;` - it could only ever promote. The new body assigns, so a cache entry with `hasMidiInput == false` now clears a `canReceiveMidi` that the project saved as true.

That matters on the load path: `TrackSerializer.cpp:744` restores the saved `canReceiveMidi`, and `TrackSerializer.cpp:807` then calls this function, which overwrites it. `PlanCompiler.cpp:47` gates MIDI delivery on `device.canReceiveMidi`, so the effect is a device that silently stops receiving MIDI after a reopen.

The invariant is written down elsewhere in the codebase, in `PluginManagerSync.cpp` just above `clearStaleFaustAudioSidechain`:

> Faust's processor owns a dynamic canSidechain flag, while the generic capability updater only ever promotes flags to true. Keep the type guard so a stale serialized flag on another plugin cannot erase valid routing.

And `updateDeviceCapabilityFlags` (`PluginManagerSync.cpp:134`) still promotes-only from the *same* snapshot, so the live sync sets true and the next add/deserialize clears it back.

Second problem on the same line: it drops the `hasMidiInputOverride` that `mergeSnapshotWithDevice` honours, so a `DeviceType::MIDI` device whose snapshot says no MIDI input now loses `canReceiveMidi` too.

magda/daw/core/TrackManager.cpp
```diff
@@ -3504,6 +3504,7 @@ void TrackManager::notifyDeviceAdded(const ChainNodePath& devicePath, const Devi
 
 DeviceInfo TrackManager::prepareNewDevice(const DeviceInfo& device) {
     DeviceInfo newDevice = device;
+    newDevice.beginNewPluginAssignment();
```
lucaromagnoli, line 3516: **`reassignChainElementIds` is the top-level counterpart of `rekeyPads` and did not get the same call.**

`TrackManagerDevices.cpp:1937-1941` is the shared walk that stamps fresh `DeviceId`s on duplicated, pasted and preset-inserted devices:

```cpp
auto& device = magda::getDevice(element);
const auto oldDeviceId = device.id;
device.id = allocateDeviceId();
```

No `beginNewPluginAssignment()`. The pad loop three lines away in this file got one; the top-level loop that is the reason the pad loop exists did not. So a duplicated track's external plugin is a separate live assignment with its own engine instance that carries the source's token.

`DeviceInfo.hpp` states the contract as "Every path that authors a replacement from a copy or mutates plugin identity in place must come through here", and nothing enforces it - no assert, no test. Right now the four call sites added here are the whole of the enforcement, and at least three copy-with-new-id paths are missing (see the other comments).

magda/daw/core/TrackManager.cpp
```diff
@@ -3530,6 +3531,7 @@ void TrackManager::rekeyPads(DeviceInfo& device, std::map<DeviceId, DeviceId>* r
 
             auto& padDevice = magda::getDevice(element);
             const auto oldId = padDevice.id;
+            padDevice.beginNewPluginAssignment();
```
lucaromagnoli: **`addDeviceToPostFx` / `addDeviceToMixerAnalysis` copy without minting, and their id spaces overlap the fx one.**

`TrackManager.cpp:2214-2216` and `2279-2281`:

```cpp
DeviceInfo newDevice = device;
newDevice.id = nextPostFxDeviceId_++;
applyCachedCapabilitiesToDevice(newDevice);
```

Copy, new id, no `beginNewPluginAssignment()`. Same in `duplicateTrack` (`TrackManager.cpp:1091` and `1093`).

That would be survivable if `(deviceId, generation)` were still distinct, but the three counters are independent and all start at 1 (`TrackManager.hpp:1269-1271`, reset together at `TrackManager.cpp:3307-3309`). Paste an external effect from the fx chain into post-fx while `nextPostFxDeviceId_` happens to be sitting on the source's id, and you have two live devices with an identical id *and* an identical token - which is exactly the pair `completeExternalPluginLoad` treats as proof the answer is still wanted.

The codebase already knows bare ids collide: `findUniqueBareDeviceIdMatch` (`TrackManagerDevices.cpp:275-289`) collects matches across fx, post-fx and mixer-analysis for master and every track and returns null unless there is exactly one.

magda/daw/audio/plugins/engine/EngineDeviceFactory.hpp
```diff
@@ -186,22 +218,25 @@ ExternalDeviceResult createEngineExternalDevice(const magda::DeviceInfo& device,
  * means the device is gone, and the load is abandoned rather than completed
  * against a project that no longer has it.
  */
-using CurrentDeviceLookup = std::function<const magda::DeviceInfo*()>;
+using CurrentDeviceLookup = std::function<const magda::DeviceInfo*(magda::DeviceId deviceId)>;
```
lucaromagnoli, line 221: **A bare `DeviceId` is not a unique address in this model, and the old signature did not need one.**

The previous `std::function<const DeviceInfo*()>` let the host close over whatever addressing it liked - a `ChainNodePath`, which *is* unique. The new signature hands it a `DeviceId` and the doc says "It resolves @p currentDevice by the captured id first", which pushes hosts toward exactly the lookup `findUniqueBareDeviceIdMatch` exists to refuse (`TrackManagerDevices.cpp:288`: `return matches.size() == 1 ? matches.front() : nullptr;`).

A pad device makes it worse: pads allocate from `nextFxDeviceId_` via `allocateDeviceId()`, so a pad device and a top-level device never collide, but a post-fx device and a top-level device do.

`ChainNodePath` in the `RequestedPlugin` would key the request on the slot the plan actually addresses, and the generation would then only have to distinguish successive assignments in one slot, which is what it is good at.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -99,6 +100,15 @@ ExternalDeviceResult adaptExternalPluginInstance(
     bool offlineRender) {
     instance->enableAllBuses();
 
+    // Scan descriptions report the format's default buses, not the instance
+    // after the host has enabled its sidechains and extra outputs. These are
+    // the only channel counts safe to compile a live plan from.
+    auto resolvedDevice = device;
+    resolvedDevice.audioInputChannels = instance->getTotalNumInputChannels();
+    resolvedDevice.audioOutputChannels = instance->getTotalNumOutputChannels();
+    resolvedDevice.canReceiveMidi = !resolvedDevice.isInstrument && instance->acceptsMidi();
```
lucaromagnoli: **Same promote-vs-assign problem as `applyCachedCapabilitiesToDevice`, and a narrower source of truth.**

`resolvedDevice.canReceiveMidi = !resolvedDevice.isInstrument && instance->acceptsMidi();` is an assignment off the raw processor flag alone. The codebase's canonical answer is `makePluginCapabilitySnapshot` (`PluginManagerSync.cpp:104-118`): `hasMidiInput = tracktionTakesMidiInput || processorAcceptsMidi`, and `updateDeviceCapabilityFlags` only ever promotes it.

Since the header documents `resolvedDevice` as something "a caller may validate and publish", a caller that does publish it will clear `canReceiveMidi` on any device whose instance reports `acceptsMidi() == false` but which MAGDA routes MIDI to today. That is a routing loss, not a metadata refresh.

`producesMidi` on the next line duplicates the `pluginProducesMidi()` formula (`producesMidi() || isMidiEffect()`) from `PluginManagerSync.cpp:85-90` rather than sharing it.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -184,6 +179,37 @@ bool isRegisteredDevice(const juce::String& pluginId) {
...
+        return resolved;
+    }
+
+    resolved.description = match.description;
+    applyInstalledRole(resolved.planDevice, match.description.isInstrument);
+
+    // Keep the project's identity and preference key. The capability cache is
+    // nevertheless queried with the exact installed identity resolution found.
+    const auto resolvedIdentifier = match.description.createIdentifierString();
+    magda::applyCachedCapabilitiesToDevice(resolved.planDevice, resolvedIdentifier);
```
lucaromagnoli, line 227: **The cache is queried with a key it is never written under, for the case this PR is about.**

The write side keys every snapshot on `PluginCapabilityCache::identifierForDevice(device)` (`PluginManagerSync.cpp:95`), which is `uniqueId` then `pluginId` then `fileOrIdentifier` (`PluginCapabilities.cpp:145-151`). This line queries with `match.description.createIdentifierString()`.

For a browser-added device those agree, because `uniqueId` was set from `createIdentifierString()`. For an imported DAWproject device - the one with an empty `uniqueId` that the surrounding comment is about - the cache entry is under the *file path*, and this lookup misses it entirely. The single-arg overload would have hit it.

A fallback to `identifierForDevice(device)` when the resolved key misses would keep both.

Separately: on the synchronous path and at async completion, `adaptExternalPluginInstance` overwrites both flags from the live instance, so this call only ever affects the returned `planDevice`. Worth saying so in the comment, otherwise it reads as if it feeds the created device.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -125,32 +136,16 @@ juce::String describeMissingPlugin(const magda::DeviceInfo& device) {
...
-    // description anyway and lets the format's own lookup have a go, which is
-    // right for a session where a user is watching and can be told; a render is
-    // not that, and a plugin resolved by a route nothing recorded is the kind
-    // of difference a null-diff corpus cannot attribute afterwards.
-    if (!match.found)
-        return {.description = {}, .failure = describeMissingPlugin(device)};
+/// Apply only a role JUCE's installed description can author. MIDI and Analysis
+/// are explicit MAGDA roles, and replacing either with Effect/Instrument would
+/// change the plan rather than enrich it.
+void applyInstalledRole(magda::DeviceInfo& device, bool isInstrument) {
```
lucaromagnoli, line 160: **Rewriting `isInstrument` / `deviceType` for the plan is a role correction the incumbent engine never makes.**

`updateDeviceCapabilityFlags` (`PluginManagerSync.cpp:131-139`) promotes `canSidechain`, `canReceiveMidi` and `producesMidi` from a live plugin and deliberately leaves `isInstrument` and `deviceType` alone. This function changes both for the plan device.

`PlanCompiler.cpp:47` reads `device.isInstrument || device.canReceiveMidi || device.deviceType == DeviceType::MIDI` to decide MIDI consumption, so for a project where the scan disagrees with the saved role, the native engine compiles a different topology than TE does from the same document. For a file whose stated job is a null-diff against the incumbent, that is a divergence introduced by the resolver rather than found by it.

There is also a reachable inconsistency: the model forbids instruments in post-fx (`addDeviceToPostFx` rejects `device.isInstrument`) and on tracks that cannot host one (`addDeviceToTrack`). A scan that reports the other half of a bundle as an instrument will produce a `planDevice` in a position the model would have refused, and nothing here validates that.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -99,6 +100,15 @@ ExternalDeviceResult adaptExternalPluginInstance(
     bool offlineRender) {
     instance->enableAllBuses();
 
+    // Scan descriptions report the format's default buses, not the instance
+    // after the host has enabled its sidechains and extra outputs. These are
+    // the only channel counts safe to compile a live plan from.
+    auto resolvedDevice = device;
+    resolvedDevice.audioInputChannels = instance->getTotalNumInputChannels();
```
lucaromagnoli: **The enrichment stops half way: widths include the sidechain, but `canSidechain` and `multiOutConfig` are not refreshed.**

`enableAllBuses()` above turns on the sidechain and the extra output buses, and these two lines then count them - the new test asserts `audioInputChannels == 3` for a 2-in plugin with a mono sidechain, and `audioOutputChannels == 6` for one with two extra stereo outs.

But `resolvedDevice.canSidechain` and `resolvedDevice.multiOutConfig.totalOutputChannels` keep whatever the project saved. The TE path sets `canSidechain` from the live plugin in the same breath as the channel counts (`PluginManagerSync.cpp:131-140`). A caller that publishes `resolvedDevice` as the header invites therefore gets a device that is three channels wide on the input with `canSidechain == false`, and a six-output device whose multi-out config still describes two - internally inconsistent in a way neither the model nor the incumbent produces.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -216,58 +242,58 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
...
-    // comparing this against that would find every unchanged plugin changed and
-    // refuse every load there is.
-    if (!magda::savedPluginIdentity(*device).namesSamePluginAs(requested.identity))
-        return {.device = {},
-                .failure =
-                    "the device changed plugin while \"" + requested.name + "\" was loading"};
-
-    return adaptExternalPluginInstance(std::move(instance), *device, offlineRender);
+    auto resolvedDevice = *device;
+    applyInstalledRole(resolvedDevice, requested.resolvedIsInstrument);
```
lucaromagnoli, line 286: **Completion applies the role but not the cached capabilities, so the completed device can disagree with the plan device compiled while it loaded.**

`resolveEngineExternalPlugin` builds `planDevice` with `applyInstalledRole` **and** `applyCachedCapabilitiesToDevice`. This path applies only the first. The gap is then partly papered over by `adaptExternalPluginInstance` overwriting both MIDI flags from the instance - but with the raw `acceptsMidi()` rather than the cache's `tracktionTakesMidiInput || processorAcceptsMidi`.

So a plugin that TE takes MIDI input for but whose `AudioProcessor::acceptsMidi()` is false gets `canReceiveMidi == true` in the plan compiled during the load and `false` in the device published at completion, and the re-prepare changes the topology under the executor. Resolution and completion should be deriving the device the same way.

lucaromagnoli (commented):
Review of the stable assignment-token change. Twelve findings, ranked; the first three are the ones I would fix before merge.

magda/daw/core/DrumGridPads.cpp
```diff
@@ -226,6 +226,7 @@ DeviceInfo deviceFromNode(const ds::Node& node) {
     // the only reason it had to read a nested plugin tree at all (#2207).
     if (const auto successor = legacy_devices::retiredDeviceSuccessor(savedType);
         successor.isNotEmpty()) {
+        device.beginNewPluginAssignment();
```
lucaromagnoli, line 229: **Dead call.** `deviceFromNode` declares `DeviceInfo device;` as a fresh local at the top of the function, so it already carries a token nothing else has - the default member initializer runs on every construction. Minting a second one here changes nothing and implies this path is doing identity work that it is not.

The same reasoning does not apply to `LegacyDeviceAliases.cpp:561`, which mutates a caller's device in place, so keep that one.

magda/daw/core/DeviceInfo.hpp
```diff
@@ -194,6 +207,24 @@ struct MeterInfo {
...
```
lucaromagnoli, line 215: **Minting at construction rather than at assignment is what leaves the contract unenforceable.**

Because the token is minted by the default member initializer, "a newly constructed replacement gets another value" is true for free, and "every path that authors a replacement from a copy must come through `beginNewPluginAssignment()`" is true only where somebody remembered - and three copy-with-new-id paths did not (see the `TrackManager.cpp` comments).

The fields that decide plugin identity are right here in the same struct (`pluginId`, `uniqueId`, `fileOrIdentifier`, `format`, `isInstrument`). Minting from a small setter that owns those, instead of from construction plus scattered manual calls, would make the property structural rather than a convention.

Minor, on the same line: every `DeviceInfo` default construction now does an atomic RMW, including throwaway temporaries, `std::vector<DeviceInfo>` growth and the `static const std::vector<PostFxChainElement> empty;` returns, and `DeviceInfo` is no longer constexpr-default-constructible.

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -216,58 +242,58 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
...
+ExternalPluginResolution createEngineExternalDeviceAsync(
+    const magda::DeviceInfo& device, const ExternalPluginServices& services, bool offlineRender,
+    CurrentDeviceLookup currentDevice, std::function<void(ExternalDeviceResult)> completed) {
     jassert(completed != nullptr);
     jassert(currentDevice != nullptr);
 
-    const auto resolved = resolvePlugin(device, services);
+    auto resolved = resolveEngineExternalPlugin(device, services);
     if (resolved.failure.isNotEmpty()) {
         completed({.device = {}, .failure = resolved.failure});
```
lucaromagnoli, line 298: **Completion timing is now inconsistent in a way the new return value makes observable.**

This branch calls `completed(...)` synchronously before returning. The success path defers: JUCE's `AudioPluginFormat::createPluginInstanceAsync` posts an `AsyncCreateMessage` and the callback arrives on a later message-loop turn.

That asymmetry was harmless while the function returned void. Now the documented usage is "the returned resolution carries the installed role and cached capabilities a plan compiled while the instance loads must use" - so a caller compiles from the return value *after* the call, and in the missing-plugin case that happens after its failure handler has already run and, presumably, already torn the binding down. Posting the failure completion the same way the success completion arrives would give callers one ordering to reason about.

PR #2263, opened 2026-08-31
Own plugin-assignment identity at runtime, not in DeviceInfo (#2261)

### LINKED ISSUE #2261: Own plugin-assignment identity at runtime, not in DeviceInfo
Follow-up to #2252.

## Context

#2252 got the principle right: an async plugin load must not infer assignment identity from mutable plugin metadata, because resolution deliberately tolerates a moved file, a renamed plugin, a missing vendor and an unreliable DAWproject role. It needs an opaque, stable identity.

PR #2255 implemented that identity as `DeviceInfo::pluginAssignmentGeneration`, a process-local counter minted in a default member initializer, preserved by copy, and re-minted by an explicit `beginNewPluginAssignment()` call at every site that turns a copy into a live device.

That placement does not hold. `DeviceInfo` is value-semantic, and C++ copying cannot distinguish the cases that matter: a snapshot, an undo record, a browser template, a duplicate, a paste, a preset import, and a new live placement are all the same copy. The invariant therefore lives only in whoever remembered to call `beginNewPluginAssignment()`, and review found missed sites one round at a time:

- `prepareNewDevice` (top-level adds)
- `addDeviceToPostFx`
- `addDeviceToMixerAnalysis`
- `duplicateTrack`'s post-FX and mixer-analysis loops
- `reassignChainElementIds` (copy/paste, preset insert)
- `rekeyPads` (direct pad devices)
- `rekeyPads` again (racks nested inside a pad)

Each was a real defect and each was fixed, but the succession is the finding: the invariant is not enforceable at this layer.

Two further problems come with it:

- **The failure direction is unsafe.** A missed mint means a copy keeps the token, so the completion guard *accepts* a stale load and restores one plugin's state onto another. A correct design should fail closed.
- **Concerns are coupled.** Re-keying ids, remapping macro/mod links, and guarding async completions became one piece of code, because re-keying now owned token minting.

## Required design

Make assignment identity runtime-owned rather than model-embedded.

```cpp
struct AssignmentHandle {};

struct ActiveAssignment {
    engine::DeviceKey key;
    std::shared_ptr<const AssignmentHandle> handle;
};

struct LoadRequest {
    engine::DeviceKey key;
    std::weak_ptr<const AssignmentHandle> handle;
};
```

A coordinator owns the live set and:

1. Creates a handle when a device becomes live.
2. Preserves it across metadata enrichment, parameter edits, state restores and ordinary plan recompilation.
3. Replaces or destroys it when the plugin is replaced or the device is deleted.
4. Issues fresh handles to duplicates, paste/import, undo reinsertion and reused ids.
5. At completion, accepts only if it still holds the same handle for the same `DeviceKey`.

`DeviceInfo` then carries plugin state and no runtime lifetime identity, so copy and re-key code never touches async tokens. Expiry rejects deletion and id reuse without a global counter, and a forgotten registration fails closed.

## Also in scope

`CurrentDeviceLookup` (`EngineDeviceFactory.hpp`) resolves by a bare `DeviceId`. `DeviceId` is allocated per section, so the main FX, post-FX and mixer-analysis sections can each hold the same integer, and `findUniqueBareDeviceIdMatch` exists precisely because of that collision. `engine::DeviceKey` (`RenderPlan.hpp`) is already section-aware, hashable and ordered. The lookup should key on it.

## Separately: pad link retargeting

Independent of assignment identity, and pre-dating #2252. `TrackManager::rekeyPads()` re-keys the ids inside a Drum Grid's pads but discards the resulting map, so macros and modulators targeting anything in that pad subtree keep their old `ChainNodePath` and stop resolving. The duplication and preset paths already do this correctly (`remapPresetLinksRecursive`, `reassignCopiedElementIds`). Fix on its own rather than as part of the identity work.

## Acceptance criteria

- Every acceptance criterion of #2252 still holds.
- `DeviceInfo` carries no runtime lifetime identity; `pluginAssignmentGeneration` and `beginNewPluginAssignment()` are gone.
- No copy, re-key or serialization path needs to know about async load identity.
- A duplicated, pasted, preset-imported or undo-reinserted device cannot accept a load requested for the device it came from.
- A device whose id is reused after `clearAllTracks()` cannot accept a load requested for its predecessor.
- A device in one section cannot accept a load requested for a device holding the same `DeviceId` in another.
- A forgotten registration rejects the load rather than accepting it.
- Production wiring of `createEngineExternalDeviceAsync()` is blocked on this issue.

lucaromagnoli:
Closes #2261. Follow-up to #2252/#2255.

## The problem with where the identity lived

#2255 put assignment identity on `DeviceInfo` as `pluginAssignmentGeneration`: a process-local counter minted in a default member initializer, preserved by copy, and re-minted by an explicit `beginNewPluginAssignment()` at every site that turns a copy into a live device.

That placement could not hold. `DeviceInfo` is value-semantic, and C++ copying cannot distinguish a snapshot from an undo record from a browser template from a duplicate from a paste from a preset import from a new live placement -- all of them are the same copy. The invariant lived only in whoever remembered to call the minting function, review found missed sites one round at a time, and a missed one failed in the unsafe direction: the copy kept the token, so the completion guard *accepted* a stale load and restored one plugin's state onto another.

## What replaces it

`magda/daw/audio/plugins/engine/PluginAssignments.hpp` -- a coordinator owning the live set, keyed on `engine::DeviceKey`:

- `AssignmentHandle` is empty on purpose: identified by its address and by nothing else, so there is no field a copy could carry and no value a caller could forge.
- A device that becomes live gets a `shared_ptr` handle; a load holds a `weak_ptr` to it, so a load in flight never keeps an assignment alive.
- `accepts()` requires both that the weak reference locks **and** that the set still holds that same handle for that key. Locking alone is not enough: a caller holding the `ActiveAssignment` it was handed keeps the handle alive past the release, so the live set is what decides.

Deletion, replacement, duplication, paste, preset import, undo reinsertion and an id handed out again after `clearAllTracks()` are now one shape: a key that is no longer this assignment. A device that was never registered has no handle to lend, so a **forgotten registration refuses the load** instead of accepting it.

`DeviceInfo::pluginAssignmentGeneration` and `beginNewPluginAssignment()` are gone, along with all eight mint sites. No copy, re-key or serialization path touches async load identity.

## Also in scope

`CurrentDeviceLookup` now keys on `engine::DeviceKey` rather than a bare `DeviceId`. `DeviceId` is allocated per section, so main FX, post-FX and mixer-analysis can each hold the same integer and a bare one names up to three devices.

## Separately: pad link retargeting

Independent of the identity work and pre-dating #2252. `prepareNewDevice` re-keyed a Drum Grid's pads and discarded the resulting map, so macros and modulators addressing anything in that pad subtree kept their old `ChainNodePath` and stopped resolving -- the macro is still there, still says it is linked, and moves nothing.

The map is now seeded with the grid's own old `DeviceId` (which is what a pad path names, in its `PadRack` step) and both the grid's own links and the pad subtree's are retargeted through it. `remapPresetLinksRecursive` gained a `PresetState` flag so the add path follows links without also rewriting saved plugin state, which is preset-only behaviour.

## Acceptance criteria

- [x] Every acceptance criterion of #2252 still holds
- [x] `DeviceInfo` carries no runtime lifetime identity
- [x] No copy, re-key or serialization path knows about async load identity
- [x] A duplicated, pasted, preset-imported or undo-reinserted device cannot accept a load requested for the device it came from
- [x] A device whose id is reused after `clearAllTracks()` cannot accept a load requested for its predecessor
- [x] A device in one section cannot accept a load requested for the same `DeviceId` in another
- [x] A forgotten registration rejects the load rather than accepting it
- [x] `CurrentDeviceLookup` keys on `engine::DeviceKey`
- [x] Pad link retargeting fixed on its own

Production wiring of `createEngineExternalDeviceAsync()` was blocked on this and is now unblocked; it is not done here.

## Tests

New `tests/test_plugin_assignments.cpp` covers the coordinator directly (9 cases). `tests/test_engine_external_device.cpp` gains three adapter-level cases: a duplicate refusing its source's load, a section collision refusing with "removed" rather than standing in, and an unregistered device refusing a load that otherwise succeeded. `tests/test_device_api.cpp` gains a case for the pad link retarget -- verified to fail without the fix.

- `magda_tests`: 3210 cases, 0 failures
- `magda_juce_tests`: all suites `0 failed`

### DIFF
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
@@ -318,6 +318,19 @@ ExternalPluginResolution createEngineExternalDeviceAsync(
                                      .resolvedIsInstrument = resolved.description.isInstrument},
          currentDevice = std::move(currentDevice), offlineRender, completed = std::move(completed)](
             std::unique_ptr<juce::AudioPluginInstance> instance, const juce::String& error) {
+            // The lifetime gate, before anything is called rather than after.
+            // This lambda is stored by JUCE and run a turn or more later, and
+            // `completed` is the runtime's own: it publishes the device into the
+            // project and reports the failures. A runtime that is gone has
+            // nowhere to publish and nobody to tell, so calling it at all is the
+            // use-after-free -- refusing inside it would already be too late.
+            //
+            // Deliberately not isStillWanted(): a device deleted while its
+            // runtime lives is a load to refuse *and say so about*, because
+            // something is still waiting to hear it.
+            if (!requested.assignment.runtimeIsAlive())
+                return;
+
             completed(completeExternalPluginLoad(std::move(instance), error, requested,
                                                  currentDevice, offlineRender));
         });
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
@@ -255,27 +255,38 @@ ExternalDeviceResult createEngineExternalDevice(const magda::DeviceInfo& device,
 ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPluginInstance> instance,
                                                 const juce::String& error,
                                                 const RequestedPlugin& requested,
+                                                const PluginAssignments& assignments,
                                                 const CurrentDeviceLookup& currentDevice,
                                                 bool offlineRender) {
-    // The model first, before anything is done with the instance. Everything
+    const auto& requestedName = requested.displayName;
+
+    // The identity boundary, and the first thing asked. Scan metadata may have
+    // changed in every field while the plugin loaded, including the role the
+    // plan compiles from; those are facts learned about this assignment, not
+    // about another one, so none of them is consulted here. What decides is
+    // whether the runtime still holds the very assignment the load was started
+    // against, which no copy of the model can carry and no unregistered device
+    // can claim.
+    if (!assignments.accepts(requested.assignment)) {
+        // Which of the two it was, for a person reading the log: a key nothing
+        // holds any more is a device that went away, and a key held under a
+        // different assignment is a slot that is now asking for something else.
+        const auto replaced = static_cast<bool>(assignments.current(requested.assignment.key));
+        return {.device = {},
+                .failure =
+                    replaced
+                        ? "the device changed plugin while \"" + requestedName + "\" was loading"
+                        : "the device was removed while \"" + requestedName + "\" was loading"};
+    }
+
+    // Then the model, before anything is done with the instance. Everything
     // below restores from it, and it is a different object from the one the
     // load was requested with: seconds have passed.
-    const auto* device = currentDevice ? currentDevice(requested.deviceId) : nullptr;
-
-    const auto& requestedName = requested.displayName;
+    const auto* device = currentDevice ? currentDevice(requested.assignment.key) : nullptr;
 
     if (device == nullptr)
         return {.device = {},
                 .failure = "the device was removed while \"" + requestedName + "\" was loading"};
-
-    // This is the identity boundary. Scan metadata may have changed in every
-    // field while the plugin loaded, including the role the plan compiles from;
-    // those are facts learned about this assignment, not another assignment.
-    // A replacement DeviceInfo carries a newly authored generation even when
-    // it is another variant in the same bundle or has the same display name.
-    if (device->pluginAssignmentGeneration != requested.assignmentGeneration)
-        return {.device = {},
-                .failure = "the device changed plugin while \"" + requestedName + "\" was loading"};
 
     if (instance == nullptr)
         return {.device = {},
@@ -288,8 +299,10 @@ ExternalDeviceResult completeExternalPluginLoad(std::unique_ptr<juce::AudioPlugi
 }
 
 ExternalPluginResolution createEngineExternalDeviceAsync(
-    const magda::DeviceInfo& device, const ExternalPluginServices& services, bool offlineRender,
-    CurrentDeviceLookup currentDevice, std::function<void(ExternalDeviceResult)> completed) {
+    const magda::DeviceInfo& device, magda::engine::DeviceKey key,
+    const ExternalPluginServices& services, bool offlineRender,
+    const PluginAssignments& assignments, CurrentDeviceLookup currentDevice,
+    std::function<void(ExternalDeviceResult)> completed) {
     jassert(completed != nullptr);
     jassert(currentDevice != nullptr);
 
@@ -301,13 +314,13 @@ ExternalPluginResolution createEngineExternalDeviceAsync(
 
     services.formats->createPluginInstanceAsync(
         resolved.description, services.context.sampleRate, services.context.maxBlockSize,
-        [requested = RequestedPlugin{.deviceId = device.id,
-                                     .assignmentGeneration = device.pluginAssignmentGeneration,
+        [requested = RequestedPlugin{.assignment = assignments.request(key),
                                      .displayName = device.name,
                                      .resolvedIsInstrument = resolved.description.isInstrument},
-         currentDevice = std::move(currentDevice), offlineRender, completed = std::move(completed)](
-            std::unique_ptr<juce::AudioPluginInstance> instance, const juce::String& error) {
-            completed(completeExternalPluginLoad(std::move(instance), error, requested,
+         &assignments, currentDevice = std::move(currentDevice), offlineRender,
+         completed = std::move(completed)](std::unique_ptr<juce::AudioPluginInstance> instance,
+                                           const juce::String& error) {
+            completed(completeExternalPluginLoad(std::move(instance), error, requested, assignments,
                                                  currentDevice, offlineRender));
         });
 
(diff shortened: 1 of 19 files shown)


magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -301,13 +314,13 @@ ExternalPluginResolution createEngineExternalDeviceAsync(
...
         resolved.description, services.context.sampleRate, services.context.maxBlockSize,
-        [requested = RequestedPlugin{.deviceId = device.id,
-                                     .assignmentGeneration = device.pluginAssignmentGeneration,
+        [requested = RequestedPlugin{.assignment = assignments.request(key),
                                      .displayName = device.name,
                                      .resolvedIsInstrument = resolved.description.isInstrument},
-         currentDevice = std::move(currentDevice), offlineRender, completed = std::move(completed)](
-            std::unique_ptr<juce::AudioPluginInstance> instance, const juce::String& error) {
-            completed(completeExternalPluginLoad(std::move(instance), error, requested,
+         &assignments, currentDevice = std::move(currentDevice), offlineRender,
```
lucaromagnoli: **P1 — The asynchronous callback captures a lifetime-unsafe reference.** `createPluginInstanceAsync()` posts completion to a later message-loop turn, but this lambda holds `PluginAssignments` by reference. If the project/runtime owner is destroyed while instance creation is pending, completion dereferences a destroyed registry before it can reject the request; `releaseAll()` does not protect destruction of the registry itself. Please make the registry state callback-safe—for example, put the live map in shared state and carry a weak reference to that state in `LoadRequest`, rejecting when either the registry or assignment has expired—and cover shutdown while a load is pending.\n\n— Codex

magda/daw/audio/plugins/engine/PluginAssignments.hpp
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * @brief A device is live under @p key, as a new assignment.
+     *
+     * Any handle already held for @p key is dropped, which expires every
+     * request outstanding against it. That is what covers a slot whose plugin
+     * was replaced and an id handed out again after the project was cleared:
+     * both are a new assignment on a key that had one, and neither can be told
+     * from the other by anything in the model.
+     */
+    ActiveAssignment assign(magda::engine::DeviceKey key);
```
lucaromagnoli: **P2 — The API needs to encode preserve versus replace.** `assign(key)` always drops the existing handle, but normal plan preparation/registration must preserve the current assignment while an actual plugin replacement must invalidate it. The current test helper already has to call `current()` before `assign()`, leaving every production caller to remember the same semantic rule; a routine re-prepare could otherwise invalidate an in-flight load repeatedly. Please expose distinct operations such as `ensureAssignment()` and `replaceAssignment()`, or model the slot with an owned assignment lease, so this invariant is structural rather than a caller convention.\n\n— Codex

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -227,32 +229,67 @@ juce::String stripPresetRuntimePluginState(const juce::String& pluginState) {
...
+        return;
+
+    PresetIdRemap remap;
+    remap.trackId = trackId;
+    remap.devices = ids.devices;
+    remap.racks = ids.racks;
+    remap.chains = ids.chains;
+
+    remapPresetLinks(device.macros, device.mods, remap);
+    for (auto& pad : device.pads->chains)
```
lucaromagnoli: **P2 — This traversal skips links owned by the pad rack itself.** `device.pads` is a `RackInfo` with its own `macros` and `mods`. Iterating only `device.pads->chains` remaps links owned by devices and nested racks inside the pads, but leaves a macro/mod stored directly on the pad rack pointing at the old `ChainNodePath`; those rack-owned controls are compiled normally. Please remap the whole rack here, e.g. `remapRackPresetLinks(*device.pads, remap, PresetState::Keep)`, and add coverage for a pad-rack-owned macro or mod targeting a re-keyed pad device.\n\n— Codex

lucaromagnoli (commented):
The runtime-owned identity direction is correct, but these lifecycle and retargeting gaps should be closed before production wiring. — Codex

magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
```diff
@@ -301,8 +313,7 @@ ExternalPluginResolution createEngineExternalDeviceAsync(
 
     services.formats->createPluginInstanceAsync(
         resolved.description, services.context.sampleRate, services.context.maxBlockSize,
-        [requested = RequestedPlugin{.deviceId = device.id,
-                                     .assignmentGeneration = device.pluginAssignmentGeneration,
+        [requested = RequestedPlugin{.assignment = assignments.request(key),
```
lucaromagnoli, line 316: **P1 — Shutdown still invokes a callback that may point into the destroyed runtime.** The raw `PluginAssignments&` is gone, but this stored JUCE lambda still unconditionally invokes `completed(...)` after `isStillWanted()` has rejected an expired table. In production `completed` will normally capture the runtime/project (`this`) it publishes into, so closing that owner while loading can still cause a use-after-free—now on the completion callback instead of the registry. The new shutdown test calls `completeExternalPluginLoad()` directly and therefore does not exercise this outer callback. Use the assignment table as a lifetime gate before invoking owner callbacks (drop completion when the table/runtime has expired), or make the completion target weak-owner-safe by construction, and cover the actual async entry point with teardown before dispatch. — Codex

PR #2267, opened 2026-08-31
Floor pencil note insertion to the containing grid cell (#2266)

### LINKED ISSUE #2266: Piano roll pencil places notes in the next cell: insert snaps to nearest grid line instead of containing cell
## Symptom

Drawing a MIDI note with the pencil (Alt+click) in the piano roll requires aiming at the left half of the target cell. Click anywhere in the right half and the note lands in the cell to the right. It feels like the note follows the middle of the pencil glyph rather than the tip, and hitting the intended cell reliably means clicking near its start line.

## Cause

The pencil insert path snaps to the **nearest** grid line instead of flooring to the **containing** cell:

- `getNoteInsertPosition()` -> `clipBeatForDisplayX()` applies `snapBeatToGrid()` to the click beat (`PianoRollGridComponent.cpp:1999`).
- `snapBeatToGrid()` uses `std::round(beat / gridResolutionBeats_) * gridResolutionBeats_` (`PianoRollGridComponent.cpp:1850`).

With round-to-nearest, the zone that lands a note on cell N is the half cell either side of N's start line, so half of every cell's visible span places the note in the next cell.

The pencil cursor itself is not the problem: its hotspot already sits on the drawn tip (`CursorManager::createNoteDrawCursor`, hotspot 5,22).

## Expected

A pencil click anywhere within a cell's span places the note on that cell: floor to the containing cell (`std::floor(beat / grid) * grid`) for note insertion.

Round-to-nearest stays correct for the other `snapBeatToGrid()` consumers (note drag/resize, playhead placement), where snapping to the closest line is the desired feel. Only the insert/draw path should floor.

lucaromagnoli:
Fixes #2266.

The pencil insert path (`getNoteInsertPosition()` → `clipBeatForDisplayX()`) snapped the click beat to the **nearest** grid line, so a click in the right half of a cell placed the note in the next cell.

## Changes

- Added `PianoRollGridComponent::snapBeatToGridFloor()`, which floors to the containing cell (with a small epsilon so a click a float-hair below a cell boundary still lands on that cell).
- `clipBeatForDisplayX()` takes a `floorToCell` flag; `getNoteInsertPosition()` passes it, covering both insert paths (Alt+click pencil and double-click).
- Drag, resize, draw-drag end beat, and playhead placement keep round-to-nearest snapping, as the issue specifies.

The drum grid already floors on both of its insert paths (empty-cell click and repeat stamp), so it needed no change.

## Testing

- Debug build compiles; `[pianoroll]` unit tests pass.
- Verified manually in the app: pencil clicks anywhere in a cell now land the note on that cell.

https://claude.ai/code/session_01VMjNEtqw2WxSK3evwdeXZG

### DIFF
--- a/magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp
+++ b/magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp
@@ -1858,9 +1858,12 @@ double PianoRollGridComponent::snapBeatToGridFloor(double beat) const {
     if (!snapEnabled_ || gridResolutionBeats_ <= 0.0) {
         return beat;
     }
-    // Epsilon absorbs float error when the click maps to a hair below a cell
-    // boundary, so a click on the line still lands on the cell it starts.
-    return std::floor(beat / gridResolutionBeats_ + 1e-6) * gridResolutionBeats_;
+    // Grid lines render at integer pixels (beatToPixel rounds), so a click on
+    // a rendered boundary can map back up to half a pixel below the true beat.
+    // Absorb that quantization so a click on the line lands on the cell the
+    // line starts.
+    const double halfPixelBeats = pixelsPerBeat_ > 0.0 ? 0.5 / pixelsPerBeat_ : 0.0;
+    return std::floor((beat + halfPixelBeats) / gridResolutionBeats_ + 1e-6) * gridResolutionBeats_;
 }
 
 bool PianoRollGridComponent::isNearGridLine(int mouseX) const {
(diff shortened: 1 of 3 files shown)


lucaromagnoli (commented):
**[P2] Half-pixel bias moves the pixel left of rounded-up lines into the next cell**

At the default 50 px/beat zoom with a 0.25-beat grid, the boundary renders at `round(12.5) = 13px`. Clicking pixel 12—visibly one pixel left of the line—produces beat `0.24`; adding `0.5 / 50 = 0.01` advances it to `0.25`, so it incorrectly inserts into the next cell. The added “one pixel left” test only covers a boundary rounded down; repeating it with 50 px/beat exposes this case. Cell selection needs to compare against the same rounded pixel boundaries used for rendering rather than applying a uniform half-pixel shift.

Issue #1899, opened 2026-07-27
Out-of-process plugin sandboxing


lucaromagnoli:
Part of #1882. Post-cutover.

Out-of-process plugin sandboxing, per-plugin opt-in.

- Sandbox host process (JUCE-based, like the existing out-of-process scanner) loading VST3/AU.
- Shared-memory audio transport implementing the same device op contract as in-process hosting; deadline policy on the live path (miss the deadline: output silence, flag the plugin offline, never stall the callback).
- Crash recovery via recompile: dead process yields a placeholder op; state restored from the last chunk snapshot on respawn.
- Anticipated-domain first (IPC cost and hangs just drain look-ahead), live-domain second.
- Editor UI: cross-process reparenting works on Windows; on macOS sandboxed editors float in their own windows for v1.

PR #2268, opened 2026-08-31
Write a plugin's state back out, and say the rule once (#2244)

### LINKED ISSUE #2244: External plugins slice 3: the state restore contract as an API
Slice 3 of #1893. Depends on #2241 and the instantiation slice.

A plugin's state, restored under the native engine, at parity with the three-step contract the app
already depends on: the saved parameter array as a baseline, the plugin's own state chunk as the
authoritative overlay, then the host's parameter cache refreshed to agree with what the chunk left
behind. Get the order wrong and a restored voice is overwritten by the defaults the host read at
construction, which is the bug that contract was written for.

MAGDA persists one thing for an external plugin: the base64 chunk TE writes from
`getStateInformation`. The engine's side of it is `setStateInformation`, applied before the adapter
takes the instance -- and the adapter reads the parameter list at construction, so a restore that
happened afterwards would be mapped against parameters the plugin no longer has.

What #1893 asks for beyond parity is that this stops being an archaeology exercise: the current
sequence lives in `ExternalPluginStateUtil` and reaches an internal Tracktion header that is not
self-contained and cannot be sorted in its include block. The native engine's version is an API a
host calls, with the ordering stated in one place rather than reconstructed from three call sites.

The other half is saving: a project bounced or saved while the native engine holds the instance has
to write the same chunk the fork would, so a project round-trips between the two engines during the
dual-engine release rather than losing a plugin's voice on the way.

---

**Half of this landed in #2249**, because leaving it out made EngineExternalDevice's header say
something untrue: the chunk is applied before the adapter takes the instance, so a saved patch
loads rather than the plugin's initialised voice. What is left here is the half that is genuinely
its own work:

- **Saving.** Writing the chunk back out of an instance the native engine holds, so a project
  round-trips between the two engines during the dual-engine release.
- ~~Precedence where the two disagree.~~ Settled in #2249 rather than left open, because it is
  already settled in this repository: the incumbent lets the chunk win and
  `test_external_plugin_state_restore_juce.cpp` pins it for every project saved by a MAGDA old
  enough to have written a stale array. The adapter declines to write a parameter still sitting
  where the project said it sits, which reaches the same place without the engine writing to the
  model from a render.
- **The API.** Both of the above stated in one place rather than reconstructed from call sites,
  which is what #1893 asks for beyond parity.
- **DeviceInfo::vst3Preset**, the portable .vstpreset a DAWproject carries. Not applied on the
  render path and its field comment says why (interchange only, native state is pluginState), but
  the import path under the native engine has to do something with it.

lucaromagnoli:
Closes #2244. Slice 3 of #1893, whose other half landed in #2249.

## What was missing

A project's record of an external plugin was only readable. `applySavedPluginState()` stated the restore order in one place; saving was not stated anywhere. So a project saved while the native engine held the instance would have lost the plugin's voice on the way back to the fork, which is exactly the round trip the dual-engine release depends on.

## Saving

`captureSavedPluginState()` is the mirror of `applySavedPluginState()`, and what it writes is what `ExternalPlugin::flushPluginStateToValueTree()` writes for the same instance:

- The chunk from `getStateInformation()`, in `juce::MemoryBlock`'s own base64 rather than the standard one, because that is the encoding the field already holds and `decodeSavedChunk()` already reads.
- Absent rather than empty when the plugin has nothing to say. The fork removes the property; a stored zero-length chunk would come back as a baseline anyway, and a previous save's chunk must not survive the plugin ceasing to have one.
- The parameter array beside it, so the two records agree: the array is the baseline the chunk overlays on the next load, and it is also what the UI draws and what automation addresses.
- The portable VST3 records refreshed, which is what the incumbent does on every capture.

The plugin is suspended across the read the way the fork suspends it, and restored either way, since a plugin left suspended by its own throw would render silence for the rest of the session. A plugin that throws describing itself writes **neither** record and reports it: a fresh array under a stale chunk restores the stale voice and calls it the project's, so a save that can only write one of them writes none and the project keeps the last pair that agreed.

## The portable preset

`DeviceInfo::vst3Preset` is the .vstpreset a DAWproject carries, and until now the native engine did nothing with it -- a comment in `applySavedPluginState()` said so and pointed here. It is now the overlay when a project has one, which is what the incumbent does, and it is **consumed**: the load that applies it is what turns it into native state, and a project that kept it would apply it again over whatever had been saved since.

`adaptExternalPluginInstance()` clears it on `resolvedDevice` only, which a caller publishes solely on success -- so a load that failed leaves the project still holding the one thing that describes its plugin.

One deliberate divergence from the incumbent, stated in the header rather than left to be found: a plugin that **refuses** the preset falls through to the chunk instead of being left on the bare array. The incumbent's `else` is a consequence of TE's async lifecycle rather than a rule anyone stated, and no project carries both records unless a preset has already failed to apply once -- at which point the fork discards the project's own native record every load thereafter.

## The API, said once

#1893 asks for this to stop being an archaeology exercise. `ExternalPluginState.hpp` now states both directions and has no Tracktion dependency at all, which was the original complaint: the incumbent's sequence reaches an internal TE header that is not self-contained and has to be fenced off from `SortIncludes`.

Three copies of the VST3 `ExtensionsVisitor` become one pair of functions, `readVst3Preset()` and `writeVst3Preset()`:

- `PluginManager::captureVst3Info` keeps its name and its TE null-checks, and delegates.
- `ExternalPluginStateUtil::applyVst3Preset` keeps only what is genuinely TE's: the async guard, the state-property mirror and the parameter cache refresh.
- `AudioBridge`'s two-mode `Vst3PresetVisitor` is gone; `getVst3DeviceId`, `loadPluginPresetFile` and `savePluginPresetFile` call the shared pair.

## Tests

Twelve cases in `tests/engine/test_engine_external_device.cpp`. The stub gains a VST3 extension, because whether an instance answers that visit is the only way a host can ask what format it is, and a real `.vstpreset` header so the class id is read at the offset it actually lives at.

Restore: an imported preset is the overlay and reaches the plugin through the format's own door rather than the chunk's; an applied one is spent rather than kept; a refused one falls through to the chunk; a plugin that is not a VST3 never sees one.

Save: the chunk matches what the fork writes and a second instance restored from it holds what the first held, including the parameter no host may automate; the array is written beside it; a plugin with nothing to say saves no chunk; a plugin that throws leaves the last good save intact and is not left suspended; the portable records are refreshed for a VST3 and left alone for anything else; and a save reads the instance through `EngineExternalDevice::instance()`, which is the seam a host actually saves through.

## Verification

Full Catch2 suite passes (650,349 assertions), `make debug` builds clean, and `magda_juce_tests` passes including the real-project corpus -- the `FAIL` lines in its report are the recorded known-diff cases and are unchanged.

### DIFF
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -804,7 +804,7 @@ juce::String AudioBridge::getVst3DeviceId(const ChainNodePath& devicePath) const
         return {};
     // Pull the current state as a .vstpreset; its header carries the 32-char
     // class id. Empty for a plugin that is not a VST3.
-    return vst3::classIdFromPreset(readVst3Preset(*pi));
+    return vst3::classIdFromPreset(readVst3Preset(*pi).preset);
 }
 
 int AudioBridge::getPluginNumPrograms(const ChainNodePath& devicePath) const {
@@ -861,7 +861,7 @@ bool AudioBridge::loadPluginPresetFile(const ChainNodePath& devicePath,
         juce::MemoryBlock raw;
         if (!presetFile.loadFileAsData(raw))
             return false;
-        applied = writeVst3Preset(*pi, raw);
+        applied = writeVst3Preset(*pi, raw) == Vst3PresetOutcome::Applied;
     } else if (extension == ".aupreset") {
         juce::MemoryBlock raw;
         if (!presetFile.loadFileAsData(raw))
@@ -886,7 +886,7 @@ bool AudioBridge::savePluginPresetFile(const ChainNodePath& devicePath,
     const auto extension = presetFile.getFileExtension().toLowerCase();
 
     if (extension == ".vstpreset") {
-        const auto preset = readVst3Preset(*pi);
+        const auto preset = readVst3Preset(*pi).preset;
         if (preset.getSize() == 0)
             return false;  // not a VST3 plugin
         presetFile.getParentDirectory().createDirectory();
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -15,6 +15,7 @@
 #include "DeviceParameterDisplayTextProvider.hpp"
 #include "Vst3Preset.hpp"
 #include "modifiers/ADSRDebugLog.hpp"
+#include "plugin_manager/ExternalPluginState.hpp"
 #include "plugins/DeviceServices.hpp"
 #include "plugins/InternalPluginRegistry.hpp"
 #include "plugins/MidiInThruSync.hpp"
@@ -792,25 +793,6 @@ te::ExternalPlugin* asExternalPlugin(te::Plugin::Ptr plugin) {
     return dynamic_cast<te::ExternalPlugin*>(plugin.get());
 }
 
-// Loads / saves a .vstpreset blob via JUCE's VST3Client extension. Two-mode
-// visitor: when `dataIn` is non-empty we apply it as a preset; otherwise we
-// pull the current state into `dataOut`.
-struct Vst3PresetVisitor : juce::ExtensionsVisitor {
-    juce::MemoryBlock dataIn;
-    juce::MemoryBlock dataOut;
-    bool ok = false;
-    bool save = false;
-
-    void visitVST3Client(const VST3Client& client) override {
-        if (save) {
-            dataOut = client.getPreset();
-            ok = dataOut.getSize() > 0;
-        } else {
-            ok = client.setPreset(dataIn);
-        }
-    }
-};
-
 }  // namespace
 
 juce::String AudioBridge::getVst3DeviceId(const ChainNodePath& devicePath) const {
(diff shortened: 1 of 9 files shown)


lucaromagnoli:
All four review findings are addressed. Head is now f05dfa5ad; the last review read 88957a5, which predates the parameter-gate fix.

## The races

The parameter one was a real hole in my own first fix. I put the gate inside `processPluginBlock()`, and `process()` calls `writeParameters()` before reaching it, so parameter writes went past while a capture was reading the values those writes were changing. The gate is now the first thing `process()` does, and everything plugin-facing is downstream of it.

I checked the test actually catches it rather than passing by accident: with the gate back where it was, `tone->writes` goes 1 to 2 and gain lands on 0.3 instead of holding. The original test missed it because it supplied the values the stub already held, so nothing was written either way; it now supplies values the plugin does not have and counts the writes that reach each parameter.

## On the ownership boundary

I agree with the diagnosis and I have split the difference deliberately, so here is the reasoning rather than a silent decision.

**Taken now**, because it is contained and because it was a defect on its own terms: capture no longer mutates a `DeviceInfo` while holding a plugin. `captureExternalPluginState()` returns an `ExternalPluginSnapshot` and `applyCapturedPluginState()` writes it. That was an asymmetry I introduced -- the restore beside it already reads the instance and hands corrections back for the caller to apply, precisely because the model is the caller's. It is also the part of the redesign that makes a remote implementation straightforward later, since a snapshot can cross a process boundary and a plugin cannot, and it gives a caller the place to stand where it can ask whether the device it read is still the device it is about to write.

**Deferred**, to #2270:

`EngineExternalDevice::instance()` is not this PR's. It came in with #2249 and the editor-window path uses it too, so removing it is a change to that slice's design rather than to this one's, and doing it here would mean this PR silently rewrites a seam it did not introduce.

The `DeviceRuntime.control` interface -- keyed by `DeviceKey`, asynchronous, fallible, with assignment-lifetime checks -- is the out-of-process boundary. That is #1899, and the engine process split was assessed in August and deferred; building its boundary speculatively inside a state-capture PR is scope this issue did not ask for. #2244 asks for the state rules stated in one place, which is what is here.

The strongest argument for doing it immediately is that capture has no production caller yet, so the shape is free to change. That argument holds for exactly as long as nothing calls it, so I have made it a gate rather than a race: #2270 says the control-plane endpoint lands before the first production caller of capture, and names both races above as what the split ownership already cost.

## Verification

Full Catch2 suite passes (650,354 assertions), `make debug` is clean, and the JUCE suite passes including the real-project corpus.

PR #2269, opened 2026-08-31
Host external plugins in the null-diff native leg, and the master chain (#2175)

lucaromagnoli:
Part of #2175, and the step it opens with. Sits on top of the external plugin slices (#2241, #2243, #2244).

## What the leg was doing

`NullDiffNativeLeg` bound a Device op through `createEngineDevice()`, which asks the two internal catalogs. An external plugin is in neither, so a project hosting one rendered a passthrough where the incumbent rendered the plugin -- and said nothing about it, because the diagnostic beside that path only fires for a `pluginId` some catalog knows. "Nothing hosts VST3 in the engine" was true when #2175 was written and stopped being true with #2249; what was left was the leg asking.

## The wiring

The blocking entry point, which is the one written for a render with nothing to do until the plugin is there. The scan comes from the runner and is the shared engine's own `KnownPluginList`: both legs resolve against one list, because two legs that each found their own copy of a plugin would be rendering two projects and calling the difference an engine bug. On a machine with no scan there is nothing to find, which is the case every CI runner is in.

Two orderings that matter, and neither is incidental:

- **Resolution runs before the plan is compiled**, against a copy of the model. For an external plugin the effect/instrument role the plan compiles from is the project's guess rather than a fact -- an imported project carries whatever the exporting host said, which is the correction #2252 exists for -- and a plan compiled from the guess routes MIDI to the wrong places before any plugin is created.
- **`restoredParameters` is written back into that copy before `resolvePlanValues`.** The chunk wins over the project's saved array, and the parameter table is what the plan writes onto the device every block, so a model left holding the stale array would put it back on the first one.

## A plugin this machine does not have

Now reported rather than passed through. The runner already turns a diagnostic the case did not declare into unmeasurable, which is the treatment #2175 asks for and the same one a proxy that never arrived already gets. The render still completes and the stand-in is still bound, because the executor refuses an unbound Device op and a leg that failed outright would report a broken harness rather than a project it could not measure.

`resolveEngineExternalPlugin`'s no-scan failure now names the device. It was the one failure sentence in that file that did not, against `ExternalDeviceResult`'s own promise that a failure says which plugin and what went wrong, and a caller collecting these holds a project's worth of them.

## The master chain, which is a real bug

`devicesIn()` collected the tracks' devices and not the master's. Every Device op the compiler emits for the master chain was therefore looked up, not found, and bound to the stand-in -- silently, since the "no native device for" diagnostic sits behind a model the lookup did not have.

`project.demo` carries a limiter there. The corpus has been comparing the incumbent's limited render against an unlimited native one.

**Measured rather than assumed.** On `dev/0.20.0` the demo reports envelope correlation **0.442** and shift 5400; with this change, **0.417** and shift -3832. That number got slightly worse, and it is the comparison getting honest: the limiter now renders on both legs and the two evidently disagree. `project.fmchain` and `project.sidechain` are unchanged to the decimal (5.4/21.9 and 3.6/16.5 dB), which fits -- neither has a master device. The demo stays under calibration for its own declared reason, and its calibration note now says what moved and that its shift has two mechanisms behind it rather than one.

## Tests

Three in `test_null_diff_native_leg.cpp`, in the model-only target:

- A gain of zero on the master silences the render, with a unity control beside it so the assertion is about the master device rather than about a case that renders nothing either way. Bound it silences; as a stand-in it passes signal, which is the difference the old lookup could not tell.
- A project naming a plugin this machine cannot resolve renders, and names it in the diagnostics.
- The same for a plugin on the master chain, where the two halves of this change meet -- a master bus limiter is where a real project puts one.

Plus the no-scan failure naming its device, in `test_engine_external_device.cpp`.

## Verification

Full Catch2 suite passes (650,288 assertions) and the whole JUCE suite passes including the real-project corpus, with no case changing pass/fail state.

## Still open on #2175

The plugin-version field on `Case`'s environment block, and the corpus projects that carry real VST3 state.

### DIFF
--- a/tests/NullDiffNativeLeg.cpp
+++ b/tests/NullDiffNativeLeg.cpp
@@ -18,6 +18,7 @@
 #include "exec/PlanExecutor.hpp"
 #include "exec/PlanValues.hpp"
 #include "io/PrefetchThread.hpp"
+#include "magda/daw/audio/plugin_manager/ExternalPluginState.hpp"
 #include "magda/daw/audio/plugins/engine/EngineDeviceFactory.hpp"
 #include "plan/PlanCompiler.hpp"
 #include "transport/TempoMap.hpp"
@@ -25,6 +26,8 @@
 namespace magda::nulldiff {
 
 using namespace magda::engine;
+
+namespace adapter = ::magda::daw::audio::engine_adapter;
 
 namespace {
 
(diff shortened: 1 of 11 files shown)


tests/NullDiffNativeLeg.cpp
```diff
@@ -327,9 +378,23 @@ NativeRender renderNative(const Case& value) {
...
+    // compiled with, and a plan compiled from the project's own guess would
+    // route MIDI to the wrong places before any plugin was created.
+    auto tracks = value.tracks;
+    auto master = value.master;
+    resolveExternalDevices(tracks, master, services);
+
     CompileOptions options;
     options.deviceMeters = false;
-    const auto plan = compileRenderPlan(value.tracks, value.master, options);
+    const auto plan = compileRenderPlan(tracks, master, options);
```
lucaromagnoli, line 498: **P1 — Compile from the instantiated plugin topology**

At this point only scan metadata has been applied. `createEngineExternalDevice()` later restores state and returns `external.resolvedDevice`, whose live bus counts and MIDI capabilities can differ from the saved model, but that value is ignored. `PlanCompiler` has already frozen those facts into port widths and MIDI outputs, so plugins whose state changes mono/stereo, sidechain, instrument, or multi-output layout can be rendered through the wrong graph.

Please instantiate and restore the active external devices before the final compile, apply `resolvedDevice` plus restored parameters to the copied model, then compile and bind those same instances. A preliminary compile can be used only to identify active `DeviceKey`s if needed.

tests/NullDiffNativeLeg.cpp
```diff
@@ -186,32 +189,80 @@ class ImpulseSynthDevice final : public EngineDevice {
...
-void collectDevices(const std::vector<magda::ChainElement>& elements,
-                    std::map<DeviceKey, const magda::DeviceInfo*>& out) {
-    for (const auto& element : elements) {
+void collectDevices(std::vector<magda::ChainElement>& elements,
+                    std::map<DeviceKey, magda::DeviceInfo*>& out) {
+    for (auto& element : elements) {
         if (magda::isDevice(element)) {
-            const auto& device = magda::getDevice(element);
+            auto& device = magda::getDevice(element);
             out[DeviceKey{ChainSegment::Fx, device.id}] = &device;
```
lucaromagnoli: **P1 — Descend into pad-rack device chains**

This records a pad-rack device itself but never walks `device.pads->chains`. `PlanCompiler::emitPadRack()` does walk those chains and emits `Device` ops for their children, so a plugin inside a Drum Grid pad is absent from `modelDevices`: it silently receives `Passthrough`, is never resolved, and produces no missing-plugin diagnostic.

Please include pad chains in this traversal, preferably by reusing `chain_walk::forEachDevice(..., Pads::Enter, ...)` so the harness and the model do not maintain different definitions of every device.

tests/NullDiffNativeLeg.hpp
```diff
@@ -75,6 +105,10 @@ struct NativeRender {
 };
 
 /// Render @p value through the native engine, over its own beat range.
-NativeRender renderNative(const Case& value);
+///
+/// @p installed is the scan an external plugin is resolved against. A case
+/// whose project names one without it renders nothing for that device and says
+/// so in the diagnostics.
+NativeRender renderNative(const Case& value, const InstalledPlugins& installed = {});
```
lucaromagnoli, line 112: **P1 — Do not let no-scan callers certify passthrough renders**

`test_null_diff_block_size.cpp::gate()` still calls this overload with the default at its reference, repeat, and alternate block sizes. For an external plugin that means resolution fails, a passthrough is bound, and only a diagnostic is returned; the gate checks `failure` but never checks `diagnostics`, so it can certify the passthrough as block-size invariant without ever running the plugin.

Please thread installed plugin services through that gate, or make undeclared render diagnostics unmeasurable there. If the model-only target cannot host external plugins yet, explicitly refusing those cases is safer than returning a green result.

tests/NullDiffNativeLeg.cpp
```diff
@@ -180,38 +184,175 @@ class ImpulseSynthDevice final : public EngineDevice {
...
+/// entered with Pads::Enter. A Drum Grid's pads are chains of devices and
+/// PlanCompiler::emitPadRack() emits an op for each of them, so a walk that
+/// stopped at the grid would leave every plugin in every pad looked up and not
+/// found -- bound to a stand-in, never resolved, and reported as nothing. Two
+/// definitions of "every device in a project" is exactly the disagreement that
+/// file exists to prevent.
+void collectDevices(std::vector<magda::ChainElement>& elements,
+                    const magda::ChainNodePath& parentPath, ChainSegment segment,
+                    std::map<DeviceKey, magda::DeviceInfo*>& out) {
+    magda::chain_walk::forEachDevice(
```
lucaromagnoli, line 205: **P2 — Use the pad-aware walk for the corpus policy checks too**

This fixes the binding lookup, but two other every-device walks still stop at the Drum Grid container: `externalDevicesIn()` in `NullDiffCorpus.cpp` and `everyDevice()` in `MgdFixture.cpp`. Consequently an external plugin inside a pad is treated as an internal project by the block-size gate, and can also bypass the MGD fixture external-plugin refusal.

Please reuse `chain_walk::forEachDevice(..., Pads::Enter, ...)` in those policy walks as part of this change, and add a pad case to the block-size attribution test. Otherwise the harness still has multiple conflicting definitions of every device.

Issue #2272, opened 2026-08-31
Honor hardware output destinations in the native engine render plan


lucaromagnoli:
Follow-up to #2264 (and part of #1882).

Track output dropdowns now store hardware destinations as plain device strings in `TrackInfo.audioOutputDevice` (`"stereo:<device>"` for a pair, bare device name for mono). The TE path honors them via `TrackOutput::setOutputToDeviceID`.

The native engine does not yet: `parseTrackRoute` classifies these strings as `External`, and the PlanCompiler routes every External output to the default/master output. The render plan needs a hardware-output destination concept so an External route naming a specific device+channel pair reaches that pair instead of master.

Also worth deciding there: how device names serialize portably across engines/hardware changes (today the names come from TE's wave output devices, e.g. "Out 1 + 2").

PR #2273, opened 2026-08-31
Route track outputs to hardware wave output devices (#2264)

### LINKED ISSUE #2264: Implement hardware output routing for track outputs
## User report

Reported by a Windows 10 user with an NI Audio8 DJ (8 in / 8 out, ASIO): all hardware outputs appear in the track output dropdown and can be selected, but audio only ever plays through the master stereo pair. Also reports crashes when changing the output configuration back and forth (likely a separate device-reconfiguration issue).

Reproduced: no output other than master can be selected effectively.

## Root cause

Hardware output selection is stubbed out in the UI. `RoutingSyncHelper::populateAudioOutputOptions` (`magda/daw/ui/components/mixer/RoutingSyncHelper.hpp:224-260`) lists all hardware stereo pairs (IDs 10+) and mono channels (IDs 100+), but every handler maps those selections straight back to master:

- `TrackHeadersPanel.cpp:837-840` — `selectedId >= 10` → `setTrackAudioOutput(trackId, "master")` ("For now, hardware channels route via device ID")
- `SessionView.cpp:1093-1094` — same
- `MixerView.cpp:918-919` — same
- `TrackInspector.cpp` — same pattern

The backend already supports it: `TrackController::setTrackAudioOutput` (`magda/daw/audio/TrackController.cpp:291-294`) routes any non-master/non-track destination via `setOutputToDeviceID()`, and the TE wrapper enables all hardware output channels (`TracktionEngineWrapperDevices.cpp:67-70`).

## Fix shape

Mirror the input path: have `populateAudioOutputOptions` emit an option-ID → wave-output-device mapping (the input side already does this via `outChannelMapping` with `"stereo:" + deviceName`, `RoutingSyncHelper.hpp:73-97`), and have the four handlers pass the device string instead of `"master"`. Verify the round-trip in `getTrackAudioOutput` (`TrackController.cpp:320`, `output.getOutputName()`) matches what the selectors expect so the dropdown doesn't snap back.

## Native engine consideration (#1882)

Design the destination representation in the model layer (track output destination as data: master / track / hardware device+channel pair), not as TE-specific plumbing. The TE `setOutputToDeviceID` path is the interim implementation; the native engine's render plan will need the same hardware-output destination concept, so the serialized/track-level representation should be engine-agnostic from the start.

## Out of scope (separate issues)

- Crash when toggling output configuration repeatedly
- Zoom in/out key shortcuts + invertible drag-zoom direction preference
- Editing toolbox

lucaromagnoli:
Closes #2264.

Hardware entries in the track output dropdowns were stubbed: every selection ≥ ID 10 mapped back to `"master"` in all four handlers, so on multi-out interfaces (e.g. the reporter's NI Audio8 DJ) audio only ever left through the master pair. The TE backend already supported device routing; the UI just never passed the destination through.

## Changes

- `RoutingSyncHelper::populateAudioOutputOptions` emits an option-ID → wave-output-device mapping, mirroring the input path: stereo pairs (ID 10+) as `"stereo:<device>"`, monos (ID 100+) as the bare device name, "Out N" fallback when TE names are unavailable.
- The four selection handlers (TrackHeadersPanel, SessionView, MixerView, TrackInspector) pass the mapped device string to `setTrackAudioOutput` instead of `"master"`. Handlers that read the shared mapping inside the callback copy the string first, since the map is rebuilt when the change notification re-syncs selectors.
- `syncSelectorsFromTrack` re-selects the stored hardware destination so the dropdown doesn't snap back to Master.
- `TrackController::setTrackAudioOutput` strips the `stereo:` marker before `setOutputToDeviceID`; `getTrackAudioOutput` round-trips the bare name.
- New `AudioBridge::getOutputDeviceNamesByChannel()`, mirroring the input-side helper.

## Engine-agnosticism (#1882)

The stored representation stays a plain device string in `TrackInfo.audioOutputDevice`; the native engine's `parseTrackRoute` already classifies it as `External`, so nothing TE-shaped reaches the model layer. TE-specific surface is confined to the bridge layer (`stereo:` strip + device-name lookup). Caveat: the native engine's PlanCompiler currently sends every External output route to the default output, so under it these selections are stored but not yet honored — tracked in #2272.

## Testing

- New `tests/routing/test_hardware_output_routing_juce.cpp`: mapping emission (TE names + fallback), selector round-trip for pair and mono selections, `stereo:` stripping through a real TE track, master round-trip.
- Verified negatively: with the sync re-selection and marker strip reverted, 4 assertions fail exactly as the bug predicts.
- Full debug build, new suite, and `[routing]` catch2 tests (135 assertions) all pass.

Out of scope (per issue): crash when toggling output configuration repeatedly.

### DIFF
--- a/magda/daw/ui/components/mixer/RoutingSyncHelper.hpp
+++ b/magda/daw/ui/components/mixer/RoutingSyncHelper.hpp
@@ -139,9 +139,14 @@ inline void populateAudioInputOptions(RoutingSelector* selector, juce::AudioIODe
 inline void populateAudioOutputOptions(RoutingSelector* selector, TrackId currentTrackId,
                                        juce::AudioIODevice* device,
                                        std::map<int, TrackId>& outTrackMapping,
-                                       juce::BigInteger enabledOutputChannels = {}) {
+                                       juce::BigInteger enabledOutputChannels = {},
+                                       std::map<int, juce::String>* outChannelMapping = nullptr,
+                                       const std::map<int, juce::String>& teDeviceNames = {}) {
     if (!selector)
         return;
+
+    if (outChannelMapping)
+        outChannelMapping->clear();
 
     std::vector<RoutingSelector::RoutingOption> options;
     options.push_back({1, technicalText(TechnicalTextToken::Master)});
@@ -237,13 +242,25 @@ inline void populateAudioOutputOptions(RoutingSelector* selector, TrackId curren
                 }
             }
 
+            // Helper: get the TE device name for a channel index, fall back to "Out N"
+            auto getDeviceName = [&](int channelIndex) -> juce::String {
+                auto it = teDeviceNames.find(channelIndex);
+                if (it != teDeviceNames.end())
+                    return it->second;
+                return "Out " + juce::String(channelIndex + 1);
+            };
+
             int id = 10;
             for (int i = 0; i < activeIndices.size(); i += 2) {
                 if (i + 1 < activeIndices.size()) {
                     int ch1 = activeIndices[i] + 1;
                     int ch2 = activeIndices[i + 1] + 1;
                     juce::String pairName = juce::String(ch1) + "-" + juce::String(ch2);
-                    options.push_back({id++, pairName});
+                    options.push_back({id, pairName});
+                    // Use actual TE device name for routing
+                    if (outChannelMapping)
+                        (*outChannelMapping)[id] = "stereo:" + getDeviceName(activeIndices[i]);
+                    ++id;
                 }
             }
 
@@ -254,7 +271,10 @@ inline void populateAudioOutputOptions(RoutingSelector* selector, TrackId curren
             id = 100;
             for (int i = 0; i < activeIndices.size(); ++i) {
                 int channelNum = activeIndices[i] + 1;
-                options.push_back({id++, juce::String(channelNum) + " (mono)"});
+                options.push_back({id, juce::String(channelNum) + " (mono)"});
+                if (outChannelMapping)
+                    (*outChannelMapping)[id] = getDeviceName(activeIndices[i]);
+                ++id;
             }
         }
     }
@@ -403,7 +423,9 @@ inline void syncSelectorsFromTrack(
     juce::BigInteger enabledOutputChannels = {},
     std::map<int, juce::String>* inputChannelMapping = nullptr,
     const std::map<int, juce::String>& teDeviceNames = {},
-    std::map<int, TrackId>* midiInputTrackMapping = nullptr) {
+    std::map<int, TrackId>* midiInputTrackMapping = nullptr,
+    std::map<int, juce::String>* outputChannelMapping = nullptr,
+    const std::map<int, juce::String>& teOutputDeviceNames = {}) {
     bool hasAudioInput = !track.audioInputDevice.isEmpty();
     bool hasMidiInput = !track.midiInputDevice.isEmpty();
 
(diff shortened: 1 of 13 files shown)


lucaromagnoli (commented):
Two inline routing correctness findings from the review.

magda/daw/ui/components/mixer/RoutingSyncHelper.hpp
```diff
@@ -254,7 +271,10 @@ inline void populateAudioOutputOptions(RoutingSelector* selector, TrackId curren
             id = 100;
             for (int i = 0; i < activeIndices.size(); ++i) {
                 int channelNum = activeIndices[i] + 1;
-                options.push_back({id++, juce::String(channelNum) + " (mono)"});
+                options.push_back({id, juce::String(channelNum) + " (mono)"});
+                if (outChannelMapping)
+                    (*outChannelMapping)[id] = getDeviceName(activeIndices[i]);
```
lucaromagnoli: **[P2] Preserve mono channel identity in hardware destinations**

With a normal TE stereo-pair configuration, `getOutputDeviceNamesByChannel()` returns the same wave-device name for both member channels (for example, channels 0 and 1 both map to `Out 1 + 2`). That makes IDs 100 and 101 identical here: selecting `2 (mono)` routes to the stereo device, and the next selector sync finds ID 100 first and visually snaps to `1 (mono)`. Conversely, when TE exposes mono devices, the stereo option above maps only the first mono device. Please preserve the actual channel/channel-set identity in the model destination and resolve it in the backend, or expose only options supported by the current wave-device configuration. A regression assertion should also verify that the two mono IDs remain distinguishable and reach different channels.

magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
```diff
@@ -819,8 +822,9 @@ void TrackHeadersPanel::setupRoutingCallbacks(TrackHeader& header, TrackId track
 
     // Capture outputTrackMapping_ by value so each header has its own snapshot
     // (the shared member is rebuilt per-header in populateAudioOutputOptions)
-    header.outputSelector->onSelectionChanged = [trackId,
-                                                 mapping = outputTrackMapping_](int selectedId) {
+    header.outputSelector->onSelectionChanged = [trackId, mapping = outputTrackMapping_,
+                                                 channelMapping =
+                                                     outputChannelMapping_](int selectedId) {
```
lucaromagnoli: **[P2] Keep the header hardware mapping in sync with rebuilt options**

This captures `outputChannelMapping_` only when the header is created. `updateRoutingSelectorFromTrack()` can later rebuild the selector options and the shared mapping (notably after an audio-device or output-pairing change), but it does not reinstall this callback. The UI can therefore show IDs from the new map while a selection routes through the old snapshot, potentially choosing the previous device or falling back to master. Please keep the hardware map per header or refresh the capture whenever options are rebuilt; the callback can still copy the selected string before calling `TrackManager` to avoid synchronous re-population invalidating it.

magda/daw/ui/components/mixer/RoutingSyncHelper.hpp
lucaromagnoli: Fixed in 487746a2. Options are now grouped by the wave device the channels belong to: a stereo pair is one option and a mono option exists only where TE actually has a mono wave device, so no offered option collapses to another's destination and none is un-honorable. Distinct mono devices keep distinct mapped strings; the test now asserts a pair produces no mono entries, that `Main L`/`Main R` mono devices stay distinguishable, and that a trailing mono device round-trips through the selector.

magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
lucaromagnoli: Fixed in 487746a2. The callback now reads the shared `outputChannelMapping_` member at selection time instead of snapshotting at header creation — the hardware mapping is device-global (unlike the per-track `outputTrackMapping_` snapshot, which stays as it was), so every rebuild via `updateRoutingSelectorFromTrack()` is picked up. The selected string is copied before the `TrackManager` call to survive synchronous re-population.

PR #2275, opened 2026-08-31
Take the projects that host a plugin, and the silence they found (#2175)

lucaromagnoli:
The rest of #2175, and it closes it: the plugin field on a case's environment block, the rule that an absent plugin is not a pass, and the three corpus projects that carry real VST3 state. Sits on top of #2269.

## The gate

`absentPluginsIn()` asks the scan, before either leg is driven, which of a project's plugins this machine has not got.

A project rendered without the plugin it names is a different project: both legs bind nothing in that slot and null against each other perfectly, which is a case passing by having tested less than it claims. So it does not run at all, and it is reported as **not run** rather than as a pass or as a complaint. It is not a fact about the engine or the harness, it is a fact about which plugins are on the desk, and every CI runner has none of them.

Strict in one direction on purpose. It walks the model rather than the compiled plan, so a plugin on a bypassed chain refuses the case too. The cost of that is a case reported not run on a machine that could have measured it, with the plugin named; the cost of the other way round is a green case that rendered a project nobody has.

## The scan, which the corpus has to do itself

Left alone, none of this would ever run. A test binary runs under a sandboxed `HOME` (the Makefile's `TEST_ENV`), so the application data directory the app keeps its `KnownPluginList` in is not this machine's, and it must not be: a suite whose verdict depends on what somebody last clicked in a scan dialog is not a suite. The list is therefore empty on every machine, and a gate that is always closed tests as little as one that is always open.

So the runner looks for itself, which is the honest source anyway. It walks the plugin folders, which is cheap, and then loads only the bundles whose name matches one a corpus project asked for: three files at most, each named in a manifest somebody wrote. That is deliberate rather than convenient. A full scan loads every bundle installed, which is minutes on a desk with a library on it and one bad plugin away from taking the suite down, which is why the app scans out of process.

Names are matched by containment either way, folded for case: a project names a plugin what its author's host called it, and FabFilter ships "Pro-L 2" inside "FabFilter Pro-L 2.vst3". A false match costs one bundle loaded that resolution then declines, because `matchInstalledPlugin` has the project's own identifier to check against; a missed match costs a case that never runs on a machine that could have measured it.

## The projects

- **project.retrospect** (0.13.0): two audio tracks each through their own Retrospect, one of them with an internal limiter after it. An external device and an internal one in one chain, and both instances carry state a released build wrote.
- **project.hostedinstrument** (0.18.0): two MIDI clips back to back into a hosted Pianoteq. A hosted instrument rather than a hosted effect, which is the other path through the plan. No audio sources at all, so whatever comes out came out of Pianoteq, and the notes that went in are compared as a stream beside it.
- **project.multitrack** (0.10.2): twenty-three stems under three group tracks, Pro-L 2 on the snare and on the master. The only project in the corpus with a track hierarchy, and the only one hosting the same plugin twice.

All three declare `AudioTier::Invariants` and the rig refuses them otherwise, because a plugin frames its own work and a residual measured across one is a number about the plugin. `MgdFixture::hostedPlugins` is checked both ways, like the sources beside it: a plugin the manifest missed is refused, a name no device claims is refused.

**envfollower stays out**, which is the one exclusion that reads like an exception and is not. Its two Retrospect instances are both on muted tracks, so an arrangement render of it is one internal filter and no plugin at all, and it carries 4OSC besides. It belongs with the seven 4OSC exclusions.

## A real bug, which is what these projects are for

**project.retrospect rendered silent.**

A device's parameter window is contiguous and indexed from zero, because that is what lets a device be handed a span and read its own parameters by the index it declared them at. An index nothing declared still costs a slot, and that slot was published as `0.0`. To the device reading it, a fabricated zero is indistinguishable from a value somebody set.

Retrovid was saved before the wet/dry pair was persisted, so it has no parameters at indices 0 and 1. The two fabricated zeros set the pair fully dry and fully attenuated: silence, out of a plugin whose own chunk had been restored perfectly.

A sparse array is the normal shape of a hosted plugin's saved parameters rather than a model to correct: the fork's list is a filtered view of the instance's, and a project saves what it had. So an undeclared slot now publishes no segments at all, which is exactly the answer a device already gets for an index past the end of its window, and `EngineExternalDevice::writeParameters` already does the right thing with it. The contiguity diagnostic stays for internal devices, where a gap is a device that skipped an index, and goes for hosted ones, where diagnosing it would make every project saved by an older build unmeasurable.

Two regression tests in `test_param_table.cpp`, one per side of that rule.

## Two things the invariants tier could not ask

**The tail.** It asks what is left running after the material stops, and a case that renders eight beats of a five-hundred-beat arrangement stops first: what is in its last fifty milliseconds is the music, and reporting it as a leak is the check misreading the case. Cases declare whether they render past their material. The tail is still measured and still printed, marked with a `?` so a number nobody asked for cannot be read as a check that held.

**The assertion watch.** It is a process-wide hook, and its premise, that everything asserting here is the engine or the harness, is true of every case in the corpus but the three that load somebody else's code into the process. Retrospect is an LV2 plugin behind a VST3 shell and hands JUCE its own URI where a path is expected, once per instantiation. On a case that hosts a plugin the watch now records rather than condemns, and the report prints it on the case's line. Nothing is dropped; what changes is what it is allowed to conclude.

## Verification

Full Catch2 suite passes (651,672 assertions) and the whole JUCE suite passes.

`project.retrospect` holds on this machine, at the invariants tier, with both Retrospect instances resolved:

```
[ 7] project.retrospect  invariant step=-3.0  tail=0.7?  length=ok  ok
     [rate=44100 blockSize=512 channels=2 seed=0 plugin="Retrospect 1.0.0 (VST3)" x2]
[ 8] project.hostedinstrument  invariant not run: this machine has not scanned "Pianoteq 8"  --
[ 9] project.multitrack        invariant not run: this machine has not scanned "Pro-L 2"     --
```

The step bound is the measured figure with its mechanism named: -3.0 dB is 0.71 of full scale in one sample, and Retrospect splices, so a splice between two phases of a tone steps by the distance between them. What the bound refuses is a step larger than a splice, which is what a graph transition without a ramp leaves behind.

The other two report not run, which is what a machine without Pro-L 2 or Pianoteq owes them, and what every CI runner will report for all three. The block-size gate skips them for the same reason and says so: it is Catch2 over the model with nothing beside it that could scan a plugin folder without turning a fast gate into a slow one.

No existing case changed pass/fail state.

### DIFF
--- a/tests/NullDiffRunner.cpp
+++ b/tests/NullDiffRunner.cpp
@@ -531,6 +531,7 @@ CaseReport runCase(const Case& value, const RunnerLog& log) {
         InvariantOptions options;
         options.sampleRate = value.sampleRate;
         options.maxStepPerSample = value.maxStepPerSample;
+        options.minPeakDb = value.minPeakDb;
         options.asksForTail = value.rendersPastItsMaterial;
 
         report.hasInvariants = true;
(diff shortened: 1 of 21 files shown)


tests/NullDiffRunner.cpp
```diff
@@ -419,6 +531,7 @@ CaseReport runCase(const Case& value, const RunnerLog& log) {
         InvariantOptions options;
         options.sampleRate = value.sampleRate;
         options.maxStepPerSample = value.maxStepPerSample;
+        options.asksForTail = value.rendersPastItsMaterial;
```
lucaromagnoli, line 535: **P1 — Require liveness from each invariant render**

All three new hosted-plugin fixtures set `rendersPastItsMaterial = false`, so this branch skips the tail verdict. What remains—finite samples, equal length, and a bounded step—is also true of a completely silent buffer. The guard above rejects only when *both* legs are silent because residual-based tiers expose one-sided silence, but this tier returns without computing a residual. Consequently, if the native Pianoteq path goes silent while the incumbent sounds (or vice versa), MIDI can still match and the case reports `ok`; a hosted effect reduced to a no-op can likewise satisfy these invariants. Please require a declared per-side liveness/energy check before awarding an invariants pass.

— Codex

tests/NullDiffRunner.cpp
```diff
@@ -533,21 +646,52 @@ SuiteRun runSuite(const std::vector<Case>& cases, const RunnerLog& log) {
...
+        //
+        // Recorded and printed either way. Nothing is dropped; what changes is
+        // what it is allowed to conclude.
         if (auto fired = watch.take(); !fired.empty()) {
-            report.passed = false;
-            report.unmeasurable =
-                "the engine asserted while rendering: " + fired.front().toStdString() +
-                (fired.size() > 1 ? " (and " + std::to_string(fired.size() - 1) + " more)" : "");
-            run.asserted.insert(value.name);
+            const auto hosted = !externalDevicesIn(value).empty();
```
lucaromagnoli: **P2 — Do not waive every assertion for a hosted case**

This predicate only says that the model contains an external device; it does not identify which code raised an assertion. Once it is true, every assertion from the engine, graph compiler, harness, or plugin discovery is converted to informational output and the case can remain green. It even covers an external device on a bypassed chain that the render never instantiated. That means a genuine engine regression occurring in one of the three newly valuable hosted cases is silently forgiven. Please isolate/attribute plugin assertions, or narrowly allowlist the known Retrospect `juce_File.cpp:219` assertion, while continuing to fail unrelated assertions.

— Codex

tests/NullDiffRunner.cpp
lucaromagnoli: Fixed in `e7d9a16e`. The invariants tier now requires a finite `minPeakDb`, measures each leg independently, includes `bothSound` in `passed()`, and has regression coverage for silence on either side with the tail waived. The three hosted fixtures declare a conservative liveness floor. This addresses the finding.

— Codex

tests/NullDiffRunner.cpp
lucaromagnoli: Fixed in `e7d9a16e`. Hosted assertions are now matched against explicit per-case substrings; every unexplained assertion still fails the case, and an expected assertion that stops firing also fails so stale exceptions cannot accumulate. Retrospect declares only the known `juce_File.cpp:219` assertion. This addresses the finding.

— Codex

PR #2278, opened 2026-08-31
A hardware insert is a send op and a return op (#2245)

### LINKED ISSUE #2245: External plugins slice 4: hardware inserts as a send and return pair
Slice 4 of #1893. Depends on the instantiation and state slices.

Hardware inserts: audio and MIDI leaving the machine, coming back, and being aligned as though it
had never left.

In the plan this is a send/return op pair rather than a device with special cases, which is the
difference #1893 asks for. The current engine special-cases an insert by where it sits in a chain;
the plan has ops for sending and receiving already, and an insert is those two with the outside
world between them. Latency is what makes it real: the round trip through the interface is measured
and compensated by the same PDC the rest of the graph uses, so a chain with an insert in it lines
up with a chain without one.

Also here: insert capture for bounce. An offline render cannot run the outside world at faster than
real time, so a bounce either runs the insert at real time or plays back a capture of it, and the
current engine does the latter.

The feature shipped in 0.15 with hardware verification still outstanding, so this slice is the one
place in the epic where the incumbent's own behaviour is not fully pinned down by a test. The
verification and the port are the same afternoon's work with the interface plugged in.

lucaromagnoli:
Slice 4 of #1893. The incumbent recognises an insert by where it sits in a chain and wires it there; the plan already had ops that consume a signal and ops that produce one, so an insert is one of each with the outside world between them.

**Scope note:** this opened with an insert-capture implementation in it as well. That came back out (see the last commit and #2279) -- it was one dual-mode object owning seven jobs, and the review's last finding was structural rather than a missed conditional. What is left is the plan slice, which is what #2245 is actually about.

## What an insert is, as model values

It lived in a `te::InsertPlugin`'s ValueTree, which meant the only thing that knew what an insert sent to was the fork. `TrackManager::getExternalInstrumentRouting` reached into the live plugin to fill in the read-only mirror the track selectors show, so a project that had loaded but whose plugins had not been built yet showed an empty pair, and so did one whose path would not resolve.

`DeviceInfo::insert` carries it now, the project serialises it, and `InsertConfigBridge` is the one place the two representations meet. Which direction one device needs is stated once, in `insertSyncDirectionFor`:

- a project that carries the config wins, because the model is what the plan compiles from and what the fork plays has to be the same insert;
- a project saved before the model carried any of this is filled from the plugin it just restored -- without that the compiler emits an ordinary Device op for an insert, for every project anybody already has;
- a plugin that resolved to nothing is not read back, because an insert whose ports this machine does not have would otherwise erase what the project says the next time it is saved.

The types are the fork's own derivation from the names rather than the model's guess. A send pointing at a device that is not enabled resolves to nothing, so a bridge that read the name and called it an audio send would describe an insert this machine cannot make. The JUCE test asserts exactly that: writing a config claiming both ends are audio leaves the plugin reporting `noDevice`, because the names it was handed do not resolve here.

## The pair

- **InsertSend** consumes the chain's audio or MIDI and declares no output, because what it writes leaves the machine. A sink, for the same reason the hardware output is.
- **InsertReturn** has no inputs and declares one port from what the insert says comes back.

That asymmetry is the point. An external effect sends audio and gets audio back; an external instrument is fed MIDI and answers with audio. The two ends were never the same shape, which is why this is two ops rather than one op used twice.

## The alignment, which is what makes it real

The return declares the round trip when the plan is prepared, in the same array a device's latency goes into, and the same pass compensates it. There is no insert-shaped code in the latency pass and there is not meant to be: a chain with an insert lines up with a chain without one because nothing doing the lining up knows what an insert is.

Pinned in both directions, because a pass that always held something back would satisfy the first test and be wrong.

## Half configured is reported

An end with a type and no device named it, and a send with nothing coming back. In the second case the chain carries on with the signal it already had, which is the failure that leaves a person still hearing their track; the diagnostic is what stops that reading as an insert that worked.

An unbound insert is reported rather than passed through, unlike a device that failed to load. There is no passing through an insert: a render that went ahead silently would be a render of a project with the outboard unplugged, with nothing in the audio to say so.

## What is not here

Insert capture for bounce, which #2245 lists and which #2279 now carries. `EngineInsert` is already the seam for it -- an implementation returning captured samples and one returning what an interface just handed it satisfy the same two calls -- and the interface says why it has no notion of an implementation that is present but unfit to render: nothing here could enforce one. The executor binds a pointer and calls it, so a validity flag is a rule some caller has to remember, and the render that forgot would be wrong invisibly. A capture that cannot be replayed has to fail to be constructed rather than fail to be valid, and that is a decision for whatever owns captures.

## Tests

- `tests/engine/test_plan_insert.cpp`: the pair rather than a Device op, the chain carrying on from the return, an external instrument's MIDI-out/audio-back shape, a MIDI return, both half-configured cases, a bypassed insert compiling to nothing, the round trip aligning a parallel path and a plan without one not being delayed, the unbound diagnostic, and an end-to-end render through a bound insert.
- `tests/devices/test_insert_config_bridge_juce.cpp`: both directions against a real `te::InsertPlugin`, including which end is which -- the fork names its ends for what they are to the machine, so `outputDevice` is the send, and a bridge that read them the other way round would compile a plan that sent audio to an input.
- `tests/devices/test_external_insert_registry.cpp`: the three cases of the sync direction rule.
- `tests/project/test_project_serialization.cpp`: the config round-tripping (with a negative manual adjustment, which is a figure somebody with a loopback really measures), and an ordinary device writing no insert block at all.

## Verification

Full Catch2 suite passes (651,855 assertions), the whole JUCE suite passes, and the app builds.

Hardware verification is still outstanding, which #2245 says up front: this is the one slice in the epic where the incumbent's own behaviour is not fully pinned down by a test, and the verification is the same afternoon's work as the port with an interface plugged in.


magda/daw/audio/plugin_manager/PluginManagerSync.cpp
```diff
@@ -2415,6 +2416,19 @@ te::Plugin::Ptr PluginManager::loadDeviceAsPlugin(const ChainNodePath& devicePat
...
+        // What the project says this insert sends to and gets back from (#2245).
+        //
+        // The model is the authority and the fork is what makes the sound, so
+        // the model is applied to it here, on the path every device takes on its
+        // way into the graph. A project saved before the model carried any of
+        // this has an inactive config and nothing is written, which leaves the
+        // plugin holding what its own state blob restored: the two
+        // representations are the same facts, and the older of them is still
+        // the one on disk.
+        if (isExternalInsert && device.insert.isActive())
```
lucaromagnoli: **P1 — Hydrate the model from legacy insert state**

For projects saved before this field existed, `device.insert` is inactive and the routing lives only in the restored `te::InsertPlugin` state. This branch deliberately skips the write, but no load/sync path calls `insertConfigOf` afterward (its only non-test caller is the UI edit callback). The model therefore stays inactive, so the native compiler emits a normal Device op instead of InsertSend/InsertReturn for every existing project. Please mirror the restored plugin state into the canonical DeviceInfo when the model has no insert config.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,134 @@
...
+    captured_ = 0;
+}
+
+void CapturedInsert::reset() {
+    if (live_ != nullptr)
+        live_->reset();
+}
+
+int CapturedInsert::latencySamples() const {
+    if (mode_ == Mode::Playing)
```
lucaromagnoli: **P1 — Preserve round-trip alignment when replaying**

The captured return is written at the current block timeline position before any graph compensation. With a round trip of L samples, the sample stored at t is actually the send from t-L; during live playback the latency pass aligns other paths to it. In Playing mode this now advertises zero latency, so those other paths are no longer delayed while the capture still returns the t-L sample at t, making the bounced insert path L samples late. Either shift the captured data to its compensated timeline position or retain equivalent latency during replay, and add a test whose live insert actually delays its input.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,134 @@
...
+    // audio through somebody's monitors at whatever speed the render is going.
+    if (mode_ == Mode::Playing || live_ == nullptr)
+        return;
+
+    live_->send(block, audio, midi);
+}
+
+void CapturedInsert::receive(const BlockInfo& block, juce::dsp::AudioBlock<float> audio,
+                             juce::MidiBuffer& midi) {
+    const auto numSamples = static_cast<std::int64_t>(audio.getNumSamples());
```
lucaromagnoli: **P1 — Capture MIDI returns as well as audio**

For a MIDI-return insert the executor calls `receive` with an empty audio block, so `numSamples` is zero here, no capture progress is recorded, and Playing mode later clears every returned MIDI event. A MIDI-return insert feeding a downstream instrument therefore works live but produces missing audio in the bounce. Use `block.numSamples` for the timeline span and retain/replay MIDI events at their absolute sample positions alongside the audio capture.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+int CapturedInsert::latencySamples() const {
+    if (mode_ == Mode::Playing)
+        return 0;
+    return live_ != nullptr ? live_->latencySamples() : 0;
+}
+
+bool CapturedInsert::covers(std::int64_t first, std::int64_t count) const {
+    return first >= 0 && count >= 0 && first + count <= captured_;
```
lucaromagnoli: **P1 — Track actual coverage instead of only the high-water mark**

`captured_` is only the greatest block end, so a seek directly to a later block makes `covers(0, end)` return true even though the entire gap was never captured. It also advances past the fixed buffer capacity when a block lands beyond ten minutes. In both cases the preflight check can approve a bounce that this class then fills with silence. Please track written intervals/pages (and clamp them to stored capacity) so `covers` can prove every requested sample was captured.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,134 @@
...
+    const auto capacity =
+        static_cast<std::size_t>(std::max(0.0, context.sampleRate * kCaptureSeconds));
+
+    // Only while capturing. A bounce is handed a capture that already exists and
+    // must not clear it, which is the one way this class could quietly turn a
+    // render into silence.
+    if (mode_ != Mode::Capturing)
+        return;
+
+    samples_.assign(static_cast<std::size_t>(numChannels_), std::vector<float>(capacity, 0.0f));
```
lucaromagnoli: **P2 — Avoid allocating ten minutes per insert up front**

This eagerly allocates about 212 MB for one stereo insert at 44.1 kHz, 230 MB at 48 kHz, and 922 MB at 192 kHz. A project with several inserts can exhaust memory merely by preparing capture, even when the requested export is short. Size storage to the requested capture window or use bounded/chunked backing storage prepared off the audio thread; the existing disk-backed capture path is a useful precedent.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+std::int64_t CapturedInsert::timelineSampleOf(const BlockInfo& block) const {
+    return static_cast<std::int64_t>(std::llround(block.startSeconds * sampleRate_));
+}
+
+void CapturedInsert::prepare(const RenderContext& context) {
+    if (live_ != nullptr)
+        live_->prepare(context);
+
+    sampleRate_ = context.sampleRate;
```
lucaromagnoli: **P1 — Keep capture and render sample rates distinct**

When a Playing instance is prepared for an export rate, this overwrites `sampleRate_` while `samples_`, `written_`, `window_`, and `capturedLatency_` are still expressed in capture-rate samples. A capture taken at 44.1 kHz and rendered at 48 kHz is therefore indexed with 48 kHz timeline positions and reports its 44.1 kHz latency as 48 kHz samples, producing shifted/stretched audio and eventually silence. Please retain the capture rate and resample/reindex the capture, window, MIDI positions, and latency for the render rate (or reject mismatched rates); the existing Tracktion capture service already performs this conversion.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,192 @@
...
+        const auto index = offset + sample;
+        if (index >= written_.size() || written_[index] == 0)
+            return false;
+    }
+
+    return true;
+}
+
+std::int64_t CapturedInsert::capturedSamples() const {
+    return std::count(written_.begin(), written_.end(), char{1});
```
lucaromagnoli: **P2 — Make progress reporting race-free and constant-time**

`capturedSamples()` is documented for progress polling, so it will be read off the audio thread while `receive()` writes the same plain `written_` bytes. That is a C++ data race, and scanning the entire export window on every UI poll is also O(window length) work per insert. Keep an atomic captured-count updated when a byte changes from 0 to 1; `covers()` can still inspect the bitmap after capture has stopped.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,192 @@
...
+                    audio.getSample(channel, static_cast<int>(fromBlock + sample));
+        }
+
+        // The MIDI too, at absolute positions, for the same reason the audio is
+        // written at its own: a MIDI-returning insert feeding an instrument
+        // downstream is silent in the bounce if this is dropped.
+        for (const auto metadata : midi) {
+            const auto at = first + metadata.samplePosition;
+            if (at >= from && at < to)
+                midi_.addEvent(metadata.data, metadata.numBytes, static_cast<int>(at));
```
lucaromagnoli: **P2 — Enforce the MIDI capacity instead of only reserving it**

`ensureSize(1 MiB)` is a reservation, not a ceiling. Once a dense/long capture or SysEx stream exceeds it, this `addEvent` grows the MidiBuffer on the audio thread, and the capture still reports success. Check capacity before appending and mark the capture invalid on overflow, or use fixed/pre-sized storage whose bound is derived off-thread, so this path cannot allocate or silently produce an incomplete bounce.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,254 @@
...
+                             juce::MidiBuffer& midi) {
+    // The block's own length rather than the audio's. A MIDI-returning insert
+    // is handed an empty audio block, and a span taken from that would be zero:
+    // nothing would be recorded as captured, and the bounce would replay a
+    // capture it believed was empty.
+    // The block's own length rather than the audio's, and in capture-rate
+    // samples: a block is a duration, and the two rates count it differently.
+    const auto seconds =
+        static_cast<double>(block.numSamples) / (renderRate_ > 0.0 ? renderRate_ : 1.0);
+    const auto numSamples = static_cast<std::int64_t>(std::llround(seconds * captureRate_));
```
lucaromagnoli: **P1 — Enforce the rate refusal before writing the render block**

`rateMismatch_` only makes `covers()` and `usable()` return false; `receive()` itself still runs when the generic PlanExecutor calls it. If a 96 kHz capture is prepared in a 48 kHz render, a 64-frame output block becomes `numSamples == 128` here, and the playback loop writes indices 0 through 127 into the 64-frame AudioBlock. The EngineInsert interface does not expose the preflight flag, so missing the external check becomes an out-of-bounds write rather than merely bad audio. In Playing mode, clear both outputs and return immediately when unusable before converting the block length.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,254 @@
...
+/// and the length. Counted into the ceiling so the check is about what will be
+/// stored rather than about what was handed over.
+constexpr int kMidiEventOverhead = 6;
+
+}  // namespace
+
+CapturedInsert::CapturedInsert(Mode mode, int numChannels, EngineInsert* live)
+    : mode_(mode), numChannels_(std::max(1, numChannels)), live_(live) {}
+
+std::int64_t CapturedInsert::captureSampleOf(const BlockInfo& block) const {
```
lucaromagnoli: **P2 — Do not prepare the live hardware during replay**

The class contract says `live_` is used only while capturing, but every Playing prepare still forwards the offline render context to it before checking the mode; `reset()` likewise resets it. An export at another rate can therefore reconfigure or reset the live hardware object even though replay sends and receives nothing through it. Gate both lifecycle calls on Capturing, or detach the live pointer when switching to Playing.

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+            // Bounded rather than merely reserved. ensureSize is a reservation
+            // and addEvent grows past it, which on this thread is an
+            // allocation; and a capture that allocated its way out of the
+            // problem would still be one with events missing from it, reporting
+            // success. So the ceiling is checked and the capture is marked
+            // unusable instead.
+            if (static_cast<int>(midi_.data.size()) + metadata.numBytes + kMidiEventOverhead >
+                kCapturedMidiBytes) {
+                midiOverflowed_ = true;
```
lucaromagnoli: **P2 — Publish MIDI overflow without a data race**

This flag is written on the audio thread, while `usable()` is a public status/preflight query intended to be polled from the capture controller; the new test even checks it while capture is in progress. Because `midiOverflowed_` is a plain bool, that read/write is a C++ data race. Make the failure flag atomic, as the progress counter is (the existing Tracktion capture path also uses an atomic failure flag).

— Codex

magda/engine/io/CapturedInsert.cpp
```diff
@@ -0,0 +1,254 @@
...
+                break;
+            }
+
+            midi_.addEvent(metadata.data, metadata.numBytes, static_cast<int>(at));
```
lucaromagnoli: **P2 — Store MIDI positions relative to the capture window**

`at` is a 64-bit absolute timeline sample, but MidiBuffer timestamps are 32-bit ints, so this cast wraps events in long sessions (about 3.1 hours at 192 kHz or 12.4 hours at 48 kHz). Playback then compares the wrapped value to the 64-bit window and silently loses the events. Store `at - window_.firstSample` in the MidiBuffer and add the window origin during lookup; the capture allocation already makes the relative span the bounded domain.

— Codex

Issue #2174, opened 2026-08-23
The device layer under both null-diff legs


lucaromagnoli:
Part of #1896. What #2081 is actually blocked on.

#2081 names #1891 and #1892 as its blockers and both are now done or all but. The blocker it does
not name is that neither leg can run a device a real project contains.

The corpus knows one device. It is a gain with one parameter, written once as a contract in
`tests/NullDiffGain.hpp` and implemented twice, at `NullDiffNativeLeg.cpp:114` and
`NullDiffTeLeg.cpp:167`, because until #2123 a `Device` op resolved to a stand-in and the incumbent
instantiated none of the model's devices. That was the right amount of device for a slice about
parameters: what a gain renders is the value of its own parameter, so a case that plays a constant
into it draws the curve directly and nine cases null.

A real project is not a parameter case. It contains the devices MAGDA ships, and on the incumbent
side those arrive through `PluginManager` and the sync layer behind it -- the boundary this corpus
has refused to cross since #2040. The refusal is on the record and it was right at the time:
writing plugins straight into the leg is a second sync, and a second sync is something that can
agree with itself while both engines are wrong.

Crossing it properly is this slice. The incumbent leg drives the app's own device path rather than
a copy of it, the native leg resolves a `Device` op through the host's factory rather than a
stand-in, and both are handed the same model devices out of the same case. What keeps it honest is
that the devices are the app's: a device implemented twice for the corpus only ever proves the
corpus can implement a device twice.

Three things section 7 of `docs/architecture/native-engine.md` names rather than covers come back
inside this, each having been tried once and stopped at this same boundary. Rack-scope macros want
a `te::RackType` that `RackSyncManager` builds out of a `PluginManager`. Sends have their levels
resolved by the value layer and their `SendTap` ops emitted pre and post fader already, and stop
because the incumbent's half lives on `te::AuxSendPlugin` instances `PluginManagerSync` creates.
The envelope follower stops because the fork feeds it through a `FollowerSourceTapPlugin` that
`PluginManager` installs, so without the device layer the fork's follower is handed nothing. Each
is a case the day the boundary moves, and each is cheaper to write then than to rediscover later.

The envelope follower carries a finding with it rather than only a case. Its transport gate is the
fork asking `TransportControl::isPlaying()`, false throughout every offline render, so a
transport-triggered envelope does nothing in a bounce in the current engine while the engine being
built plays it. That is a divergence to pin with its mechanism, not a case to calibrate.

[...]

PR #2280, opened 2026-09-01
The corpus hosts a plugin, and two hosts stop disagreeing about it (#2246)

### LINKED ISSUE #2246: External plugins slice 5: parity cases for external plugin hosting
Slice 5 of #1893, and the last of it. Depends on every slice above.

The parity cases for external plugin hosting, written here rather than in #1896, for the reason the
scope note on this issue gives: a parity case is worth what its author's knowledge of the feature is
worth, and that knowledge is at its peak on the day the port lands. Written later by someone reading
the code back, a case asserts the code rather than the behaviour.

What the cases have to cover is what the port claimed:

- The wet/dry pair at values a project can actually save, including fully dry, which still runs the
  plugin.
- A mono plugin in a stereo chain and a stereo plugin in a mono slot, which are the two channel
  adaptations with arithmetic in them.
- A plugin reporting latency, aligned against the same chain without it.
- A sidechained plugin, where the key arrives on the bus after the plugin's own.
- A plugin that is asked where the transport is, which is the case a null cannot be asserted for
  without a tempo-synced plugin to ask.
- MIDI in and out of an instrument.

These are the invariant tier rather than the null tier wherever the plugin is entitled to frame its
own work (#2075), and a project hosting one buys the block-size epsilon the gate refuses to
projects with nothing to attribute a difference to (#2078). The real projects that host a plugin are
#2175 and land against this corpus.

lucaromagnoli:
Slice 5 of #1893, and the last of it. Every other device in the corpus is written twice, once for each leg, from a contract that says what the two owe each other. An external plugin is the one device that cannot be written twice, and that is the whole reason these cases exist: neither engine wrote it and neither can ask it anything, so what the two hosts do around it is the entire difference between them.

## One implementation, two hosts

`NullDiffHostedPlugin` is a `juce::AudioPluginInstance` behind a `juce::AudioPluginFormat`, registered with the engine's own format manager and listed in the same `KnownPluginList` both legs resolve against. It arrives as a `te::ExternalPlugin` on one side and an `EngineExternalDevice` on the other, by the ordinary path, and nothing in either engine knows the plugin is ours.

It is in the binary rather than on the machine, and that is not a shortcut around #2175. The two questions are different. A real project hosting a real plugin asks whether this machine's Pro-Q renders the same under both engines, and it can only be asked where that plugin is installed. These ask what the host does -- which channels a narrow plugin is handed, what the wrapper pair means, where a key lands, what a plugin is told about the transport. Every one of those is an assertion about a number the host chose, and a plugin that reports exactly what it was given answers it on CI, where every real plugin is absent.

Each role is shaped so the answer is readable off the render rather than inferred from it: the polarity plugin makes the wrapper pair's arithmetic the whole signal, the key plugin returns its sidechain instead of its input so a render of the wrong bus is silence, and the transport plugin renders where it thinks it is.

## The nine cases

| case | what it pins |
| --- | --- |
| `plugin.wetdry` | the wrapper pair at dry 0.6 / wet 0.4 over an inverting plugin |
| `plugin.wetdry.dry` | a fully dry slot in front of a plugin that still runs |
| `plugin.mono` | a mono plugin folded into and out of a stereo chain |
| `plugin.narrow.slot` | a stereo plugin in a mono slot |
| `plugin.latency` | a reported latency, taken back out, against a track without one |
| `plugin.sidechain` | a key on the bus after the plugin's own |
| `plugin.transport` | the position a plugin is told it is at |
| `plugin.instrument` | MIDI into a hosted instrument |
| `plugin.instrument.midiout` | an instrument's MIDI output reaching the device behind it |

All nine null exactly (peak `-inf`), and all nine hold **bit identity** through the block-size gate (#2078) rather than taking the epsilon a project hosting a plugin may declare. The plugin is ours and nothing it does depends on how a render was framed, so there is nothing here to attribute a difference to and the cases do not spend the allowance on the host.

## Two of them found bugs

**A mono plugin was handed one side of the bus.** The fork averages a stereo bus into a mono plugin and spreads the result back over both sides, and `EngineExternalDevice` transcribes every step of that -- but the plan narrowed the port to one channel first, so the device got the left channel and the average was never taken. The arithmetic was in the adapter, unreachable. Measured before the fix: a peak residual of -6.0 dB, which is exactly `|L - R| / 2` on the case's two-channel noise.

`PlanCompiler` now leaves an external plugin's widths at the bus. The adaptation is the adapter's, it is arithmetic rather than routing, and every project hosting a plugin that is not stereo was mixed against the fork's version of it.

**A render came back late by the plugin's latency.** A plan whose devices report latency is that far behind the timeline, which is right during playback and wrong in a file: the whole render sat 333 samples late, so a bounce would drift against every other bounce of the same project, and against the project itself the moment it is imported back in.

`renderOffline` pulls the plan's latency in extra samples at the end and drops the same number off the front, so the file starts where the range does and material still inside a delay line when the range ends comes out rather than being cut off. `OfflineRenderResult` now reports what the sink received rather than what the render pulled.

The incumbent compensates by losing its first 333 samples instead -- the impulse at beat zero is simply not in its file, while every impulse after it sits at its own sample. That is why the case's clips start a beat in, and the case says so where it does it: it is about alignment, and the fork's own head loss is not what it is measuring.

## What the corpus's own plugins changed around them

- The block-size gate and the native-leg test build a scan of their own instead of reporting these cases as not gated for want of a plugin no scan here could find. A real project's plugin is still absent, still skipped, still named.
- The DAWproject round trip maps **device** ids back by name, the way it already maps tracks and clips. The comment that said this would be needed "the day a case carries a VST3" was right; this is that day. It also declares two losses the format really has: a device's sidechain source, which is a reference from a device to a track, and what its MIDI ports do, which is a capability read off a live plugin.
- The plan goldens and the compiler and executor tests now say `Internal` on the devices they stand up. `DeviceInfo::format` starts at VST3, and a fixture about declared widths has to be a device that declares them -- the same correction #2174 made in the corpus.

## Tests

- `tests/NullDiffHostedPlugin.{hpp,cpp}`: the plugin, its format and the model device that names it.
- The nine `plugin.*` cases in the corpus, through both legs, the block-size gate and the DAWproject round trip.
- `test_render_plan_compiler.cpp`: an external plugin is handed the bus whatever it declares.
- `test_offline_render.cpp`: a compensated pure delay renders what the same project renders without it, asserted as an identity rather than as an offset.

## Verification

Full Catch2 suite passes (740,358 assertions), the whole JUCE suite passes, and the app builds.

### DIFF
--- a/magda/engine/exec/OfflineRender.cpp
+++ b/magda/engine/exec/OfflineRender.cpp
@@ -218,11 +218,19 @@ OfflineRenderResult renderOffline(PlanExecutor& executor, const PlanValues& valu
     // same request a moment later.
     clock.advance(snapshot, context.sampleRate, 0);
 
-    // The latency goes on whichever span is last, so that the samples it flushes
+    // A request with nothing in it renders nothing, latency or not. The flush
+    // below exists to push the last of a range through the graph, and a range
+    // with no samples in it has nothing to push: rendering it anyway would run
+    // every bound device and ask the caller whether to continue, for blocks the
+    // sink is never handed, so an empty request could come back cancelled or
+    // leave a delay line holding something.
+    const auto flushSamples = rangeSamples > 0 || tailSamples > 0 ? latencySamples : 0;
+
+    // The flush goes on whichever span is last, so that the samples it produces
     // are the continuation of what came before them: with a tail, the graph is
     // still ringing out with the transport stopped, and without one the sources
     // play on past the range while the kept window ends where it was asked to.
-    renderSpan(rangeSamples + (tailSamples > 0 ? 0 : latencySamples));
+    renderSpan(rangeSamples + (tailSamples > 0 ? 0 : flushSamples));
 
     if (result.cancelled || tailSamples <= 0)
         return result;
@@ -241,7 +249,7 @@ OfflineRenderResult renderOffline(PlanExecutor& executor, const PlanValues& valu
     snapshot.request.playing = false;
     snapshot.request.locate = false;
 
-    renderSpan(tailSamples + latencySamples);
+    renderSpan(tailSamples + flushSamples);
 
     return result;
 }
--- a/magda/engine/exec/OfflineRender.cpp
+++ b/magda/engine/exec/OfflineRender.cpp
@@ -15,6 +15,79 @@ std::int64_t samplesBetween(double startSeconds, double endSeconds, double sampl
     return static_cast<std::int64_t>(std::llround((endSeconds - startSeconds) * sampleRate));
 }
 
+/**
+ * @brief A sink with the plan's own latency taken off the front (#2246).
+ *
+ * A prepared plan's output is however far behind the timeline the devices on
+ * its longest path report between them, and during playback that is the whole
+ * point: the graph is aligned with itself and the interface's own buffer is
+ * what the listener hears it through. A file has no such excuse. A bounce whose
+ * first sample is not the range's first sample is a bounce that drifts against
+ * every other bounce of the same project, and against the project itself the
+ * moment it is imported back in.
+ *
+ * So the render pulls the plan's latency in extra samples at the end and this
+ * drops the same number off the front. What the sink receives is exactly the
+ * range it asked for, at the position it asked for it, with nothing lost:
+ * material still inside a delay line when the range ends comes out during those
+ * extra samples rather than being cut off.
+ *
+ * Nothing at all when the plan reports no latency, which is every project
+ * without a latency-reporting plugin in it.
+ */
+class LatencyTrimmedSink final : public OfflineRenderSink {
+  public:
+    LatencyTrimmedSink(OfflineRenderSink& target, int latencySamples)
+        : target_(target), remaining_(latencySamples) {}
+
+    void write(const juce::AudioBuffer<float>& block, int numSamples) override {
+        if (remaining_ <= 0) {
+            forward(block, numSamples);
+            return;
+        }
+
+        if (remaining_ >= numSamples) {
+            remaining_ -= numSamples;
+            return;
+        }
+
+        // The block the trim ends inside, handed on as the part of itself that
+        // is past it. A view rather than a copy: the sink's contract already
+        // says the buffer is the renderer's and is reused, so a sink that wants
+        // to keep these samples was going to copy them anyway.
+        const auto skipped = remaining_;
+        remaining_ = 0;
+
+        juce::AudioBuffer<float> rest(const_cast<float* const*>(block.getArrayOfReadPointers()),
+                                      block.getNumChannels(), skipped, numSamples - skipped);
+        forward(rest, numSamples - skipped);
+    }
+
+    /// What the sink behind this one actually received. The result reports
+    /// these rather than what the render pulled: a caller writing a file is
+    /// owed the length of the file, and the samples that went into the trim
+    /// never reached it.
+    std::int64_t forwardedSamples() const {
+        return forwardedSamples_;
+    }
+
+    int forwardedBlocks() const {
+        return forwardedBlocks_;
+    }
+
+  private:
+    void forward(const juce::AudioBuffer<float>& block, int numSamples) {
+        target_.write(block, numSamples);
+        forwardedSamples_ += numSamples;
+        ++forwardedBlocks_;
+    }
+
+    OfflineRenderSink& target_;
+    int remaining_ = 0;
+    std::int64_t forwardedSamples_ = 0;
+    int forwardedBlocks_ = 0;
+};
+
 }  // namespace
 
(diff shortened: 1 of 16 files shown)


tests/CMakeLists.txt
```diff
@@ -383,6 +383,7 @@ if(MAGDA_BUILD_NATIVE_ENGINE)
     list(APPEND TEST_SOURCES NullDiffMaterial.cpp)
     list(APPEND TEST_SOURCES NullDiffCompare.cpp)
     list(APPEND TEST_SOURCES NullDiffCorpus.cpp)
+    list(APPEND TEST_SOURCES NullDiffHostedPlugin.cpp)
```
lucaromagnoli, line 386: **P1 — Exercise the hosted plugins in the DAWproject round trip**

Adding the fixture implementation to this target only satisfies the linker. `test_dawproject_round_trip.cpp::renderHolds` still calls the no-scan `renderNative(value)` overload, so every new `plugin.*` case binds a passthrough and both sides report the same missing-plugin diagnostic. I verified the consequence by supplying `HostedScan`: `plugin.wetdry` and `plugin.wetdry.dry` immediately diverge because DAWproject drops the host-owned `wrapperParameters`, and those losses are not declared here. Please pass the in-binary scan through the round-trip renders and declare/restore the wrapper pair (or make it portable), so this test actually hosts the plugin it claims to round-trip.

— Codex

magda/engine/exec/OfflineRender.cpp
```diff
@@ -135,7 +218,11 @@ OfflineRenderResult renderOffline(PlanExecutor& executor, const PlanValues& valu
     // same request a moment later.
     clock.advance(snapshot, context.sampleRate, 0);
 
-    renderSpan(rangeSamples);
+    // The latency goes on whichever span is last, so that the samples it flushes
+    // are the continuation of what came before them: with a tail, the graph is
+    // still ringing out with the transport stopped, and without one the sources
+    // play on past the range while the kept window ends where it was asked to.
+    renderSpan(rangeSamples + (tailSamples > 0 ? 0 : latencySamples));
```
lucaromagnoli: **P2 — Keep an empty render side-effect free**

When `rangeSamples == 0`, `tailSamples == 0`, and the plan reports latency, this now calls `renderSpan(latencySamples)`: it processes every bound device and consults `shouldContinue`, then trims every produced sample. An empty request can therefore mutate device state or return `cancelled` even though the sink receives no blocks. Preserve the existing empty-render contract with an early return after resolving the range and tail when both are zero.

— Codex

tests/DawProjectRoundTrip.cpp
```diff
@@ -740,6 +830,74 @@ bool takeClips(ProjectDocument& imported, const std::map<TrackId, TrackId>& trac
...
+ * Top-level chain devices only. It is where the corpus's hosted plugins sit,
+ * and the sections the format has no shape for arrive empty rather than
+ * renumbered.
+ */
+bool mapDeviceIds(Case& value, const Case& original, std::string& refusal) {
+    const auto mapTrack = [&refusal](TrackInfo& track, const TrackInfo& source) {
+        std::map<juce::String, DeviceId> byName;
+        for (const auto& element : source.chain.fxChainElements)
+            if (isDevice(element) &&
+                !byName.emplace(getDevice(element).name, getDevice(element).id).second) {
```
lucaromagnoli: **P2 — Do not use display names as device identity**

This refuses any project containing two top-level instances with the same display name, which is normal for chains containing two copies of one plugin. Unlike tracks and clips, the corpus has no device-name uniqueness contract, and adding one would encode a harness limitation into otherwise valid projects. Map surviving devices by ordered portable identity plus occurrence (using the name only for diagnostics), so duplicate plugin instances remain representable and a real reorder is still detected.

— Codex

lucaromagnoli (commented):
Architecture review focused on responsibility boundaries and whether the new corpus exercises the production paths it claims to cover.

— Codex

lucaromagnoli (commented):
Follow-up review of commit `76a34936f`: two findings are addressed cleanly; one regression assertion is currently inert.

— Codex

tests/engine/test_offline_render.cpp
```diff
@@ -116,6 +116,56 @@ class DecayDevice final : public EngineDevice {
...
+                stored = input;
+            }
+
+            position_ = (position_ + 1) % std::max(1, latency_);
+        }
+    }
+
+    /// Blocks it was handed, which is how a render that produced nothing for the
+    /// sink can still be caught having run the graph.
+    int blocksProcessed = 0;
```
lucaromagnoli, line 162: **P2 — Increment the counter this regression test relies on**

`blocksProcessed` is initialized and later checked, but `LatentDevice::process()` never increments it. The new empty-range test therefore passes even if the production flush fix is reverted and the device processes latency blocks. Increment this at the start of `process()` so the assertion actually pins the side effect it describes.

— Codex

PR #2281, opened 2026-09-01
Expose device parameter discovery and control over MCP (#2274)

### LINKED ISSUE #2274: Expose device parameter discovery + control over MCP
# Feature request: expose device parameter discovery + control over MCP

## Summary

The MCP server (`magda-mcp`) currently exposes rich read/write access to
tracks, clips, devices (list-only), racks, selection, transport, session,
and automation — but there's no way for an external MCP client to:

1. **Discover** what parameters a device (native or third-party VST3/AU)
   actually has (names, indices, ranges, current values).
2. **Set a parameter's value directly**, right now, the way turning a
   knob would (as distinct from writing an automation point at a beat
   position, which is a separate concept and doesn't help for "just
   change this setting").
3. **Open a device's editor UI**, for cases where direct value-setting
   isn't precise enough and a human needs to see/tweak it themselves
   after the AI gets it close.

This means an external MCP client (e.g. Claude Desktop connected via
`magda-mcp.exe`) can see that a device like `OP-X PRO-II` exists on a
track, but can't inspect or change any of its actual sound-design
parameters — even though MAGDA's own **Per-Device AI Panel** can do
exactly this today, just through MAGDA's internal AI configuration
(Settings → AI Settings) rather than through the MCP surface.

## Why this matters

The in-app AI Sound Designer (`Configure Parameters…` → tick **AI
Agent** per parameter → per-device **AI** panel or `/design`) proves
the underlying plumbing already exists: MAGDA can already enumerate a
plugin's AI-eligible parameters and set their values. Exposing that
same capability as MCP operations would let any MCP-connected AI
client (not just MAGDA's own configured model) do real sound design
work — matching what's already possible for tracks/clips/transport/etc.

## Proposed operations

Modeled on the existing `devices.list` shape (`devicePath`, `id`,
`trackId`, etc.) so it composes naturally with what's already exposed:

### `devices.listParameters`
- **access**: `read`, **requiredScope**: `read`
- **input**: `{ deviceId: integer }` (or `devicePath`, matching the
  existing convention in `devices.list`'s output)
- **output**: array of
  ```json
  {
    "index": 0,
    "name": "Filter Cutoff",
    "aiAgentEnabled": true,
    "unit": "Hz",
    "min": 20,
    "max": 20000,
    "currentValue": 800,
    "normalizedValue": 0.34
  }
  ```
  Ideally reflects the same `AI Agent` opt-in flag from **Configure
  Parameters…**, so a client can tell up front which parameters it's
  actually allowed to touch versus which are locked out by the user.

### `devices.setParameter`
- **access**: `write`, **requiredScope**: `edit`
- **input**: `{ deviceId: integer, parameterIndex: integer, value: number }`
  (raw or normalized — whichever `listParameters` reports back)
- **output**: the updated parameter object (same shape as above)
- Should respect the same `AI Agent` per-parameter allowlist as the
  in-app Sound Designer — i.e. reject (with a clear error) any
  parameter that hasn't been explicitly enabled, rather than silently
  allowing MCP clients to bypass a safeguard the user set up for the
  in-app AI.

### `devices.openEditor` *(nice-to-have, lower priority)*
- **access**: `write`, **requiredScope**: `edit` (or a new `ui` scope)
- **input**: `{ deviceId: integer }`
- **output**: `{ accepted: boolean }`
- Opens the plugin's native editor window in the MAGDA UI — useful as
  a fallback when an AI-driven change needs a human to fine-tune by
  ear afterward, without them having to hunt for the right device slot
  themselves.

## Context / how this was found

While testing the MCP connection from Claude Desktop, I could list
devices (`devices.list`) and confirm `OP-X PRO-II` was loaded on a
track, but had no way to either inspect its parameters or change them
— only `automation.createLane`/`automation.addPoint` exist for
parameter-adjacent writes, and those require a `parameterIndex` with
no way to discover what that index even corresponds to. The in-app
docs (Panels → AI Assistant → Per-Device AI Panel / third-party plugin
support) confirmed MAGDA already does this internally; this request is
just to expose that same capability over MCP for external clients.

lucaromagnoli:
Closes #2274.

Adds the three operations the issue asks for, addressed by `devicePath` like the rest of the device surface (a bare `deviceId` is ambiguous across a track's fx / post-fx sections):

- **`devices.listParameters`** (read) — every parameter in real units (`index`, `stableId`, `name`, `unit`, `minValue`/`maxValue`/`defaultValue`/`currentValue`), plus `normalizedValue` so a client can compose with `automation.addPoint`'s 0–1 domain, and the Configure Parameters customization flags: `visible`, `miniMixer`, `aiAgentEnabled`. A client can therefore both discover what `parameterIndex` means and audit how well a plugin's parameters are curated.
- **`devices.setParameter`** (write, `edit` scope) — real-unit write that rejects rather than clamps out-of-range values, and refuses (`permission_denied`) any external-plugin parameter the user has not ticked as **AI Agent** — the same allowlist the in-app sound designer honours, so MCP is not a way around it. Internal devices accept writes on every parameter (the dialog offers no AI opt-in for them). The write path goes through `TrackManager::setDeviceParameterValue`, so it reaches the engine via `setParameterFromHost` and works with modifiers active.
- **`devices.openEditor`** (write, `edit` scope) — new `DeviceApi::openDeviceEditor` over `AudioBridge::showPluginWindow`; returns `accepted` reflecting whether a window is actually open afterwards (false when headless or the device has no native editor), and does not advance the project revision.

Also updated:
- Scope policy table + `docs/architecture/remote-api-permissions.md`.
- DevicePanel agent surface: allowlists the three tools and drops the "parameter mutation is advisory" prompt fragment (preset mutation keeps its caveat).
- `MockDeviceApi`: `setDeviceParameter` now updates `currentValue` like the live manager, and records editor opens.

Tests: contract projection + round-trip (external allowlist semantics, internal always-enabled), dispatch tests for all three handlers (success, `permission_denied`, `validation_failed`, `not_found`, revision behaviour), console-routing allowlist assertions. The allowlist gate was verified negatively — removing it makes the refusal test fail.

Follow-up (agreed with @lucaromagnoli): a persisted `devices.setParameterConfig` needs the per-plugin XML store extracted out of `ParameterConfigDialog` and will be tackled separately.

### DIFF
--- a/magda/daw/api/remote_api_projection.cpp
+++ b/magda/daw/api/remote_api_projection.cpp
@@ -4,6 +4,7 @@
 #include "../core/AutomationInfo.hpp"
 #include "../core/ClipInfo.hpp"
 #include "../core/DeviceInfo.hpp"
+#include "../core/ParameterUtils.hpp"
 #include "../core/RackInfo.hpp"
 #include "../core/TrackInfo.hpp"
 #include "../project/ProjectInfo.hpp"
@@ -336,6 +337,39 @@ DeviceCatalogEntryDto makeDeviceCatalogEntryDto(const DeviceCatalogEntry& entry)
     dto.type = deviceTypeName(entry.type);
     dto.instrument = entry.isInstrument;
     return dto;
+}
+
+std::vector<DeviceParameterDto> makeDeviceParameterDtos(const DeviceInfo& device) {
+    const auto contains = [](const std::vector<int>& positions, int position) {
+        return std::find(positions.begin(), positions.end(), position) != positions.end();
+    };
+    // The customization lists key on position in `parameters`, while the wire
+    // `index` is `paramIndex` — the address parameter writes take. The two are
+    // normally equal, but only the position indexes the user's selections.
+    std::vector<DeviceParameterDto> dtos;
+    dtos.reserve(device.parameters.size());
+    for (size_t i = 0; i < device.parameters.size(); ++i) {
+        const auto& info = device.parameters[i];
+        const auto position = static_cast<int>(i);
+        DeviceParameterDto dto;
+        dto.index = info.paramIndex >= 0 ? info.paramIndex : position;
+        dto.stableId = info.stableId;
+        dto.name = info.name;
+        dto.unit = info.unit;
+        dto.minValue = info.minValue;
+        dto.maxValue = info.maxValue;
+        dto.defaultValue = info.defaultValue;
+        dto.currentValue = info.currentValue;
+        dto.normalizedValue = ParameterUtils::realToNormalized(info.currentValue, info);
+        dto.visible = contains(device.visibleParameters, position);
+        dto.miniMixer = contains(device.miniMixerParameters, position);
+        // Configure Parameters offers the AI opt-in only for external plugins;
+        // internal devices accept agent writes on every parameter.
+        dto.aiAgentEnabled = device.format == PluginFormat::Internal ||
+                             contains(device.aiSoundDesignerParameters, position);
+        dtos.push_back(std::move(dto));
+    }
+    return dtos;
 }
 
 SelectionDto makeSelectionDto(MagdaApi& api) {
--- a/magda/daw/api/remote_api_projection.cpp
+++ b/magda/daw/api/remote_api_projection.cpp
@@ -358,9 +358,14 @@ std::vector<DeviceParameterDto> makeDeviceParameterDtos(const DeviceInfo& device
         dto.unit = info.unit;
         dto.minValue = info.minValue;
         dto.maxValue = info.maxValue;
-        dto.defaultValue = info.defaultValue;
-        dto.currentValue = info.currentValue;
-        dto.normalizedValue = ParameterUtils::realToNormalized(info.currentValue, info);
+        // currentValue/defaultValue are model values: for an external plugin
+        // whose saved config gave it a display range, they stay in TE's native
+        // domain while min/max describe real units. Project through the
+        // normalized domain so the wire always carries display units.
+        const auto normalized = ParameterUtils::modelToNormalizedValue({info.currentValue}, info);
+        dto.defaultValue = ParameterUtils::modelToRealValue({info.defaultValue}, info);
+        dto.currentValue = ParameterUtils::normalizedToReal(normalized.value, info);
+        dto.normalizedValue = normalized.value;
         dto.visible = contains(device.visibleParameters, position);
         dto.miniMixer = contains(device.miniMixerParameters, position);
         // Configure Parameters offers the AI opt-in only for external plugins;
--- a/tests/remote/test_remote_api_contract.cpp
+++ b/tests/remote/test_remote_api_contract.cpp
@@ -40,6 +40,9 @@ TEST_CASE("Remote API registry is versioned, discoverable, and unique", "[remote
     REQUIRE(registry.find("system.describe") != nullptr);
     REQUIRE(registry.find("project.get") != nullptr);
     REQUIRE(registry.find("devices.list") != nullptr);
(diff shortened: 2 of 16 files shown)


magda/daw/api/remote_api_projection.cpp
```diff
@@ -338,6 +339,39 @@ DeviceCatalogEntryDto makeDeviceCatalogEntryDto(const DeviceCatalogEntry& entry)
...
+        DeviceParameterDto dto;
+        dto.index = info.paramIndex >= 0 ? info.paramIndex : position;
+        dto.stableId = info.stableId;
+        dto.name = info.name;
+        dto.unit = info.unit;
+        dto.minValue = info.minValue;
+        dto.maxValue = info.maxValue;
+        dto.defaultValue = info.defaultValue;
+        dto.currentValue = info.currentValue;
+        dto.normalizedValue = ParameterUtils::realToNormalized(info.currentValue, info);
```
lucaromagnoli: [P2] Preserve the model/display conversion here. For a configured external parameter, `currentValue` can remain in the host/model normalized domain while `minValue` and `maxValue` describe real display units. Calling `realToNormalized` directly then normalizes the model value a second time, and returning the raw model value as `currentValue` violates this operation real-unit contract. Please derive the normalized value with `modelToNormalizedValue`, project the current value with `normalizedToReal`, and convert real-unit set requests back with `normalizedToModelValue` before writing the model. The native-engine path preserves this distinction, so the TE cutover alone does not remove the issue.

tests/remote/test_remote_api_contract.cpp
```diff
@@ -489,6 +492,56 @@ TEST_CASE("devices.catalog lists addable devices without their file locations",
                                              "description", "format", "type", "instrument"});
 }
 
+TEST_CASE("devices.listParameters projects real units and customization flags",
+          "[remote-api][contract]") {
+    DeviceInfo device;
+    device.format = PluginFormat::VST3;
+    device.parameters.emplace_back(0, "Cutoff", "Hz", 20.0f, 20000.0f, 800.0f);
+    device.parameters.emplace_back(1, "Resonance", "%", 0.0f, 100.0f, 10.0f);
```
lucaromagnoli, line 500: This fixture masks the configured-external-parameter case: with a 0..1 backing range and no display provider, `ParameterUtils` classifies the value as display-mapped, so storing `800` happens to agree with the asserted display units. Please add a realistic external override case with a normalized model value and a different display range (for example model `0.75`, display range `-24..24`) and assert that discovery reports current `12` plus normalized `0.75`, and that setting real `12` stores the corresponding model value. That regression test should cover both halves of the conversion.

magda/daw/api/remote_api_projection.cpp
lucaromagnoli: Fixed in 872446ac. Added `ParameterUtils::modelToRealValue` / `realToModelValue` and applied them at every surface promising real units: the projection derives `normalizedValue` via `modelToNormalizedValue` and projects `currentValue` (and `defaultValue`, which comes from TE's `getDefaultValue()` and has the same domain) with `normalizedToReal`; `DeviceApiLive::setDeviceParameter` converts real-unit writes back with `realToModelValue` before writing the model, and `getDeviceParameters` projects on the way out. `MockDeviceApi` mirrors both so tests exercise the same contract.

tests/remote/test_remote_api_contract.cpp
lucaromagnoli: Added in 872446ac: `configured external parameters project model values into display units` (model 0.75, display -24..24, `displayText` set so the value classifies as TE-native → discovery reports current 12 / normalized 0.75), and `devices.setParameter converts display units to the model domain` (writing real 12 stores model 0.75, echo reads back in display units). Reverting the projection conversion makes the first test fail.

Issue #2270, opened 2026-08-31
The device op is location-transparent for rendering and not for anything else


lucaromagnoli:
Part of #1893, and a prerequisite for #1899.

## The requirement it is against

#1893 asks for one thing beyond parity: "The device op contract is location-transparent from day one: in-process and out-of-process are two implementations of the same op, so sandboxing later is an add-on, not a rework."

The realtime half honours that. `magda::engine::EngineDevice` says nothing about where a device lives, and `EngineExternalDevice` is one implementation of it.

The control half does not. `EngineExternalDevice::instance()` hands out a `juce::AudioPluginInstance&`, and everything that is not rendering reaches the plugin through it: the editor window, and now the state capture. A plugin in another process cannot produce one, cannot lend its callback lock, and cannot be suspended by a caller holding a reference to it.

## What the split cost already

Two races found in review on #2268, one after the other, and both of the same shape: an operation reached the plugin from the control side while the audio side was reaching it too, and no single object owned the rule that says those must not overlap.

- The capture suspended the plugin, and the adapter ran `processBlock()` without holding the callback lock or reading the suspended flag, so nothing waited and nothing stopped.
- The gate then sat around the `processBlock()` call, and `writeParameters()` runs before it, so parameter writes still went past while a capture was reading the values those writes were changing.

Both are fixed, and the second only existed because the policy is spread across two APIs that do not know about each other. Editor windows, program changes and preset loading are the same shape and are not covered.

## What to build

A control-plane endpoint owned by whatever owns the instance, keyed the way the realtime side is keyed:

- Processing, parameter writes, suspension and state capture all behind the object that owns the plugin, so the serialisation rule has one owner rather than a convention.
- Capture addressed by `engine::DeviceKey` rather than by a reference to an instance, which is the identity the plan and `PluginAssignments` already use. Something of the shape `captureState(DeviceKey, callback)`.
- Fallible and asynchronous, so an out-of-process implementation can have IPC, a timeout and a device that went away mid-call.
- `EngineDevice` stays the realtime endpoint and says nothing about any of it.
- `EngineExternalDevice::instance()` goes from the save path, and the editor-window path moves to the same endpoint.

[...]

PR #2283, opened 2026-09-01
A device answers for its plugin, and hands out no pointer to it (#2270)

lucaromagnoli:
Part of #1893, and the prerequisite that makes #1899 an add-on rather than a rewrite.

#1893 asks for one thing beyond parity: the device op contract is location-transparent from day one, so running a plugin in another process later is not a rework. The realtime half honoured it already — `magda::engine::EngineDevice` says nothing about where a device lives. The control half did not: everything that was not rendering reached the plugin through `EngineExternalDevice::instance()`, and a `juce::AudioPluginInstance&` is the one thing a plugin in another process cannot hand over.

It is also how the serialisation rule lost its owner. The two races found in the #2268 review were the same shape twice — an operation reaching the plugin from the control side while the audio side was reaching it too, with the policy spread across two APIs that do not know about each other.

## The device answers, and hands out no pointer

`EngineExternalDevice::captureState()` reads what the plugin holds; `instance()` is gone. The read itself is unchanged and still stated once in `ExternalPluginState.hpp` — what changes is that only the object holding the instance can perform it, and that object is the one already honouring the suspension inside `process()`. There is nothing left for a caller to remember.

## The endpoint

`DeviceControl.hpp`:

```cpp
class DeviceControlPlane {
    using CaptureCallback = std::function<void(CaptureOutcome)>;
    virtual void captureState(magda::engine::DeviceKey key, CaptureCallback completed) = 0;
};
```

Keyed by `DeviceKey`, which is the identity the plan and `PluginAssignments` already use, rather than by a reference to a live object. Asynchronous and fallible, because a remote implementation has IPC, a timeout and a device that can go away mid-call.

`LocalDeviceControlPlane` answers before it returns — the plugin is right here — and answers through the callback all the same: a caller that only worked when the answer arrived synchronously would only work in this process. It owns nothing; which device sits at a key is the runtime's business, so the lookup is handed in, the same arrangement the load path has with `CurrentDeviceLookup`.

A key with nothing bound is a **failure with a reason**, not an empty snapshot. Those are different findings — a slot whose plugin has not arrived, against a plugin that answered with nothing — and a host told the second when the first happened writes that absence into the project and loses the patch the slot is about to load.

## The commit

`commitCapturedState()` writes a snapshot onto a device only while the assignment it was read from still holds — the same boundary `completeExternalPluginLoad()` applies to the other direction, and for the same reason: between asking and answering, a slot can have been given a different plugin, or emptied, or the whole runtime can have gone.

The token both directions carry is therefore no longer called `LoadRequest`. It is an **`AssignmentRequest`**: one type, one check, rather than a convention each call site reimplements.

## What is deliberately not here

The editor window is the other caller that reaches for an instance. It does not exist on this path yet — MAGDA's plugin windows are still the fork's — and the header says where it goes when it arrives, which is behind this endpoint rather than a second accessor beside it.

## Tests

Six cases in `test_engine_external_device.cpp`: a capture through the plane answers with what the plugin holds; a key with nothing bound is a named failure; a plugin that will not describe itself is another; a snapshot is written onto the assignment it was read from, is refused when that assignment has been replaced, and is refused when the runtime that read it is gone. The existing save case now reads through the device rather than through its instance.

## Verification

Full Catch2 suite passes (740,373 assertions), the whole JUCE suite passes, and the app builds.

### DIFF
--- a/magda/daw/audio/plugins/engine/ControlExecutor.cpp
+++ b/magda/daw/audio/plugins/engine/ControlExecutor.cpp
@@ -1,0 +1,91 @@
+#include "ControlExecutor.hpp"
+
+#include <utility>
+
+namespace magda::daw::audio::engine_adapter {
+
+void MessageThreadControlExecutor::run(Work work) {
+    if (!work)
+        return;
+
+    if (isCurrent()) {
+        work();
+        return;
+    }
+
+    juce::MessageManager::callAsync(std::move(work));
+}
+
+bool MessageThreadControlExecutor::isCurrent() const {
+    return juce::MessageManager::existsAndIsCurrentThread();
+}
+
+SerialControlThread::SerialControlThread() : thread_([this] { loop(); }) {}
+
+SerialControlThread::~SerialControlThread() {
+    {
+        const std::lock_guard<std::mutex> held(lock_);
+        stopping_ = true;
+
+        // Dropped rather than drained. Work that has not started is work whose
+        // answer nobody is left to receive: this executor is destroyed before
+        // the devices its work would reach, and running it now would be
+        // reaching for plugins that are about to go.
+        queued_.clear();
+    }
+
+    wake_.notify_all();
+
+    if (thread_.joinable())
+        thread_.join();
+}
+
+void SerialControlThread::run(Work work) {
+    if (!work)
+        return;
+
+    // Already here: run it now rather than queue it behind ourselves, which is
+    // what an operation asked for from inside another one would otherwise wait
+    // forever for.
+    if (isCurrent()) {
+        work();
+        return;
+    }
+
+    {
+        const std::lock_guard<std::mutex> held(lock_);
+        if (stopping_)
+            return;
+
+        queued_.push_back(std::move(work));
+    }
+
+    wake_.notify_one();
+}
+
+bool SerialControlThread::isCurrent() const {
+    return std::this_thread::get_id() == thread_.get_id();
+}
+
+void SerialControlThread::loop() {
+    for (;;) {
+        Work work;
+
+        {
+            std::unique_lock<std::mutex> held(lock_);
+            wake_.wait(held, [this] { return stopping_ || !queued_.empty(); });
+
+            if (stopping_)
+                return;
+
+            work = std::move(queued_.front());
+            queued_.pop_front();
+        }
+
+        // Outside the lock, because the work is what takes the time and because
+        // it is entitled to queue more of itself.
+        work();
+    }
+}
+
+}  // namespace magda::daw::audio::engine_adapter
--- a/magda/daw/audio/plugins/engine/ControlExecutor.cpp
+++ b/magda/daw/audio/plugins/engine/ControlExecutor.cpp
@@ -1,9 +1,18 @@
 #include "ControlExecutor.hpp"
 
 #include <utility>
-#include <vector>
 
 namespace magda::daw::audio::engine_adapter {
+
+MessageThreadControlExecutor::MessageThreadControlExecutor()
+    : posted_(std::make_shared<Posted>()) {}
+
+MessageThreadControlExecutor::~MessageThreadControlExecutor() {
+    // Everything still in flight becomes a cancellation. It arrives on the
+    // message thread like every other answer, which is what keeps a caller from
+    // having to ask which thread it is on before it can act on one.
+    posted_->cancelled = true;
+}
 
 bool MessageThreadControlExecutor::run(Work work) {
     if (!work)
@@ -12,8 +21,13 @@ bool MessageThreadControlExecutor::run(Work work) {
     // Posted even from the message thread itself. Running it here would let a
     // nested operation into a plugin the outer one has open, and would put it
     // ahead of everything already waiting (ControlExecutor.hpp).
-    return juce::MessageManager::callAsync(
-        [work = std::move(work)]() mutable { work(ExecutionState::Ran); });
+    //
+    // The flag travels with the post rather than being read off this object: by
+    // the time it arrives, this executor may be gone, and what the call needs
+    // to know is whether it still means anything rather than who it came from.
+    return juce::MessageManager::callAsync([work = std::move(work), posted = posted_]() mutable {
+        work(posted->cancelled ? ExecutionState::Cancelled : ExecutionState::Ran);
+    });
 }
 
 bool MessageThreadControlExecutor::isCurrent() const {
@@ -27,58 +41,56 @@ SerialControlThread::SerialControlThread() : shared_(std::make_shared<Shared>())
         for (;;) {
             Work work;
 
+            auto state = ExecutionState::Ran;
+
             {
                 std::unique_lock<std::mutex> held(shared->lock);
                 shared->wake.wait(
                     held, [&shared] { return shared->stopping || !shared->queued.empty(); });
 
-                // Stopping wins over what is left: whatever is still queued has
-                // already been cancelled by the destructor, and running it now
-                // would be reaching for devices that are going away.
-                if (shared->stopping)
-                    return;
+                if (shared->stopping) {
+                    // Drained rather than abandoned, and drained here rather
+                    // than by whoever is destroying the executor: an answer owed
+                    // on this thread is owed on this thread whichever of the two
+                    // answers it turns out to be.
+                    if (shared->queued.empty())
+                        return;
+
+                    state = ExecutionState::Cancelled;
+                }
 
                 work = std::move(shared->queued.front());
                 shared->queued.pop_front();
             }
 
             // Outside the lock, because the work is what takes the time and
-            // because it is entitled to queue more of itself.
-            work(ExecutionState::Ran);
+            // because it is entitled to queue more of itself -- which run()
+            // refuses once stopping, so a drain cannot feed itself.
+            work(state);
         }
     });
 }
 
 SerialControlThread::~SerialControlThread() {
-    std::deque<Work> abandoned;
-
     {
         const std::lock_guard<std::mutex> held(shared_->lock);
         shared_->stopping = true;
-        abandoned.swap(shared_->queued);
     }
 
     shared_->wake.notify_all();
-
-    // Answered rather than dropped. Every one of these is an operation somebody
-    // is still waiting on, and an executor that let them go would leave a
-    // promise of exactly one answer unkept by whoever made it.
-    //
-    // Before the wait below, so that a caller destroying this executor knows
-    // that when the destructor returns, every operation it accepted has been
-    // accounted for.
-    for (auto& work : abandoned)
-        work(ExecutionState::Cancelled);
 
     if (thread_.get_id() == std::this_thread::get_id()) {
         // Destroyed by its own work, which is what a completion that closes the
         // project it belongs to looks like from here. A thread cannot wait for
         // itself, so it is let go instead: the loop's own copy of the shared
-        // state keeps it alive, and the loop stops the moment this work returns.
+        // state keeps it alive, and it drains what is left as cancellations and
+        // stops the moment this work returns.
         thread_.detach();
         return;
     }
 
+    // Waited for, so that a destroyed executor has accounted for everything it
+    // accepted: the worker answers what is left before it stops.
     if (thread_.joinable())
         thread_.join();
 }
--- a/magda/daw/audio/plugins/engine/ControlExecutor.cpp
+++ b/magda/daw/audio/plugins/engine/ControlExecutor.cpp
@@ -1,91 +1,108 @@
 #include "ControlExecutor.hpp"
 
 #include <utility>
+#include <vector>
 
 namespace magda::daw::audio::engine_adapter {
 
-void MessageThreadControlExecutor::run(Work work) {
+bool MessageThreadControlExecutor::run(Work work) {
     if (!work)
-        return;
+        return false;
 
-    if (isCurrent()) {
-        work();
-        return;
-    }
-
-    juce::MessageManager::callAsync(std::move(work));
+    // Posted even from the message thread itself. Running it here would let a
+    // nested operation into a plugin the outer one has open, and would put it
+    // ahead of everything already waiting (ControlExecutor.hpp).
+    return juce::MessageManager::callAsync(
+        [work = std::move(work)]() mutable { work(ExecutionState::Ran); });
 }
 
 bool MessageThreadControlExecutor::isCurrent() const {
     return juce::MessageManager::existsAndIsCurrentThread();
 }
 
-SerialControlThread::SerialControlThread() : thread_([this] { loop(); }) {}
+SerialControlThread::SerialControlThread() : shared_(std::make_shared<Shared>()) {
+    // The worker holds the state as well, so that this object can be destroyed
+    // by work running on the worker without taking the worker's world with it.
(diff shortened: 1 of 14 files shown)


magda/daw/audio/plugins/engine/DeviceControl.hpp
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * One of the two, never both and never neither: a snapshot the caller may write
+ * down, or a reason it has nothing to write. A failure is a string rather than
+ * an enumeration because every one of them is something to tell a person and
+ * nothing to branch on -- there is no device here, this plugin would not
+ * describe itself, the answer did not come back in time.
+ */
+struct CaptureOutcome {
+    std::optional<magda::ExternalPluginSnapshot> snapshot;
+    juce::String failure;
```
lucaromagnoli: **P2 — Make the outcome invariant representable by the type**

The contract says exactly one of snapshot or failure, but this public aggregate permits both and neither, and `ok()` silently treats the both-present case as success. That is fragile for the IPC boundary this API is explicitly preparing: every implementation and decoder must recreate an invariant the type does not enforce. Please use a discriminated result (`variant`/expected-style type), or private construction with success/failure factories, so malformed outcomes cannot cross the endpoint.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.cpp
```diff
@@ -0,0 +1,70 @@
...
+bool commitCapturedState(const AssignmentRequest& request,
+                         const magda::ExternalPluginSnapshot& snapshot, magda::DeviceInfo& device) {
+    // The same question the other direction asks of a plugin that has finished
+    // loading, and the same answer: a snapshot is only worth writing down onto
+    // the assignment it was read from. A key that is live again under a
+    // different assignment is not that, and neither is a runtime that has gone.
+    if (!request.isStillWanted())
+        return false;
+
+    magda::applyCapturedPluginState(device, snapshot);
```
lucaromagnoli: **P1 — Bind the commit target to the validated key**

`isStillWanted()` proves that `request.key` still owns this assignment handle, but it establishes no relationship with the arbitrary `device` reference passed here. A caller can supply another slot model and this will apply the snapshot while returning true, so the helper does not actually enforce the ownership boundary described above. Please mirror `completeExternalPluginLoad()`: after validating the token, resolve the current mutable model by `request.key` inside this boundary and apply only to that result.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (device == nullptr) {
+        // Named rather than reported as an empty state. A key with nothing
+        // bound to it is a slot whose plugin has not arrived or has gone, and a
+        // caller told "no state" would write that absence into the project.
+        completed(
+            {.snapshot = std::nullopt, .failure = "no plugin is bound for " + describeKey(key)});
+        return;
+    }
+
+    auto snapshot = device->captureState();
```
lucaromagnoli: **P1 — Uphold the message-thread contract here**

This runs the lookup, plugin state read, and completion on whichever thread invoked `captureState()`, while the interface promises completion on the message thread and `EngineExternalDevice::captureState()` itself is documented as message-thread-only. A save dispatched from a worker would therefore touch third-party plugin state on the wrong thread and observe different callback semantics from a remote implementation. Please either marshal this operation to the message thread with lifetime-safe ownership, or make message-thread invocation an explicit interface precondition and enforce it at this boundary.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.cpp
```diff
@@ -0,0 +1,120 @@
...
+    //
+    // Asked as "is there a message thread, and is this it" rather than as "is
+    // this the message thread", because those differ where the engine is
+    // supposed to work: a headless render has no message manager at all, so
+    // there is no thread to be off, and a check that refused there would refuse
+    // every offline host on the grounds that it is not an application.
+    if (auto* messages = juce::MessageManager::getInstanceWithoutCreating();
+        messages != nullptr && !messages->isThisTheMessageThread()) {
+        JUCE_ASSERT_MESSAGE_THREAD
+        completed(CaptureOutcome::failed("a capture was asked for off the message thread"));
```
lucaromagnoli: **P1 — Keep the rejection callback on the promised thread**

This branch detects a worker-thread call but immediately invokes `completed` on that same worker. The interface still promises that completion runs on the message thread, and the new prose explicitly defines rejection behavior for a violated precondition, so this is observable contract breakage in release builds regardless of the assertion. Please dispatch the failure callback onto the message thread, or redefine a precondition violation as having no callback/undefined behavior and adjust the contract and test accordingly.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.cpp
```diff
@@ -0,0 +1,120 @@
...
+    // asked for from the wrong thread is a caller to fix and not a call to
+    // rescue -- and because a remote implementation would answer the same way
+    // rather than quietly acquiring a thread to hop from.
+    //
+    // Asked as "is there a message thread, and is this it" rather than as "is
+    // this the message thread", because those differ where the engine is
+    // supposed to work: a headless render has no message manager at all, so
+    // there is no thread to be off, and a check that refused there would refuse
+    // every offline host on the grounds that it is not an application.
+    if (auto* messages = juce::MessageManager::getInstanceWithoutCreating();
```
lucaromagnoli: **P1 — Headless still needs a single control owner**

With no `MessageManager`, this condition admits every calling thread. Two headless callers can therefore enter `captureExternalPluginState()` concurrently: both set suspension, both read plugin state, and the first one to finish clears suspension while the other is still reading, allowing `process()` back into the plugin mid-capture. That reopens the serialization race this endpoint is meant to close. Please enforce one designated control thread in headless mode, or serialize the entire capture transaction with a device-owned guard; absence of a message manager cannot mean unrestricted thread access.

— Codex

magda/daw/audio/plugins/engine/ControlExecutor.cpp
```diff
@@ -0,0 +1,91 @@
...
+        // answer nobody is left to receive: this executor is destroyed before
+        // the devices its work would reach, and running it now would be
+        // reaching for plugins that are about to go.
+        queued_.clear();
+    }
+
+    wake_.notify_all();
+
+    if (thread_.joinable())
+        thread_.join();
```
lucaromagnoli, line 95: **P1 — Avoid joining the executor from its own callback**

A plane may be the last owner of its `shared_ptr<ControlExecutor>`. If a completion running on `SerialControlThread` closes that plane/runtime, releasing the last reference runs this destructor on `thread_` itself and `join()` throws/terminates for a self-join. Callback-driven teardown is a normal control-plane outcome, so the worker state/lifetime must be arranged so final release on the executor thread is safe, or the ownership contract must enforce an external owner that outlives every plane and callback.

— Codex

magda/daw/audio/plugins/engine/ControlExecutor.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * @brief The executor for a host that has no message thread.
+ *
+ * An offline render, a command-line tool, a test binary. It owns a thread and
+ * runs what it is given on that one, in order, which is the same guarantee the
+ * message thread gives an application and the reason a headless host is not
+ * left deciding per call site.
+ *
+ * Work queued but not yet run is dropped when this is destroyed. That is the
```
lucaromagnoli: **P1 — Reconcile executor shutdown with exactly-once completion**

Dropping queued work means every `captureState()` represented by that work loses the callback that `DeviceControlPlane` promises exactly once. The message-thread implementation has the same hole because it ignores the `bool` returned by `MessageManager::callAsync()` when posting fails during shutdown. Please make submission/cancellation part of the executor contract—either complete accepted operations with a cancellation failure on the control context, or explicitly expose cancellation/no-callback semantics from `DeviceControlPlane`. The current two contracts cannot both hold.

— Codex

magda/daw/audio/plugins/engine/ControlExecutor.hpp
```diff
@@ -0,0 +1,131 @@
...
+    ControlExecutor& operator=(const ControlExecutor&) = delete;
+    ControlExecutor(ControlExecutor&&) = delete;
+    ControlExecutor& operator=(ControlExecutor&&) = delete;
+
+    using Work = std::function<void()>;
+
+    /**
+     * @brief Run @p work on this executor.
+     *
+     * Immediately when the caller is already on it, which is what keeps an
```
lucaromagnoli: **P1 — Do not make serial control execution re-entrant**

Running nested work inline means this is not logically “one piece of work at a time.” If control operation A has suspended the plugin and asks for operation B, B can run and resume the plugin before A finishes, letting audio back into A mid-transaction—the same suspension bug this executor is meant to eliminate. It also violates FIFO when B was already queued and nested C jumps ahead of it. `run()` is fire-and-forget, so nested work can be queued safely; if synchronous composition is required, it needs a separate transaction-aware API rather than re-entering the executor.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.cpp
```diff
@@ -0,0 +1,123 @@
...
+    // Everything past here runs on the executor: the lookup, the plugin's own
+    // state read, and the answer. Nothing checks which thread asked, because
+    // asking is allowed from any of them -- what is not allowed is two of these
+    // being inside one plugin at once, and one serial executor is what makes
+    // that impossible for every control operation rather than for this one.
+    //
+    // The lookup is copied rather than reached for through `this`: work outlives
+    // the call that queued it, and a plane destroyed in between would otherwise
+    // be read from a thread that is still running.
+    executor()->run([devices = devices_, key, completed = std::move(completed)]() mutable {
```
lucaromagnoli: **P1 — Give queued capture work a lifetime-safe target**

Copying the `std::function` only copies whatever it captured; a normal runtime lookup such as `[this] { ... }` still contains a raw pointer. Because this path is now genuinely asynchronous, the plane/project can be destroyed after queuing and the message/serial executor will later invoke both a dangling lookup and potentially a dangling completion. Please carry a weak/shared runtime state or assignment lifetime into the queued operation and validate it on the executor before lookup, with an explicit cancellation policy. This needs a teardown-before-dispatch test like the asynchronous load path already has.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.hpp
```diff
@@ -0,0 +1,279 @@
...
+     * Locked on the executor for the length of the operation, which is more
+     * than a check: it is what stops the runtime being destroyed halfway
+     * through a capture rather than only before one. An expired lifeline is a
+     * failure with a reason, and the lookup is never called.
+     *
+     * The same shape the load path uses for the same problem, where an
+     * assignment is held weakly and a completion that finds it gone is refused
+     * (PluginAssignments.hpp).
+     */
+    LocalDeviceControlPlane(std::shared_ptr<ControlExecutor> executor, std::weak_ptr<void> lifeline,
```
lucaromagnoli: **P1 — Bind the lifetime lease to the lookup target**

This repeats the original commit-boundary problem in another form: the weak `lifeline` and `DeviceLookup` are independent arguments, so locking one proves nothing about the object captured by the other. The tests even pass a standalone `shared_ptr<int>` while the returned device is owned elsewhere; keeping that token alive would not prevent the device/lookup owner from being destroyed, and the queued call would still dereference stale state. Please make lookup return a lease that pairs the strong owner with the device pointer, or make the plane hold a typed weak registry and perform lookup through the locked registry, so the safety relationship is enforced by the type rather than by caller convention.

— Codex

magda/daw/audio/plugins/engine/ControlExecutor.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * @brief The executor for a host that has a message thread.
+ *
+ * Which is every host with a window in it. Control work belongs on the message
+ * thread there for reasons older than this file: it is the thread a plugin's
+ * editor lives on, the one its own host callbacks expect, and the one the model
+ * a callback writes into is edited from.
+ *
+ * The one guarantee it cannot make is the one JUCE does not: a message posted
```
lucaromagnoli: **P2 — Do not exempt one implementation from the base guarantee**

`ControlExecutor::run()` promises that accepted work is called exactly once, but this implementation explicitly admits that accepted work can disappear when the message loop is torn down. `DeviceControlPlane` then relies on the stronger base promise for its own exactly-once contract. Process teardown may justify cancellation, but it cannot silently violate the polymorphic contract: track/cancel posted work, scope the guarantee to a live dispatch loop in the base API, or model message-loop shutdown as a first-class cancellation state.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.hpp
```diff
@@ -0,0 +1,279 @@
...
+     *
+     * Asked from any thread, answered on the executor, always later than this
+     * returns. A caller writing a project's model in that callback is therefore
+     * on the thread it is entitled to write one from, whichever thread asked
+     * and whichever process answered.
+     *
+     * @p completed is called exactly once. Three things can happen to it:
+     *
+     * - the capture ran, and it carries a snapshot or a reason there is none;
+     * - the plane's executor went away before its turn, and it carries that as
```
lucaromagnoli: **P1 — Keep the capture callback executor-affine**

Cancellation now invokes `completed` on the executor destructor thread, and refusal invokes it on the arbitrary calling thread. That breaks the central abstraction above: a caller cannot safely update the model/UI from `CaptureCallback`, and `CaptureOutcome` does not even distinguish an off-executor cancellation from an ordinary failure. Every consumer would need `isCurrent()` checks and its own marshaling—the patching this control plane is meant to prevent. Please keep this callback exclusively executor-affine; report submission failure synchronously via the return value/a request handle, and define teardown cancellation separately rather than delivering the same callback on a different thread.

— Codex

magda/daw/audio/plugins/engine/DeviceControl.hpp
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * @brief The device at @p key, or null.
+     *
+     * Null for a key nothing is bound to, which includes one whose plugin has
+     * not finished loading. Called on the control executor, with the registry
+     * held alive for the length of the operation, so what it returns is safe to
+     * use until that operation ends.
+     */
+    virtual EngineExternalDevice* find(magda::engine::DeviceKey key) const = 0;
```
lucaromagnoli: **P1 — Make the lookup return a device lease, not a raw pointer**

Locking the registry only proves that the registry object is alive; this interface does not require it to own the returned device or otherwise tie the device lifetime to that lock. In fact, `OneDeviceRegistry` in this commit stores a raw pointer to a device owned independently by `result.device`, demonstrating that the claimed relationship is not enforced. A device can therefore be removed or destroyed while the registry remains alive, leaving `captureState()` with a dangling pointer after `find()`.

Please make the lifetime part of the result (for example, return a `shared_ptr` or explicit lease), or make the registry a concrete owner whose access and removal protocol guarantees the device remains alive through the operation. The boundary should encode the lease it relies on instead of documenting it.

— Codex

PR #2289, opened 2026-09-01
Report scale and choices in devices.listParameters (#2286)

### LINKED ISSUE #2286: Report scale and choices in devices.listParameters
`DeviceParameterDto` (#2281) omits `ParameterInfo::scale` and `choices`, so a discrete parameter reaches an MCP client as nothing more than a number in a range. Live example from testing: Surge XT's "A Filter 1 Type" reports `0..1, no unit` — a client has to guess that 0.5 lands mid-way through the filter-model list instead of asking for a model by name. The in-app sound-design agent's snapshot already includes scale and choices, so the external surface is strictly poorer than the internal one here.

Additive DTO change: `scale` (the existing wire strings: linear/logarithmic/exponential/discrete/boolean/fader_db) and `choices` (array of labels, empty for continuous parameters), plus schema, projection (`choiceIndexForModelValue` handles the TE-native spreading), and round-trip decode. Complements #2285: overrides written there should read back here.

lucaromagnoli:
Closes #2286.

Adds `scale` (linear / logarithmic / exponential / discrete / boolean / fader_db) and `choices` (labels, empty for continuous parameters) to `DeviceParameterDto`, its schema, JSON codecs, and the projection. The scale↔string mapping moves from `PluginParameterConfigStore`'s anonymous namespace to its public API (`scaleToString` / `scaleFromString`) so the config XML and the remote wire share one vocabulary owner.

Live motivation from the #2274 test session: Surge XT's "A Filter 1 Type" reported `0..1, no unit`, forcing a client to set 0.5 blind instead of asking for a filter model by name.

Contract test extended with a discrete parameter (choices + scale asserted, round-tripped).

### DIFF
--- a/magda/daw/api/remote_api.cpp
+++ b/magda/daw/api/remote_api.cpp
@@ -4,6 +4,7 @@
 #include <cmath>
 #include <limits>
 
+#include "../core/PluginParameterConfigStore.hpp"
 #include "remote_handlers.hpp"
 
 namespace magda::remote {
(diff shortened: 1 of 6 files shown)


lucaromagnoli:

PR #2290, opened 2026-09-01
Expose device add, remove, and move over MCP (#2284)

lucaromagnoli:
Closes #2284. Stacked on #2289.

- **`devices.add`** (write, edit): `{catalogId, trackId | parentPath, index?}` → `{id, devicePath}`. The catalogue id comes from `devices.catalog`; `trackId` targets the main FX chain, `parentPath` a rack chain at any depth. The returned `devicePath` is the canonical spelling (top-level id on the path root, chain-hosted as a device step) so the client can immediately address the new device.
- **`devices.remove`** (write, edit): `{devicePath}` → `{accepted}`.
- **`devices.move`** (write, edit): `{devicePath, toIndex}` → `{accepted}`.

All three are thin handlers over the pre-existing `DeviceApi::addDevice` / `removeDevice` / `moveDevice`, which were implemented (live + mock, undoable commands) but never registered. The DevicePanel agent surface allowlists them plus `devices.catalog` — previously it had no way to discover what could be added.

Motivation from the #2274 live session: deleting and recreating a Surge XT instance over MCP was impossible; the user had to do it in the UI.

Dispatch tests cover create-with-address, unknown catalogue id, move/remove success, and unresolvable-path refusal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2291, opened 2026-09-01
Accept both spellings of a top-level device in engine lookups (#2288)

### LINKED ISSUE #2288: Parameter writes report success when no processor consumes them
Found while tracing a false alarm during #2274 live testing (the writes did land; the trace surfaced this adjacent hazard).

`TrackManager::setDeviceParameterValue` updates the model and notifies; `AudioBridge::deviceParameterChanged` (AudioBridge.cpp:567) is the sole engine write, and it **returns silently** when `getDeviceProcessor(devicePath)` misses. The remote handler then re-reads the model and echoes the new value — so an MCP client (or any host-write caller) sees success while the plugin never heard the write.

Two real ways to hit it:

1. **Null processor forever**: `SyncedDevice::processor` stays null when an external plugin's async load fails (`PluginManagerSync.cpp:1313-1330` only creates the processor `if (loaded)`), but the map entry survives.
2. **Path-spelling asymmetry**: `TrackManager::getDeviceInChainByPath` accepts a top-level FX device as either `topLevelDeviceId=N` or `steps=[{device,N}]` (TrackManagerDevices.cpp:1342), and `PluginManager::getPlugin` has a lenient fallback — but `PluginManager::getDeviceProcessor` is a strict map lookup keyed only on the `topLevelDeviceId` spelling (PluginManager.cpp:231). A hand-constructed path with the `steps` spelling validates, mutates the model, echoes success, and is never heard by the engine.

Suggested fixes, independently useful:
- Make `getDeviceProcessor` resolve both spellings (or normalize `ChainNodePath` before keying/lookup).
- Have `deviceParameterChanged` log (or surface) the miss, distinguishing "no map entry" from "entry with null processor".
- Consider propagating a "consumed by engine" signal back so `devices.setParameter` can report `Conflict`/degraded status instead of unqualified success when the processor is missing.

lucaromagnoli:
Closes #2288. Stacked on #2290.

`TrackManager::getDeviceInChainByPath` accepts a top-level FX device as either `topLevelDeviceId=N` or `steps=[{device,N}]`, but `PluginManager::syncedDevices_` keys only the first spelling, and `getDeviceProcessor` was a strict map lookup. A host parameter write addressed with the step spelling therefore validated, mutated the model, and echoed success while `AudioBridge::deviceParameterChanged` silently dropped it.

- New `ChainNodePath::normalized()` collapses the step spelling onto the canonical `topLevelDevice` form (identity for canonical, nested, sectioned, and track-level paths — unit-tested).
- `PluginManager::findSyncedDevice` normalizes before lookup, so `getPlugin`, `getDeviceProcessor`, and every other consumer accept either spelling uniformly.
- The two silent-drop sites now log, distinguishing "no map entry" from "entry whose async plugin load failed and never produced a processor" — the second reachable case from the issue.

Not included: propagating a "consumed by engine" signal into the `devices.setParameter` response — left open on the issue as a possible follow-up, since it changes the operation's success semantics.

### DIFF
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -569,6 +569,10 @@ void AudioBridge::deviceParameterChanged(const ChainNodePath& devicePath, int pa
     // A single device parameter changed - sync only that parameter to processor
     auto* processor = getDeviceProcessor(devicePath);
     if (!processor) {
+        // The model updated and the caller was told ok, but nothing reached
+        // the engine. Say so rather than dropping the write silently (#2288).
+        DBG("AudioBridge: dropping parameter change for device " << devicePath.getDeviceId()
+                                                                 << " - no processor");
         return;
     }
 
--- a/magda/daw/audio/plugin_manager/PluginManager.cpp
+++ b/magda/daw/audio/plugin_manager/PluginManager.cpp
@@ -165,12 +165,14 @@ PluginManager::PluginManager(te::Engine& engine, te::Edit& edit, TrackController
 
 PluginManager::SyncedDeviceMap::iterator PluginManager::findSyncedDevice(
     const ChainNodePath& devicePath) {
-    return syncedDevices_.find(devicePath);
+    // Entries are keyed by the canonical spelling; the model layer accepts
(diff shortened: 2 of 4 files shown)


lucaromagnoli:

PR #2292, opened 2026-09-01
Retry rate-limited requests in the MCP bridge with bounded backoff (#2287)

### LINKED ISSUE #2287: Seventh sequential write on one MCP connection fails
Reproduced twice during live testing of #2281 against a running MAGDA (macOS debug build, `magda-mcp` stdio bridge, single connection, client `claude-code` with edit scope):

- A burst of sequential `tools/call` → `devices.setParameter` requests (one at a time, waiting for each response) fails on **exactly the seventh call**, both times.
- The response arrives with the request's id but without a usable result payload (the test harness only captured that `result.structuredContent` was absent; raw payload not logged — first thing to capture when reproducing).
- The same write immediately succeeds on a fresh bridge connection.
- Calls 1–6 succeed and are audible/persisted; the failure is deterministic by position, not by parameter.

Suspects: per-connection request limits applied via `applyRequestLimits` / the request-limit machinery in the remote service or MCP server, or something in the bridge's HTTP session handling. If it is an intentional rate limit, the refusal should be a structured, self-describing error (like `permission_denied` is) so a client can back off and retry rather than seeing a malformed-looking response.

Repro sketch: initialize one stdio bridge session, issue 7+ sequential `devices.setParameter` calls, log every raw response line.

lucaromagnoli:
Closes #2287. Stacked on #2291.

Root cause (details on the issue): the MCP server's token bucket has burst capacity `maxConcurrentRequests` (8) with 50/s refill, so a fresh session's `initialize` + `initialized` + 6 tool calls exhaust the burst and the 7th tool call gets HTTP 429 — a structured refusal, but one most stdio MCP hosts surface as a terminal error.

Fix at the bridge, per the issue's option 2: `magda-mcp` retries 429s with bounded exponential backoff (100/200/400/800 ms). Bursts from agent workflows are absorbed transparently; a genuinely saturated server still surfaces after ~1.5s of refusals. The server-side bound is unchanged.

Regression test runs the real bridge binary against a server configured with a burst of 2 and 20/s refill, and requires all 8 rapid sequential calls to answer without error. Verified negatively: disabling the retry loop fails the test.

### DIFF
--- a/magda/mcp_bridge/main.cpp
+++ b/magda/mcp_bridge/main.cpp
@@ -32,6 +32,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <chrono>
 #include <cstdlib>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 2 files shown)


lucaromagnoli:

Issue #2305, opened 2026-09-01
Session launcher slice 6: handle lifetime on the plan-swap epoch


lucaromagnoli:
Slice 6 of #1894. Depends on slice 2. Worth doing early despite the number, because it is expensive to retrofit.

## The requirement

From #1894: "Handle lifetime via the plan-swap epoch reclamation; no hand-rolled deferred-release protocol."

The engine already has the answer. Plan swap is the reclamation point for audio-thread object lifetime, epoch based, and that was settled in the engine core (#2012). A launch handle is an audio-thread object with a message-thread writer, which is exactly the category epoch retirement was built for.

## What would go wrong without this slice

The fork's handle owns its own threading: an atomic `PlayState` and an `std::optional<NextState>` behind a `crill::spin_mutex`, with the message thread pushing and the audio thread peeking and clearing. That is a per-object protocol, and it works, and it is precisely the "hand-rolled deferred release" the epic rules out -- not because a spin mutex is wrong, but because every object that invents its own lifetime rule is a place the engine's one rule does not apply, and the reason the engine has one rule is that the fork had a dozen.

The retrofit cost is what makes this early. A handle whose lifetime is its own becomes a handle every caller holds a raw pointer to, and moving it onto the epoch afterwards means changing every one of those call sites plus the slot ops that bind them.

## Scope

- Handle allocation, ownership and retirement on the plan-swap epoch.
- What happens to a playing handle when the plan is recompiled underneath it, which is a track being added while a scene plays, and which must not restart the clip.
- The message-to-audio launch request path, decided here rather than inherited: the fork's spin mutex is one option and the engine's existing lanes are another.

## Verification

The differ property tests from #2077 extended over edit sequences that include launches, asserting that a structural edit during playback does not restart, drop or double a playing slot.

lucaromagnoli:
Narrowing this after slice 1 (#2307), because most of what it asked for turns out to be already built.

`RuntimeStateStore` does the ownership half generically: it owns runtime objects keyed by section-aware model identity, `realise()` binds them into each published plan, `releaseDeleted()` retires what the model no longer names, and objects survive a recompile by design -- "the same edit that recompiles the plan leaves the objects it names untouched". That is the same guarantee this issue asks for when it says a structural edit during playback must not restart a playing clip, and devices, clip sources, meters and value taps all already have it.

[...]

Issue #2304, opened 2026-09-01
Session launcher slice 5: follow actions, and an engine-initiated launch reaching the app


lucaromagnoli:
Slice 5 of #1894. Depends on slices 1 and 4.

## What it is

`ClipInfo` already carries the whole vocabulary: `followAction`, `followActionDelayBeats`, `followActionLoopCount` (`core/ClipInfo.hpp:880-882`). What is missing is the engine acting on them, which means the engine launching a clip that nobody asked it to launch.

## The hard half

#1894 states it as a requirement rather than a feature: engine-initiated launches must be "reflected back into app state without an adoption dance". Today user intent lives in `TrackInfo::activeSessionClipId` and the scheduler maintains it, because every launch originates on the message thread. A follow action originates on the audio thread, at a beat, and the model has to end up agreeing without the message thread having been the one to decide.

The failure mode to design against is the model and the engine disagreeing about what is playing, then the next user launch resolving the disagreement by stopping the wrong clip. That is the shape of bug this requirement exists to prevent, and it is the reason follow actions come after the reporting slice rather than before it: if the engine's state is already reported as data, the model updating from it is a subscription rather than an adoption.

## Scope

- The actions themselves, evaluated against the played range at the point the handle would loop or end.
- The delay and the loop count, which mean a follow action does not necessarily fire on the first pass.
- An engine-initiated launch reaching `activeSessionClipId` through the same path a user launch takes, or a stated reason why it cannot.
- What a follow action does when the clip it names has been deleted or moved since the project loaded.

## Verification

Offline, over a driven transport: a chain of follow actions produces a known sequence of played ranges, and the model's idea of what is playing matches the engine's at every step rather than only at the end.

lucaromagnoli:
Merged in #2375.

Issue #2302, opened 2026-09-01
Session launcher slice 3: per-track session and arranger switching with de-click


lucaromagnoli:
Slice 3 of #1894. Depends on slice 2.

## What it is

A track plays the arranger or the session, not both, and the switch happens on a launch. Switching sources mid-render is a discontinuity, and the fork ramps it: `ClipInfo::launchFadeSamples` defaults to 256 (`core/ClipInfo.hpp:838`) and is the ramp on the stopped-to-playing transition.

## Why this is small

`OpKind::Crossfade` already exists and already does the thing: "an edge as it was and as it is, ramped from one to the other" (`engine/plan/RenderPlan.hpp:117`). It was built for #2019, to ramp changed ops across a plan swap. A session/arranger switch is the same operation at a different trigger.

So this slice is mostly about binding an existing op to a new cause, and the work is in the cases where that is not enough:

- A switch that happens mid-block, at the split point slice 1 reports, rather than at a block boundary.
- A stop with nothing to switch to, where the ramp is to silence.
- Two switches inside one block, which a 256-sample ramp cannot both complete.
- Whether the ramp is per track or per source, which decides what a scene launch sounds like when eight tracks switch at once.

## What would be wrong

Reusing the plan-swap crossfade path wholesale, because a launch is not a plan swap. A launch is a data change inside a stable plan, and if it compiled a new plan every time a clip was triggered then a scene launch would recompile the graph eight times and the differ would be doing structural work on the audio thread's critical path. The ramp is shared; the mechanism that triggers it is not.

## Verification

Offline renders asserting no discontinuity at the switch, at the sample. A switch at a block boundary and a switch mid-block produce the same ramp, which is the assertion that catches a ramp anchored to the wrong clock.

Issue #2301, opened 2026-09-01
Session launcher slice 2: session sources in the plan, and a slot's voice across a swap


lucaromagnoli:
Slice 2 of #1894. Depends on slice 1 and on the clip snapshot (#2034).

## What is missing

`OpKind` (`engine/plan/RenderPlan.hpp:106`) has `ClipAudio` and `ClipMidi`, and `PlanCompiler` emits one of each per track from the arranger. There is no session concept anywhere in the plan: no slot, no scene, no second source competing for the same chain head. A session clip is invisible to the native engine.

## The shape

A session source is not a new kind of clip playback. It reads the same clip snapshot as the arranger source and it feeds the same `TrackAudioInput` and `TrackMidiInput` roles. What differs is what decides whether it produces anything in a given block, which is the handle from slice 1.

So the plan gains a session source op per track alongside the arranger one, keyed on the track the way `ClipAudio` already is, with `OpRole` gaining the session pair so the differ can tell the two apart at the same model location. That is the same problem `OpRole` was introduced for: one model ID, several ops, distinguished by structural role.

## What has to be decided in it

- **One op per track or one per slot.** A track plays at most one session clip at a time, which argues for one op holding whichever slot is live. Against that, a slot's voice has to survive being launched, stopped and relaunched while a different slot plays in between, and an op that changes which clip it reads is an op whose identity the differ cannot use. This is the decision the slice exists to make, and it is the one that determines whether a launch reuses a voice or builds one.
- **What the source does with a split block.** Slice 1 hands down two ranges. The source renders the playing sub-ranges and nothing else, and the fade at the boundary belongs to slice 3 rather than here.
- **Where the clip's own loop meets the handle's loop.** `ClipInfo` has a loop, and `setLooping` on the handle is a separate re-trigger of the play duration. Both exist in the fork and they are not the same thing. The distinction has to be stated here because a source that conflated them would loop correctly in the common case and wrongly whenever a clip's loop length and its launch duration differ.

## Verification

Plan goldens for a project with session clips, and offline renders against a driven handle. A session source fed a handle that never launches renders silence and costs nothing; one fed a handle launched mid-block starts at the beat.

PR #2307, opened 2026-09-01
A launch is a beat, not a block, and the clock grows a third face (#2300)

### LINKED ISSUE #2300: Session launcher slice 1: the launch handle, and what a block straddling a launch point means
Slice 1 of #1894. Depends on the engine core and on nothing else, which is why it is first.

## What the fork owns today

`SessionClipScheduler` says it plainly in its own header: "This is NOT a parallel scheduler -- TE owns playback state via LaunchHandle." Everything MAGDA has on the session side is a command layer. The state machine is `te::LaunchHandle` (`model/clips/tracktion_LaunchHandle.h`), and replacing it is the whole of this slice.

What it is:

- `PlayState` of `stopped` or `playing`, written by the audio thread and read by the message thread through an atomic.
- An optional `QueueState` of `playQueued` or `stopQueued` with an optional `MonotonicBeat` for when it fires, written by the message thread under a spin mutex and read and cleared by the audio thread.
- `advance(SyncRange)` returning a `SplitStatus`, called on the audio thread once per block.
- Played range reporting, both as an Edit `BeatRange` and a monotonic one, unlooped and monotonically increasing.
- `playSynced`, `setLooping(std::optional<BeatDuration>)` and `nudge(BeatDuration)`.

## What makes it a slice rather than a detail

`SplitStatus` is the reason. A block can straddle the beat a launch is quantized to, so advancing does not return one play state, it returns two ranges with possibly different states and their own start times:

```
struct SplitStatus {
    bool playing1 = false, playing2 = false;
    BeatRange range1, range2;
    bool isSplit = false;
    std::optional<BeatPosition> playStartTime1, playStartTime2;
};
```

A handle that answered "playing or not" per block would launch on a block boundary rather than on the beat, which at 512 samples and 120 bpm is up to 11 ms of slop on every clip in a scene, and the slop would differ per track. That is audible and it is exactly what quantized launch exists to prevent. Anything downstream that consumes the handle has to be able to act on two sub-ranges of one block, so the shape has to be right here before a source or a chain reads it.

## Scope

The handle and its state machine, as engine-owned model-level code with no plan, no executor and no audio in it:

- The two-state play status and the queued transition, resolved against monotonic beats.
- The split advance, including the case where a launch and a stop both land inside one block.
- Looping as a re-trigger of the play duration, which is what `setLooping` means, and which is not the same as the clip's own loop.
- `playSynced` against another handle, which is what makes a scene launch one event rather than N.
- Nudge.
- The played range, unlooped, in both domains.

The model already carries the vocabulary this reads: `LaunchMode`, `LaunchQuantize`, `FollowAction`, `followActionDelayBeats`, `followActionLoopCount` and `launchFadeSamples` are all on `ClipInfo` (`core/ClipInfo.hpp:878-882`). Nothing here needs new model fields.

## Not in this slice

Follow actions themselves (#1894's own slice), the de-click ramp, anything that reads a clip, and the threading contract for how a launch request gets from the message thread to the handle in the first place. The fork uses a spin mutex over a trivially copyable `NextState`; whether that survives is a decision for the lifetime slice, not this one.

## Verification

Entirely offline and entirely deterministic, which is unusual for this subsystem and worth spending. A handle is fed synthetic `SyncRange`s and asserted against:

- A launch quantized to a beat inside a block splits the block at that beat, not at its start or end.
- A launch and a stop inside one block produce the right two ranges and the second range's own start time.
- A launch quantized to a beat that is exactly a block boundary is not double counted.
- Looping re-triggers at the duration, and the played range keeps increasing across the wrap.
- `playSynced` against a handle mid-play starts at the synced position rather than at zero.
- The played range after a stop and a relaunch does not carry the first play's length.

`tracktion_LaunchHandle.test.cpp` exists in the fork and is the reference for what the cases should cover, not a file to copy.

lucaromagnoli:
Closes #2300. Slice 1 of #1894, which had not been decomposed and is, with #1895, the whole of what is left to build before the cutover.

The launcher's state machine. `SessionClipScheduler`'s own header is explicit about what MAGDA has today: "This is NOT a parallel scheduler -- TE owns playback state via LaunchHandle." Everything on the session side is a command layer over `te::LaunchHandle`. This is the state it has been sending commands to.

## The clock grew a third face

The handle could not be written without one, which is why a transport change is in a launcher PR.

A queued launch names a future beat. The timeline beat goes backwards every time the loop wraps and every time somebody locates, and `TransportClock` had no quantity that did not: `samplesSinceAnchor_` re-anchors at exactly the events a launch has to survive. Resolved against the timeline, a launch quantized to the next bar fires at the wrong moment, or twice, or never.

`BlockInfo` carries a monotonic beat range now, alongside the beats and the seconds it already had, and the header says why the three are siblings rather than two facts and a helper. It is accumulated from what each playing segment covered rather than read off the cursor, because the cursor is the thing a wrap is about to move. A stopped block does not advance it, for the same reason it does not advance the timeline.

`test_launch_handle.cpp` has the case that makes this concrete: a launch queued for a monotonic beat the timeline visits in neither block still fires, on the right sample. A handle resolving against the timeline waits for ever there.

## Why the answer is two ranges

`advance` does not return a play state. It returns up to two sub-ranges of the block with possibly different states, which is the fork's `SplitStatus` and is the reason this is a slice rather than a detail.

A block can straddle the beat a launch is quantized to. A handle answering once per block would start every clip on a block boundary: at 512 samples and 120 bpm that is up to 11 ms of slop, it differs per track, and removing exactly that slop is what quantized launch is for. Everything downstream has to be able to act on two pieces of one block, so the shape has to be right here, before a source or a chain reads it.

Half-open, and tested as such: a launch quantized to a beat that is exactly a block boundary belongs to the next block and is not counted twice.

## Anything quantized lands on its beat

At most one thing happens to a handle in a block, and when two are due the request wins outright rather than the earlier one winning.

A handle holds a single pending request, so a later `play` replaces an unfired `stop` rather than queueing behind it: two launches arriving before either fires means the second is what was asked for, the same way two locates leave the cursor where the second one said. The only other thing that can fall due is a loop re-trigger, and it gives way.

That is the rule rather than an optimisation. A wrap is the handle's own bookkeeping and is not entitled to move an instant somebody asked for. A clip launched off the grid has its wraps off the grid too, and an earlier-wins rule would carry that drift into the one action explicitly put on the beat, landing a quantized stop a block late. An unquantized launch accepts drift by definition; a quantized stop does not inherit it.

Giving way costs the wrap nothing. A stop ends the run and a launch restarts it, so either way the run it would have restarted is gone before the block is over, and there is no phase left to preserve.

Measured over 4000 launch positions, off-grid launches included: no stop is deferred at a 512 buffer or at 2048. Under the earlier-wins rule it was 0.05% and 1.23%. The comparison that picked between the two events is gone rather than reordered, so this is less code as well.

What a block still cannot express is two transitions, which would need a third sub-range `SplitStatus` has nowhere to put and the caller nowhere to act on. Nothing reaches that: a request cannot compete with another request, and a wrap no longer competes at all.

## playSynced joins a run, and copying the origin is not enough

The one test that failed first, and worth reading as part of the review.

`playSynced` is what makes a scene launch one event instead of N. The first implementation carried the leader's origin beat into the follower's run, which is what the fork's signature suggests, and it is wrong: a handle joining a leader four bars into its run then reports a played length of zero, so a scene launch puts every clip that was already playing back at its start.

What a follower has to adopt is the position, not the origin. The elapsed amount is derived at the instant the launch fires rather than captured when it was requested, because those are different numbers whenever the launch was queued for a later beat and only the first one is true. It falls out of the monotonic domain: elapsed is the fire beat minus the origin, and nothing has to be recorded in between.

It also makes looping fall out correctly. A synced handle's loop origin is the leader's, so a scene wraps together rather than each clip wrapping on its own anniversary.

## Loop is not the clip's loop

`setLooping` re-triggers the play duration: same handle, new run, played range restarts, and whatever reads it seeks. A clip looping inside its own length is the clip's business and happens without the handle knowing. Both exist in the fork, and conflating them is right whenever the two lengths agree and wrong whenever they do not, which is the kind of bug that survives a demo.

## What is deliberately not here

- **Threading.** Requests come from one thread and `advance` runs on another. How they meet is #2305's, along with who owns a handle's lifetime, and the epic rules out the fork's answer: a spin mutex over a trivially copyable `NextState`, per object, which works and is the hand-rolled deferred-release protocol #1894 says not to build. Until that slice this is a plain object.
- **Anything that reads a clip**, which is #2301.
- **The de-click ramp**, which is #2302 and mostly a matter of binding `OpKind::Crossfade` to a new cause.
- **Follow actions**, which are #2304. `ClipInfo` already carries `followAction`, `followActionDelayBeats` and `followActionLoopCount`; none of it is touched here.

## Tests

Sixteen cases, all offline and all deterministic, which is unusual for this subsystem and worth spending while it is available. Quantized launch and stop landing inside a block, on a boundary, and not at all; a request replacing an unfired one; the played range surviving a timeline wrap; a launch resolved in the monotonic domain the timeline never reaches; loop re-trigger and a queued stop beating a wrap in the same block; `playSynced` in phase with its leader; two handles launched as a scene starting on the same beat; nudge; and a stopped handle advancing eight blocks and changing nothing.

Full Catch2 suite green locally: 3293 passed, 13 skipped for plugins absent from this machine, 0 failures.

### DIFF
--- a/magda/engine/launch/LaunchHandle.cpp
+++ b/magda/engine/launch/LaunchHandle.cpp
@@ -186,6 +186,12 @@ SplitStatus LaunchHandle::advance(const SyncRange& range) {
         if (wrap < range.monotonic.end && (!event || wrap < *event)) {
             event = wrap;
             fromPending = false;
+
+            // A second wrap inside the same block is one this cannot report.
+            // Counted rather than swallowed: the run carries on past where it
+            // was due to restart, and nothing downstream can see that happen.
+            if (wrap + *loopBeats_ < range.monotonic.end)
+                ++loopRetriggerOverflows_;
         }
     }
 
--- a/magda/engine/launch/LaunchHandle.cpp
+++ b/magda/engine/launch/LaunchHandle.cpp
@@ -1,5 +1,6 @@
 #include "launch/LaunchHandle.hpp"
 
+#include <algorithm>
 #include <cmath>
 
 namespace magda::engine {
@@ -18,6 +19,20 @@ double project(const BeatRange& from, const BeatRange& to, double value) {
 }
 
 }  // namespace
+
+double LaunchHandle::virtualStart(const SyncRange& piece) const {
+    // Where the run would have begun on the timeline as it stands now, which
+    // is not where it did begin once the timeline has wrapped or been located
+    // since. A source works out how far into the material it is by subtracting
+    // this from the block it was handed, so the two have to be in the same
+    // cycle of the timeline; the beat the run actually started on stops being
+    // in any cycle the moment the loop takes it back.
+    //
+    // Derived from monotonic elapsed, which is the one quantity that survives
+    // the wrap. It may sit before the block or before zero, and that is what it
+    // means: the run began that far back.
+    return piece.timeline.start - (piece.monotonic.start - playedMonotonic_->start);
+}
 
 std::optional<LaunchHandle::QueueState> LaunchHandle::queuedState() const {
     if (!pending_)
@@ -69,11 +84,16 @@ void LaunchHandle::nudge(double beats) {
     // The origin moves, not the end: the played length is what a source reads
     // its position from, so shifting where the run began is what moves the
     // playhead through the material without interrupting the run.
-    if (played_)
-        played_->start -= beats;
+    if (!played_ || !playedMonotonic_)
+        return;
 
-    if (playedMonotonic_)
-        playedMonotonic_->start -= beats;
+    // Backwards only as far as the run has got. Further would put the origin
+    // in the future, which is a run of negative length and a loop whose next
+    // wrap has already passed; a handle cannot be nudged to before it started.
+    const auto effective = std::max(beats, -played_->length());
+
+    played_->start -= effective;
+    playedMonotonic_->start -= effective;
 }
 
 std::optional<BeatRange> LaunchHandle::playedRange() const {
@@ -185,9 +205,17 @@ SplitStatus LaunchHandle::advance(const SyncRange& range) {
     if (!event && playState_ == PlayState::playing && loopBeats_ && playedMonotonic_) {
         const auto origin = playedMonotonic_->start;
         const auto elapsed = range.monotonic.start - origin;
-        auto wrap = origin + (std::floor(elapsed / *loopBeats_) + 1.0) * *loopBeats_;
 
-        while (wrap <= range.monotonic.start)
+        // The first wrap at or after this block begins, which is ceil rather
+        // than floor-plus-one: a wrap due exactly on the first sample was
+        // passed over by the previous block, whose range ended before it, and
+        // rounding up again here would skip it a second time and every
+        // block-aligned wrap after it. Never the origin itself, which is where
+        // the run started rather than somewhere it repeats.
+        const auto turns = std::max(1.0, std::ceil(elapsed / *loopBeats_));
+        auto wrap = origin + turns * *loopBeats_;
+
+        while (wrap < range.monotonic.start)
             wrap += *loopBeats_;
 
         if (wrap < range.monotonic.end) {
@@ -204,8 +232,8 @@ SplitStatus LaunchHandle::advance(const SyncRange& range) {
 
     if (!event) {
         if (playState_ == PlayState::playing) {
+            status.playStartTime1 = virtualStart(range);
             extendRun(range);
-            status.playStartTime1 = played_->start;
         }
 
         return status;
@@ -222,8 +250,8 @@ SplitStatus LaunchHandle::advance(const SyncRange& range) {
         status.playing2 = status.playing1;
 
         if (playState_ == PlayState::playing) {
+            status.playStartTime1 = virtualStart(range);
             extendRun(range);
-            status.playStartTime1 = played_->start;
         }
 
         return status;
@@ -237,8 +265,8 @@ SplitStatus LaunchHandle::advance(const SyncRange& range) {
                            BeatRange{*event, range.monotonic.end}};
 
     if (playState_ == PlayState::playing) {
+        status.playStartTime1 = virtualStart(first);
         extendRun(first);
-        status.playStartTime1 = played_->start;
     }
 
     applyEvent(fromPending, second.timeline.start, second.monotonic.start);
@@ -246,8 +274,8 @@ SplitStatus LaunchHandle::advance(const SyncRange& range) {
     status.playing2 = playState_ == PlayState::playing;
 
     if (playState_ == PlayState::playing) {
+        status.playStartTime2 = virtualStart(second);
         extendRun(second);
-        status.playStartTime2 = played_->start;
     }
 
     status.isSplit = true;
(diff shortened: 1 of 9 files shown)


magda/engine/launch/LaunchHandle.cpp
```diff
@@ -0,0 +1,260 @@
...
+        } else if (*position < range.monotonic.end) {
+            event = *position;
+            fromPending = true;
+        }
+    }
+
+    if (!event && playState_ == PlayState::playing && loopBeats_ && playedMonotonic_) {
+        const auto origin = playedMonotonic_->start;
+        const auto elapsed = range.monotonic.start - origin;
+        auto wrap = origin + (std::floor(elapsed / *loopBeats_) + 1.0) * *loopBeats_;
```
lucaromagnoli: **P1 — Retrigger a loop that lands exactly at block start**

This calculation always selects a wrap strictly after `range.monotonic.start`, so a due wrap at the first sample is skipped. A handle started at 0 with a two-beat loop, advanced over `[0, 2)` and then `[2, 3)`, returns an unsplit run from 0 through 3 instead of beginning the new run at 2. Please recognize the boundary as an event at block start (without treating the initial launch origin as a wrap) and add the corresponding half-open boundary test.

magda/engine/launch/LaunchHandle.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+    loopBeats_ = beats;
+}
+
+void LaunchHandle::nudge(double beats) {
+    // The origin moves, not the end: the played length is what a source reads
+    // its position from, so shifting where the run began is what moves the
+    // playhead through the material without interrupting the run.
+    if (played_)
+        played_->start -= beats;
```
lucaromagnoli: **P2 — Clamp a backward nudge before it inverts the played range**

A negative nudge larger than the elapsed duration moves `start` past `end` while the handle remains playing. After two played beats, `nudge(-3.0)` produces `{3, 2}`, makes the range empty, and also puts the loop origin in the future. Clamp the new duration to zero and shift both origins by only the effective nudge, as the incumbent does; please add a backward-nudge test.

magda/engine/launch/LaunchHandle.cpp
```diff
@@ -0,0 +1,260 @@
...
+            // was due to restart, and nothing downstream can see that happen.
+            if (wrap + *loopBeats_ < range.monotonic.end)
+                ++loopRetriggerOverflows_;
+        }
+    }
+
+    if (!event) {
+        if (playState_ == PlayState::playing) {
+            extendRun(range);
+            status.playStartTime1 = played_->start;
```
lucaromagnoli: **P1 — Derive the playback origin in the current timeline cycle**

`played_->start` remains the original timeline beat, so it stops being a usable `playStartTime` after the arrangement timeline wraps or is located. For example, launching over timeline/monotonic `[6, 8]` and then advancing timeline `[0, 2]` with monotonic `[8, 10]` returns `range1.start == 0` and `playStartTime1 == 6`; a source subtracting them seeks to `-6` even though the run is two beats in. The incumbent already derives the virtual start from monotonic elapsed. Please use the equivalent of `range.timeline.start - (range.monotonic.start - playedMonotonic_->start)` here and in the other `playStartTime` assignments, and cover the returned `SplitStatus` across a timeline wrap—the current test only checks `playedRange()`.

Issue #2294, opened 2026-09-01
Expose device mods and macros over MCP


lucaromagnoli:
Follow-up to #2274/#2293, found while doing live sound design on Surge XT over MCP: "apply some modulations" is currently impossible for an external agent.

## The gap

MAGDA's modulation system is invisible to the remote API:

- No `mods.*` or `macros.*` operations exist in the registry — a client cannot create, list, edit, remove, or link an LFO/envelope mod or a macro.
- No facade exists either: `MagdaApi` has no `ModApi`; the UI drives modulation directly through `TrackManager` (`TrackManagerModulation.cpp`), so there is nothing for a remote handler to call.
- The only touchpoint is indirect: automation targets already carry `modId`/`modParameterIndex`, so a lane can automate a *pre-existing* mod's own parameter — but `devices.list` does not project `DeviceInfo::mods`, so a client cannot discover a `modId` to point at.

## Why it matters

Third-party instruments typically do not expose their internal mod routing as automatable parameters (Surge XT's mod matrix is the live example — its filter-EG depth and LFO routings are unreachable over VST3). MAGDA-side mods on the plugin's parameters are therefore *the* way an agent adds movement to an external instrument, and macros complete the loop: bind cutoff to M1 inside Surge once, and an agent can drive M1 with a MAGDA LFO despite the matrix being closed.

## Proposed shape (mirrors the #2274 parameter work)

1. **Facade**: `ModApi` (or extend `DeviceApi`) over `TrackManagerModulation` — list mods/macros per device, create (shape, rate, depth, target parameter), update, remove, link/unlink; mock implementation alongside.
2. **Projection**: mods and macros in the device surface — either on `DeviceDto`/a `devices.listMods` read, including `modId`s so automation targets become discoverable.
3. **Operations** (edit scope): `mods.list`, `mods.create`, `mods.update`, `mods.remove`, and macro equivalents (`macros.list`, `macros.setValue`, `macros.link`).
4. Scope-policy rows, permissions-doc note, DevicePanel allowlist, contract/dispatch tests.

Respecting the AI-agent allowlist for link targets (a mod may only be linked to parameters the user opted in, same as `devices.setParameter`) keeps the safeguard consistent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1863, opened 2026-07-25
Agent runtime: bounded observe-act loop over RemoteApiService


lucaromagnoli:
Part of #1861. Depends on #1862 and the Remote API foundations #1854/#1855.

## Summary

Implement the reusable runtime that turns a contextual agent definition into a bounded observe → model → tool → observation loop. Keep it independent of `AIChatConsoleContent` and DAW UI components.

## Runtime model

A run owns:

- Agent definition and active surface
- Conversation and baseline context
- Project/revision snapshot
- Tool allowlist and permission context
- Step, time, token, and mutation budgets
- Cancellation/deadline state
- Per-step trace and audit data
- Optional staged transaction/draft handle

## Scope

- [ ] Define `AgentDefinition`, `AgentRun`, run states, and terminal outcomes
- [ ] Build model requests from baseline context plus allowed Remote API tool schemas
- [ ] Execute tool calls only through `RemoteApiService`
- [ ] Append structured results, errors, and updated revisions to the next model step
- [ ] Support final-text completion and explicit stop conditions
- [ ] Enforce maximum steps, wall time, tokens, mutations, and repeated-call detection
- [ ] Propagate cancellation through model calls and queued message-thread work
- [ ] Add permission/approval hooks before mutations
- [ ] Define sequential semantics first; reject unsafe parallel mutations
- [ ] Add deterministic runtime tests with a scripted fake model and mock tools
- [ ] Add run-level tracing suitable for UI diagnostics

## Acceptance

- A scripted test model can read state, perform two dependent actions, observe both results, and finish.
- Execution errors can be repaired on a subsequent step without user resubmission.
- State revisions are refreshed after every operation.
- A looping or repeatedly failing model terminates predictably with a useful reason.
- Cancellation and MAGDA shutdown cannot leave an in-flight tool touching destroyed state.
- The runtime contains no direct references to `TrackManager`, `ClipManager`, or UI classes.

PR #2308, opened 2026-09-01
Unify agent and MCP access to MagdaApi (#2295, #2296, #2297)

### LINKED ISSUE #2295: Route in-app agents through RemoteApiService and enforce the ConsoleRouting allowlists
## Current state

In-app agents and the MCP server act on the same engine-owned `MagdaApi` instance, but through different layers:

- MCP goes through the `OperationRegistry` contract (51 ops in `remote_api.cpp`): schema-validated, permission-scoped, audit-logged, dispatched by `RemoteApiService`.
- In-app agents bypass all of it. `AIChatConsoleContent.cpp:1110` grabs `engine->getMagdaApi()` and hands the raw facade to the DSL/IR interpreters (`dsl_interpreter`, `InstructionExecutor`, `AutomationExecutor`) — no registry, no scopes, no audit log.

The unification layer was designed but never wired:

- `agent_runtime.hpp:133` defines `ToolExecutor`, whose comment says production adapters "must delegate this boundary to RemoteApiService". The only implementation in the tree is `FakeExecutor` in `tests/agents/test_agent_runtime.cpp:55`; `AgentRuntime` is never constructed outside tests (#1863 built it, nothing adopted it).
- `ConsoleRouting.cpp:39-200` defines per-surface tool allowlists drawn from remote op names, but they are prompt/routing metadata only: `invalidSurfaceTools()` is called from `tests/routing/test_console_routing.cpp` and nowhere else. `AgentRunPolicy.approveMutations` (`ConsoleRouting.hpp:62`) is likewise only consumed by the unwired runtime.

## Expected

Agent tool calls execute through `RemoteApiService` (a production `ToolExecutor`), so agents get the same validation, scoping, and audit trail as MCP clients, and the `ConsoleRouting` surface allowlists are actually enforced at dispatch time.

## Notes

Full parity is blocked on the remote op set covering what the DSL uses today (grooves, plugins, note editing, etc.) — tracked separately. This issue is the plumbing: executor, enforcement, audit.

https://claude.ai/code/session_01VMjNEtqw2WxSK3evwdeXZG

### LINKED ISSUE #2296: Move the setParameter user-opt-in gate from the remote handler into DeviceApi
## Symptom

The guard that refuses parameter writes the user hasn't opted in lives only in the MCP/remote layer: `devices.setParameter` checks the opt-in flag in `remote_handlers.cpp:588` (flag projected at `remote_api_projection.cpp:376`; internal devices always allowed).

`DeviceApi::setDeviceParameter` (`device_api.hpp:147`) has no such check, and the in-app agents call the facade directly (`AIChatConsoleContent.cpp:1110` → DSL/IR executors constructed with `MagdaApi&`). So the same agent action that an MCP client would be refused goes through unguarded when issued from the in-app console.

## Expected

The opt-in gate is enforced at the `DeviceApi` boundary (or another layer every caller shares), so remote and in-app callers see the same policy. The remote handler keeps only its transport-level concerns.

## Notes

Related to #2295 (routing agents through RemoteApiService would also close this hole, but the gate belongs with the API regardless — LuaController, OSC, and the CLI hold the same facade).

https://claude.ai/code/session_01VMjNEtqw2WxSK3evwdeXZG

### LINKED ISSUE #2297: Remote API parity: cover the MagdaApi surfaces and DSL verbs that have no MCP operations
## Current state

The remote `OperationRegistry` (51 ops, `remote_api.cpp`) covers `tracks/clips/devices/project/selection/session/transport/automation/undo`, plus MCP-only `subscriptions.*` and `system.describe`. The in-app agents use `MagdaApi` surfaces and DSL verbs that have no remote equivalent:

**MagdaApi sub-interfaces with no remote handlers:**
- `grooves()` — used 4× by `dsl_interpreter.cpp`
- `plugins()` — used 2× (plugin catalogue)
- `aliases()`, `midi()`, `focused()` — no remote ops at all

**DSL verbs with no MCP equivalent** (`dsl_interpreter.hpp:196-240`): `addChord`, `addArpeggio`, `transpose`, `setPitch`, `setVelocity`, `quantize`, `resizeNotes`, `deleteNotes`, `selectNotes`/`selectClips`/`forEach` filtering, `groupTracks`, `moveTrack`, `renameClip`, `setClip`, `addFx`, rack-chain create/delete/set, `groove.new/extract/set/list`.

## Why it matters

This gap blocks #2295: agents can't be routed through `RemoteApiService` until the op set covers what the DSL actually does. It also means MCP clients (Claude etc.) simply can't do these things.

## Expected

Remote operations for the missing surfaces and verbs, with the usual DTO schemas, scopes, and projection support, so the `ConsoleRouting` allowlists can name real ops for every surface.

## Notes

#2294 (device mods and macros over MCP) is a sibling slice of the same parity effort. Probably best split per surface (notes/MIDI editing, grooves, plugins, track organization) rather than one mega-PR.

https://claude.ai/code/session_01VMjNEtqw2WxSK3evwdeXZG

lucaromagnoli:
Closes #2295, closes #2296, closes #2297.

One PR for the three issues from the agent/MCP access audit: in-app agents and remote clients acted on the same `MagdaApi` instance through different layers, with different capabilities and different safeguards. Three commits, one per issue.

## #2296 — the parameter opt-in gate moved into DeviceApi

The per-parameter AI opt-in was checked only in the MCP handler, so any facade consumer (agents, Lua, OSC, CLI) bypassed it. The gate now lives in `DeviceApiLive::setDeviceParameter`; the handler keeps a pre-check solely to classify the refusal as `PermissionDenied` with a useful message. Negative-verified: the new opt-in test fails with the gate reverted.

## #2297 — 14 new remote operations

Closing the widest capability gaps between the DSL and the registry:

- `clips.update` / `clips.transpose` / `clips.quantize` / `clips.sliceNotes`
- `tracks.group` / `tracks.move`
- `grooves.list` / `grooves.upsert`
- `focused.get` / `focused.setMacro` / `focused.cycleDevice`
- `midi.listOutputPorts` / `midi.send` / `midi.sendSysEx` — the **first operations behind the `hardware-midi` scope**, which #1860 declared ahead of time for exactly this moment

Scope policy table, permissions doc, service tests, and cross-transport conformance vectors updated. Deliberately not added: `AliasApi` (registry references, not remotable as-is), `PluginApi`'s sequencer/Faust appliers (sibling of #2294's device-surface work), and DSL sugar verbs like `addChord` (composable client-side from `clips.addMidiNote`).

## #2295 — agents routed through RemoteApiService

The production half the agent runtime (#1863) was missing:

- **`magda/agents/agent_tool_bridge`** — `agentToolsForSurface()` resolves each `ConsoleRouting` allowlist against `OperationRegistry` into the agent's tool set; `agentScopesForSurface()` grants only what the allowlist needs; `RemoteAgentToolExecutor` executes every call through `RemoteApiService`, so agent actions get schema validation, scope enforcement, optimistic concurrency, undo grouping, and the audit log — same as an MCP client. The allowlists are now enforced at execution time (definition membership in the runtime + an independent executor check), pinned by a test that resolves every registered surface against the live registry.
- **`magda/agents/llm_tool_model`** — `agent::Model` over `llm::LLMClient` using juce_llm's native tool-calling surface; one adapter for all configured providers.
- **Console wiring** — an opt-in tool-loop path in `AIChatConsoleContent` (`Config` `ai.toolLoop`, default **off**): when enabled, the surface's agent runs `AgentRuntime` end to end; preflight failure (flag off, no remote host, provider without tool calling) falls through to the existing DSL workflows unchanged. Off by default because the loop needs live-LLM soak time; the DSL paths remain the shipped behaviour. The approval hook is in place and currently auto-approves, matching what the DSL executors do today — a per-step approval UI is the follow-up that reads `AgentRunPolicy.approveMutations`.

## Testing

- Full unit suite: 3321 cases pass.
- New: 21 test cases across `[device-api]` gate, `[remote][service]` content ops, conformance vectors for `hardware-midi`, and `[agent-bridge]` (surface resolution, executor enforcement, end-to-end runtime run, LLM mapping).

Note: overlaps with #2293 only in `tests/remote/test_remote_service.cpp` (both append test cases); everything else merges cleanly.

https://claude.ai/code/session_01VMjNEtqw2WxSK3evwdeXZG

### DIFF
--- a/magda/agents/agent_tool_bridge.cpp
+++ b/magda/agents/agent_tool_bridge.cpp
@@ -59,7 +59,10 @@ remote::ScopeSet agentScopesForSurface(const AgentSurface& surface) {
 
 RemoteAgentToolExecutor::RemoteAgentToolExecutor(remote::RemoteApiService& service,
                                                  const AgentSurface& surface)
-    : service_(service), surface_(surface), scopes_(agentScopesForSurface(surface)) {}
+    : service_(service),
+      surface_(surface),
+      scopes_(agentScopesForSurface(surface)),
+      runNamespace_(juce::Uuid().toDashedString()) {}
 
 ToolResult RemoteAgentToolExecutor::execute(const ToolExecutionRequest& request,
                                             const CancellationToken& cancellation) {
@@ -88,7 +91,7 @@ ToolResult RemoteAgentToolExecutor::execute(const ToolExecutionRequest& request,
     context.clientId = "agent:" + juce::String(surface_.name);
     context.clientName = "magda-agent";
     context.transport = "in-app";
-    context.requestId = request.call.id;
+    context.requestId = runNamespace_ + "/" + request.call.id;
     // The runtime threads the revision it last observed; passing it through
     // gives agent writes the same optimistic concurrency a remote client gets.
     // Zero means the run has observed nothing yet and expects nothing.
--- a/magda/agents/agent_tool_bridge.cpp
+++ b/magda/agents/agent_tool_bridge.cpp
@@ -1,0 +1,152 @@
+#include "agent_tool_bridge.hpp"
+
+#include <algorithm>
+#include <chrono>
+#include <memory>
+
+#include "../daw/api/remote_service.hpp"
+
+namespace magda::agent {
+namespace {
+
+const remote::OperationDescriptor* resolvableOperation(const std::string& name) {
+    const auto* operation = remote::OperationRegistry::instance().find(juce::String(name));
+    if (operation == nullptr) {
+        juce::Logger::writeToLog("Agent surface allowlist names an unknown operation: " + name);
+        jassertfalse;
+        return nullptr;
+    }
+    // Transport-scoped operations (subscriptions) are per-connection state and
+    // an in-process agent has no connection; a surface listing one would hand
+    // the model a tool nothing can execute.
+    if (operation->transportScoped) {
+        juce::Logger::writeToLog("Agent surface allowlist names a transport-scoped operation: " +
+                                 name);
+        jassertfalse;
+        return nullptr;
+    }
+    return operation;
+}
+
+}  // namespace
+
+std::vector<ToolDefinition> agentToolsForSurface(const AgentSurface& surface) {
+    std::vector<ToolDefinition> tools;
+    tools.reserve(surface.toolAllowlist.size());
+    for (const auto& name : surface.toolAllowlist) {
+        const auto* operation = resolvableOperation(name);
+        if (operation == nullptr)
+            continue;
+        tools.push_back({.name = operation->name,
+                         .description = operation->summary,
+                         .inputSchema = operation->inputSchema.clone(),
+                         .access = operation->access == remote::OperationAccess::Write
+                                       ? ToolAccess::Mutation
+                                       : ToolAccess::Read});
+    }
+    return tools;
+}
+
+remote::ScopeSet agentScopesForSurface(const AgentSurface& surface) {
+    remote::ScopeSet scopes{remote::Scope::Read};
+    for (const auto& name : surface.toolAllowlist) {
+        const auto* operation = resolvableOperation(name);
+        if (operation != nullptr)
+            scopes.add(operation->requiredScope);
+    }
+    return scopes;
+}
+
+RemoteAgentToolExecutor::RemoteAgentToolExecutor(remote::RemoteApiService& service,
+                                                 const AgentSurface& surface)
+    : service_(service), surface_(surface), scopes_(agentScopesForSurface(surface)) {}
+
+ToolResult RemoteAgentToolExecutor::execute(const ToolExecutionRequest& request,
+                                            const CancellationToken& cancellation) {
+    const auto failure = [&](juce::String code, juce::String message,
+                             juce::var details = {}) -> ToolResult {
+        return {.callId = request.call.id,
+                .toolName = request.call.name,
+                .success = false,
+                .content = {},
+                .error = ToolError{.code = std::move(code),
+                                   .message = std::move(message),
+                                   .details = std::move(details)},
+                // The current revision, never something older: the runtime
+                // treats a regressed revision as a broken executor.
+                .projectRevision = service_.currentRevision(),
+                .mutated = false};
+    };
+
+    const bool allowed = std::find(surface_.toolAllowlist.begin(), surface_.toolAllowlist.end(),
+                                   request.call.name.toStdString()) != surface_.toolAllowlist.end();
+    if (!allowed)
+        return failure("tool_not_allowed", "Operation '" + request.call.name + "' is not in the " +
+                                               surface_.name + " surface's allowlist");
+
+    remote::RequestContext context;
+    context.clientId = "agent:" + juce::String(surface_.name);
+    context.clientName = "magda-agent";
+    context.transport = "in-app";
+    context.requestId = request.call.id;
+    // The runtime threads the revision it last observed; passing it through
+    // gives agent writes the same optimistic concurrency a remote client gets.
+    // Zero means the run has observed nothing yet and expects nothing.
+    if (request.expectedRevision != 0)
+        context.expectedRevision = request.expectedRevision;
+    context.scopes = scopes_;
+    context.deadline = request.deadline;
+
+    const auto revisionBefore = service_.currentRevision();
+
+    // The service runs the completion inline when this is already the message
+    // thread (or there is none), and posts it to the message thread otherwise —
+    // so one dispatch-and-wait covers every calling context, and the inline
+    // cases signal before the first wait. Shared state keeps the completion
+    // valid even if this thread abandons the wait at the deadline.
+    struct Pending {
+        juce::WaitableEvent done;
+        remote::Response response;
+    };
+    auto pending = std::make_shared<Pending>();
+    service_.dispatch(request.call.name, request.call.arguments, context,
+                      [pending](remote::Response completed) {
+                          pending->response = std::move(completed);
+                          pending->done.signal();
+                      });
+
+    constexpr int pollMs = 50;
+    while (!pending->done.wait(pollMs)) {
+        if (cancellation.isCancellationRequested())
+            return failure("cancelled", "Run cancelled while the operation was in flight");
+        if (std::chrono::steady_clock::now() >= request.deadline)
+            return failure("time_limit", "Operation did not complete before the deadline");
+    }
+    const remote::Response response = std::move(pending->response);
+
+    if (!response.ok) {
+        juce::var details;
+        if (!response.error.issues.empty()) {
+            juce::Array<juce::var> issues;
+            for (const auto& issue : response.error.issues) {
+                auto* entry = new juce::DynamicObject();
+                entry->setProperty("path", issue.path);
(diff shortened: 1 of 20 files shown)


magda/daw/api/remote_handlers.cpp
```diff
@@ -1084,4 +1089,315 @@ HandlerResult automationClearLane(MagdaApi& api, const juce::var& input, const R
...
+            api, clipId, enabled ? juce::String("Enable Clip") : juce::String("Disable Clip"),
+            [enabled](auto& manager, ClipId id) { manager.setClipEnabled(id, enabled); });
+        mutated = true;
+    }
+    if (has(input, "grooveTemplate")) {
+        const auto templateName = input["grooveTemplate"].toString();
+        // An empty name clears the assignment; anything else must name a
+        // template that exists, or the clip would carry a groove that grooves
+        // nothing.
+        if (templateName.isNotEmpty() && !api.grooves().getTemplateNames().contains(templateName)) {
```
lucaromagnoli: [P1] Please validate the complete patch before executing any commands. `name` and `enabled` are mutated above this point, so `{ name: "Hook", grooveTemplate: "missing" }` renames the clip and then returns `NotFound`. Because failed handlers do not advance the revision or roll back the compound, callers see a failed request even though project state changed. A regression test should combine an earlier changed field with an invalid groove and assert no mutation.

magda/agents/agent_tool_bridge.cpp
```diff
@@ -0,0 +1,152 @@
...
+                                   request.call.name.toStdString()) != surface_.toolAllowlist.end();
+    if (!allowed)
+        return failure("tool_not_allowed", "Operation '" + request.call.name + "' is not in the " +
+                                               surface_.name + " surface's allowlist");
+
+    remote::RequestContext context;
+    context.clientId = "agent:" + juce::String(surface_.name);
+    context.clientName = "magda-agent";
+    context.transport = "in-app";
+    context.requestId = request.call.id;
```
lucaromagnoli: [P1] The idempotency namespace needs to be unique per agent run. Tool-call IDs are provider-generated and may be reused in later responses (for example `call_0`), while `clientId` is constant for the surface. Since the service caches writes by `clientId + requestId`, a later run can silently replay an earlier write with different arguments. Please add a per-run UUID to the client/request namespace, or leave `requestId` empty for this in-process executor.

lucaromagnoli (commented):
Inline review findings from the agent/MCP integration pass.

magda/daw/api/remote_handlers.cpp
```diff
@@ -1084,4 +1089,315 @@ HandlerResult automationClearLane(MagdaApi& api, const juce::var& input, const R
...
+HandlerResult tracksGroup(MagdaApi& api, const juce::var& input, const RequestContext&) {
+    std::vector<TrackId> trackIds;
+    for (const auto& item : *input["trackIds"].getArray()) {
+        const auto trackId = static_cast<TrackId>(static_cast<int>(item));
+        if (api.tracks().getTrack(trackId) == nullptr)
+            return notFound("track", trackId);
+        trackIds.push_back(trackId);
+    }
+
+    const auto groupId = api.tracks().groupTracks(trackIds, input["name"].toString());
```
lucaromagnoli: [P1] Please execute grouping and movement through `GroupTracksCommand` and `MoveTrackCommand` rather than calling `TrackApi` directly. The service opens an undo compound, but direct manager mutations add no command to it, so these remote edits cannot be undone. The same issue applies to `tracksMove` below.

magda/daw/api/remote_handlers.cpp
```diff
@@ -1084,4 +1089,315 @@ HandlerResult automationClearLane(MagdaApi& api, const juce::var& input, const R
...
+        return HandlerResult::fail(ErrorCode::ValidationFailed,
+                                   "status byte " + juce::String(bytes.front()) + " expects " +
+                                       juce::String(expected) + " bytes, got " +
+                                       juce::String(static_cast<int>(bytes.size())));
+
+    const juce::MidiMessage message(bytes.data(), static_cast<int>(bytes.size()));
+    if (!api.midi().sendMidi(input["port"].toString(), message))
+        return HandlerResult::fail(ErrorCode::NotFound, "MIDI output '" + input["port"].toString() +
+                                                            "' cannot be opened");
+    return HandlerResult::ok(acceptedResult());
```
lucaromagnoli: [P2] This should return `HandlerResult::unchanged(acceptedResult())`, as should `midiSendSysEx` below. `HandlerResult::ok` declares a project mutation, so every hardware MIDI message currently advances the project revision, creates an empty undo step, and publishes a project change even though project state did not change.

magda/daw/ui/panels/content/AIChatConsoleContent.cpp
```diff
@@ -65,6 +70,93 @@ namespace {
...
+    // console request thread — it is created at startup and destroyed at
+    // shutdown, after this panel (and its RequestThread) are gone.
+    auto* host = magda::remote::activeHost();
+    if (host == nullptr)
+        return std::nullopt;
+
+    const auto llmConfig = config.getAgentLLMConfig(magda::role::COMMAND);
+    // The on-device command model is an intent tagger, not a tool-calling
+    // model; the DSL path is the one that knows how to drive it.
+    if (llmConfig.provider == magda::provider::FAST_INFERENCE)
```
lucaromagnoli: [P2] Please also preflight `LLAMA_LOCAL` (or expose a tool-capability check) and fall through to the DSL path. `LlamaLocalClient` explicitly rejects every request containing tools, so enabling this flag with the embedded provider starts the loop only to return a model error, contrary to the fallback contract above.

magda/daw/api/remote_handlers.cpp
lucaromagnoli: Fixed in 7256320d — the groove template is validated before any field is applied, so a bad groove rejects the whole patch. Regression test asserts a rename + unknown groove mutates nothing (verified failing without the fix).

magda/agents/agent_tool_bridge.cpp
lucaromagnoli: Fixed in 7256320d — the executor now namespaces requestId with a per-run UUID minted in its constructor. Test reproduces the cross-run replay with a reused call_0 (verified failing without the fix).

magda/daw/api/remote_handlers.cpp
lucaromagnoli: Fixed in 7256320d — tracks.group and tracks.move run GroupTracksCommand / MoveTrackCommand through the undo stack. Live WebSocket tests prove one undo dissolves the group and restores the order; mock tests assert no direct facade writes.

magda/daw/api/remote_handlers.cpp
lucaromagnoli: Fixed in 7256320d — both handlers return HandlerResult::unchanged(acceptedResult()); tests assert the revision stays put and no undo step is created.

magda/daw/ui/panels/content/AIChatConsoleContent.cpp
lucaromagnoli: Fixed in 7256320d — LLAMA_LOCAL is preflighted alongside FAST_INFERENCE, so the request falls through to the DSL path instead of starting a loop the embedded client will refuse.

PR #2309, opened 2026-09-01
Give the engine a session domain: slots in the snapshot, handles in the store (#2301)

lucaromagnoli:
Part of #2301, and the half of it that stands on its own. The plan emission and the source that binds it are the other half, and the last section says why they are not here.

Session clips did not reach the engine at all before this. They were filtered out where lanes are built and the snapshot compiler diagnosed one that slipped through, so what is added here is a domain the engine did not have rather than an op.

## A slot has no position, so it is compiled at the origin

Nothing writes `placement.startBeat` for a session clip. The scene index is where it lives, and the beat field holds whatever it last was, so a slot compiled from what the model happens to carry would start wherever the clip used to be in the arrangement. The test uses a leftover 18.5 for exactly that reason.

Each slot is normalised to the origin and compiled by **recursing with a single-clip lane** rather than by a second implementation. That is the point rather than a shortcut: a clip dragged out of a slot onto the timeline has to sound the same, and there is now one answer about fades, events, warp and stretch instead of two that can drift apart. `plugin.wetdry` and the rest of #2246 exist because two hosts disagreeing about one plugin is a real bug; two compilers disagreeing about one clip would be the same bug a layer up.

It costs one compile per slot on the publishing thread and nothing on the audio thread.

`ClipLane` carries the two views in separate lists, so the existing guard keeps meaning what it said. A session clip in an arrangement lane is still a diagnostic, and the reverse now is too.

## Handles per slot, not per track

`RuntimeStateStore` owns them, keyed by `SlotKey`, retired by the rule it already applies to devices, clip sources, meters and value taps.

That rule is also the answer to #1894's requirement that a structural edit during playback must not restart a playing clip. The store already exists so that "the same edit that recompiles the plan leaves the objects it names untouched", so a handle needed no new lifetime mechanism. This is most of what #2305 asked for, and that issue is narrowed to the request lane accordingly.

Per slot rather than per track because a slot's state has to survive another slot playing in between. Launch scene 0, then scene 1, then scene 0 again: the first slot's loop phase, played range and last-played range are still its own. One handle per track would have slot 1 overwrite them, and follow actions (#2304) need exactly that survival to decide when to advance.

The op stays per track for the opposite reason. An op is a place in the graph and wants to be stable so the differ has nothing to do; per-slot ops would put a silent node in the graph for every scene on every track, and adding a scene would change the plan's shape everywhere instead of costing nothing.

## The op is defined and deliberately not emitted

`SessionAudio` and `SessionMidi` exist as kinds and roles, wired through `PlanBindings`, the executor, `PlanValues` and `PlanLayout`. The compiler does not emit them.

Emitting them is one line and it was written, run, and taken back out: it changed **54 test cases**, because every track gains ops and so every plan and every golden moves. Paying that for ops nothing can bind is churn with no behaviour behind it. They land in the slice that binds them, so the plan moves once.

That slice also has to answer how a slot's clips get streams, since `ClipAudioSource` matches against a table keyed by `(trackId, clipId, eventId)`. The answer is a section selector on `ClipAudioSource` rather than a parallel session source, for the same reason the snapshot recurses: stretch, warp, fades and voice management should not be written twice either.

## A bug this introduced, and the tests that caught it

The session loop went in between `if (!track.audio.empty() || !track.midi.empty())` and the push it guarded, which made the push unconditional and gave every empty lane a track entry. Three existing cases failed on it.

Worth recording because the fix is not a revert: the guard now covers the push **and** admits a session-only track, which is a real one. Nothing in its arrangement, and its slots waiting to be launched.

## Tests

Five cases, all offline:

- a slot compiles at the origin despite a leftover placement of 18.5
- a slot and the same clip in an arrangement lane produce the same spans, the same event count and the same event seconds
- slots are found by scene and held in scene order, and a scene nobody filled is not a slot rather than an empty one
- both diagnostics, plus a disabled slot skipped without comment
- a track carrying an arrangement clip and a session slot keeps them apart, with no diagnostics

Full Catch2 suite green locally: 3312 passed, 13 skipped for plugins absent from this machine, 0 failures.

### DIFF
--- a/magda/engine/clip/ClipSnapshot.hpp
+++ b/magda/engine/clip/ClipSnapshot.hpp
@@ -270,10 +270,43 @@ struct MidiClipPlayback {
 };
 
 /** Everything one track plays. */
+/**
+ * @brief One session slot's material, as if it began at beat zero.
+ *
+ * A session clip has no position on the timeline. Its scene index says which
+ * slot it sits in and nothing about when it sounds, because when it sounds is
+ * decided by a launch at runtime rather than by the model (#2301). So it is
+ * compiled at the origin, describing the material and nothing else, and the
+ * launch handle's virtual origin is what maps a block onto it.
+ *
+ * Compiled through the same path the arrangement's clips go through, so a clip
+ * dragged from a slot to the timeline sounds the same in both places.
+ */
+struct SessionSlotPlayback {
+    int sceneIndex = -1;
+
+    /// The slot's material, at most one of each. Empty means the slot holds a
+    /// clip that compiled to nothing playable, which is a diagnostic rather
+    /// than a slot that does nothing.
+    std::vector<AudioClipPlayback> audio;
+    std::vector<MidiClipPlayback> midi;
+
+    /// What one pass through the slot is worth, which is what a launch handle
+    /// re-triggers on. The clip's own loop is a separate thing that happens
+    /// inside this.
+    double lengthBeats = 0.0;
+};
+
 struct TrackClipPlayback {
     TrackId trackId = INVALID_TRACK_ID;
     std::vector<AudioClipPlayback> audio;
     std::vector<MidiClipPlayback> midi;
+
+    /// Sorted by scene index, so two compiles of one model agree.
+    std::vector<SessionSlotPlayback> session;
+
+    /// The slot at @p sceneIndex, or null.
+    const SessionSlotPlayback* slot(int sceneIndex) const;
 };
 
 /**
--- a/magda/engine/exec/RuntimeStateStore.cpp
+++ b/magda/engine/exec/RuntimeStateStore.cpp
@@ -109,6 +109,27 @@ template <typename Map, typename Ids> std::size_t eraseUnnamed(Map& map, const I
     return removed;
 }
 
+/// Handles whose slot the model no longer names.
+///
+/// Keyed by slot and retained by track, because that is the identity the live
+/// plan carries: the session op is per track, so what the plan says is which
+/// tracks still have a session, and every slot on a track that is gone goes
+/// with it. A slot emptied while its track remains reaches here as that track's
+/// own slots changing rather than as a plan edit.
+std::size_t eraseUnnamedSlots(std::map<SlotKey, std::unique_ptr<LaunchHandle>>& map,
+                              const std::set<TrackId>& tracks) {
+    std::size_t removed = 0;
+    for (auto entry = map.begin(); entry != map.end();) {
+        if (tracks.contains(entry->first.trackId)) {
+            ++entry;
+            continue;
+        }
+        entry = map.erase(entry);
+        ++removed;
+    }
+    return removed;
+}
+
 }  // namespace
 
 PlanBindings RuntimeStateStore::realise(const RenderPlan& plan, const RenderContext& context) {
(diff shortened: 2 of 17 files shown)


magda/engine/exec/RuntimeStateStore.cpp
```diff
@@ -109,6 +109,27 @@ template <typename Map, typename Ids> std::size_t eraseUnnamed(Map& map, const I
...
 }
 
+/// Handles whose slot the model no longer names.
+///
+/// Keyed by slot and retained by track, because that is the identity the live
+/// plan carries: the session op is per track, so what the plan says is which
+/// tracks still have a session, and every slot on a track that is gone goes
+/// with it. A slot emptied while its track remains reaches here as that track's
+/// own slots changing rather than as a plan edit.
+std::size_t eraseUnnamedSlots(std::map<SlotKey, std::unique_ptr<LaunchHandle>>& map,
```
lucaromagnoli, line 129: **P1 — Make slot identity part of handle retirement**

This only tests the track ID, so deleting or moving a slot while its track remains can never retire that slot's handle. The stale state is observable, not just leaked: if scene 0 was playing, is emptied, and later receives another clip, `handle({track, 0})` returns the old playing handle and the replacement inherits its phase/played ranges. Clip snapshots publish independently of plan topology, and the live plan names only the per-track session op, so neither current keep set can express which slot handles are still model-named or epoch-reachable. Please carry exact `SlotKey`s through the model retention set and through whichever published epoch holds the audio-side handle pointers, and add a delete/move-and-reuse test; simply narrowing this predicate without an epoch keep set would trade the stale state for a use-after-free.

— Codex

magda/engine/clip/ClipSnapshot.hpp
```diff
@@ -270,10 +270,43 @@ struct MidiClipPlayback {
...
+    double lengthBeats = 0.0;
+};
+
 struct TrackClipPlayback {
     TrackId trackId = INVALID_TRACK_ID;
     std::vector<AudioClipPlayback> audio;
     std::vector<MidiClipPlayback> midi;
+
+    /// Sorted by scene index, so two compiles of one model agree.
+    std::vector<SessionSlotPlayback> session;
```
lucaromagnoli, line 306: **P2 — Include the session domain in the canonical snapshot dump**

`dumpClipSnapshot()` still prints only `track.audio` and `track.midi`. A session-only track therefore dumps as `audio=0 midi=0`, and none of its scene keys, lengths, clips, events, or source data appears, even though the dump's contract calls it the snapshot's canonical/testable surface and the primary diagnostic for what a track plays. Please extend the track counts and body with these sorted slots and cover at least one session slot in the golden dump; otherwise two materially different session snapshots can compare identical here.

— Codex

magda/engine/exec/RuntimeStateStore.cpp
lucaromagnoli: **P1 — Route slot-aware retirement through the clip publish path**

The precise predicate and tests here are good, but the new three-argument `collectRuntimeStateIds` overload is used only by `test_launch_handle_lifetime.cpp`. Every existing session/host caller still builds the two-argument IDs, where `slots == nullopt`, and `EngineSession::publishClips()` only swaps the snapshot. Moving, emptying, or refilling a slot is explicitly a clip-snapshot update rather than a topology change, so it need not cause the structural `publish()` that calls `releaseDeleted`; in the real path this therefore continues to keep the stale handle by track. The direct store test bypasses the missing integration. Please make the clip publication path update/retire against the exact slot set (while preserving the documented epoch keep once handles become audio-bound), and cover this through `EngineSession::publishClips` rather than only by calling `releaseDeleted` directly.

— Codex

Issue #2299, opened 2026-09-01
Ten hand-written devices are still te::Plugin only, so a project using one renders as passthrough


lucaromagnoli:
Part of #1882. What is left of #2192 after its compiled-Faust slice, plus what the sampler issue set aside.

## What is missing

#2192 measured that most shipped devices could not run natively and named the compiled Faust effects as the slice worth doing first. Twenty-two of them crossed in #2195. What it did not cover, and what nothing tracks now, is the hand-written C++ devices. Ten remain `te::Plugin` with no `createDevice` in `BaseDevicePack.cpp`:

| device | class | header |
| --- | --- | --- |
| Arpeggiator | `ArpeggiatorPlugin : MidiDevicePlugin` | `plugins/ArpeggiatorPlugin.hpp:16` |
| MIDI Chord Engine | `MidiChordEnginePlugin : te::Plugin` | `plugins/MidiChordEnginePlugin.hpp:29` |
| Strum | `MidiStrumPlugin : MidiDevicePlugin` | `plugins/MidiStrumPlugin.hpp:27` |
| Step Sequencer | `StepSequencerPlugin : MidiDevicePlugin` | `plugins/StepSequencerPlugin.hpp:19` |
| Poly Step Sequencer | `PolyStepSequencerPlugin : MidiDevicePlugin` | `plugins/PolyStepSequencerPlugin.hpp:23` |
| Convolution | `MagdaConvolutionPlugin : te::Plugin` | `plugins/MagdaConvolutionPlugin.hpp:49` |
| Faust Instrument | `FaustInstrumentPlugin : te::Plugin` | `plugins/FaustInstrumentPlugin.hpp:35` |
| Mutable Elements | `MutableElementsPlugin : te::Plugin` | `plugins/mutable/MutableElementsPlugin.hpp:25` |
| Mutable Rings | `MutableRingsPlugin : te::Plugin` | `plugins/mutable/MutableRingsPlugin.hpp:26` |
| Mutable Clouds | `MutableCloudsPlugin : te::Plugin` | `plugins/mutable/MutableCloudsPlugin.hpp:28` |

`MidiDevicePlugin` is itself `te::Plugin` (`plugins/MidiDevicePlugin.hpp:19`), so the five MIDI devices are not a shorter job than the rest.

[...]

Issue #2271, opened 2026-08-31
The native engine has no sampler, so no Drum Grid project renders through it


lucaromagnoli:
Part of #1882. Found while wiring external plugins into the null-diff native leg (#2175).

## What is missing

`MagdaSamplerPlugin` has no `createDevice` in its registry entry (`BaseDevicePack.cpp`, the spec at `MagdaSamplerPlugin::xmlTypeName`), so `createSdkDevice()` returns null for `magdasampler` and the native engine cannot build one. It has a `createPlugin` and a `createProcessor`; what it does not have is the SDK device the plan binds.

## Why this one first

Every Drum Grid is made of them. `DrumGridPads.cpp` fills each pad with a `magdasampler` when the grid restores, so a Drum Grid project is not one missing device but one per pad, and a drum kit built the ordinary way renders nothing under the native engine.

The grid itself does not need one: `PlanCompiler::emitPadRack()` treats a pad-rack device as a rack rather than a device, and emits Device ops for the pads' contents instead. So the sampler is the whole of what stands between the native engine and a Drum Grid project.

## How it went unnoticed

The null-diff harness collected devices from a track's chain tree without descending into a device's pads, so no pad's sampler was ever looked up, and the executor's stand-in was bound in its place without a diagnostic. Both legs then rendered, the incumbent through real samplers and the native one through passthroughs, and any project whose pads had nothing to play nulled perfectly.

`project.v1bounces` is that project. Its Drum Grid has nothing to play, so the case passed while silently substituting a passthrough for every pad. #2175 makes the harness walk pads (`chain_walk::forEachDevice` with `Pads::Enter`), which turns the substitution into a reported diagnostic, and the fixture now declares it in `expectedDiagnostics` with a note pointing here.

**That declaration is the marker for this issue.** The day the sampler runs natively the diagnostic stops appearing, the declaration fails in the direction it is asserted in, and the line comes out of `MgdFixtures.cpp`.

## Scope

The sampler alone. It is one of roughly twenty-five entries in `BaseDevicePack.cpp` with no `createDevice` -- 4OSC, the step sequencers, the Mutable ports, the meter and monitor taps, and the rest -- and those are their own work under #1882 rather than this issue's. What makes this one worth doing on its own is that it is the only thing between the corpus and every Drum Grid project in it.

## Not affected

The tone generator already has one (`createDevice<ToneGeneratorPlugin>`), so projects carrying tone generators render natively today.

[...]

PR #2312, opened 2026-09-02
Port six hand-written devices to MagdaDevice (#2299)

lucaromagnoli:
Part of #2299 (epic #1882). Six of the ten hand-written `te::Plugin`-only devices cross to the native engine: **Rings (Halo), Elements (Materia), Clouds (Nimbus), Convolution (IR Reverb), Strum, Arpeggiator**. Each becomes an engine-neutral `MagdaDevice` hosted by `TracktionMagdaDevicePlugin` in the current engine and `EngineMagdaDevice` in the native one; the DSP and schedulers are untouched, and slot ids, order, and display ranges are pinned to what the retired plugins registered (skewed JUCE ranges reproduced exactly, e.g. skew 4 → `Exponential` with `skewFactor 0.25`).

## Shared infrastructure the remaining ports reuse

- **`MagdaDeviceProcessor`** (processors/base): wrapper-aware `DeviceProcessor` that surfaces the device's own `ParameterInfo` and converts display↔normalized at the slot boundary, the way `CompiledFaustProcessor` does for the compiled pack. The model keeps seeing real display units, which is what projects store and what the faceplates are written against.
- **`MidiMagdaDevice`**: engine-neutral base for the MIDI FX (note-strip atomics, note-off helper). The retired `MidiDevicePlugin` base stays until its last subclass crosses.
- **`transport/RampCurve.hpp`**: the Time Bend curve split out of `StepClock` so the neutral pack computes it without engine dependencies; `StepClock` delegates.
- Faceplates move off CachedValue writes and state-watching onto the model path (`updateFromParameters` / `onParameterChanged`) — StrumUI and ArpeggiatorUI in this PR, matching HaloUI.

## Adapter fixes that fell out of the suites

- `TracktionMagdaDevicePlugin::initialise()` now syncs parameters into the device before `prepare()` (same rule `reset()` already followed): the convolution parks its wet/dry smoothers on the current mix in prepare, and seeded from defaults it audibly ramped over the first block.
- `getTailLength()` reads the device live rather than the construction-time snapshot — the convolution's tail is the length of whatever IR is loaded.
- Convolution `flushState()` no longer removes an `irFileData` blob it does not hold.
- `pluginProducesMidi()` recognises a wrapped MIDI device.

## Survey corrections (#2299's open claims)

- **Faust Instrument has not crossed** — #2192's claim was wrong; it is still `createPlugin`-only and remains on the list.
- **The survey no longer counts fork plumbing** — the only "no native device" reasons left are `4osc` (7 projects, licence decision) and `magdasampler` (#2271), both explicitly out of scope. No survey fix needed.

## Verification

- Full Catch2 suite: 3,329 cases / 740k assertions, 0 failures, including a new engine render test for Rings and the generic SDK-device sweep now covering all six.
- JUCE suites: Device Param Schema Freeze (slot ids/order unchanged), Device State Schema, Legacy Corpus Device State Migration, Convolution Device (rewritten against the wrapper's slots, still asserting the retired normalised curves), Faust/Master Sidechain.
- Manual click-through of the Halo faceplate confirmed in-app.

## Also on this branch: #2317 (merged via #2319)

The model is now the sole authority for internal-device parameters and authored state: state documents carry authored state only, old documents hydrate into `DeviceInfo::parameters` once at load with the era chronology moved out of the bridge, save never copies values back off the engine (`populateParameters` takes a mandatory `ValueSource`), and arp settings / IR loads are model-first edits (IR load undoable). See #2319 for the full review trail.

## Not in this PR (follow-ups agreed on #2299)

Step Sequencer + Poly Step Sequencer (pattern state stays in the model under the existing STEP ids, edits become one snapshot-style undoable command, engine consumes compiled pattern data), MIDI Chord Engine (values-not-lookups: compile-time inputs + snapshot data instead of `trackContext`), and Faust Instrument (own PR: compile at publish off the audio thread, factory+voices in RuntimeStateStore, null-diffed against the TE leg).


magda/daw/audio/plugins/BaseDevicePack.cpp
```diff
@@ -335,26 +335,26 @@ void registerNativeDevices(InternalPluginRegistry& registry) {
...
          .createMode = InternalPluginCreateMode::SavedStateOrFresh,
-         .matchesPlugin = matches<ArpeggiatorPlugin>,
+         .matchesPlugin = matchesDevice<ArpeggiatorPlugin>,
          .createProcessor = makeProcessor<ArpeggiatorProcessor>,
          .showInBrowser = true,
          .tags = kArpeggiatorTags,
          .tagCount = static_cast<int>(std::size(kArpeggiatorTags)),
          .createInSession = createValueTreePlugin,
-         .createPlugin = createPlugin<ArpeggiatorPlugin>});
+         .createDevice = createDevice<ArpeggiatorPlugin>});
```
lucaromagnoli, line 341: [P1] Preserve the saved parameter domain when adopting the wrapper. Existing v2 device-state documents were captured from the old plugins, so their `params` values are in real units (for example Strum length 90 ms or Arpeggiator fixed velocity 100). `applyDeviceStateParameters()` still writes those values directly, but this registration now creates `[0,1]` wrapper parameters, causing them to clamp and restore as 400 ms / velocity 127. Pre-v2 Arpeggiator and Strum trees also use properties such as `arpRate` and `strumLength`, which the wrapper does not bind. Please add a migration/domain-aware restore before switching these saved-state devices to `createDevice`.

magda/daw/ui/components/chain/custom_ui/ArpeggiatorUI.cpp
```diff
@@ -135,7 +124,7 @@ ArpeggiatorUI::ArpeggiatorUI() {
     cyclesSlider_.setValueParser([](const juce::String& t) { return t.getDoubleValue(); });
     cyclesSlider_.onValueChanged = [this](double value) {
         if (plugin_)
-            plugin_->rampCycles = juce::roundToInt(value);
+            plugin_->rampCycles.store(juce::roundToInt(value), std::memory_order_relaxed);
```
lucaromagnoli: [P1] Route these non-slot settings through model state. This writes only the Tracktion wrapper device atomics. The native renderer owns a separate `EngineMagdaDevice` instance built from `DeviceInfo.pluginState`, and this write neither updates that state nor republishes a plan, so Ramp Cycles, Quantize, Subdivision, and Hard Angle do not affect native-engine playback. It also leaves the project clean, so autosave will not capture the edit. These controls need a model command/state update path.

magda/daw/ui/components/chain/custom_ui/ArpeggiatorUI.cpp
```diff
@@ -170,101 +159,113 @@ ArpeggiatorUI::ArpeggiatorUI() {
...
+    if (plugin_ == nullptr)
         return;
 
-    // Read modulated values from AutomatableParams (includes macro modulation)
-    float depth = plugin_->rampParam ? plugin_->rampParam->getCurrentValue() : plugin_->ramp.get();
-    float skew = plugin_->skewParam ? plugin_->skewParam->getCurrentValue() : plugin_->skew.get();
+    // Modulated values: the host pushes the modulated slot positions into the
+    // device every block, so its parameter mirror includes macro modulation.
+    const auto slotDisplay = [this](int slot) {
+        return magda::ParameterUtils::normalizedToReal(plugin_->parameterValue(slot),
```
lucaromagnoli: [P2] Avoid reading the device parameter mirror from the message thread. `parameterValue()` reads `values_`, which is a plain `std::array<float>`, while the audio thread writes the same entries from `syncParametersToDevice()` every block. The 30 Hz timer therefore creates a C++ data race. Please expose these live modulated values through atomics or a telemetry snapshot.

lucaromagnoli (commented):
Review findings from the device-porting pass.

magda/daw/audio/plugins/BaseDevicePack.cpp
lucaromagnoli: Fixed in a8400a14. Documents for internal-registry devices behind the wrapper are now display-domain by convention, converted at the bridge boundary in both directions: `applyDeviceStateParameters()` converts display -> normalized through the device's ParameterInfo, and `captureInternalDeviceState()` converts back, so the convention is stable across save/load. Released (0.19) docs — which are display-domain because the retired plugins' parameters ran in display ranges — restore unclamped; a new Device State Schema case pins the fixed-velocity-100 scenario end to end, and the legacy corpus migration suite now passes with live values actually matching the documents (the failures it previously masked were this exact clamp). The compiled pack is excluded and keeps its normalized convention, since its parameters were 0..1 host slots before and after crossing.

Scoped out per Luca: the pre-v2 engine-XML property spellings (`arpRate`, `strumTrigger`, ...) don't gain a migration table — those trees predate the corpus and restore parameter defaults. The non-parameter settings (`arpQuantizeSub` etc.) still restore, since the device reads those spellings directly.

magda/daw/ui/components/chain/custom_ui/ArpeggiatorUI.cpp
lucaromagnoli: Fixed in a8400a14. The four settings now report through `ArpeggiatorUI::onSettingsEdited`, which the manager routes to `capturePluginState(devicePath_)` — the same model-state path the IR load uses — so an edit flushes the device state into `DeviceInfo.pluginState`, dirties the project for autosave, and reaches the native engine's own device instance rather than only the live wrapper.

magda/daw/ui/components/chain/custom_ui/ArpeggiatorUI.cpp
lucaromagnoli: Fixed in a8400a14. `process()` publishes the modulated Time Bend pair into `displayedRamp_` / `displayedSkew_` atomics each block, and the 30 Hz timer reads those; nothing on the message thread touches the parameter mirror anymore.

magda/daw/audio/plugins/tracktion/TracktionDeviceStateBridge.cpp
```diff
@@ -137,6 +138,22 @@ juce::ValueTree legacyPluginTree(const juce::String& savedState) {
...
+/// Documents for these devices carry parameter values in the DEVICE'S DISPLAY
+/// DOMAIN. That is what every document captured from the retired host-native
+/// plugins already holds - their automatable parameters ran in display ranges -
+/// so released projects restore unchanged. The wrapper's parameters are
+/// normalised [0,1] slots, so capture and apply convert at this boundary.
+const MagdaDevice* displayDomainDeviceFor(te::Plugin& plugin, const juce::String& deviceType) {
+    auto* adapter = dynamic_cast<TracktionMagdaDevicePlugin*>(&plugin);
+    if (adapter == nullptr || findInternalPluginSpec(deviceType) == nullptr)
+        return nullptr;
+    return &adapter->device();
```
lucaromagnoli: [P1] Preserve v2 documents produced by devices that were already behind this wrapper on the target branch. Before this commit, `captureInternalDeviceState()` saved `getCurrentBaseValue()` unchanged, so Test Tone, Sidechain, Faust, and the analyzers can already have normalized v2 documents. This helper now classifies all of them as display-domain; for example, a saved normalized Test Tone frequency is interpreted as sub-1 Hz and clamps to 20 Hz on restore. The model parameter array may mask this for a full project, but state-only presets/imported chains use this fallback. Please encode/detect the parameter domain (or otherwise migrate the existing normalized form) and add a regression using a v2 document produced by the target branch.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -931,25 +937,38 @@ bool DeviceCustomUIManager::createMidiUtilityUI(const magda::DeviceInfo& device,
...
         arpeggiatorUI_ = std::make_unique<ArpeggiatorUI>();
+        forwardParameterChanges(*arpeggiatorUI_, callbacks);
+        // Non-slot settings live in the device state, so an edit has to be
+        // captured into the model: that is what persists it, dirties the
+        // project, and hands the native engine's device the new values.
+        arpeggiatorUI_->onSettingsEdited = [this] {
+            auto* audioEngine = magda::TrackManager::getInstance().getAudioEngine();
+            auto* bridge = audioEngine != nullptr ? audioEngine->getAudioBridge() : nullptr;
+            if (bridge != nullptr)
+                bridge->getPluginManager().capturePluginState(devicePath_);
```
lucaromagnoli: [P1] This still does not perform the model update described above. `PluginManager::capturePluginState()` only flushes the live wrapper and assigns `DeviceInfo::pluginState`; it does not call `ProjectManager::markDirty()`, emit a `TrackManager` notification, or recreate/republish the native `EngineMagdaDevice`. Consequently autosave still does not see the edit and the native instance keeps its old Ramp Cycles/Quantize/Subdivision/Hard Angle state. Please route this through a model command/state-change path that marks dirty and updates the native plan, with a test for both effects.

lucaromagnoli (commented):
Re-reviewed a8400a14. The atomic telemetry change fixes the UI data race, and the focused tests pass. Two persistence/model-state issues remain inline.

magda/daw/audio/plugins/tracktion/TracktionDeviceStateBridge.cpp
lucaromagnoli: Fixed in e80614cd. The document now declares its domain: wrapper captures write display values under a `paramsDomain: "display"` marker, and apply converts only when the marker is present — or for the six devices this branch ported, whose unmarked documents can only be display-domain (they were never behind the wrapper before the marker existed). Unmarked documents for the devices already wrapped on the target branch apply raw again, pinned by a regression using a target-branch-style normalised sidechain document (attack 0.5 restores as half the knob, not 0.5 ms). Readers that predate the field ignore it, so the marker costs nothing on the way down.

One observation from writing the regression: a raw `PluginCache::createNewPlugin` for `toneGenerator` resolves the stock Tracktion built-in that `TracktionEngineWrapperInit` also registers, not MAGDA's wrapper — which is why the test uses the sidechain. That resolution order predates this PR (#2196), but it may be worth a look at whether the built-in registration is still wanted.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
lucaromagnoli: Partially fixed in e80614cd, with the rest scoped deliberately. The edit now marks the project dirty alongside the capture, so autosave sees it; and a new engine test pins the native half — a device built from the model's saved document (`adapter::createEngineDevice` from `DeviceInfo.pluginState`) holds the edited ramp-cycles/quantize/subdivision/hard-angle values. Pre-cutover there is no long-lived native instance to republish to: engine devices are built from the model per render, so capture + dirty is the complete hand-off today.

What this does not add is an undoable command, on purpose: these four settings are authored state, the same category as the step-sequencer patterns, and #2313 already carries the decision that authored device state moves to the model with snapshot-style undoable commands. The arpeggiator settings should join that machinery when it lands rather than get a bespoke one-off command here.

magda/daw/audio/plugins/tracktion/TracktionDeviceStateBridge.cpp
```diff
@@ -137,6 +138,39 @@ juce::ValueTree legacyPluginTree(const juce::String& savedState) {
...
+
+/// Whether an UNMARKED document for @p deviceType holds display-domain values.
+///
+/// A document without the `paramsDomain` marker predates it, so its values are
+/// in whatever range the capturing parameter ran in. For the devices that
+/// crossed to the wrapper in #2312 that can only be the retired plugin's
+/// display range - they were never behind the wrapper before the marker
+/// existed. For everything that was already wrapped on the target branch (Test
+/// Tone, Sidechain, the runtime Faust device, the analyzers) an unmarked
+/// document is normalised and must apply raw.
```
lucaromagnoli: [P1] This assumption conflates two unmarked formats for the same device IDs. In v0.19, Sidechain/Test Tone/runtime Faust were still host-native and `captureInternalDeviceState()` saved their display-ranged parameter values; #2192 later moved them behind the normalized wrapper, after which target-branch captures became normalized. Both documents are schema-v2 and unmarked. Returning false for Sidechain therefore fixes the target-branch `0.5` case but breaks a released document such as release `15 ms`: it is applied as normalized `15`, clamps to `1`, and restores as `500 ms`. The new marker cannot disambiguate documents already written before it. Please preserve both historical forms (using reliable provenance or a safe legacy-domain discriminator) and add the corresponding v0.19-shaped Sidechain regression alongside the normalized one.

lucaromagnoli (commented):
Re-reviewed e80614cd. The dirty-state change is sufficient for the current per-render native-device lifecycle, and both focused suites pass. One released-state ambiguity remains inline.

magda/daw/audio/plugins/tracktion/TracktionDeviceStateBridge.cpp
lucaromagnoli: Fixed in the follow-up commit. You're right that the marker cannot reach back, and no provenance survives in the old documents — but the values discriminate safely in one direction: nothing normalised sits outside [0, 1], so an unmarked document containing any such value applies as display. A 0.19 sidechain document always carries one, because capture wrote every parameter and release defaults to 15 ms. The residue — an unmarked document whose every value is inside the unit interval — reads as normalised, which is exact where the display range is the unit interval and matches the target-branch form for the rest; for the released sidechain that residue requires every time constant to have been set below 1 ms, which I'm treating as acceptable.

Both eras are pinned side by side now: the normalised 0.5 attack restores as half the knob, and a v0.19-shaped document (gain 0.8, attack 1 ms, release 15 ms) restores in ms.

lucaromagnoli: Superseded by #2317 (merged into this branch as f3ef49f9, via #2319): the value-sniff era discrimination is gone from the bridge entirely. Documents no longer carry a parameter record at all; old records are consumed once at load by `DeviceStateHydration`, where the two unmarked eras resolve by container provenance (a project/preset `magdaVersion` < 0.20 can only be display-domain) with the released-state-first value sniff kept as the fallback for provenance-less state. Covered per era in `test_device_state_hydration.cpp`.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
lucaromagnoli: The deliberately-scoped remainder landed with #2317 (f3ef49f9 on this branch): the arp faceplate no longer writes wrapper atomics at all — settings edits travel as model document patches through `TrackManager::updateDeviceAuthoredState`, and the projection updates the live device. The capture direction this thread describes no longer exists for these settings.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1897,6 +1897,97 @@ void TrackManager::setDeviceParameterValue(const ChainNodePath& devicePath, int
...
+}
+
+bool TrackManager::setDeviceAuthoredState(const ChainNodePath& devicePath,
+                                          const juce::String& docText) {
+    auto* device = getDeviceInChainByPath(devicePath);
+    if (device == nullptr || device->format != PluginFormat::Internal)
+        return false;
+    // Same contract as updateDeviceAuthoredState: state this build cannot read
+    // must not be replaced blind.
+    if (device_state::isFutureDeviceState(device->pluginState))
```
lucaromagnoli: [P2] Validate the incoming snapshot, not only the state being replaced. If the device currently holds v2 state, `setDeviceAuthoredState(path, futureDoc)` passes this guard; `decode(futureDoc)` then returns nullopt and the future document falls through verbatim. The model accepts state this build cannot read, while projection turns the failed decode into a bare typed tree. Today the IR command only restores a snapshot captured from the same device, so that caller cannot produce this case, but this public write boundary now claims the invariant and does not actually enforce it. Please reject `isFutureDeviceState(docText)` as well (and ideally a decoded device-type mismatch), with a direct setter regression.

lucaromagnoli (commented):
Re-review complete at f3ef49f9. The previous parameter-domain, authored-state ownership, migration-order, convolution undo, and snapshot-canonicalization findings are resolved. One P2 remains inline: setDeviceAuthoredState validates the destination's current state rather than the incoming replacement, so its public write-boundary contract still accepts an unreadable future-schema payload. Focused authored-state/hydration tests pass (99 assertions), as do Device State Schema, Legacy Corpus Device State Migration (484 assertions), and Convolution Device. The combined head builds cleanly; current CI is green except Windows, which is still running.

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed in aebcc84f. The write boundary now enforces what it claims on both sides: an incoming `isFutureDeviceState(docText)` snapshot is refused (not stored unreadable while the projection restores a bare tree), and a readable document naming a different device is refused too — the saved type is resolved through the registry's load aliases before comparing, so an older alias for the same device still passes. Direct-setter regressions for both cases in `test_device_authored_state.cpp`.

lucaromagnoli (commented):
Re-review at aebcc84f: no remaining code findings. The incoming future-schema payload is now rejected, readable documents are checked against the destination device after load-alias resolution, and legacy XML/empty snapshots retain their intended passthrough. The exact head builds; [device-authored-state] passes 39 assertions across 5 cases; diff check passes. macOS, code quality, security, and release gate are green; Linux, Windows, and JUCE Linux are still running.

PR #2318, opened 2026-09-02
Add session sources to the plan, with voices and readers behind them (#2301)

lucaromagnoli:
The other half of #2301. The domain landed in #2309 (slots in the snapshot, handles in the store); this is the plan emitting the ops, the sources binding them, and the pool keeping a reader open for material that has no position on the timeline.

## The block moves, not the material

A slot is compiled at the origin and a launch handle says where its run began, so playing one is the arrangement's own playback over a block whose beat axis has been moved onto that origin (`clip/SessionPlayback.hpp`). Everything derived from the block's own two ends keeps working, which is the point: a material beat resolves to the sample the timeline beat it came from would have resolved to, so MIDI goes on writing into the callback's own buffer and audio into a sub-block of it.

That is why there is no second source. `ClipAudioSource` and `ClipMidiSource` each gained a constructor that takes a handle feed, and the feed is the section selector: a source given one plays the track's slots, one without plays its arrangement. Below the choice, the spans, fades, stretching, warping, readers, voices and the note-off invariant are the same code. A clip dragged out of a slot onto the timeline has to sound the same in both places, and now it goes through one implementation of that rather than two.

A split block, one a launch or a stop lands inside, is two such shifts over two sub-blocks of the output. That is what makes a clip start on its beat instead of on a callback boundary: at 512 samples and 120 bpm the difference is up to 11 ms and it differs per track. What is on the far side of a split is not rendered; the ramp across that boundary is #2302's.

## Advanced once, read twice

A slot has two sources and `LaunchHandle::advance` is a state machine that must see each block exactly once, so nothing that renders advances anything. `advanceLaunchHandles` runs over the whole table in `EngineSession::process`, before the plan, and the sources read `LaunchHandle::blockStatus`. That ordering is also what makes a scene coherent: every handle sees the same block and the same monotonic range, so eight clips launched for one beat fire in the same sub-range rather than a block apart depending on where their tracks sit in the plan.

The handles reach the audio thread the way everything else of this shape does, as an immutable table published whole (`LaunchHandleFeed`). The store builds and publishes it, because what is published is also the keep set eviction may not waive: the caveat #2309 left at `eraseUnnamedSlots` is now discharged rather than deferred. Nothing is swapped when the slots have not moved, since a clip dragged across the timeline republishes its snapshot at gesture speed and a publish waits for the block the callback is in.

## The ops, and what they cost

`SessionAudio` and `SessionMidi` are emitted for every track that carries clips, on the same terms as the arrangement's ops and for the same reason: whether a track has clips in it is a property of the snapshot, so a clip put into a scene must no more recompile a plan than one dropped on the timeline does. The compiler is handed tracks and not clips, so it could not have known either way.

Every plan therefore grows: a session op per track, and the two elided delays a two-input mix now carries. That is the 54 test cases #2309 declined to pay for, and every plan golden moves with them. A delay holding no samples is not a buffer, not a copy and not an op, so what the delays cost at render time is nothing; the session op costs a buffer clear, which is what an empty arrangement already costs on every track.

An unbound session source is reported only by a host that binds any of them. A host with no launcher, an offline bounce or a rig asserting topology, is an ordinary configuration rather than a track that has lost its source; one that binds some and not others has lost a track's session, and that is the message.

## Readers for material that is never due

The pool provisions every slot on every round, whatever the transport is doing. A slot has no position, so there is no window it comes into and no moment it is due: a launch can arrive on any block and nothing warns the pool first. They are cued at their own origin and never moved, which is what makes the launch itself cost no seek, and they rank as sounding without displacing a clip the transport is actually inside. A slot stopped part way through and launched again does seek; the cue that would avoid it belongs with the request that knows a launch is coming, which is #2305's.

## Tests

Sixteen new cases plus the ones the ops moved. The block arithmetic on its own, then silence for a slot nobody launched, material from the origin nine beats down the timeline, a launch and a stop landing on their sample inside a block, a handle advanced once over two blocks rather than twice, MIDI played from the origin and every note it started ended by the stop, the pool holding a reader per slot wherever the transport is, and the whole path once through `EngineSession`: publish the clips, take the handle, launch, hear it.

Full Catch2 suite green locally: 3363 passed, 13 skipped for plugins absent from this machine, 0 failures. The JUCE suite passes and the app builds.


magda/engine/clip/SessionPlayback.hpp
```diff
@@ -0,0 +1,145 @@
...
+ * of it, which is the audio side: the sample arithmetic then lands inside the
+ * piece it was given instead of inside the callback.
+ */
+inline BlockInfo materialSubBlock(const BlockInfo& block, const BeatRange& range, double playStart,
+                                  int count) {
+    auto material = block;
+    material.numSamples = count;
+    material.startBeat = range.start - playStart;
+    material.endBeat = range.end - playStart;
+    material.startSeconds = timeOfBeat(block, material.startBeat);
```
lucaromagnoli: **P1 — Keep material seconds aligned with rendered samples**

This rebases the seconds range through the tempo map at the material origin while retaining the sample count produced at the launch position. If beat 0 is 120 BPM and the slot launches in a 60 BPM section, a quarter-beat sub-block contains 12,000 samples but is labelled as 0.125 seconds. ClipVoice reads all 12,000 samples, while the next block reading position advances by only 6,000, so half of the material repeats.

Please derive the rebased seconds from the actual sub-range and launch instant, or carry a seconds-domain run origin, so the seconds span remains consistent with the rendered sample duration.

magda/engine/exec/EngineSession.cpp
```diff
@@ -142,6 +142,14 @@ void EngineSession::publishClips(std::shared_ptr<const ClipSnapshot> clips) {
     if (voices_ != nullptr)
         voices_->setSnapshot(clips);
 
+    // Handles before clips, and both before anything can look for them. A
+    // handle table that arrives first names slots the callback cannot see yet,
+    // and a handle nobody has launched is silent; a snapshot that arrives first
+    // would name a slot with no handle behind it, which is a launch the block
+    // in between cannot honour.
+    if (clips != nullptr)
```
lucaromagnoli: **P1 — Retire handles when clip publication removes their slots**

Publishing a snapshot without a slot removes it from the live handle table, but the LaunchHandle remains in RuntimeStateStore. The only releaseDeleted call is on plan publication, while clip edits intentionally do not recompile the plan. If the slot is refilled before another structural publish, handle(key) therefore republishes the old handle, including its playing state, loop phase, and played range. Publishing a null snapshot also leaves the previous handle table live.

Please retire handles after the replacement table has been safely published, including publishing an empty table for a null snapshot, so an emptied and refilled slot starts with fresh state.

magda/engine/clip/ClipVoicePool.cpp
```diff
@@ -329,10 +335,36 @@ void ClipVoicePool::service() {
...
+                // knows a launch is coming, which is #2305's.
+                //
+                // Ranked as sounding, because a slot is always about to be. It
+                // still sorts behind an arrangement clip the transport is
+                // actually inside: those started earlier, and this stands at
+                // the window's own start.
+                for (const auto& slot : track.session)
+                    for (const auto& clip : slot.audio)
+                        for (const auto& event : clip.events)
+                            candidates.push_back(Candidate{clip.clipId, event.eventId, &clip,
```
lucaromagnoli: **P2 — Do not permanently exclude slots beyond the reader budget**

These session candidates are always positioned at windowStart and sorted by stable IDs before the existing resize to kMaxReadersPerTrack. Unlike arrangement candidates, earlier slots are never passed by the transport, so a track with more than 32 session audio events always provisions the same first 32 and later slots never rotate into the table. Launching one of those later slots therefore produces permanent silence.

Please budget session readers separately, remove the cap for permanently provisioned slots, or add launch-aware prioritisation so every launchable slot can acquire a reader.

Issue #2315, opened 2026-09-02
Port the Faust Instrument to MagdaDevice


lucaromagnoli:
Part of #2299, and its own PR: the largest single port (~1300 lines of voice management around libfaust), and the one #2192 wrongly recorded as already done — #2312 established it is still `createPlugin`-only.

The runtime Faust FX crossed in #2196 and is the template. Three things to hold onto from it:

- **libfaust compilation stays off the audio thread.** Compile at publish; swap the instance with the same nonRealtimeMutatable pattern everything else uses. The dsp source is state, not a parameter — the engine factory already restores the saved document before the adapter snapshots parameter metadata, which is the ordering bug #2196 fixed for the FX and this port must not reintroduce.
- **The compiled factory and voices live in RuntimeStateStore, keyed on model identity**, so a plan recompile never rebuilds voices mid-note.
- **Null-diff it against the TE leg** like the other ports. 1300 lines of voice management is exactly where a silent behavioral drift hides; the render either matches the incumbent or the diff says where it does not.

lucaromagnoli:
Done in #2350, merged into `dev/0.20.0`. GitHub did not auto-close it because the merge target is not the default branch.

Issue #2314, opened 2026-09-02
Port the MIDI Chord Engine to MagdaDevice: values, not lookups


lucaromagnoli:
Part of #2299. The Chord Engine is the one MIDI device that could not follow #2312's recipe directly: it takes `services.trackContext` at creation, and an engine-built device gets an empty session key, so a straight port would hand the native leg a device that silently answers nothing.

## The decision: replace the service with values

No engine device holds a session pointer — devices get values, not lookups. Whatever `trackContext` answers at runtime splits two ways:

- **Edit-rate facts** (key, scale, chord-track assignment) become compile-time input: the compiler resolves them into the device's compiled state, and model edits already trigger a republish, so they stay fresh for free.
- **Positional facts** (the chord under the playhead) belong in the snapshot the audio thread already reads. If the chord track's content is not in ClipSnapshot yet, that is the real work item here — a lane feature, not a device port — and it goes first.

The port itself then follows the #2312 shape: slots pinned to the retired plugin's registration, ChordPanelContent moved onto the parameter path, `createMidiChordEnginePlugin` and its service plumbing deleted.

lucaromagnoli:
One thing for the port, from #2427 (closed into this): the Chord Engine reads MIDI and emits none, but it is declared a generator (`TrackManager.cpp:389` sets `DeviceType::MIDI`, `BaseDevicePack.cpp:193` tags it `midi-generator`), so `emitsMidi()` is true, it shows a thru toggle that does nothing, and the plan would put a `ChainMidiMerge` behind it and double every note. Under the #2347 contract its `DeviceProperties` should say `takesMidiInput` without `producesMidi`, and the model's type should agree so it gets `RawInputOnly` and no toggle.

lucaromagnoli:
Ported in #2452.

Worth recording that the premise above did not hold: the device reads exactly one host fact — `isChordTrackMuted()`, the chord track's own mute — and nothing anywhere reads a chord under the playhead, so no ClipSnapshot lane was needed. What actually blocked it was the plan skipping chord tracks (#2446) and the engine's declared role (#2427).

The fact is a host push now, delivered to every engine on the chord track at any rack depth. Detection stayed on the device, which two MagdaDevices already do.

Issue #2313, opened 2026-09-02
Port the step sequencers to MagdaDevice with pattern state in the model


lucaromagnoli:
Part of #2299. The last two `MidiDevicePlugin` subclasses, Step Sequencer and Poly Step Sequencer, follow the Strum/Arpeggiator shape from #2312 — with one design decision that the other ports did not need, settled here so the port does not make it implicitly.

## The decision: pattern state and undo live in the model, not the device

The engine owns compiled state, never authored state: compilers are pure functions barred from the message-thread singletons, and undo lives entirely in the model layer. A step pattern is authored state, the same category as MIDI notes. So:

- The pattern lives in the model under the engine-neutral device-state v2 schema (#1887), keeping the existing `STEP` tree ids as the persistence format so saved projects carry over unchanged.
- Faceplate and step-sequencer-agent edits go through UndoableCommands. The undo is cheaper than per-field commands: a pattern is small, so one generic whole-pattern snapshot command (capture the STEP subtree before/after, the SnapshotCommand pattern already in the codebase) covers every edit both surfaces make. Step toggles are exactly the kind of edit users spam and expect to undo.
- The engine device receives the pattern as plain compiled data on each publish. The engine never knows undo exists.

## The mechanics

`StepClock` is nearly neutral already: only its `processBlock` signature touches Tracktion (isPlaying plus two beat lookups). Refactor it to take a `BlockTiming{startBeat, endBeat, isPlaying, numSamples}` struct — the only `processBlock` callers are these two plugins, and every other user (agents, UIs, MidiNoteCommands, clip export) touches enums and statics only. The ramp curve already moved to `transport/RampCurve.hpp` in #2312.

Slots, registration, processor, and faceplate rewiring follow the recipe in #2312's commits; `MidiDevicePlugin` is deleted when the second sequencer crosses, together with the legacy branch in `pluginProducesMidi()`.

lucaromagnoli:
One structural constraint on the port, so the sequencer logic is extractable as its own module and reusable outside MAGDA (#2239's direction):

The sequencing core gets its own directory — pattern types, clock, and scheduling as a self-contained library with no MagdaDevice, no ValueTree, no engine types, and nothing above juce_core (ideally plain C++ so the eventual magda-sdk wasm target takes it as-is):

- **Pattern types**: `Step` / pattern as plain data structs. Serialization to the STEP tree ids is an adapter that lives with the model schema, not in the core.
[...]

PR #2319, opened 2026-09-02
Make the model the sole authority for internal-device parameters (#2317)

### LINKED ISSUE #2317: Make the model the sole authority for internal-device parameters and authored state
Follow-up to #1887, #2312, and the authored-state decision in #2313. This should not block merging #2312.

## Problem

The current internal-device bridge has two parameter authorities:

- `DeviceInfo::parameters`, which the model and plan value layer carry;
- `DeviceState::Doc::params` inside `DeviceInfo::pluginState`, captured back from a live Tracktion plugin.

Those copies have used different domains over time. Host-native plugins wrote display-ranged values, while `TracktionMagdaDevicePlugin` stores normalized slots. The same schema-v2 device ID can therefore have unmarked documents in either domain depending on which build captured it. #2312 exposes the ambiguity directly: a device-history allowlist can preserve one generation but cannot identify both forms reliably.

The update direction is also inverted for authored state: some faceplates mutate the live Tracktion wrapper and then call `capturePluginState()` to copy the result back into the model. The native engine is built from the model, so the incumbent adapter temporarily becomes the source of truth rather than a projection of it.

## Architectural decision

Use one direction everywhere:

`UI/agent command -> canonical model -> Tracktion projection / native-engine projection / serialization`

Concretely:

- `DeviceInfo::parameters` is the sole canonical parameter store for internal devices, in one documented domain.
- New internal-device `pluginState` documents do not duplicate automatable parameter values. They carry only non-parameter authored/device state that cannot live in the parameter array (for example IR data, Faust source, or sequencer patterns while those fields are migrated into typed model slices).
- UI and agent edits update model state through commands. They do not write a live wrapper and recapture it.
- Tracktion and native devices are disposable projections built or updated from the same model snapshot.
- Authored state uses the generic snapshot-style undo path established by #2313 rather than device-specific capture callbacks.
- Any remaining serialized state contract has an explicit schema/domain marker; runtime bridge code must not encode device migration chronology.

External-plugin native chunks are out of scope: they remain opaque plugin-owned state.

## Migration

Add a one-time load migration for existing internal-device state:

1. Read legacy engine XML and schema-v2 documents before constructing either engine projection.
2. Hydrate missing canonical `DeviceInfo::parameters` from the old document exactly once.
3. Resolve the historical display/normalized split using reliable project/preset provenance where available. Define and test a released-state-first fallback for standalone state that lacks provenance.
4. Remove the duplicate parameter payload when the project/preset is next saved, writing the new canonical form.
5. Preserve future-schema documents without destructive downgrade, as the current bridge does.

The migration must cover both sides of earlier wrapper cutovers, particularly Test Tone, Sidechain, runtime Faust, and the analyzers: released host-native display-domain documents and later wrapper-normalized documents share schema v2 and device IDs.

## Acceptance criteria

- A newly saved internal device has exactly one persisted automatable-parameter source of truth.
- Loading and saving does not copy parameters from Tracktion back into the model.
- Tracktion and native projections receive identical parameter and authored state from one model snapshot.
- Faceplate/agent authored-state edits are model commands, dirty the project, participate in undo where appropriate, and reach both projections.
- Project and device-preset fixtures cover:
  - pre-v2 engine XML;
  - released schema-v2 display-domain state;
  - post-wrapper schema-v2 normalized state;
  - the new canonical format;
  - state-only presets/imported chains with no accidental defaulting.
- `TracktionDeviceStateBridge` contains no per-device cutover allowlist such as `unmarkedDocIsDisplayDomain()`.
- Downgrade/future-schema behavior is explicit and tested.

lucaromagnoli:
Closes #2317. Stacked on #2312.

`DeviceInfo::parameters` (display domain) is now the sole persisted authority for an internal device's automatable parameters, and authored state flows one way: **UI/agent command → model → Tracktion projection / native projection / serialization**.

## What changed

**Documents stop carrying parameters.** `captureInternalDeviceState` writes authored (non-parameter) state only; `encode` skips the empty record, so a v2 document without `params` is the new canonical form — old builds read it gracefully (their doc-param apply no-ops and `syncFromDeviceInfo` still seats the model array). The live parameter list is still walked at capture so the engine's per-parameter properties stay filtered out of the root.

**The bridge lost its chronology.** `applyDeviceStateParameters`, `displayDomainDeviceFor`, and the `unmarkedDocIsDisplayDomain` allowlist are deleted. Restore paths seat parameters from the model array alone (`syncFromDeviceInfo`); `TracktionDeviceStateBridge` no longer encodes any device migration history.

**One-time load hydration** (`audio/plugins/DeviceStateHydration`). Old documents that do carry `params` are consumed exactly once, at load, before either engine projection exists: entries the model array is missing are hydrated from the record, converted to display domain by the chronology rules that used to live in the bridge — the document's `paramsAreDisplayDomain` marker first, then per-device era resolution (the #2312 six read display; the pre-marker wrapped five resolve by container provenance — a project/preset `magdaVersion` < 0.20 — falling back released-state-first on the value sniff; the compiled pack is always normalised; everything else is display). Slot metadata comes from a throwaway SDK device so the plan's value layer converts through real ranges; the minimal fallback is reserved for devices the native engine can't build anyway. Model entries are never overwritten. Wired into both project staging paths (with project provenance), and device/chain/rack preset loads (with preset-envelope provenance). The duplicate record disappears on the next save because capture no longer writes one.

**Loading and saving no longer copy parameter values Tracktion → model.** `DeviceProcessor::populateParameters(info, ValueSource)` now takes a mandatory `ValueSource::Engine | Model` argument, so every call site states which way values flow — the decision this issue exists to make explicit. `Model` keeps the model's `currentValue` for any entry that still names the same parameter; only entries the model lacked (new device, added parameter, Faust recompile that changed a slot's meaning) take the engine's value, and an external device always reads `Engine` because its chunk is authoritative. The per-processor engine read became the protected `populateParametersFromEngine()` hook (NVI split across the six processors), so the policy lives in one base method. All internal registration/save/capture sites pass `Model`; seeding and external paths pass `Engine`.

**Authored-state edits are model edits.** New `TrackManager::updateDeviceAuthoredState` patches the state document *in the model* and projects it into the running engine (never capture-back). The arpeggiator faceplate's settings edits now publish `(property, value)` patches in the device's own vocabulary (`ArpeggiatorPlugin::SettingIDs`, now public) instead of writing device atomics; the convolution IR load becomes an undoable `LoadImpulseResponseCommand` (snapshot of the whole document, so undo restores the previous IR) with the FLAC encoding factored into a pure static so validation happens before anything is committed.

## Explicitly deferred (owned by their ports)

Faust source publishes, step-sequencer pattern edits, and the 4OSC agent's non-parameter extras still write the live device and capture root-only authored state back. Their inversions need the typed model slices / publish-compile designs settled in #2313, #2314, and #2315, and their parameter edits already go through the model. The 4osc agent's parameter writes were verified to go through `setDeviceParameterValue`.

## Tests

- New Catch2 suite `test_device_state_hydration.cpp`: every document era — marked display, unmarked ported-with-marker, the two unmarked sidechain eras (out-of-range proof, all-unit-interval residue, provenance override), compiled-pack normalised, the params-less canonical form, future schema, pre-v2 XML, external untouched — plus `magdaVersion` provenance parsing.
- `Device State Schema` (JUCE): capture asserts *no* parameter record and no parameter-property leak into root; new model-authority test drives display values through `syncFromDeviceInfo` into the wrapper's normalised slots and proves `populateParametersModelFirst` doesn't read values back; round-trip proves a parameter does *not* travel through the document.
- `Legacy Corpus Device State Migration` and `Convolution Device` rewritten against the authored-state-only contract (conversion writes no duplicate record; second capture still byte-stable).
- Full Catch2 suite: 3,330 cases / 740,651 assertions, 0 failures. JUCE suites green: Device State Schema, Device Param Schema Freeze, Legacy Corpus Device State Migration, Convolution Device, DrumGrid Pad Chain Serialization (its two reload tests now tweak through `setDeviceParameterValue`, the only write path that reaches a save), Master Sidechain, External Plugin State Restore, MIDI Signal Routing.

### DIFF
--- a/magda/daw/audio/plugins/DeviceStateHydration.cpp
+++ b/magda/daw/audio/plugins/DeviceStateHydration.cpp
@@ -5,6 +5,7 @@
 #include <memory>
 
 #include "core/ChainWalk.hpp"
+#include "core/DeviceParamMigrations.hpp"
 #include "core/DeviceState.hpp"
 #include "core/ParameterUtils.hpp"
 #include "core/TrackInfo.hpp"
@@ -147,8 +148,32 @@ bool hydrateParametersFromDeviceState(DeviceInfo& device, const Provenance& prov
     if (!doc || doc->params.empty())
         return false;
 
+    // A record saved against a parameter order this build has since renumbered
+    // is identified by ITS length, not the model array's - the model array is
+    // exactly what an old preset or imported chain does not have. Remap the
+    // frozen indices first, or every old index would be read as a current one
+    // and land on the wrong slot (the pre-Enabled EQ, the 7-slot limiter).
+    // Dropped indices are dropped for the same reason the project-level pass
+    // drops them: re-pointing a value at a neighbour is corruption.
+    auto savedParams = doc->params;
+    const auto* migration = device_param_migrations::findMigrationForSavedCount(
+        device.pluginId, static_cast<int>(savedParams.size()));
+    if (migration != nullptr) {
+        std::vector<ds::ParamValue> remapped;
+        remapped.reserve(savedParams.size());
+        for (auto saved : savedParams) {
+            const auto newIndex =
+                device_param_migrations::migratedParamIndex(*migration, saved.index);
+            if (!newIndex.has_value())
+                continue;
+            saved.index = *newIndex;
+            remapped.push_back(std::move(saved));
+        }
+        savedParams = std::move(remapped);
+    }
+
     std::vector<const ds::ParamValue*> missing;
-    for (const auto& saved : doc->params)
+    for (const auto& saved : savedParams)
         if (saved.index >= 0 && !modelCarries(device, saved))
             missing.push_back(&saved);
     if (missing.empty())
@@ -184,6 +209,33 @@ bool hydrateParametersFromDeviceState(DeviceInfo& device, const Provenance& prov
         }
         // A normalised value without slot metadata is meaningless and is
         // dropped: there is no device to run it on either.
+    }
+
+    // Parameters the migrated order ADDED, seeded so the device keeps doing
+    // what the old one did (an old EQ's bands were always running; today they
+    // default off). Seed values are display-domain by contract
+    // (DeviceParamMigrations.hpp), whatever domain the record itself used.
+    if (migration != nullptr) {
+        for (const auto& seed : migration->seeded) {
+            if (device.findParameterByIndex(seed.index) != nullptr)
+                continue;
+            if (metadata != nullptr && seed.index < metadata->parameterCount()) {
+                auto info = metadata->parameterInfo(seed.index);
+                info.paramIndex = seed.index;
+                info.currentValue = seed.value;
+                device.parameters.push_back(std::move(info));
+            } else {
+                ParameterInfo info;
+                info.paramIndex = seed.index;
+                info.name = seed.name != nullptr ? juce::String(seed.name) : juce::String();
+                info.minValue = std::min(0.0f, seed.value);
+                info.maxValue = std::max(1.0f, seed.value);
+                info.defaultValue = seed.value;
+                info.currentValue = seed.value;
+                device.parameters.push_back(std::move(info));
+            }
+            added = true;
+        }
     }
 
     // A fully hydrated array keeps the device's own display order; entries
--- a/magda/daw/audio/plugins/DeviceStateHydration.cpp
+++ b/magda/daw/audio/plugins/DeviceStateHydration.cpp
@@ -71,19 +71,6 @@ ParamDomain resolveDomain(const DeviceInfo& device, const ds::Doc& doc,
     return ParamDomain::Display;
 }
 
-/// The saved root as the tree `MagdaDevice::restoreState()` takes - same shape
-/// the native engine's device factory builds (`EngineDeviceFactory.cpp`).
-juce::ValueTree treeFromNode(const ds::Node& node) {
-    juce::ValueTree tree(node.type.isNotEmpty() ? juce::Identifier(node.type)
-                                                : juce::Identifier("PLUGIN"));
-    for (int i = 0; i < node.props.size(); ++i)
-        tree.setProperty(node.props.getName(i), node.props.getValueAt(i), nullptr);
-    for (const auto& child : node.children)
-        if (child.type.isNotEmpty())
-            tree.appendChild(treeFromNode(child), nullptr);
-    return tree;
-}
-
 /// A throwaway SDK device for the parameter metadata a normalised value needs
 /// to become a display one. Restored from the document's own root first, for
 /// the devices whose parameter set depends on it (the runtime Faust device
@@ -106,7 +93,7 @@ std::unique_ptr<MagdaDevice> metadataDevice(const juce::String& pluginId, const 
 
     auto device = create();
     if (device != nullptr) {
-        auto tree = treeFromNode(doc.root);
+        auto tree = ds::toValueTree(doc.root);
         tree.setProperty(juce::Identifier("type"), doc.deviceType, nullptr);
         device->restoreState(tree);
     }
--- a/magda/daw/audio/plugins/DeviceStateHydration.cpp
+++ b/magda/daw/audio/plugins/DeviceStateHydration.cpp
@@ -1,0 +1,245 @@
+#include "plugins/DeviceStateHydration.hpp"
+
+#include <algorithm>
+#include <array>
+#include <memory>
+
+#include "core/ChainWalk.hpp"
+#include "core/DeviceState.hpp"
+#include "core/ParameterUtils.hpp"
+#include "core/TrackInfo.hpp"
+#include "plugins/DevicePluginHandle.hpp"
+#include "plugins/InternalPluginRegistry.hpp"
+#include "plugins/MagdaDevice.hpp"
+#include "plugins/compiled/CompiledPluginRegistry.hpp"
+
+namespace magda::daw::audio::device_state_hydration {
+
+namespace {
+
+namespace ds = magda::device_state;
+
+/// The domain a pre-#2317 document's parameter record used.
+enum class ParamDomain { Display, Normalized };
+
+/// Devices that were already behind `TracktionMagdaDevicePlugin` before the
+/// `paramsAreDisplayDomain` marker existed. Their unmarked documents have TWO
+/// eras: 0.19 captures from the retired host-native plugins hold display
+/// values, later captures from the wrapper hold normalised slots. Everything
+/// wrapped since the marker always writes it, so an unmarked document for any
+/// other device predates its wrapper and can only be display-domain.
+bool wrappedBeforeDomainMarker(const juce::String& deviceType) {
+    static constexpr std::array<const char*, 5> kPreMarkerWrapped{
+        "toneGenerator", "sidechain", "faust-fx", "oscilloscope", "spectrumanalyzer",
+    };
+    return std::any_of(kPreMarkerWrapped.begin(), kPreMarkerWrapped.end(),
+                       [&deviceType](const char* id) { return deviceType == id; });
+}
+
+/// Released-state-first fallback for a two-era document with no provenance:
+/// any value outside [0, 1] can only be display-domain (a 0.19 sidechain
+/// document always carries one - its release defaults to 15 ms). The residue -
+/// every value inside the unit interval - is read as normalised, which is
+/// exact for slots whose display range IS the unit interval and correct for
+/// wrapper-era captures of the rest.
+bool anyValueOutsideUnitInterval(const ds::Doc& doc) {
+    return std::any_of(doc.params.begin(), doc.params.end(), [](const ds::ParamValue& saved) {
+        return saved.value < -1.0e-4f || saved.value > 1.0f + 1.0e-4f;
+    });
+}
+
+ParamDomain resolveDomain(const DeviceInfo& device, const ds::Doc& doc,
+                          const Provenance& provenance) {
+    if (doc.paramsAreDisplayDomain)
+        return ParamDomain::Display;
+
+    // The compiled Faust pack's parameters were always normalised slots, and
+    // so were its documents, in every era.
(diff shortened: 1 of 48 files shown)


magda/daw/core/DeviceStateCommands.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return device != nullptr && irData_.getSize() > 0;
+}
+
+juce::String LoadImpulseResponseCommand::captureState() {
+    const auto* device = TrackManager::getInstance().getDeviceInChainByPath(devicePath_);
+    return device != nullptr ? device->pluginState : juce::String();
+}
+
+void LoadImpulseResponseCommand::restoreState(const juce::String& state) {
+    TrackManager::getInstance().setDeviceAuthoredState(devicePath_, state);
```
lucaromagnoli, line 39: [P1] Clear the live convolution when the restored snapshot has no IR. Undoing the first IR load restores an empty/default `pluginState`; `setDeviceAuthoredState` either projects no tree (for an empty string) or projects a tree without `irFileData`, and `MagdaConvolutionPlugin::restoreState` leaves `irData_` and the current convolution kernel untouched in both cases. The model therefore says the load was undone while playback still uses the new IR, and a later plugin-state capture can write that IR back into the model. Please make the absent-IR state explicitly reset the live device and cover the first-load undo path with a test.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1897,6 +1897,62 @@ void TrackManager::setDeviceParameterValue(const ChainNodePath& devicePath, int
...
+        return false;
+
+    device_state::Doc doc;
+    if (auto decoded = device_state::decode(device->pluginState))
+        doc = std::move(*decoded);
+    doc.deviceType = device->pluginId;
+
+    patch(doc);
+
+    device->pluginState = device_state::encode(doc);
```
lucaromagnoli, line 1956: [P1] Do not re-serialize the retired parameter record here. `decode()` preserves an older document’s `doc.params`, and `device_state::encode()` still writes that array whenever it is non-empty, so an Arp/IR authored-state edit writes the stale duplicate back into `DeviceInfo::pluginState`. A later Tracktion capture may happen to strip it, but preset serialization and native/no-live-plugin saves need not pass through that capture, so the document remains a second persisted authority and can hydrate stale values again on the next load. Please clear the migrated record before encoding (or make the production encoder omit it unconditionally) and add a decoded-old-document → patch → encode regression test.

magda/daw/core/DeviceStateCommands.cpp
```diff
@@ -0,0 +1,49 @@
...
+                                                       juce::MemoryBlock irData)
+    : devicePath_(devicePath), irName_(irName), irData_(std::move(irData)) {}
+
+juce::String LoadImpulseResponseCommand::getDescription() const {
+    return "Load Impulse Response";
+}
+
+bool LoadImpulseResponseCommand::canExecute() const {
+    const auto* device = TrackManager::getInstance().getDeviceInChainByPath(devicePath_);
+    return device != nullptr && irData_.getSize() > 0;
```
lucaromagnoli: [P2] Reject future-schema state in `canExecute()`. `updateDeviceAuthoredState()` deliberately returns false for that state, but `performAction()` ignores the result; `SnapshotCommand::execute()` then marks the command executed and `UndoManager` records a dirty undo step anyway, while the caller updates the IR label and returns success even though neither model nor engine changed. Checking the same future-state precondition here (and ideally the internal convolution type) prevents this false-success path.

magda/daw/audio/plugins/DeviceStateHydration.cpp
```diff
@@ -0,0 +1,232 @@
...
+    // any live processor could refresh it. The minimal fallback is reserved for
+    // devices without a factory - exactly the ones the native engine refuses to
+    // build, so nothing downstream converts through the placeholder.
+    const auto metadata = metadataDevice(device.pluginId, *doc);
+
+    bool added = false;
+    for (const auto* saved : missing) {
+        const bool haveSlot = metadata != nullptr && saved->index < metadata->parameterCount();
+        if (haveSlot) {
+            auto info = metadata->parameterInfo(saved->index);
```
lucaromagnoli, line 197: [P1] Run known index migrations on document-only parameter state before interpreting it with current metadata. The load order calls `applyParamIndexMigrations()` first, but that pass detects the old EQ/limiter layouts from `DeviceInfo::parameters.size()`. If that model array is empty—the exact old-preset/imported-chain case hydration is intended to recover—a 33-slot `magda_eq` document does not match the 33→41 migration. This loop then treats every old index as a current index, converts normalized values through the wrong slot metadata, omits the new Enabled seeds, and leaves the preset shifted. Please include the decoded record in migration discovery/remapping (before domain conversion) and add an empty-model, old-EQ/limiter document regression test.

lucaromagnoli (commented):
Review summary: three blocking correctness issues remain.\n\n- Document-only legacy EQ/limiter parameters bypass index migration and can hydrate into the wrong current slots.\n- Authored-state patching can re-encode legacy doc.params, preserving the stale second authority in preset and native-device paths.\n- Undoing the first convolution IR load restores the model snapshot but does not clear the live convolution kernel.\n\nI also left a non-blocking correctness comment for future-schema IR loads being reported as successful. The detached build, focused hydration/schema/corpus suites, convolution suite, and all current CI checks pass. GitHub does not permit the PR author to submit a changes-requested review, so these P1 threads should be treated as blockers.

magda/daw/audio/plugins/DeviceStateHydration.cpp
lucaromagnoli: Fixed in 991f89ec. Hydration now runs the record through the migration table itself, matched on the record's own length (`findMigrationForSavedCount`, exported for this): old indices remap via `migratedParamIndex` before any domain conversion, dropped slots are dropped rather than re-pointed, and the migrated order's seeded parameters (the EQ's eight Enabled switches) hydrate in display units per the SeededParam contract. Regression tests: a 33-slot `magda_eq` document and a 7-slot `magda_limiter` document hydrating onto an empty model, in `test_device_state_hydration.cpp`.

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed in 991f89ec. `updateDeviceAuthoredState` clears `doc.params` (and the domain flag) right after decode, before the patch — the edit is the moment a pre-#2317 document goes canonical, whatever save path follows. The production encoder keeps writing a non-empty record on purpose: the hydration tests use `encode` to fabricate old-era documents. Regression test: decoded old document with params → patch → encoded result carries none, in `test_device_authored_state.cpp`.

magda/daw/core/DeviceStateCommands.cpp
lucaromagnoli: Fixed in 991f89ec. Absent (or empty) `irFileData` now MEANS no impulse response: `restoreState` clears `irData_` and the name, and the device unloads by loading the single-sample identity impulse (dsp::Convolution has no unload; the dirac IS pass-through — a bypass flag turned out not to survive the render path). `flushState` now removes the blob from the engine tree when the device holds none, so a capture after the undo cannot resurrect it — the schema suite's binary-property test moved to a neutral property since `irFileData` is owned exactly by the device now. `setDeviceAuthoredState` projects an empty snapshot as a bare typed tree so the first-load undo actually reaches the device. Regression test: load → restore stateless document → name cleared, zero tail, no blob on flush, and a render peaking at zero delay, in the Convolution Device suite.

magda/daw/core/DeviceStateCommands.cpp
lucaromagnoli: Fixed in 991f89ec. `canExecute()` now rejects future-schema state, non-internal devices, and non-convolution ids (a literal, same layering reason as the property names). And since `UndoManager::executeCommand` records even a command whose `execute()` declines, the UI checks `canExecute()` BEFORE handing the command over — a refused load leaves no undo step, no dirty flag, and returns failure. Covered in `test_device_authored_state.cpp` (future-schema and wrong-device sections).

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -1897,6 +1897,80 @@ void TrackManager::setDeviceParameterValue(const ChainNodePath& devicePath, int
...
+    return true;
+}
+
+bool TrackManager::setDeviceAuthoredState(const ChainNodePath& devicePath,
+                                          const juce::String& docText) {
+    auto* device = getDeviceInChainByPath(devicePath);
+    if (device == nullptr || device->format != PluginFormat::Internal)
+        return false;
+
+    device->pluginState = docText;
```
lucaromagnoli: [P1] Canonicalize snapshots before restoring them. `LoadImpulseResponseCommand::captureState()` can snapshot a pre-#2317 document that still contains `doc.params`; execute canonicalizes it through `updateDeviceAuthoredState()`, but undo reaches this verbatim setter and reinstates the retired parameter record. That makes a simple load/undo cycle recreate the second authority this patch is meant to eliminate, especially in the preset/native paths called out above. This is the recurring pattern here: the invariant is implemented in one mutation path rather than at the authored-state write boundary. Please make replacement enforce the same decode/type/future-schema/canonicalization contract (or snapshot the canonical form), and extend the regression test through command undo.

lucaromagnoli: Fixed in 5158da1e — and taken as the general point, not the instance: the canonicalization now lives at the write boundary. `setDeviceAuthoredState` strips the retired record from any v2 document it is handed (so no caller, present or future, can put one back), refuses to replace future-schema state under the same contract as `updateDeviceAuthoredState`, and passes through verbatim only what decode refuses — legacy engine XML and empty snapshots, where there is no record to strip and no readable document to mangle. Regression tests: a full `LoadImpulseResponseCommand` execute → undo cycle over a pre-#2317 convolution document asserts the undone state is canonical (no `params`, no `irFileData`, authored props intact), plus a legacy-XML passthrough check, in `test_device_authored_state.cpp`.

PR #2325, opened 2026-09-02
A fourth face for the instant: monotonic seconds, and no run measured through the map (#2324)

### LINKED ISSUE #2324: Add a monotonic-seconds clock; never reconstruct elapsed runtime through the tempo map
## Problem

The engine has timeline seconds and monotonic beats, but it has no monotonic seconds domain.

That omission makes session playback reconstruct elapsed run time from beats. PR #2318 currently computes a synthetic virtual launch beat after a transport discontinuity and passes it through TempoMap::beatToTime:

    origin = beatToTime(virtualPlayStart)
    elapsed = block.startSeconds - origin

This is not elapsed time. Timeline seconds reset with the timeline on a loop or locate, while virtualPlayStart is only a beat projection into the new cycle. Converting that synthetic beat through the project tempo map silently changes the run position whenever tempo differs across the projected interval.

Example: with 120 BPM before beat 4 and 60 BPM from beats 4–8, launch at beat 7 and loop from 8 back to 4. One beat/one second of material has elapsed at the wrap. The projected virtual origin is beat 3, so beatToTime(4) - beatToTime(3) reports 0.5 seconds instead. Plain audio seeks backwards; later wraps compound the error. A locate has the same class of failure.

This is the same architectural rule monotonic beats already enforce: runtime state must not be reconstructed by round-tripping through a timeline coordinate that can jump.

## Required primitive

Add monotonic elapsed seconds derived directly from rendered sample counts in TransportClock, alongside startMonotonicBeat/endMonotonicBeat.

Carry that domain through BlockInfo and the launch range/status state so LaunchHandle can retain a run origin in monotonic seconds. Session playback must receive elapsed material seconds directly. It must not derive them with beatToTime or timeToBeat.

Synced launches need to share the same monotonic-seconds origin so beat-positioned and seconds-positioned material join the same run.

## Acceptance criteria

- Monotonic seconds advance from rendered sample duration, not from tempo conversion.
- They do not advance while stopped.
- They survive transport loop wraps, locates, count-in boundaries, and tempo edits.
- Session audio remains sample-continuous through a transport loop whose two sides have different tempos.
- Session audio retains its elapsed position across a locate.
- Synced handles agree in both monotonic beats and monotonic seconds.
- No runtime elapsed duration is reconstructed through a beat-to-time or time-to-beat round trip.
- PR #2318 either depends on this primitive or implements it before merging.

Found while re-reviewing #2318 after discussion_r3913923864.

lucaromagnoli:
Closes #2324. Stacked on #2318, because the bug it fixes is in that branch and the issue asks for the primitive before it merges.

The engine had timeline beats, timeline seconds and monotonic beats. It had no monotonic seconds, and session playback was reconstructing elapsed run time without one: it projected a virtual launch beat into the block's own cycle and put that through the tempo map.

That is not elapsed time. A map answers where a beat sits, and how long a run has been going is a different question. With 120 bpm to beat 4 and 60 bpm from 4 to 8, a slot launched at beat 7 and looping from 8 back to 4 has one beat and one second of material behind it at the wrap; the projected origin is beat 3, which is in the fast half, and the map reports half a second. Plain audio seeks backwards, and every later wrap compounds it. A locate is the same failure.

## The primitive

`TransportClock` accumulates monotonic seconds from the samples each segment actually rendered, beside the monotonic beats it already accumulated, and `BlockInfo` carries both ends. Nothing derives it from the beat faces: a block's own two faces agree, so the line itself is unremarkable, and what matters is what it refuses. A caller holding two monotonic beats and a tempo map cannot get seconds out of them, because after a wrap the two beats are not in the same cycle and after an edit the map between them is not the map it was.

It does not advance while stopped, and it survives wraps, locates, count-ins, tempo edits and a change of block size, for the reason the monotonic beat does: it is a sum of what was rendered rather than a reading of a position.

## Through the launcher

`SyncRange` carries all four faces of the block and `SyncPoint` carries the three a run's origin is remembered in, so a run cannot be started with three faces of one instant and one of another. A handle keeps its origin in monotonic seconds beside the one in monotonic beats, and `SplitStatus` reports both projected onto the block: `playStartTime` for the beat axis and `playStartSeconds` for the seconds one, each from its own monotonic domain.

`playSynced` shares all three, or the agreement would be only musical: a scene of one MIDI clip and one audio clip joined in beats alone would put the notes in the right bar and the samples somewhere else. `nudge` takes both faces from the caller for the same reason, rather than converting one into the other.

## And into the material

`materialBlock` and `materialSubBlock` now subtract on both axes instead of subtracting on one and converting for the other, so `timeOfBeat` is gone and with it the last round trip in the session path.

The project's tempo map does not travel onto a material block either. Both of its axes have moved and by different amounts, so asking the map where a material second falls answers about the timeline; for an auto tempo event, whose position is a question about beats, that is a beat the material never had. Dropping it leaves `beatAtTime` on the block's own two ends, which are the material's own two ends and are exact there.

## Tests

The clock's own domain gets the case list the monotonic beat has, plus the one it does not need: a tempo edit halves the musical time in the next block and leaves the wall-clock time alone. Then #2324's own example end to end, driven through a real `TransportClock`: a slot launched at beat 7 of a 4-to-8 loop whose two halves are at different tempos reads one block further into its file per block, straight across the wrap. The same across a locate to the top of the project. And synced handles agreeing in both monotonic domains and in both reported origins.

Both new playback cases fail on the old derivation, reading silence where the material should have continued.

## Acceptance criteria

- Monotonic seconds advance from rendered sample duration, not from tempo conversion: `TransportClock` sums `samples / sampleRate` per segment.
- They do not advance while stopped: covered.
- They survive loop wraps, locates, count-in boundaries and tempo edits: covered, one case each.
- Session audio remains sample-continuous through a loop whose two sides have different tempos: covered, and it is the issue's own example.
- Session audio retains its elapsed position across a locate: covered.
- Synced handles agree in both monotonic beats and monotonic seconds: covered.
- No runtime elapsed duration is reconstructed through a beat-to-time or time-to-beat round trip: `timeOfBeat` is deleted and the material block carries no map, so `ClipVoice::beatAtTime` reads the block's own line.

Full Catch2 suite green locally: 3370 passed, 13 skipped for plugins absent from this machine, 0 failures. The JUCE suite passes and the app builds.


magda/engine/clip/SessionPlayback.hpp
```diff
@@ -3,174 +3,101 @@
...
-    material.continuous = runsOn(block, range, playStart);
+    material.beats = BeatRange{block.beats.start - origin.beat, block.beats.end - origin.beat};
+    material.seconds =
+        SecondsRange{block.seconds.start - origin.seconds, block.seconds.end - origin.seconds};
+    material.continuous = runsOn(block, range, origin);
+
+    // Both axes moved, by different amounts, so the project's map would answer
+    // about the timeline. Dropped, so beatAtTime falls back to this block's own
+    // two ends, which are the material's.
+    material.tempo = nullptr;
```
lucaromagnoli: **P2 — Preserve tempo-aware beat lookup for session material**

Clearing `material.tempo` forces `BlockInfo::beatAtTime()` onto its two-end straight-line fallback. That is not safe for the caller this block reaches: `ClipVoice::renderThroughCells()` asks `beatAtTime()` at fixed 128-sample cell boundaries, and a cell may extend past the current block. Across a tempo step or ramp, a launched auto-tempo or warped clip therefore renders that part of the cell using the previous block’s slope, and the held tail is consumed by the next continuous block. `RenderContext.hpp` already documents this exact constraint: cells read at instants beyond the block and cannot afford the straight-line fallback.

Monotonic seconds should stay independent, but the material beat lookup still needs an exact tempo-aware mapping—for example, a view translated by the run’s seconds and beat origins—or the cell path needs an equivalent way to map those material instants correctly.

magda/engine/exec/RenderContext.hpp
```diff
@@ -190,20 +203,44 @@ struct BlockInfo {
...
         // instants beyond the block that produced them (ClipVoice's cells), so
         // it is exactly the caller that cannot afford the straight line.
+        //
+        // Through the origin in both directions on a shifted block, which is
+        // what keeps the map usable there: the moment goes back onto the
+        // timeline to be asked about, and the answer comes back onto the run's
+        // own axis. Zero on a timeline block, where this is the map itself.
         if (tempo != nullptr)
-            return tempo->timeToBeat(seconds);
+            return tempo->timeToBeat(moment + runOrigin.seconds) - runOrigin.beat;
```
lucaromagnoli, line 212: **P2 — Keep the shifted map aligned for intra-block launches**

This assumes `runOrigin.beat` and `runOrigin.seconds` name the same instant in the tempo map. They do for the whole-block case in the new test, but not necessarily for `materialSubBlock`: `LaunchHandle::advanceOver()` projects the split beat and split seconds independently along the block’s straight line. When a launch falls inside a block that crosses a tempo step, the projected pair does not satisfy `tempo->timeToBeat(runOrigin.seconds) == runOrigin.beat`.

A fresh auto-tempo or warped clip can therefore begin with `beatAtTime(0) != 0` and skip its leading material. Using this test file’s 6000-sample/48 kHz block across a centered 120→60 BPM step, the linear split puts beat 4 at 2.020833 s; this lookup then returns about 0.020833 material beat at material second zero—roughly 500 source samples at 120 BPM. The added test ends its block exactly on the step, so it avoids the intra-block split case.

Please either make the split’s beat/seconds pair map-consistent (while preserving the sample/count contract), cut blocks at tempo discontinuities, or represent the additional correction needed by the shifted lookup, and cover a launch inside a block crossing a tempo step.

magda/engine/transport/TempoMap.hpp
```diff
@@ -135,6 +135,26 @@ class TempoMap {
...
+     * That cut is what makes the rest of the engine's arithmetic true. Inside a
+     * block every position is worked out on a straight line between its two
+     * ends: where a launch lands, where a note goes, which instant a run began
+     * on. A straight line is exact within one tempo and, across a jump, out by
+     * hundreds of samples rather than the fraction of one the block's own
+     * documentation promises (exec/RenderContext.hpp).
+     *
+     * Steps only, not every section: a ramp is baked into many short sections
+     * and cutting at each would cut every block it covers, while the kink
+     * between two of them is small enough to stay inside that fraction.
```
lucaromagnoli: **P2 — Cut at baked ramp boundaries too**

A ramp is implemented as constant-tempo sections, so its section boundaries also change the beat/time slope. Excluding them preserves the same inconsistent projection this patch fixes for explicit steps. For example, a valid 20→300 BPM ramp over one beat, with a 512-sample block centered on the first boundary at beat 0.25, places that beat at sample 93 through the block’s linear projection although the tempo map places it at sample 256—an error of 163 samples. A launch there consequently starts early and its `RunOrigin` still does not identify one map instant.

Please cut blocks at every baked tempo-section boundary, or make the split conversion map-consistent. The assumption that ramp-boundary error remains below one sample does not hold for valid steep ramps.

— Codex

PR #2331, opened 2026-09-02
One canonical block coordinate, derived, not projected (#2330)

### LINKED ISSUE #2330: Session launch: one canonical block coordinate, derived, not projected
`SyncRange` carries four axes (timeline beats, monotonic beats, timeline seconds, monotonic seconds) and `LaunchHandle::advanceOver` projects an event onto each one independently by linear interpolation. That yields a single coherent instant only while every mapping is affine across the whole block.

Two review findings on #2325 were the same leak surfacing twice: an explicit tempo step, then a baked ramp-section boundary. Both are now handled by cutting blocks at every tempo section, which makes "this block is affine" a real enforced invariant. That closes the current cases but does not remove the class: a future nonlinear mapping reintroduces it.

## The fix

One canonical event coordinate, sample offset within the block, with every other coordinate derived from it through the tempo map.

- A `BlockInstant` built only by `BlockInfo::instantAtSample` / `instantAtMonotonicBeat`, so a `RunOrigin`'s two faces are one instant by construction rather than by luck.
- `LaunchHandle::advance` then needs the map, which is the signature change at the centre of this.

Once every coordinate derives from one, the section cut becomes scaffolding and this can delete it, with the clock's affine tests proving nothing regressed when it goes.

lucaromagnoli:
Closes #2330. Stacked on #2325, and the structural answer to the two review findings there rather than a third patch on top of them.

## The leak

`SyncRange` carries four axes and `advanceOver` read the event off each of them separately, by linear interpolation. That produces one coherent instant only while every mapping is affine across the whole block, which is why the same defect surfaced twice in review: first at an explicit tempo step, then at a baked ramp-section boundary. Cutting blocks closed both cases without closing the class.

## One coordinate

`BlockInstant` is built from a sample and nothing else. Every other face is derived from that one number:

- the two seconds faces exactly, since a block runs at one second per sample rate whatever the tempo does;
- the timeline beat through the map, rather than along the block's own straight line;
- the monotonic beat off the timeline beat, which is exact inside a block because what makes the two differ is a wrap, and a wrap ends a block.

`SyncRange` carries the map and the sample count so it can build one. A launch quantized to a beat is snapped to the sample it sounds on *before* any face is read: a launch happens on a sample or it does not happen, and the faces of the beat that asked belong to a moment between two samples that nothing plays.

## What it buys

The origin. A fresh run's `RunOrigin` is now the two faces of one sample instead of two projections that agree only under one tempo, so material second zero is material beat zero and an auto-tempo clip starts at the head of its file.

On a block spanning a 120 to 60 step the old pair put the run's origin at beat 4 and 2.020833 s, where the map puts beat 4 at 2. The material began a fiftieth of a beat in, about 500 samples at 120 bpm, and that much of every launched clip's attack was skipped.

`SyncPoint` goes, having been the three-faced version of the same idea. `materialSubBlock` crops by samples rather than by a beat read off the block's line, and `splitSample` reads the instant the launcher already resolved instead of working the position out a second time and differently.

## On the section cut

#2325's cut stays. It is no longer what makes the launch path correct, but it is still what keeps `sampleForBeat` honest for everything else placed inside a block, MIDI note placement above all. Removing it is a separate question with its own evidence, and the clock's affine tests are there to answer it when someone asks.

## Testing

The new case asserts the invariant on a block built by hand to span a step, which is a shape the clock no longer produces, so it holds whatever the clock does. Under the old projection it fails at 4.020833 against 4, and `beatAtTime(0)` at 0.020833 against 0.

Full Catch2 suite green locally: 3376 passed, 13 skipped for plugins absent from this machine, 0 failures. The JUCE suite passes, including the real-project corpus through both engines, and the app builds.

### DIFF
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -104,25 +104,26 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
 
     // The status is the launcher's, worked out before anything rendered
     // (SessionLauncher.hpp), so this source and the MIDI one see the same block.
-    forEachSlot(
-        *handles.get(), track, [&](const SessionSlotPlayback& slot, const SplitStatus& status) {
-            const auto play = [&](const BlockPiece& piece, int offset, int count) {
-                if (count <= 0 || !piece.origin)
-                    return;
+    forEachSlot(*handles.get(), track,
+                [&](const SessionSlotPlayback& slot, const SplitStatus& status) {
+                    const auto play = [&](const BlockPiece& piece, int offset, int count) {
+                        if (count <= 0 || !piece.origin)
+                            return;
 
-                gather(slot.audio, materialSubBlock(block, piece.range, *piece.origin, count),
-                       offset, streams, sounding, soundingCount);
-            };
+                        gather(slot.audio,
+                               materialSubBlock(block, piece.range, *piece.origin, offset, count),
+                               offset, streams, sounding, soundingCount);
+                    };
 
-            // What is on the far side of the event is not rendered. The ramp
-            // across that boundary is #2302's.
-            const auto split = splitSample(block, status);
+                    // What is on the far side of the event is not rendered. The ramp
+                    // across that boundary is #2302's.
+                    const auto split = splitSample(block, status);
 
-            play(status.beforeEvent, 0, split);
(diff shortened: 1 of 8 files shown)


lucaromagnoli:
Closed by the base branch being deleted when #2325 merged, not by anything wrong with it. Continued in #2334: same branch, same commits, rebased onto `feat/2301-session-playback`.

Issue #2333, opened 2026-09-02
Drop the tempo-section block cut once sampleForBeat goes through the map


lucaromagnoli:
Depends on #2332.

#2325 made `TransportClock` end a segment at every tempo-section boundary, so no block spans one. That was needed because everything placed inside a block is put on the straight line between its two ends, and across a section boundary that line is a guess: over a 20 to 300 bpm ramp a beat the map puts at sample 256 of a block lands at sample 93 on the line, 163 samples out.

It is a workaround for `sampleForBeat` interpolating rather than deriving. Once a block carries an absolute sample range and `sampleForBeat` answers through the tempo map (#2332), the line stops being load-bearing and the cut stops paying for itself: it costs an extra segment per tempo change and, on a ramp, one per baked section the callback crosses.

The affine tests added with the cut are what should decide this. `A steep ramp leaves every block affine to within a sample` asserts the property the cut exists to provide, without asserting the cut itself, so it should keep passing once the derivation replaces the cutting. If it does, the cut goes.

Worth confirming rather than assuming: the cut also keeps a block inside one time signature, which the metronome's grid may be relying on.

https://claude.ai/code/session_013CDpSHas8bog33vv4o6HLv

lucaromagnoli:
Folded into #2336, which asks the same question as one of its steps: decide whether the tempo-section callback cuts remain necessary once no consumer depends on an affine beat projection.

The note worth keeping: `A steep ramp leaves every block affine to within a sample` in test_transport_clock.cpp asserts the property rather than the cut, so it is the test that decides whether the cut can go. And the cut incidentally keeps a block inside one time signature, which the metronome grid may rely on.

Issue #2332, opened 2026-09-02
The sample is the authority at the transport and session boundary


lucaromagnoli:
Most of the engine is already sample-authoritative: render buffers, graph execution, latency and delay lines, parameter segments, prefetching, file reads, clip cells and source positions all end up in integer samples. The gap is the transport and session boundary, where beats are still treated as a coordinate rather than as musical intent.

#2331 made every face of an instant derive from one sample, which closed the class of independently-projected origins. It did not make the sample durable, and that is the next thing.

## What is still wrong

- `BlockInfo` has no absolute sample range, only `numSamples` plus beat and second ranges. `TransportClock::samplesSinceAnchor_` exists but resets on a re-anchor and is not exposed, so there is no durable coordinate to hand anyone.
- `sampleForBeat` interpolates between the block's beat endpoints instead of deriving the sample through the tempo map. That is why the clock has to cut a segment at every tempo section: an affine-block workaround standing in for sample authority.
- Session state keeps redundant authorities: `RunOrigin{beat, seconds}`, `played_`, `playedMonotonic_`, `playedMonotonicSeconds_`. `nudge()` takes independent beat and second deltas. These can disagree again.
- `BlockInstant` does not enforce the invariant it claims. Its fields are public, and `applyEvent` builds one and then overwrites three derived faces with another handle's origin for a synced launch.
- `beginRun` converts the sampled instant back into beat and second ranges, so retriggers are scheduled from an already-rounded beat and the error accumulates. At 48 kHz and 123 bpm, a thousand one-beat retriggers finish about 366 samples late.

## The shape of the fix

1. Give every `BlockInfo` an `int64_t` monotonic transport-sample range.
2. Store `RunOrigin` as that absolute sample position.
3. Keep beats as musical intent only: quantization target, loop length, model placement.
4. Resolve an intended beat once to an event sample, and feed the rounded derived beat back into scheduling rather than re-resolving from the original beat.
5. Separate the two conversions properly: an event sample is `0..N-1`, a range edge is `0..N`. #2331 has a first cut of this in `eventAtMonotonicBeat`; it wants to be the general rule, not a launcher-local one.
6. Derive the beat and second views from the sample and the block, rather than storing mutable copies on `BlockInstant`.

Not a codebase-wide rewrite. The migration surface is `TransportClock` to `BlockInfo` to `LaunchHandle` and `SessionPlayback`, then consolidating the existing beat-to-sample consumers.

[...]

PR #2334, opened 2026-09-02
One canonical block coordinate, derived, not projected (#2330)

### LINKED ISSUE #2330: Session launch: one canonical block coordinate, derived, not projected

lucaromagnoli:
Closes #2330. Replaces #2331, which GitHub closed when its base branch was deleted on merging #2325; same branch, same commits, rebased onto the merged base.

The structural answer to the two review findings on #2325 rather than a third patch on top of them.

## The leak

`SyncRange` carries four axes and `advanceOver` read the event off each of them separately, by linear interpolation. That produces one coherent instant only while every mapping is affine across the whole block, which is why the same defect surfaced twice in review: first at an explicit tempo step, then at a baked ramp-section boundary. Cutting blocks closed both cases without closing the class.

## One coordinate

`BlockInstant` is built from a sample and nothing else. Every other face is derived from that one number:

- the two seconds faces exactly, since a block runs at one second per sample rate whatever the tempo does;
- the timeline beat through the map, rather than along the block's own straight line;
- the monotonic beat off the timeline beat, which is exact inside a block because what makes the two differ is a wrap, and a wrap ends a block.

`SyncRange` carries the map and the sample count so it can build one. A launch quantized to a beat is snapped to the sample it sounds on *before* any face is read: a launch happens on a sample or it does not happen.

An event and a range edge are separate questions and get separate conversions. `atSample` answers for an edge, where one past the end of the block is legal; `eventAtMonotonicBeat` never answers past the block's last sample, because an event placed there is written nowhere and a stop would clear its note state while its note-offs went outside the buffer.

## What it buys

The origin. A fresh run's `RunOrigin` is now the two faces of one sample instead of two projections that agree only under one tempo, so material second zero is material beat zero and an auto-tempo clip starts at the head of its file.

On a block spanning a 120 to 60 step the old pair put the run's origin at beat 4 and 2.020833 s, where the map puts beat 4 at 2. The material began a fiftieth of a beat in, about 500 samples at 120 bpm, and that much of every launched clip's attack was skipped.

`SyncPoint` goes, having been the three-faced version of the same idea. `materialSubBlock` crops by samples rather than by a beat read off the block's line, and `splitSample` reads the instant the launcher already resolved.

## What this does not do

It makes the sample canonical *within a block*; it does not make it durable. The origin is still stored as beat and seconds, so a retrigger is scheduled from an already-rounded beat and the error accumulates: about 366 samples over a thousand one-beat retriggers at 48 kHz and 123 bpm. `BlockInstant`'s fields are public and `applyEvent` overwrites three of them for a synced launch, so the invariant is a convention rather than an enforcement.

Both are #2332, along with the absolute monotonic sample range on `BlockInfo` that they need. #2333 is the follow-on: once `sampleForBeat` derives through the map, #2325's tempo-section block cut stops paying for itself.

## Also here

The first step of #2332, added after the review that prompted it: the transport now counts its own samples and every block carries an absolute `monotonicSamples` range. Additive, nothing reads it yet, but it is the durable coordinate the origin has to move onto and it can be verified on its own. A locate, a loop wrap, a tempo edit and a re-anchor all leave it alone.

## Testing

The new case asserts the invariant on a block built by hand to span a step, which is a shape the clock no longer produces, so it holds whatever the clock does. Under the old projection it fails at 4.020833 against 4, and `beatAtTime(0)` at 0.020833 against 0. The event-clamp case fails at 24000 against 23999.

Full Catch2 suite green locally after the rebase: 3377 passed, 13 skipped for plugins absent from this machine, 0 failures.

### DIFF
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -104,25 +104,26 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
 
     // The status is the launcher's, worked out before anything rendered
     // (SessionLauncher.hpp), so this source and the MIDI one see the same block.
-    forEachSlot(
-        *handles.get(), track, [&](const SessionSlotPlayback& slot, const SplitStatus& status) {
-            const auto play = [&](const BlockPiece& piece, int offset, int count) {
-                if (count <= 0 || !piece.origin)
-                    return;
+    forEachSlot(*handles.get(), track,
+                [&](const SessionSlotPlayback& slot, const SplitStatus& status) {
+                    const auto play = [&](const BlockPiece& piece, int offset, int count) {
+                        if (count <= 0 || !piece.origin)
+                            return;
 
-                gather(slot.audio, materialSubBlock(block, piece.range, *piece.origin, count),
-                       offset, streams, sounding, soundingCount);
-            };
+                        gather(slot.audio,
+                               materialSubBlock(block, piece.range, *piece.origin, offset, count),
+                               offset, streams, sounding, soundingCount);
+                    };
 
-            // What is on the far side of the event is not rendered. The ramp
-            // across that boundary is #2302's.
-            const auto split = splitSample(block, status);
+                    // What is on the far side of the event is not rendered. The ramp
+                    // across that boundary is #2302's.
+                    const auto split = splitSample(block, status);
 
-            play(status.beforeEvent, 0, split);
+                    play(status.beforeEvent, 0, split);
 
-            if (status.afterEvent)
-                play(*status.afterEvent, split, block.numSamples - split);
-        });
+                    if (status.afterEvent)
+                        play(*status.afterEvent, split, block.numSamples - split);
+                });
 }
 
 void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float> out) {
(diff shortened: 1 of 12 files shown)


lucaromagnoli:
Note for whoever picks up #2336: the event clamp in `eventAtMonotonicBeat` here is a stopgap. It stops the stuck note by capping an event at the block's last sample, which fires it up to one sample early. #2336 has the correct semantics: an event resolving to `numSamples` belongs at sample zero of the next callback.

Issue #2239, opened 2026-08-30
magda-sdk: the JUCE-free core (schemas, DSP and UI cores, C ABI, hosts) as its own repo before 1.0


lucaromagnoli:
## Why

The device SDK boundary exists in-repo today: `magda_device_sdk` (INTERFACE) and `magda_devices` (STATIC) in `magda/daw/CMakeLists.txt`, from the decouple epic (#1836). The second half of that epic, moving the library to its own repo, has not happened, and meanwhile the set of things that want to be shared has grown past "devices": other projects, native and browser, are already copying code out of this repo by hand (analysis front ends, curve handling, waveform peaks) and re-implementing the same UI behaviour twice.

1.0 is the point where these pieces become contracts that other products depend on. They should be in one place, JUCE-free by construction, before the tag.

## What

A new repo, `magda-sdk`. `magda-core` stays the DAW, `magda-cloud` stays the web; no renames. C++ namespace stays `magda::`.

Consumed by `magda-core` as a submodule (the pattern already used for Faust, eurorack and the TE fork), by other native projects the same way, and by browser hosts as an npm package built from its wasm target.

## Coverage

**Schemas (JSON contracts)**
- Curve: points, interpolation kinds, shape parameters, domain and range. One format for every curve editor on every surface.
- Device and patch state, ValueTree-free (the state contract the native engine work already calls for).
- Parameter manifest: id, name, range, default, skew, units. Generated once, consumed by every host.
- Preset envelope and metadata.

**DSP cores (JUCE-free, allocation-free on process paths)**
- Primitives: biquad, SVF, envelopes, mipmapped wavetables, RNG, math, expression evaluation, SPSC queue, triple buffer.
- Curve evaluation at audio rate from the schema.
- Analysis: FFT, log-mel front end, onset, pitch tracking, envelope extraction, spectral tilt, transient separation. The measurement code behind Levels and masking lives here too.
- Peaks and thumbnails: min/max per pixel from a buffer.

**UI cores (logic and display lists, no pixels)**
- Curve editor: interaction state machine, hit-testing, snapping, undo grouping, `render()` to a display list.
- Waveform view: zoom, selection, playhead.
- Meter models: ballistics, peak hold, clip detection.
- The display list format and the colour-role vocabulary shells map onto a theme.

**C ABI**
- `create / prepare / process / set_param / midi / get_state / set_state / analyze / destroy`. The door every host uses.

**Hosts (the only directories allowed to depend on anything)**
[...]

PR #2335, opened 2026-09-02
Port the step sequencers to MagdaDevice with pattern state in the model (#2313)

### LINKED ISSUE #2313: Port the step sequencers to MagdaDevice with pattern state in the model

lucaromagnoli:
Closes #2313. Part of #2299 — the last two `MidiDevicePlugin` subclasses cross to `MagdaDevice`, and the base they shared is deleted with them.

## The pattern belongs to the model

A step pattern is authored state, the same category as MIDI notes, so it moves out of the device and into `DeviceInfo::pluginState` under the device-state v2 schema — keeping the `STEP` / `NOTE` element names saved projects already carry, so a project written by any released build loads unchanged.

- `core/StepPatternState` is the adapter between those elements and the pattern the sequencers play.
- `core/StepPatternCommands` makes every edit one undoable whole-pattern snapshot. A pattern is small, so one command covers every edit either surface makes — and it is the cheaper undo, since step toggles are exactly what users spam and expect to undo in one go. A continuous gesture (a velocity or probability drag) says so, and its run of edits coalesces into the entry that undoes back to where the drag began.
- The engine never learns that undo exists: it is handed the pattern with the rest of its state and plays what it is given.

## A sequencing core of its own

The sequencing moves out of the devices into `magda/daw/audio/sequencer/`, on the engine-neutral pack's source list so the existing configure-time validator rejects a Tracktion reference in it (#2239's direction):

| | |
|---|---|
| `StepClock` | moved from `transport/`, now takes a `BlockTiming` rather than asking a host for a tempo sequence |
| `StepPattern.hpp` | the patterns as plain data — no device, no ValueTree, no engine |
| `MonoStepSequencer` / `PolyStepSequencer` | `(pattern, params, block) → note events`, with the tie / glide / gate / probability rules |
| `NoteSink` | plain `{time, note, velocity, on/off}` records, so the core names no MIDI container |

The device shells are what is left: slots in, core events into a `DeviceMidiBuffer` out, MIDI thru around them.

## Two defects the extraction made visible

Each has a test that fails without its fix.

- **Swing above roughly 8% silently dropped half the pattern.** A tick swung past the end of the block that scheduled it was never emitted, while the step cursor advanced regardless. Such a tick is now held for the block that contains it.
- **A tie was cut after about four blocks** — 0.17 s at 48 kHz and a 512-sample buffer, so shorter than a tie at 120 bpm. A held tie has a sounding note and nothing scheduled to release it, which is exactly the shape the stuck-note guard exists to catch; it is now exempt, and the step that ends the tie is what releases the note.

## Faceplates, agents and export

Both faceplates are rewritten against the model: slot values in through `updateFromParameters`, slot edits out through `onParameterChanged`, the pattern in through `setPattern` and every edit out through `onPatternEdited`. They poll the device's published pattern — the model's projection — so an undo, an agent write or a preset load shows up without the faceplate knowing where the edit came from. Step recording cannot write the model from the audio thread, so the device records what it heard and the faceplate commits it.

The agent and MCP write paths and the clip export read and write the model too, which is what lets an export work on a device the engine has not instantiated.

## Frozen surfaces held

Slot ids, order and display ranges are the ones the retired plugins registered, so saved links and saved values load unchanged — `Device Param Schema Freeze` passes with no fixture edit. The persisted property names are unchanged, and `Legacy Corpus Device State Migration` and `Device State Schema` pass.

## Deferred

The clip export reads the model but still walks the pattern itself rather than driving the engine offline, so an exported clip ignores swing and time bend. With the core extracted that is now a small piece of work rather than a duplication of the stepping rules.

## Tests

- `magda_tests`: 3400 pass — 18 new for the sequencing core's rules, 9 new for the pattern in the model's document.
- `MIDI Signal Routing Tests`, `Device Param Schema Freeze`, `Device State Schema`, `Legacy Corpus Device State Migration`: pass.
- `Real Project Corpus` fails 2 on my machine (`v1bounces` asserts in `tracktion_DeviceManager.cpp:432`, `demo` misses a spectral calibration). Neither project contains a step sequencer and nothing in the render path changed, so this is not from this branch — worth a look separately.


magda/daw/audio/plugins/StepSequencerPlugin.cpp
```diff
@@ -1,628 +1,380 @@
...
-        if (noteOffCountdown_ <= 0) {
-            // Note-off falls within this block
-            double offTimeInBlock =
-                evt.timeInBlock + static_cast<double>(gateSamples) / sampleRate_;
-            offTimeInBlock = std::min(offTimeInBlock, blockDurationSecs);
-            killNote(midi, offTimeInBlock);
+    // --- Hold incoming MIDI aside, so the sequencer owns the buffer ---
+    int thruCount = 0;
+    if (midiThru.load(std::memory_order_relaxed)) {
+        thruCount = std::min(midi.size(), kMaxThruMessages);
```
lucaromagnoli: This truncates MIDI-through to 64 events and then clears the source buffer, so every later event is silently lost. Dense controller data or larger/offline blocks can exceed this, and a discarded note-off can leave a downstream instrument stuck. The engine-side MIDI contract permits roughly 450 short events, so please preserve the full input bound (the poly sequencer has the same truncation).

magda/daw/audio/plugins/StepSequencerPlugin.cpp
```diff
@@ -1,628 +1,380 @@
...
+        step.noteNumber = std::clamp(static_cast<int>(child.getProperty(kStepNote, 60)), 0, 127);
+        step.octaveShift = std::clamp(static_cast<int>(child.getProperty(kStepOctave, 0)), -2, 2);
+        step.gate = child.getProperty(kStepGate, true);
+        step.accent = child.getProperty(kStepAccent, false);
+        step.glide = child.getProperty(kStepGlide, false);
+        step.tie = child.getProperty(kStepTie, false);
     }
+
+    // Fill the slot the audio thread is not reading, then hand it over.
+    const int next = 1 - livePattern_.load(std::memory_order_relaxed);
```
lucaromagnoli: The atomic index does not prove that this slot is no longer being read. If audio has retained a reference to slot 0, two rapid restoreState calls publish slot 1 and then rewrite slot 0 while that callback is still using it. Continuous pattern edits make that sequence realistic, producing a C++ data race and corrupted playback. This needs reader acknowledgement or an immutable/thread-safe publication strategy; the poly implementation has the same issue.

magda/daw/core/StepPatternCommands.cpp
```diff
@@ -0,0 +1,242 @@
...
+    return description_;
+}
+
+bool SetMonoStepPatternCommand::canExecute() const {
+    return editableSequencer(devicePath_, kMonoSequencerId) != nullptr;
+}
+
+bool SetMonoStepPatternCommand::canMergeWith(const UndoableCommand* other) const {
+    const auto* next = dynamic_cast<const SetMonoStepPatternCommand*>(other);
+    return next != nullptr && next->gesture_ == StepPatternGesture::Continuous &&
```
lucaromagnoli: This has no boundary between distinct continuous gestures. Every later continuous command with the same path and description merges into the previous one, while mouseUp only clears the UI drag flag. Two separate velocity or probability drags therefore become one undo entry if no other undoable command intervenes. Please include a per-gesture identity or explicitly terminate merging at mouse up/down.

magda/daw/audio/plugins/StepSequencerPlugin.hpp
```diff
@@ -1,169 +1,174 @@
...
+    /// Incoming MIDI held back while the sequencer writes its own notes, then
+    /// appended so it still reaches the instrument downstream. A member so the
+    /// block allocates nothing.
+    ///
+    /// Sized to the whole input bound rather than a convenient number: a MIDI
+    /// port's budget is 4096 bytes and a short message costs nine, so about
+    /// 450 notes and controller changes can reach one block. Truncating below
+    /// that dropped the tail of a dense block silently, and a note-off lost
+    /// that way leaves the instrument downstream holding the note (#2335).
+    static constexpr int kMaxThruMessages = 512;
```
lucaromagnoli: **[P1] Size MIDI thru for the cheapest legal event**

This still does not cover the engine's full input bound. `EngineMagdaDevice::midiEventsWithin()` deliberately sizes the input scratch as `4096 / (overhead + 1)`, because one-byte real-time messages cost 7 bytes rather than the 9 bytes used by a note; that admits 585 events. For example, 511 timing messages followed by several note messages is still below 4096 bytes, but a tail note-off beyond index 511 is discarded by the `std::min`, recreating the stuck-note case this change is meant to close. The Tracktion adapter has no count bound at all. Please avoid truncating the existing buffer when thru is enabled (or derive this capacity from a shared, engine-neutral maximum that covers the cheapest event).

magda/daw/audio/sequencer/StepClock.cpp
```diff
@@ -201,6 +204,36 @@ int StepClock::processBlock(const te::PluginRenderContext& fc, te::Edit& edit, R
...
+    for (int i = 0; i < pendingCount_; ++i) {
+        const auto step = pending_[static_cast<size_t>(i)];
+        if (step.beat < blockEndBeat && eventCount < maxEvents)
+            emit(step.stepIndex, step.beat);
+        else
+            pending_[static_cast<size_t>(stillPending++)] = step;
+    }
+    pendingCount_ = stillPending;
+
     while (nextStepBeat_ < blockEndBeat && eventCount < maxEvents) {
```
lucaromagnoli: **[P1] Reconcile the clock position when the pattern shrinks**

`sequenceStep_` is emitted before it is wrapped to the current `numSteps`. If playback is at step 20 of a 32-step pattern and the user shortens it to 8, the next event still plays hidden step 20; in PingPong mode `advanceStep(20, 8, ...)` then returns 19, 18, etc., so inactive steps keep sounding until the index eventually reaches 7 (Forward leaks one hidden step as well). Pending events can likewise retain indices removed by the shrink. Please normalize the sequence/cycle position and reconcile pending entries when `numSteps` decreases; a live-shrink test in each direction would catch this.

tests/midi/test_step_sequencer_midi_thru.cpp
```diff
@@ -0,0 +1,147 @@
...
+    };
+    device.process(context);
+}
+
+}  // namespace
+
+TEST_CASE("A step sequencer passes a dense block through whole", "[sequencer][midithru]") {
+    // 600 events: past both the 64 the device used to keep and the ~450 a
+    // note-sized division of the byte budget suggests.
+    constexpr int kClocks = 600;
```
lucaromagnoli: **[P1] Exercise the real output bound; this payload is still truncated there**

`TestMidiBuffer` is unbounded, but the native adapter rewrites the post-device buffer with a hard `kMaxMidiBytesPerPort` check and breaks at 4096 bytes (`EngineMagdaDevice.cpp:402-406`). These 600 clocks plus note-on/off cost 4218 bytes, so when such an input arrives from a merge (the case cited above), the real adapter still stops before the final note-off; leaving the events in place merely moves the truncation downstream. With playback enabled, the sequencer's own notes are appended after the input and are the first events refused as well. Please test this through `EngineMagdaDevice` and resolve the input/output-bound contract (or an explicit overflow policy) so the asserted pass-through guarantee holds in the actual engine.

magda/daw/audio/sequencer/StepClock.cpp
```diff
@@ -201,6 +222,36 @@ int StepClock::processBlock(const te::PluginRenderContext& fc, te::Edit& edit, R
...
```
lucaromagnoli: **[P2] Preserve ordinary due ticks when the event array is full**

The new capacity test preserves a *previously pending* step, but it still loses every ordinary tick due in the squeezed block. In that test, step 2 consumes the sole output slot, this loop never runs, and `nextStepBeat_` remains at beat 1; on the following block the catch-up loop advances through beats 1, 1.25, 1.5, and 1.75 without emitting them. The test only checks that pending step 3 survived, so those four missing steps go unnoticed. This is reachable with a large/offline block exceeding the sequencers' 16-event array. Please retain overdue clock ticks too, or otherwise guarantee and enforce that `maxEvents` can cover the block.

Issue #2336, opened 2026-09-02
Engine time: make samples authoritative for runtime operations


lucaromagnoli:
## Problem

The engine still treats beats, seconds, and samples as competing runtime authorities.

#2330 and #2331 expose the immediate symptom: a launch is resolved to a sample, but `LaunchHandle::beginRun` discards that coordinate and persists the run as beat/seconds ranges. Operations later derive new event positions from those rounded values.

This is observably unstable. At 48 kHz and 123 BPM, one beat is 23,414.634 samples. If each one-beat retrigger is scheduled from the previous snapped beat, every interval becomes 23,415 samples; after 1,000 retriggers the run is 366 samples late.

The same missing primitive appears outside session playback:

- `BlockInfo` has a sample count but no absolute monotonic sample range.
- `sampleForBeat()` linearly projects block endpoints, so the clock must cut callbacks at every tempo section to keep that projection valid.
- event positions and half-open range edges use different valid sample domains, but plain integers do not encode the difference; #2331 can resolve an event to `numSamples` and apply it in the wrong callback.
- arrangement snapshots carry beats plus derived seconds and a tempo fingerprint, but playback does not enforce that the snapshot and transport maps match.

## Goal

Samples are authoritative for engine operations. Beats remain the authority for musical intent—model placement, quantization, tempo-synced durations—and seconds remain a derived view where wall-clock or source-time semantics require one. Once an operation is resolved to a sample, derived beat/second values must never become a new operational authority.

## Direction

- Add strong 64-bit sample-position/range types.
- Give every `BlockInfo` an absolute monotonic transport-sample range that does not rewind on loops or locates and advances according to the transport's rolling semantics.
- Represent a persistent `RunOrigin` with that sample position, not a beat/seconds pair.
- Keep an exact musical scheduling phase separately where a beat-quantized or beat-length operation requires it; do not feed a rounded derived beat back into the next schedule.
- Make `BlockInstant` retain only its canonical sample identity. Expose derived faces through accessors requiring the block/map context rather than public mutable fields.
- Centralise beat/time-to-sample conversion at the block boundary.
- Distinguish an event sample (`0..N-1`) from a half-open range edge (`0..N`). An event resolving to `N` belongs to sample zero of the next block.
- Migrate session launch, MIDI placement, automation, click placement, and arrangement playback to the shared conversion.
- Decide whether tempo-section callback cuts remain necessary once no consumer depends on an affine beat projection.
[...]

PR #2338, opened 2026-09-03
Count a snapshot rendered against a map it was not compiled for (#2337)

### LINKED ISSUE #2337: A clip snapshot must not render against a tempo map it was not compiled for
Split out of #2336, where this is the last acceptance criterion. Separable from the rest: it needs none of the sample-authority work and can land ahead of it.

## The gap

`ClipSnapshot` carries `tempoFingerprint`, the map its seconds were derived through. Its own comment says what a mismatch means:

> A snapshot whose fingerprint is not the transport's was compiled against a tempo that has since changed, and every second in it is wrong by however much the map moved.

Nothing compares it. The only reads in the tree are one test assertion in `test_clip_snapshot.cpp`; the compiler writes it and playback ignores it. The field documents an invariant it does not enforce.

The exposure is a tempo edit while the transport rolls. `TransportClock::followTempo` re-anchors on a fingerprint change, keeping the beat and recomputing seconds, so the transport moves onto the new map immediately. Every span and fade in the published snapshot is still in seconds derived from the old one until a recompile is published. The error is not drift; it is a discrete position error for the width of that window, scaling with how far into the timeline the cursor is and how large the tempo change was. A tempo change at bar 200 is worse than one at bar 2.

## Latent, not live

`EngineSession::publishClips` is called only from tests today. The native clip-snapshot path is not wired to the app, so nothing in production recompiles or publishes on a tempo edit and no user hits this yet.

That is the argument for doing it now rather than later: the contract should exist before a caller does, so the wiring is written against an enforced invariant instead of inheriting a silent one.

## Direction

- Compare the published snapshot's `tempoFingerprint` against the transport's map at the point of render. `BlockInfo::tempo` is already there, so `block.tempo->fingerprint()` is available without new plumbing. The clip sources do not currently read `block.tempo` at all.
- Decide what a mismatch does. Rendering stale seconds silently is the one option to rule out. Candidates: hold the previous coherent snapshot, render silence for the affected tracks, or treat it as a publish-ordering bug and make it impossible by construction.
- Prefer making it impossible: publish the snapshot and the transport map as one atomic unit, so a mismatch cannot be observed rather than being detected after the fact. Worth weighing against the cost on the publishing thread.
- Whichever way, count it. A mismatch that resolves itself in one block is different from one that persists, and silence about it is what makes a publish-ordering bug look like an audio-device fault.

## Acceptance

- A snapshot compiled against one map cannot render spans while the transport is on another.
- The case is a tempo edit mid-playback with a snapshot published late, asserting the chosen behaviour rather than the incidental one.
- Whatever the mismatch does, it is observable rather than silent.

Related: #2336.

lucaromagnoli:
Closes #2337. Stacked on #2318 because it touches the clip sources, which that stack rewrites; based on `dev/0.20.0` it would have been written against types #2318 replaces.

## The gap

`ClipSnapshot` carries `tempoFingerprint`, the map its seconds were derived through, and its own comment says what a mismatch means:

> A snapshot whose fingerprint is not the transport's was compiled against a tempo that has since changed, and every second in it is wrong by however much the map moved.

Nothing compared it. The only reads in the tree were one assertion in `test_clip_snapshot.cpp`. The field documented an invariant it did not enforce.

## Counted, not acted on

Silence was the first thing I wrote, on the reasoning that the path should be unreachable so its behaviour did not matter. That is a bad bet on a should. If the publish ordering is ever slightly wrong, the failure it chooses is an audible hole in the middle of a set, and it buys little: the stale spans usually stop overlapping the block anyway, so refusing them mostly turns an accidental gap into a deliberate one.

Re-deriving the seconds from the beats was the other candidate. The beats do survive a tempo edit, since an edit moves when a beat happens rather than which beat is playing, so it would even be correct. It is also compiling on the audio thread, which is the one thing a snapshot exists to have already done.

So it counts and plays. The count is the deliverable: it turns an invisible wrong-position bug into a visible publish-ordering one at the moment someone writes the wiring.

## Coarser than the question, on purpose

The fingerprint is the whole map. An arrangement clip's seconds depend on the map at its placement; a session slot's depend on it only over beats zero to its length, because a slot compiles at the origin (`setPlacementBeats(0.0, lengthBeats)`). A tempo change at bar 200 moves neither a slot nor a clip before it, and still changes the fingerprint.

So the count has false positives by construction. Tolerable for a counter, and it would not have been for a behaviour, which is the second reason this does not act. The comment in both sources says so rather than leaving the next reader to work it out.

## Fixtures that were already lying

Four cases built a block carrying a real map and published a snapshot fingerprinted zero. Nothing could notice before. `AudioRig::publish` now takes the map its blocks will carry and stamps it.

## Latent

`EngineSession::publishClips` is called only from tests, so nothing in the app compiles or publishes a snapshot on a tempo edit yet, and no user can reach this. That is the argument for the count existing before the caller does, rather than a reason to defer it.

The fix for a live tempo change is still the atomic publish in #2337's direction: swap the map and the snapshot compiled for it together, so the window never exists. Whoever does it should decide whether invalidation stays whole-map or becomes per-clip over the range its seconds actually came from.

## Testing

Three cases, each verified to fail with the check disabled: a mismatched snapshot is counted, a matching one is not and still sounds, and a mismatch does not cost a MIDI note already sounding.

Full Catch2 suite green locally: 3383 passed, 13 skipped for plugins absent from this machine, 0 failures. The JUCE suite passes and the app builds.

### DIFF
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -159,6 +159,31 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
     if (!snapshot)
         return silence();
 
+    // Compiled against a tempo map that has since changed: every second in it
+    // is wrong by however much the map moved (ClipSnapshot.hpp).
+    //
+    // Counted, and then played anyway. Not because playing it is right, but
+    // because the alternatives are worse in front of an audience: silence is a
+    // hole in the middle of a set, and the stale spans usually stop overlapping
+    // the block anyway, so refusing them mostly turns an accidental gap into a
+    // deliberate one. Re-deriving the seconds from the beats, which do survive
+    // a tempo edit, would be compiling on the audio thread, which is the one
+    // thing a snapshot exists to have already done.
+    //
+    // The count is the point. Zero is the only right answer and reaching it is
+    // the publish's job: the map and the snapshot compiled for it are meant to
+    // swap together, and this says when they did not (#2337).
+    //
+    // Coarser than the question it stands in for, and knowingly. The
+    // fingerprint is the whole map, while an arrangement clip's seconds depend
+    // on the map at its own placement and a session slot's depend on it only
+    // over beats zero to its length, because a slot compiles at the origin
+    // (ClipSnapshotCompiler.cpp). A tempo change at bar 200 moves neither a
+    // slot nor a clip before it, and still changes the fingerprint. Which is
+    // another reason this counts rather than acts.
+    if (block.tempo != nullptr && snapshot->tempoFingerprint != block.tempo->fingerprint())
+        staleSnapshots_.fetch_add(1, std::memory_order_relaxed);
+
     const auto* track = snapshot->find(trackId_);
     if (track == nullptr)
         return silence();
--- a/magda/engine/clip/ClipAudioSource.hpp
+++ b/magda/engine/clip/ClipAudioSource.hpp
@@ -119,6 +119,23 @@ class ClipAudioSource final : public EngineAudioSource {
         return starved_.load(std::memory_order_relaxed);
     }
 
+    /**
+     * @brief Blocks that rendered nothing because the snapshot was stale.
+     *
+     * A snapshot carries the fingerprint of the tempo map its seconds were
+     * derived through. One that is not the transport's was compiled against a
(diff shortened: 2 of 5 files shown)


lucaromagnoli:
Folding back into #2336. This sits inside its scope and the snapshot is about to move off seconds, so the check and its fixtures would be reworked as soon as that lands. Doing it once, there.

PR #2339, opened 2026-09-03
Samples are what runtime operations are in: one origin, one conversion, and no drift (#2336)

### LINKED ISSUE #2336: Engine time: make samples authoritative for runtime operations

lucaromagnoli:
Closes #2336, and with it #2332, #2333 and #2337, which were folded into it. Stacked on #2318 because it rewrites the launcher and the clip sources that stack owns.

## The sample is what plays

Most of the engine was already sample-authoritative. The gap was the transport and session boundary, where beats were still a coordinate rather than musical intent, and #2331 had made every face of an instant derive from one sample without making that sample durable.

**Two things a sample offset inside a block can mean.** An event happens on one sample the block plays, `0` to `N-1`: a note-on, a click, the start of a parameter segment. An edge bounds a stretch the block covers, `0` to `N`: a fade, a region, a hole. The split used to fall along the axis by accident, since beats mostly placed events and seconds mostly placed bounds, so `sampleForBeat` clamped to `N-1` and `sampleForTime` allowed `N`.

A note-off is where the two come apart. It arrives as a beat, so it went through the beat conversion; it is where a note's stretch ends, so it is an edge; and it still has to be written into this buffer, so an edge at `N` has to become `N-1` or the note hangs. That rule was an unnamed clamp and a comment in one test. It is `BlockInfo::soundsAt` now, and `EventSample` and `EdgeSample` are what stop the next consumer getting it wrong: a caller that uses an event as a range end drops that event's sample, one that emits at an edge writes past the buffer, and both are compile errors.

**Events floor.** A block covers a half-open stretch, so a moment inside it is somewhere in `[0, N)` and the sample it is inside is `0..N-1`, always, with no case to clamp away. Nearest has one: a moment in the last half sample rounds to `N`, a buffer index this block does not have. The block that owns that sample is the next one, where the same moment is sample zero, so no event is emitted at `numSamples` and nothing has to be carried to get it there. Guarded by a hundredth of a sample, for the reason the clock guards its own ceiling: a beat asked for on a shifted block has had an origin taken off it and put back on, and `(a - o) + o` is not `a`.

Both conversions go through the map rather than along the block's own line. Blocks carry the rate they run at so nothing else is needed, and a block assembled from beats alone still places along its own line, which is every resolver fixture in the tree.

## A run begins on a sample

Three authorities collapse into one. `played_`, `playedMonotonic_` and `playedMonotonicSeconds_` were three ranges maintained in parallel, and #2331 had only made them agree at the instant of the launch. What a run keeps now is an origin, which is a sample, plus the one face a map cannot recover afterwards: the monotonic beat that sample fell on. A map converts a position, not an elapsed duration, and after a wrap the origin is not a position on this cycle at all. Everything else comes off the samples, and elapsed seconds is a subtraction of two counts rather than an accumulation of per-block seconds.

The type carrying a run between blocks was doing two jobs under one name. What the launcher keeps is where a run began, durably; what a block carries is what its axes were shifted by, per block and derived. Those are `RunOrigin`, private to the handle, and `MaterialOrigin`, which is what a source is handed.

`BlockInstant` keeps only its canonical identity: the sample it sounds on, and the same sample on the transport's count. Every face is derived from the block that knows the map, and the split is `upTo` and `from` rather than two range literals assembled by hand. That deletes what #2332 named: `applyEvent` used to build an instant and then overwrite three of its faces with another handle's origin. A synced launch adopts the whole run now, which is what it always meant.

## The drift

The wrap was already computed phase-exactly from the origin. The bug was that a retrigger re-anchored that origin to the snapped instant, so every interval restarted from a rounded beat. The fix is that the schedule and the sound are different questions: a wrap sounds on a sample, because that is the only thing that can sound, and the next wrap is counted from the beat the run was asked for, which no rounding has touched.

The regression is a thousand one-beat retriggers at 48 kHz and 123 bpm, where a beat is 23,414.634 samples. Every wrap lands within a sample of where the beat is. Re-anchoring instead, which is what this replaces, puts the thousandth 634 samples adrift and fails the case from about the second wrap on.

## A rate change has semantics

A run's origin is a count of samples and a rate change changes what a sample is worth, so holding the number would move the run through its material: a clip a second into its file would be half a second in, or two. The origin is re-counted behind the block at the new rate, keeping the time the run has actually had. That depends on the handle seeing every block, which is an invariant rather than a hope, and the comment says which one.

`BlockInfo::monotonicSeconds` stays, with its comment corrected. It is the one thing a sample count cannot answer, because a count spanning a rate change counts samples that were not all worth the same amount of time. What it no longer is, is how a run measures itself.

## The section cut stays, for a reason it did not have before

#2325 cut a segment at every tempo-section boundary because everything inside a block was placed on the line between its two ends. Placement goes through the map now, so that justification is gone and the question was whether anything else was paying for it.

Something is. `ModLfo::modTimingFor` takes a single bpm and signature for the whole block it renders, so a block spanning a step would run at the tempo it opened on for the rest of itself. That is asserted now rather than assumed: every sample a block renders is at one tempo, over a ramp cut four ways to the beat. The end beat is excluded and has to be, since a cut lands on the first sample at or after the boundary and can sit a fraction of a sample past it.

Two consumer cases decide the placement half, both on blocks built by hand so they hold whatever the clock does about cutting. A note inside a ramp lands where the map puts it, to the sample, on a block where the map and the line are 130 samples apart at the far end; the case asserts that divergence too, so it cannot pass on a block that could not tell the two answers apart. And the metronome follows a signature change inside one block, walking the map's own ticks rather than a grid worked out from the block's ends. #2333 flagged the metronome as the thing that might need one signature per block. It does not.

## One acceptance criterion answered differently from its wording

"Arrangement audio cannot render spans compiled against a different tempo map" is a count and a play here, not a refusal, following the finding on #2337 that a dropout mid-set is a worse failure than the bug it reports. #2338's commit rides at the front of this branch unchanged rather than being rebuilt, since its branch was deleted when it was closed. The atomic publish is still the real fix and is not here: `EngineSession::publishClips` is called only from tests, so there is no production publisher to make atomic yet.

## Testing

Full Catch2 suite green locally: 3396 passed, 13 skipped for plugins absent from this machine, 0 failures. The JUCE suite passes, including the real project corpus through both engines, and the app builds.

The two new cases were each verified to fail without their fix: the retrigger case against a re-anchored schedule, and the note-off edge rule against the test that already named the old clamp.

### DIFF
--- a/magda/engine/clip/ClipAudioSource.hpp
+++ b/magda/engine/clip/ClipAudioSource.hpp
@@ -119,6 +119,23 @@ class ClipAudioSource final : public EngineAudioSource {
         return starved_.load(std::memory_order_relaxed);
     }
 
+    /**
+     * @brief Blocks that rendered nothing because the snapshot was stale.
+     *
+     * A snapshot carries the fingerprint of the tempo map its seconds were
+     * derived through. One that is not the transport's was compiled against a
+     * tempo that has since changed, and playing it would put clips somewhere
+     * the transport is not.
+     *
+     * Zero is the only right answer, and reaching it is the publish's job
+     * rather than this one's: the map and the snapshot compiled for it are
+     * meant to swap together (#2337). Non-zero says they did not, which is a
+     * publish-ordering bug wearing the costume of an audio dropout.
+     */
+    int staleSnapshots() const {
+        return staleSnapshots_.load(std::memory_order_relaxed);
+    }
+
   private:
     /// One entry that sounds in this block, and the reader it sounds through.
     struct Sounding {
@@ -173,6 +190,7 @@ class ClipAudioSource final : public EngineAudioSource {
     juce::AudioBuffer<float> scratch_;
 
     std::atomic<int> starved_{0};
+    std::atomic<int> staleSnapshots_{0};
 };
 
 }  // namespace magda::engine
--- a/magda/engine/clip/ClipAudioSource.hpp
+++ b/magda/engine/clip/ClipAudioSource.hpp
@@ -120,17 +120,24 @@ class ClipAudioSource final : public EngineAudioSource {
     }
 
     /**
-     * @brief Blocks that rendered nothing because the snapshot was stale.
+     * @brief Blocks rendered against a map the snapshot was not compiled for.
      *
      * A snapshot carries the fingerprint of the tempo map its seconds were
      * derived through. One that is not the transport's was compiled against a
-     * tempo that has since changed, and playing it would put clips somewhere
-     * the transport is not.
+     * tempo that has since changed, so every second in it is wrong by however
+     * much the map moved.
+     *
+     * Counted and then played anyway. Refusing to play it is a hole in the
+     * middle of a set to report a bug that should not happen, and it buys
+     * little, because stale spans usually stop overlapping the block in any
+     * case: what that mostly converts is an accidental gap into a deliberate
+     * one.
      *
      * Zero is the only right answer, and reaching it is the publish's job
      * rather than this one's: the map and the snapshot compiled for it are
      * meant to swap together (#2337). Non-zero says they did not, which is a
-     * publish-ordering bug wearing the costume of an audio dropout.
+     * publish-ordering bug that would otherwise be inaudible until somebody
+     * wondered why a clip was in the wrong place.
      */
     int staleSnapshots() const {
         return staleSnapshots_.load(std::memory_order_relaxed);
--- a/magda/engine/clip/ClipMidiSource.hpp
+++ b/magda/engine/clip/ClipMidiSource.hpp
@@ -155,15 +155,15 @@ class ClipMidiSource final : public EngineMidiSource {
     /// Whether @p bytes more will fit, leaving room for every off still owed.
     bool fits(int bytes) const;
 
-    void emit(juce::MidiBuffer& out, int sample, const MidiClipEvent& event);
+    void emit(juce::MidiBuffer& out, EventSample sample, const MidiClipEvent& event);
 
     /// Note-off now, or owed if it will not fit. Always leaves the note not
     /// sounding.
-    void endNote(juce::MidiBuffer& out, int sample, int channel, int note);
+    void endNote(juce::MidiBuffer& out, EventSample sample, int channel, int note);
 
     /// End everything, or everything one clip owns.
-    void endAll(juce::MidiBuffer& out, int sample);
-    void endClip(juce::MidiBuffer& out, int sample, ClipId clipId);
+    void endAll(juce::MidiBuffer& out, EventSample sample);
+    void endClip(juce::MidiBuffer& out, EventSample sample, ClipId clipId);
 
     /// The notes of @p clip actually sounding at @p timelineBeat, into @ref
     /// scratch_. Grooved edges, not written ones, and one implementation for
@@ -197,7 +197,7 @@ class ClipMidiSource final : public EngineMidiSource {
     void endUnexpected(juce::MidiBuffer& out, const NoteMask& expected);
 
     /// Every note one slot started, ended at @p sample. What a stop owes.
-    void endSlot(juce::MidiBuffer& out, int sample, const SessionSlotPlayback& slot);
+    void endSlot(juce::MidiBuffer& out, EventSample sample, const SessionSlotPlayback& slot);
 
     /// What a launched slot plays and what a stopped one owes. False when no
     /// handles have been published yet.
--- a/magda/engine/clip/ClipMidiSource.hpp
+++ b/magda/engine/clip/ClipMidiSource.hpp
@@ -107,17 +107,24 @@ class ClipMidiSource final : public EngineMidiSource {
     }
 
     /**
-     * @brief Blocks that rendered nothing because the snapshot was stale.
+     * @brief Blocks rendered against a map the snapshot was not compiled for.
      *
      * A snapshot carries the fingerprint of the tempo map its seconds were
      * derived through. One that is not the transport's was compiled against a
-     * tempo that has since changed, and playing it would put clips somewhere
-     * the transport is not.
+     * tempo that has since changed, so every second in it is wrong by however
+     * much the map moved.
+     *
+     * Counted and then played anyway. Refusing to play it is a hole in the
+     * middle of a set to report a bug that should not happen, and it buys
+     * little, because stale spans usually stop overlapping the block in any
+     * case: what that mostly converts is an accidental gap into a deliberate
+     * one.
      *
      * Zero is the only right answer, and reaching it is the publish's job
      * rather than this one's: the map and the snapshot compiled for it are
      * meant to swap together (#2337). Non-zero says they did not, which is a
-     * publish-ordering bug wearing the costume of an audio dropout.
+     * publish-ordering bug that would otherwise be inaudible until somebody
+     * wondered why a clip was in the wrong place.
      */
     int staleSnapshots() const {
         return staleSnapshots_.load(std::memory_order_relaxed);
--- a/magda/engine/clip/ClipMidiSource.hpp
+++ b/magda/engine/clip/ClipMidiSource.hpp
@@ -106,6 +106,23 @@ class ClipMidiSource final : public EngineMidiSource {
         return overflowed_.load(std::memory_order_relaxed);
     }
 
+    /**
+     * @brief Blocks that rendered nothing because the snapshot was stale.
+     *
+     * A snapshot carries the fingerprint of the tempo map its seconds were
+     * derived through. One that is not the transport's was compiled against a
+     * tempo that has since changed, and playing it would put clips somewhere
+     * the transport is not.
+     *
+     * Zero is the only right answer, and reaching it is the publish's job
+     * rather than this one's: the map and the snapshot compiled for it are
+     * meant to swap together (#2337). Non-zero says they did not, which is a
+     * publish-ordering bug wearing the costume of an audio dropout.
+     */
+    int staleSnapshots() const {
+        return staleSnapshots_.load(std::memory_order_relaxed);
+    }
+
   private:
     /// One note owed an off that would not fit in the block that owed it.
     struct OwedNote {
@@ -212,6 +229,7 @@ class ClipMidiSource final : public EngineMidiSource {
 
     std::atomic<int> dropped_{0};
     std::atomic<int> overflowed_{0};
+    std::atomic<int> staleSnapshots_{0};
 };
 
 }  // namespace magda::engine
--- a/magda/engine/launch/LaunchHandle.cpp
+++ b/magda/engine/launch/LaunchHandle.cpp
@@ -16,13 +16,46 @@ double LaunchHandle::virtualStart(const SyncRange& piece) const {
     // Derived from monotonic elapsed, which is the one quantity that survives
     // the wrap. It may sit before the block or before zero, and that is what it
     // means: the run began that far back.
-    return piece.timeline.start - (piece.monotonic.start - playedMonotonic_->start);
+    return piece.timeline.start - (piece.monotonic.start - run_->originBeat);
+}
+
+double LaunchHandle::elapsedSecondsAt(const SyncRange& piece) const {
+    // A subtraction of two sample counts, which is exact, rather than an
+    // accumulation of per-block seconds. Between rate changes: a count that
+    // spans one counts samples that were not all worth the same amount of time,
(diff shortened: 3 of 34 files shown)


magda/engine/transport/TransportClock.cpp
```diff
@@ -176,10 +175,19 @@ std::span<const TransportClock::Segment> TransportClock::advance(const Transport
...
-        // placed on the line between its two ends, and that line is the map
-        // within a section and a guess across a boundary (TempoMap.hpp).
+        // and the next segment carries on continuous.
+        //
+        // What the cut is for has changed. It arrived to keep the block's own
+        // straight line honest, because everything inside a block was placed on
+        // the line between its two ends; placement goes through the map now, so
+        // that is no longer what pays for it (#2336).
+        //
+        // What it still buys is that a block is one tempo and one signature. A
```
lucaromagnoli, line 185: The guarantee the cut is kept for has a hole, and it is the epsilon's. When the next section boundary lies within a hundredth of a sample after the cursor, `samplesUntil` returns 0, the `untilBoundary > 0` test below skips the cut, and nothing asks for the boundary after that one. The segment then runs to whatever else ends it (the callback, the loop end), through that boundary and any later ones inside the callback. So on that alignment a block opens a fraction of a sample before a boundary and spans one or more sections uncut, which is exactly the block `modTimingFor` reads one bpm for.

Pre-existing from #2325, but this PR is what makes it load-bearing. The fix is small: when `untilBoundary <= 0`, take `nextSectionBoundaryAfter(boundary)` and cut there instead.

tests/engine/test_transport_clock.cpp
```diff
@@ -786,3 +800,36 @@ TEST_CASE("The monotonic seconds count rendered time, not converted beats",
...
+        // excluded and can sit a fraction of a sample past the boundary, since
+        // a cut lands on the first sample at or after it; what has to be one
+        // tempo is every sample the block actually renders.
+        const auto last =
+            snapshot.tempo.timeToBeat(block.seconds.start + ((block.numSamples - 1) / kSampleRate));
+
+        const auto opening = snapshot.tempo.bpmAt(block.beats.start);
+        const auto closing = snapshot.tempo.bpmAt(last);
+
+        CHECK(closing == approx(opening));
```
lucaromagnoli: Alignment-dependent, and independently of the clock comment. On the alignment where the epsilon absorbs a boundary, `beats.start` sits a fraction of a sample before it, `sectionForBeat` puts that in the old section, and `bpmAt(beats.start)` is the old tempo while samples 1 to N-1 are in the new one. This check would fail there, and `modTimingFor` (ModLfo.cpp:141) reads the old bpm for the whole block. Reading the tempo at the block's midpoint, in both places, is inside the section on every alignment.

magda/engine/clip/ClipMidiSource.hpp
```diff
@@ -106,6 +106,23 @@ class ClipMidiSource final : public EngineMidiSource {
         return overflowed_.load(std::memory_order_relaxed);
     }
 
+    /**
+     * @brief Blocks that rendered nothing because the snapshot was stale.
```
lucaromagnoli: Still the silence version's wording: "rendered nothing" here and "playing it would put clips somewhere the transport is not" below. The behaviour is count and play, and the comment at the check in the .cpp says so. Same block in ClipAudioSource.hpp.

magda/engine/clip/ClipAudioSource.hpp
```diff
@@ -119,6 +119,23 @@ class ClipAudioSource final : public EngineAudioSource {
         return starved_.load(std::memory_order_relaxed);
     }
 
+    /**
+     * @brief Blocks that rendered nothing because the snapshot was stale.
```
lucaromagnoli: Same as ClipMidiSource.hpp: this describes the silence behaviour that was replaced by count and play.

magda/engine/launch/LaunchHandle.cpp
```diff
@@ -74,77 +101,124 @@ void LaunchHandle::setLooping(std::optional<double> beats) {
...
+void LaunchHandle::joinRun(const SyncRange& range, const BlockInstant& at, const Run& other) {
+    if (const auto played = playedRange())
+        lastPlayed_ = played;
+
+    playState_ = PlayState::playing;
+
+    // Adopting the position, not just the origin. Derived here rather than at
+    // the request, which is a different number whenever the launch was queued
+    // for a later beat.
+    auto run = other;
```
lucaromagnoli, line 203: `other` is the leader's run as it was at `playSynced`, and a run is rate-dependent now where the old three faces were not. A device rate change between the request and the launch beat rescales the leader's live run through `followRate` but not this copy, so the follower adopts an origin counted at the old rate with `sampleRate` still the old one. `elapsedSecondsAt` then divides a span that includes new-rate samples by the old rate, and the next block's `followRate` bakes that elapsed into the re-count. At 44.1 to 48 kHz with a second between the change and the launch, the follower ends up about 90 ms off the leader for the rest of the run.

Narrow trigger. There is no correct re-count from the copy alone, since it cannot know where the leader's change fell: the run to join has to be read at the launch instant rather than at the request.

lucaromagnoli (commented):
Read-only review of the diff. The shape is what was agreed: typed event and edge samples with `soundsAt` as the one named exception, a sample origin with the schedule kept unrounded, `BlockInstant` down to its sample, and 2338 at the front unchanged.

Two things I would act on before merge. The invariant the cut is now kept for has an alignment hole in the clock (pre-existing from #2325, but this PR is what makes it load-bearing and asserts it), and the two stale-snapshot headers still describe the silence behaviour. The synced-launch rate case is real but narrow.

tests/engine/test_block_samples.cpp
```diff
@@ -0,0 +1,158 @@
...
+    block.beats.end = ramp.timeToBeat(block.seconds.end);
+    block.tempo = &ramp;
+
+    for (const auto through : {0.25, 0.5, 0.75}) {
+        const auto beat = block.beats.start + (through * block.beats.length());
+
+        const auto exact =
+            static_cast<int>((ramp.beatToTime(beat) - block.seconds.start) * kSampleRate);
+
+        CHECK(block.eventForBeat(beat).value == exact);
```
lucaromagnoli: `exact` truncates and `eventForBeat` floors offset plus a hundredth of a sample, so they disagree whenever the fractional part is above 0.99. The equality holds for these three points by the luck of the ramp. Either apply the same rule (`eventAt(offset, N)` on the unrounded offset) or allow the plus or minus one the MIDI ramp case uses.

magda/engine/param/ModLfo.cpp
```diff
@@ -136,9 +136,18 @@ ModTiming modTimingFor(const BlockInfo& block, double sampleRate) {
...
+    // rate for the whole block, which is sound because the transport cuts a
+    // callback at every tempo section (TransportClock.cpp); what the start is
+    // not, is reliably inside that section. A cut lands on the first sample at
+    // or after the boundary, so a block can open a fraction of a sample before
+    // one and read the tempo it is leaving for every sample it renders. The
+    // middle is inside the section on every alignment.
     if (block.tempo != nullptr) {
-        const auto signature = block.tempo->barsAndBeatsAt(block.beats.start);
-        timing.bpm = block.tempo->bpmAt(block.beats.start);
+        const auto inside = block.beats.start + (0.5 * block.beats.length());
```
lucaromagnoli: **P2 — Use the same effective section for the transport-synced phase**

Moving modTimingFor to the midpoint fixes the rate/signature chosen for this block, but the transport-synced path still calls modBarPosition, whose map-backed branch reads barsAndBeatsAt(block.beats.start). On the alignment this commit handles—where a signature boundary is less than kSampleEpsilon ahead—the start is still in the old signature while the midpoint and the rendered block are treated as the new one. For an irregular change such as 4/4 to 3/4 at beat 1, a bar-rate LFO can therefore render this whole block at the old bar fraction instead of restarting on the new bar. Please derive the phase from the same effective section, projected back to sample zero, and cover a swallowed signature boundary.

magda/engine/transport/TransportClock.cpp
```diff
@@ -176,15 +175,33 @@ std::span<const TransportClock::Segment> TransportClock::advance(const Transport
...
+        // extra segment per tempo change (#2333).
+        //
+        // Past a boundary the epsilon has already swallowed, or the guarantee
+        // has a hole exactly where it is hardest to see. A boundary within a
+        // hundredth of a sample ahead of the cursor counts as zero samples
+        // away, and a zero cut is no cut: the segment would then run to
+        // whatever else ends it, through that boundary and every later one in
+        // the callback. The cursor is on that boundary for all practical
+        // purposes, so the one to cut at is the next one after it.
+        auto boundary = tempo.nextSectionBoundaryAfter(beatAfter(tempo, samplesSinceAnchor_));
```
lucaromagnoli, line 199: **P2 — Update the hosted-plugin playhead for blocks that skip this boundary**

This deliberately permits the resulting block to start fractionally before the swallowed boundary and treats its interior as belonging to the following section. EngineExternalDevice::PlayHead::setBlock, however, still publishes both bpmAt(block.beats.start) and barsAndBeatsAt(block.beats.start). A VST/AU processed over this block therefore receives the previous BPM or time signature for the entire call, exactly the block-wide-consumer failure this cut is meant to prevent. Please make the playhead choose the same representative interior section as modTimingFor while preserving its sample-zero time/PPQ position.

PR #2342, opened 2026-09-03
Modifiers read the bars the block covers, and the section cut goes (#2340)

### LINKED ISSUE #2340: Modifiers read beats off the block, not a bpm, and the section cut goes
Follow-up to #2336, which kept the tempo-section callback cut for one reason: three modifiers read a single bpm per block, so a block spanning a section would run at the tempo it opened on (#2333).

Every one of those reads is a conversion the block already answers. `ModLfo` and `ModRandom` advance their phase by block seconds over period seconds, which is beats elapsed in the block over beats per cycle. `ModAdsr` turns a stage length in beats into seconds. `block.beats.length()` is exactly the beats a block covers, derived through the map by the clock, and it is right across a step or a baked ramp boundary inside the block. A stage of N beats from a trigger beat is `beatToTime(start + N) - beatToTime(start)`. None of these needs a tempo value, and a tempo edit while rolling reaches them on the next block through the block's beat range, the same way it reaches notes and automation.

The one block that still needs a bpm is a stopped one: the transport covers no beats, and a free-running LFO has to keep its own time at the cursor's tempo. A stopped cursor sits on one beat, so one bpm is exact there.

Once nothing reads a per-block bpm while rolling, the cut has nothing left to protect. It also has a hole: when a section boundary lies within the epsilon after the cursor, `samplesUntil` returns zero, the cut is skipped, and nothing asks for the boundary after it, so the block runs through that boundary and any later ones in the callback. Dropping the cut removes the hole with it.

## Steps

- `ModLfo`, `ModRandom`: advance by `block.beats.length() / beatsPerCycle` on a playing block; keep the seconds-times-bpm advance for a stopped block only.
- `ModAdsr`: stage durations through the map from the trigger beat.
- Drop the section cut from `TransportClock::advance`, and `nextSectionBoundaryAfter` if nothing else calls it.
- The signature read stays per block, at the first sample, which is the sample the modifier's value is for.

## Tests

- An LFO across a block that spans a tempo step advances by the map's beats, not the opening tempo's.
- A free-running tempo-synced LFO keeps cycling on stopped blocks.
- "A steep ramp leaves every block affine to within a sample" and "A block is one tempo" in `test_transport_clock.cpp` go, since they assert the property the cut provided.

Related: #2325, #2333, #2336.

lucaromagnoli:
Closes #2340. Follows #2339, which kept the tempo-section callback cut for one reason: three modifiers read a single bpm per block.

## Modifiers read the bars the block covers

`ModLfo`, `ModRandom` and `ModAdsr` each used one bpm and one signature per block to turn block seconds into a fraction of a cycle. The block already knows how far it moved: `block.beats.length()` is derived through the map by the clock and is exact across a tempo step or a baked ramp boundary inside the block. A new shared `modBarsElapsed` turns that into bars off the map, walking the signature spans inside the block and summing what each is worth in its own bar length, so a block that crosses a 4/4 to 3/4 change covers one beat over four plus one beat over three rather than two over four. On a stopped block it returns the block's seconds at the tempo and signature the cursor sits on, since a stopped block covers no beats and a free-running modifier still has to keep time. A stopped cursor is on one beat, so one bpm and one bar length are exact there. `TempoMap::signatureEndAfter` is the one new map accessor it needs.

The LFO and the random walk advance their tempo-synced ramp by bars over the rate's bar fraction. The ADSR's stage machine is unit-agnostic now, comparing an elapsed amount against stage lengths without asking which unit; a synced envelope runs in bars, and the millisecond path is unchanged. The state records which unit its accumulator counts in, so toggling tempo sync mid-stage converts through the stage phase instead of reinterpreting seconds as bars.

## The opening beat replaces the midpoint

`sectionBeat()` was a midpoint read that only made sense while the cut guaranteed one section per block. `openingBeat()` is the block's start nudged forward by the clock's own tolerance, a hundredth of a sample in seconds converted through the map at the tempo the block opens on: a block that opens that close before a boundary reads the section on the far side of it, and every other block reads the section its start is in. The block-average slope would over-nudge across a step up and under-nudge across a step down, which is why it goes through the map now that a block can span a change. `modTimingFor`, the bar position and the hosted-plugin playhead read there. The playhead's tolerance and bar-start walk-back are gone, since the opening beat already carries the tolerance.

The playhead is the one block-wide reader that cannot use a bar length: JUCE hands a plugin one bpm and one signature per call. A block spanning a step reports its first sample's section for all of it, which is what other hosts do.

## The section cut goes

Nothing reads a per-block tempo or signature while rolling any more, so `TransportClock::advance` no longer ends a segment at a section boundary, and `TempoMap::nextSectionBoundaryAfter` is removed with it; nothing else called it. What still ends a segment is a loop wrap and the end of a count-in. Dropping the cut also removes the epsilon hole found in review on #2339.

## Tests

- `test_transport_clock.cpp`: the step case asserts one continuous segment covering beats 1.0 to 2.5. "A steep ramp leaves every block affine to within a sample", "A block is one tempo" and "A boundary the epsilon swallows is still cut at" are gone; they asserted the property the cut provided.
- New `[2340]` cases in `test_mod_lfo.cpp`, `test_mod_random.cpp` and `test_mod_adsr.cpp`: a hand-built block from beat 1.5 to 2.5 across a 120-to-60 step, where the opening-tempo formula would overshoot by half; a block across a 4/4 to 3/4 change at a bar rate, where the opening-signature formula would say half a bar and the map says a quarter plus a third; and an ADSR whose tempo sync is toggled halfway through an attack and keeps its place. The stopped-transport LFO case gains a tempo-synced section.
- `test_block_samples.cpp`: `openingBeat()` across a 20-to-300 step stays in the old section when the boundary is farther than the tolerance, and across a 300-to-20 step crosses when it is within it; the block-average slope gets both wrong.
- `test_session_playback.cpp`: the launched-slot case reads the material from the launch sample inside the now uncut block.
- Full Catch2 suite: 3406 passed, 13 skipped for plugins absent from this machine, 0 failures.

https://claude.ai/code/session_01WnFtPLfo2TUoSynbysJRri

### DIFF
--- a/magda/engine/exec/RenderContext.hpp
+++ b/magda/engine/exec/RenderContext.hpp
@@ -187,26 +187,31 @@ struct BlockInfo {
     }
 
     /**
-     * @brief Which beat to ask the map about this block's own tempo, signature
-     * or bar.
+     * @brief The beat sample zero sounds on, to the tolerance the clock uses.
      *
-     * Not the block's start, which is where anything reading one of those would
-     * naturally look. A block covers one tempo section, because the transport
-     * cuts a callback at every boundary (TransportClock.cpp), but a cut lands
-     * on the first sample at or after the boundary, so the block can begin a
-     * fraction of a sample before it. The beat at its start is then in the
-     * section it is leaving while every sample it renders is in the next, and a
-     * consumer that takes one rate for the whole block takes the wrong one: an
-     * LFO runs the block at the tempo it just left, a hosted plugin is told the
-     * old signature for the whole call.
+     * Not quite the block's start, which is where anything asking the map about
+     * this block's tempo, signature or bar would naturally look. A musical
+     * position within a hundredth of a sample of the cursor is the cursor as
+     * far as the clock is concerned (TransportClock::samplesUntil), so a block
+     * can open that close before a boundary while every sample it renders is
+     * past it. The beat at its start is then in the section it has already
+     * left, and a consumer that takes one signature or one bar for the whole
+     * block takes the wrong one.
      *
-     * The middle is inside the section on every alignment. This answers "which
-     * section is this block in" and nothing else; where the block *is* stays
-     * @ref beats and @ref seconds, and a position derived under this section is
-     * projected back to them by the caller (#2336).
+     * So this is the start nudged forward by that same hundredth of a sample,
+     * in beats at the block's own rate: a block opening that close before a
+     * boundary reads the section on the far side of it, and every other block
+     * reads the section its start is in.
+     *
+     * Where the block *is* stays @ref beats and @ref seconds. This only says
+     * which side of a boundary its first sample is on, and a position derived
+     * under that section is projected back by the caller (#2336).
      */
-    double sectionBeat() const {
-        return beats.start + (0.5 * beats.length());
+    double openingBeat() const {
+        if (numSamples <= 0 || beats.empty())
+            return beats.start;
+
+        return beats.start + (kSampleEpsilon * (beats.length() / numSamples));
     }
 
     /**
--- a/magda/engine/exec/RenderContext.hpp
+++ b/magda/engine/exec/RenderContext.hpp
@@ -198,10 +198,15 @@ struct BlockInfo {
      * left, and a consumer that takes one signature or one bar for the whole
      * block takes the wrong one.
      *
-     * So this is the start nudged forward by that same hundredth of a sample,
-     * in beats at the block's own rate: a block opening that close before a
-     * boundary reads the section on the far side of it, and every other block
-     * reads the section its start is in.
+     * So this is the start nudged forward by that same hundredth of a sample of
+     * seconds, converted to a beat at the tempo the block opens on rather than
+     * at the block's own average slope: through the map, the way @ref
+     * beatAtTime reads any other moment, which is what matters once a block may
+     * span a tempo change. The block-average slope would over-nudge across a
+     * step up and under-nudge across a step down, either of which can land the
+     * nudge on the wrong side of the very boundary it exists to cross (#2340).
+     * A block with no seconds face at all is one assembled by hand from beats,
+     * and the average slope is the only line there is to nudge along.
      *
      * Where the block *is* stays @ref beats and @ref seconds. This only says
      * which side of a boundary its first sample is on, and a position derived
@@ -211,7 +216,10 @@ struct BlockInfo {
         if (numSamples <= 0 || beats.empty())
             return beats.start;
 
-        return beats.start + (kSampleEpsilon * (beats.length() / numSamples));
+        if (seconds.empty() || rate() <= 0.0)
+            return beats.start + (kSampleEpsilon * (beats.length() / numSamples));
+
+        return beatAtTime(seconds.start + (kSampleEpsilon / rate()));
     }
 
     /**
--- a/magda/engine/param/ModAdsr.cpp
+++ b/magda/engine/param/ModAdsr.cpp
@@ -100,33 +100,74 @@ void enterStage(AdsrState& state, AdsrStage stage, float startValue) {
     state.stageStartValue = startValue;
 }
 
+/// Whether an envelope's stages are a musical length rather than a duration. A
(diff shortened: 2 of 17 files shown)


magda/engine/transport/TransportClock.cpp
```diff
@@ -173,36 +173,14 @@ std::span<const TransportClock::Segment> TransportClock::advance(const Transport
...
             samples = std::min(samples, samplesUntil(tempo, countInUntilBeat_));
 
-        // A block never spans a tempo section, which is the stretch the map is
-        // linear over. Not a jump: audio flows straight through a tempo change
-        // and the next segment carries on continuous.
+        // A tempo section boundary is not cut at. It used to be, so that the
+        // block's own straight line stayed honest and so that a modifier
+        // reading one bpm per block read the right one; placement goes through
+        // the map now (#2336) and a modifier reads the beats the block covers
+        // (#2340), so nothing is left that a section boundary inside a block
```
lucaromagnoli, line 180: **P1 — Preserve time-signature boundaries for bar-relative modifiers**

Removing every section cut also lets a block cross a time-signature change, but the modifier path still reads one signature at the opening. For example, a free-running Bar LFO over a 4/4-to-3/4 boundary now advances the entire block by total beats / 4; before this change the two segments advanced pre-change beats / 4 plus post-change beats / 3. The same single opening signature feeds Random and synced ADSR, so each keeps a lasting phase or stage error after the crossing.

Please either retain cuts at signature boundaries or integrate modifier advancement over each signature span; exact total beat length only removes the need for tempo-section cuts.

magda/engine/param/ModAdsr.cpp
```diff
@@ -265,10 +306,18 @@ float advanceAdsr(AdsrState& state, const AdsrSettings& settings, const BlockInf
...
-        static_cast<double>(std::max(block.numSamples, 0)) / std::max(timing.sampleRate, 1.0);
+    const auto lengths = stageLengthsFor(settings, timing);
+
+    // A synced envelope's stages are a number of beats, so the block moves it on
+    // by the beats it covered, which the map answered for exactly this stretch.
+    // A millisecond envelope is a duration and moves on by how long the block
+    // lasted (#2340).
+    const double elapsed =
+        runsInBeats(settings)
+            ? modBeatsElapsed(block, timing)
```
lucaromagnoli: **P2 — Migrate the ADSR accumulator when its unit changes**

`timeInStage` survives settings publishes, but this branch changes what that stored number means without converting it. Toggling tempo sync during a stage therefore reinterprets accumulated seconds as beats, or beats as seconds. At 120 BPM, for example, changing halfway through a 2-second attack to a one-bar attack (also 2 seconds) turns the stored `1.0` second into `1.0` beat and moves the envelope from 50% back to 25%.

Please track the active unit and convert or phase-migrate the accumulator when `runsInBeats` changes, or explicitly restart the stage.

magda/engine/exec/RenderContext.hpp
```diff
@@ -187,26 +187,31 @@ struct BlockInfo {
...
+     * which side of a boundary its first sample is on, and a position derived
+     * under that section is projected back by the caller (#2336).
      */
-    double sectionBeat() const {
-        return beats.start + (0.5 * beats.length());
+    double openingBeat() const {
+        if (numSamples <= 0 || beats.empty())
+            return beats.start;
+
+        return beats.start + (kSampleEpsilon * (beats.length() / numSamples));
```
lucaromagnoli: **P2 — Convert the opening tolerance at the opening tempo**

`beats.length() / numSamples` is the average slope of the whole block. Now that a block may span a tempo change, it is not the beat distance of a hundredth of a sample at the opening. Across a 20-to-300 step it can over-nudge into the new section even when the boundary is farther than the clock tolerance; across 300-to-20 it can under-nudge and leave a swallowed boundary in the old section. That gives modifiers and hosted plugins the wrong opening tempo/signature/bar.

Please project `seconds.start + kSampleEpsilon / rate()` through the tempo map (including `materialOrigin`) instead of scaling by the block-average beat length.

PR #2343, opened 2026-09-03
Propagate a port's MIDI byte bound through the ops that pass MIDI through (#2341)

### LINKED ISSUE #2341: Propagate a port's MIDI byte bound through the ops that pass MIDI through
## Problem

A device is told how much MIDI can reach it and nothing about how much it may emit, and the two figures disagree whenever its input is a merge.

The budget is bytes per port, not events: `kMaxMidiBytesPerPort = 4096` (`magda/engine/exec/EngineDevice.hpp:62`), and an event costs six bytes of overhead plus its data — seven for a clock, nine for a note. One port therefore carries about 585 clocks or 455 notes.

**Input.** `PlanExecutor` sums the budget forward through the MIDI graph and tells each device the real figure through `setMidiInputBoundBytes` (`PlanExecutor.cpp:691-707`). Two MIDI sources merging onto a track means that device is told 8192. This is deliberate, commented, and covered by "the adapter carries a merged input, not one producer's worth".

**Output.** Nothing tells it anything. `EngineMagdaDevice` caps its write-back at the flat constant and drops the rest with `jassertfalse` (`EngineMagdaDevice.cpp:404`), and `EngineDevice` has no output counterpart to `setMidiInputBoundBytes` (`EngineDevice.hpp:161`).

The adapter is not using the wrong number. It is using exactly what the plan reserved: the forward pass starts every port at `kMaxMidiBytesPerPort` (`PlanExecutor.cpp:636`) and recomputes only for `MergeMidi`, `Fader` and `Delay` (`PlanExecutor.cpp:658`). `OpKind::Device` is not in that list. The pass models a device as a *producer* — writes at most one producer's worth — and never as a *conduit*.

Devices are conduits whenever they pass MIDI through. `OpKind::MidiNoteGate` (`RenderPlan.hpp:122`) is a pure conduit and has the identical gap.

## Consequence

A passthrough device handed a legal 8192-byte merged input can only emit 4096. Half the stream dies at the adapter, and nothing the device does changes it — a device that buffers its own input to choose what survives just moves which events are lost, which is what #2335 removed from the step sequencers for that reason.

The one-event version is reachable without a merge at all: with an input at the full 4092 bytes, a step sequencer's own all-notes-off on its first block is the event that tips the total past 4096 and is refused. `tests/engine/test_engine_device_layer.cpp` currently warms the device up before the dense block to keep that out of the assertion.

## Direction

- Propagate the bound through the ops that pass MIDI through, so a `Device` op's MIDI output port is at least its MIDI input port's, and the same for `MidiNoteGate`. A device that also emits its own MIDI needs headroom above that; decide whether that is a flat producer's worth added on top or something a device declares.
- Give `EngineDevice` a `setMidiOutputBoundBytes` alongside the input one, and have `EngineMagdaDevice` cap on it rather than on the constant. The slot is already sized from `midiByteBounds_`, so the storage is there — only the figure the adapter enforces is wrong.
- The Tracktion adapter has no count bound at all, so nothing is needed there; this is the native engine's contract.

## Acceptance criteria

- A device with MIDI thru enabled, fed a merged input of several producers' worth, emits all of it.
- The same for `MidiNoteGate`.
- A device that emits its own MIDI on top of a full passthrough input is not silently truncated, and whatever bound it does have is a figure something told it rather than a constant it assumed.
- `jassertfalse` at `EngineMagdaDevice.cpp:404` fires only for a device genuinely past what its port reserved.

## Non-goals

- Changing the per-producer budget, or the byte-rather-than-event accounting. Both are right.
- The Tracktion path.

## Background

Found reviewing #2335, which ported the step sequencers. Not reachable from inside a device and not specific to those two: it predates the port and applies to any device that passes its MIDI input through.

lucaromagnoli:
Closes #2341.

## The gap

A device was told how much MIDI could reach it and nothing about how much it could emit.

`PlanExecutor::prepare` sums the byte budget forward through the MIDI graph, but it recomputed the figure for `MergeMidi`, `Fader` and `Delay` only — every other port kept the flat `kMaxMidiBytesPerPort` it started at. That models a device as a *producer*: writes at most one producer's worth. Devices are *conduits* whenever they pass MIDI through, and `MidiNoteGate` is a pure conduit with the identical gap.

So a passthrough device handed a legal 8192-byte merged input could emit 4096 of it. Half the stream died at `EngineMagdaDevice`, and nothing the device did changed which half.

## The fix

- **`Device` and `MidiNoteGate` carry their MIDI input's bound to their output port.** A gate emits only what arrived, so its port is its input's. A device is a conduit *and* a producer, so its port is its input's plus one producer's worth of headroom for what it adds of its own — a pattern, an arpeggio, the all-notes-off it owes after a prepare. That is the same budget every other producer is held to, and the figure `EngineExternalDevice` already sized its own buffer by.
- **`EngineDevice::setMidiOutputBoundBytes`** beside the input one. The executor tells every bound device its port's own figure; both adapters cap their write-back on that instead of on the constant. The input stays the *slot's* bound (the right size for a buffer to read from); the output is the *port's* own (everything downstream reserved its room from this port's sum, not its neighbour's).
- **`EngineMagdaDevice`'s scratch was the same bug one layer down.** One vector serves both directions — the device reads its input out of it and writes its own events back into it — and sized from the input bound alone, a device handed a block that filled its input port had nowhere to put its own all-notes-off. Sized from the wider of the two bounds now.

The Tracktion path has no count bound at all, so nothing changes there.

## Tests

Three new cases, each verified negatively (reverted the fix, watched it fail):

| Test | Without the fix |
|---|---|
| `a device passing a merged input through emits all of it` | 585 of 1170 clocks arrive, and the device's own event is refused by the scratch |
| `the executor tells a device what it may write, its input carried through` | output bound is 4096 for an input of 8192 |
| `A note gate carries its input's MIDI bound to what it feeds` | the pad's device is told 4096 for an input carrying 8192 |

`a step sequencer passing MIDI through keeps the whole legal block` no longer spends the all-notes-off on a warm-up block to stay out of the assertion's way — that warm-up existed only because of this asymmetry — and now asserts the device's own event survives a block that already fills the input port.

Full suite: 3422 passed, 13 skipped, 0 failed.

### DIFF
--- a/magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
+++ b/magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
@@ -250,6 +250,10 @@ void EngineExternalDevice::setMidiInputBoundBytes(int bytes) {
                          magda::engine::kMaxMidiBytesPerPort);
 }
 
+void EngineExternalDevice::setMidiOutputBoundBytes(int bytes) {
+    midiOutputBoundBytes_ = bytes;
+}
+
 int EngineExternalDevice::latencySamples() const {
     return latencySamples_;
 }
@@ -326,12 +330,14 @@ void EngineExternalDevice::writeMidiOut(juce::MidiBuffer& out, int numSamples) c
     // is in it afterwards, and a chain that wanted the raw input as well asks
     // the plan for it (DeviceInfo::midiInThru) rather than asking the plugin.
     //
-    // Counted in bytes against the port's budget, like every other producer.
+    // Counted in bytes against what the executor reserved for this port, which
+    // for a plugin handing its input back is that input's bound plus a
+    // producer's worth rather than the flat constant (#2341).
     int bytesWritten = 0;
 
     for (const auto metadata : midi_) {
         const auto cost = kMidiEventOverheadBytes + metadata.numBytes;
-        if (bytesWritten + cost > magda::engine::kMaxMidiBytesPerPort) {
+        if (bytesWritten + cost > midiOutputBoundBytes_) {
             jassertfalse;  // a plugin past the port's budget
             break;
         }
--- a/magda/daw/audio/plugins/engine/EngineExternalDevice.hpp
+++ b/magda/daw/audio/plugins/engine/EngineExternalDevice.hpp
@@ -124,6 +124,7 @@ class EngineExternalDevice final : public magda::engine::EngineDevice {
     void prepare(const magda::engine::RenderContext& context) override;
     void reset() override;
     void setMidiInputBoundBytes(int bytes) override;
+    void setMidiOutputBoundBytes(int bytes) override;
     int latencySamples() const override;
 
     /// One block through the plugin: its parameters written, its audio
@@ -276,6 +277,11 @@ class EngineExternalDevice final : public magda::engine::EngineDevice {
 
     int latencySamples_ = 0;
     int midiInputBoundBytes_ = magda::engine::kMaxMidiBytesPerPort;
+
+    /// What the executor reserved on the output port: this device's input
+    /// carried through plus a producer's worth of its own. A plugin handed a
+    /// merged input hands one back, and the flat constant would cut it (#2341).
+    int midiOutputBoundBytes_ = magda::engine::kMaxMidiBytesPerPort;
     bool offlineRender_ = false;
     bool prepared_ = false;
 };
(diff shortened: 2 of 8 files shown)


lucaromagnoli:
Review finding:

**[P1] Avoid counting raw MIDI twice for MIDI-through devices** (`magda/engine/exec/PlanExecutor.cpp:679`)

The compiler already follows every MIDI-producing device with `midiInThru` enabled by a `MergeMidi` of `{signal.midi, deviceMidi}`. This new calculation also includes the incoming bound in `deviceMidi`, so each stage is budgeted as `B_n = 2 * B_(n-1) + 4096`.

With one MIDI source, 16 chained MIDI-through devices reserve about 512 MiB per affected slot; the 19th computes 2,147,483,648 bytes and overflows the signed `int`, after which conversion to `size_t` can request an enormous allocation during `prepare()`. This is a valid, currently unbounded chain shape.

The raw stream needs to be counted only once: either keep the device port bound to the plugin-produced budget when the compiler supplies the raw-input merge, or remove that merge and encode device passthrough explicitly in the plan.

PR #2344, opened 2026-09-03
A track sounds one section, and both edges of the switch are ramped (#2302)

lucaromagnoli:
Slice 3 of #1894. Depends on slice 2, which is #2318.

A track has an arrangement and a session and sounded both at once: the compiler emits `ClipAudio` and `SessionAudio` into the same track mix (`PlanCompiler.cpp:1296-1310`) and nothing decided between them. Launching a slot played it over the top of whatever the timeline was already doing.

## What decides

A slot holds its track from the sample its run begins, and goes on holding it after it stops. The hand-back is not the clip stopping: silence after a stop is the session still holding the track, which is the incumbent's `playSlotClips` and is what Back to Arrangement is for. `LaunchHandle::releaseSection` is the engine side of that; the lane carrying it from off the audio thread is #2305's, so nothing in the app reaches it yet.

Per slot for a question that is per track, because the track's answer is the fold and the fold is over the table both of its sources already read (`sectionHold`, `SessionPlayback.hpp`). Two sources folding one table over one block reach the same answer, so a track's audio and its MIDI end up on the same side of the switch with nothing published between them. Set when the run begins rather than when it is queued, so a launch quantized to the next bar leaves the arrangement playing until the bar.

The incumbent's equivalent is a message-thread timer noticing that a slot has started (`ArrangerLauncherSwitchingNode::sharedTimerCallback`), which is a race it covers with a fade. This is decided on the audio thread, in the pass that has already advanced every handle, and is exact.

**The plan does not change.** A launch is a data change inside a stable plan: both ops stay in the mix and one of them renders silence, so a scene launch recompiles nothing and the differ does no structural work on the callback.

## Not a crossfade

The issue proposed binding `OpKind::Crossfade` to a new cause. That is the wrong shape here, and `FadeCurves.hpp` already says why: a gain fade would attenuate the incoming clip's opening transient, which is the thing the incumbent's de-click is deliberately built to avoid ("A conventional gain fade attenuates the entire first attack, which is especially obvious after a phase-vocoder has spread the transient across the opening window").

The two edges are not one ramp between two signals. They are two corrections:

- the outgoing section carries its last sample down to zero — `StopDeClick`, the mirror of the `StartDeClick` that was already there — which removes the step without touching anything else
- the incoming one simply starts, de-clicked by the ramp it already had

So the incoming clip is never attenuated, and what overlaps during the ramp is the outgoing step being taken out rather than a second signal. There is a test for exactly that.

**32 samples, not the 256 of a launch ramp.** A start de-click subtracts an offset from material that keeps sounding underneath, so its length is free and a longer one is gentler. A stop de-click has no material underneath: what it spends its length on is a held constant, and 5 ms of one is a thump in place of a click. The incumbent spends between 10 and 40 on this edge, in two places that disagree with each other; this is one number for one job.

## What the sources do

Both sources take the handle feed now, and an explicit `Section` says which of them a source is rather than the presence of the feed implying it.

The arrangement renders every block and drops the output while the session holds the track. That costs the CPU of a track nobody is hearing, and it is what makes handing the track back land where the timeline says rather than where the arrangement was when it lost it — the incumbent pays the same, for the same reason. Its MIDI plays up to the hand-over sample and owes note-offs there, so a note due just before a launch still sounds and none is left holding under the session's own.

## The cases the slice existed to settle

- **A switch mid-block**, at the split point slice 1 reports, rather than at a block boundary. At 512 samples and 120 bpm that is up to 11 ms, and it differs per track.
- **A stop with nothing to switch to.** The ramp is to silence and the arrangement stays away, because a stopped slot still holds the track.
- **Two switches inside one block.** Not a case: the ramp decays whatever the signal actually is now rather than resuming an earlier one, so a second switch inside a running ramp steps down from where the listener last heard it.
- **Per track or per source.** The decision is per track and the ramp is per source. Each source de-clicks its own signal with its own last sample, and both act on the one instant, so eight tracks in a scene switch on the same sample.

## Verification

`test_de_click.cpp` covers the pair on its own: both edges, carried across blocks so block size stays an I/O concept, a second stop inside a running ramp, a ramp of zero, and the two read against each other as mirrors.

`test_session_playback.cpp` covers the switch: the arrangement stops at the launch sample and not a block boundary, it is carried down rather than cut off, a launch on a block's first sample is the same switch, a stopped slot goes on holding the track, releasing gives it back mid-material, a slot stopping mid-block carries its own last sample down, the sum of a track's two sections never carries both their materials, and the MIDI hand-over sounds what was due and owes offs for what it started.

One existing assertion changed: a session stop asserted a hard cut to zero at the stop sample, and is now the ramp. That is the slice.

Full engine suite green: 809 cases, 522k assertions.

### DIFF
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -24,8 +24,8 @@ ClipAudioSource::ClipAudioSource(TrackId trackId, ClipSnapshotFeed& clips, ClipS
     : trackId_(trackId), clips_(clips), streams_(streams) {}
 
 ClipAudioSource::ClipAudioSource(TrackId trackId, ClipSnapshotFeed& clips, ClipStreamFeed& streams,
-                                 LaunchHandleFeed& handles)
-    : trackId_(trackId), clips_(clips), streams_(streams), handles_(&handles) {}
+                                 LaunchHandleFeed& handles, Section section)
+    : trackId_(trackId), clips_(clips), streams_(streams), handles_(&handles), section_(section) {}
 
 void ClipAudioSource::prepare(const RenderContext& context) {
     // Longer than a block, because a clip playing faster than its file consumes
@@ -104,6 +104,11 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
 
     // The status is the launcher's, worked out before anything rendered
     // (SessionLauncher.hpp), so this source and the MIDI one see the same block.
+    // Whether anything is still sounding when the block ends, gathered in the
+    // one walk that renders: a track with a slot still going has not stepped,
+    // whatever its other slots did.
+    auto stillSounding = false;
+
     const auto renderSlot = [&](const SessionSlotPlayback& slot, const SplitStatus& status) {
         const auto play = [&](const BlockPiece& piece, int offset, int count) {
             if (count <= 0 || !piece.origin)
@@ -114,20 +119,103 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
             gather(slot.audio, material, offset, streams, sounding, soundingCount);
         };
 
-        // What is on the far side of the event is not rendered. The ramp
-        // across that boundary is #2302's.
         const auto split = splitSample(block, status).value;
 
         play(status.beforeEvent, 0, split);
 
         if (status.afterEvent)
             play(*status.afterEvent, split, block.numSamples - split);
+
+        // Where this slot stopped sounding, for the ramp that takes the step
+        // out of it. A slot still sounding at the end of the block leaves no
+        // step, and one that stopped in an earlier block has none left here.
+        if (status.beforeEvent.playing() && !status.playingAtEnd())
+            sessionSilentFrom_ = std::max(sessionSilentFrom_.value_or(0), split);
+
+        stillSounding = stillSounding || status.playingAtEnd();
     };
 
     forEachSlot(*handles.get(), track, renderSlot);
+
+    // The ramp is bound to the same grain as the signal it corrects, which is
+    // the track: the op is per track because a track sounds one session clip at
+    // a time, so a sum that did not step needs no correction.
+    if (stillSounding)
+        sessionSilentFrom_.reset();
 }
 
 void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float> out) {
+    // Cleared for the block rather than by whatever gathers, because every path
+    // through the render can return before the gather does: a stale edge would
+    // de-click a second time, blocks after the stop it belongs to.
+    sessionSilentFrom_.reset();
+
+    renderMaterial(block, out);
+
+    switch (section_) {
+        case Section::Arrangement:
+            // Only a source that shares its track with a session has a section
+            // to lose it to. One built without the feed is the whole track.
+            if (handles_ != nullptr)
+                applySectionHold(out);
+            break;
+
+        case Section::Session:
+            if (sessionSilentFrom_) {
+                sessionStop_.begin(out, *sessionSilentFrom_, kSectionDeClickSamples);
+            } else {
+                sessionStop_.advance(out);
+
+                // What sounded, remembered for a stop that lands on the first
+                // sample of some later block.
+                sessionStop_.push(out);
+            }
+            break;
+    }
+}
+
+void ClipAudioSource::applySectionHold(juce::dsp::AudioBlock<float> out) {
+    const LaunchHandleFeed::Reader handles(*handles_);
+    const auto hold = handles ? sectionHold(*handles.get(), trackId_) : SectionHold{};
+
+    if (!hold.session) {
+        // The track is the arrangement's. Coming back to it lands mid-material,
+        // which is the same step a voice starting mid-file leaves and comes out
+        // the same way.
+        if (wasHeld_)
+            handBack_.begin(out, kSectionDeClickSamples);
+        else
+            handBack_.advance(out);
+
+        // A hand-over taken back inside its own ramp finishes into the material
+        // rather than being cut off, which would be the click it exists to
+        // remove.
+        handOver_.advance(out);
+        handOver_.push(out);
+
+        wasHeld_ = false;
+        return;
+    }
+
+    // The session has the track. What the arrangement rendered still advanced
+    // every voice, stream and stretcher, which is what makes handing the track
+    // back land where the timeline says rather than where the arrangement was
+    // when it lost it. Only the output is dropped.
+    const auto numSamples = static_cast<int>(out.getNumSamples());
+    const auto from = wasHeld_ ? 0 : std::clamp(hold.takenAt.value, 0, numSamples);
+
+    out.getSubBlock(static_cast<std::size_t>(from)).clear();
+
+    if (wasHeld_)
+        handOver_.advance(out);
+    else
+        handOver_.begin(out, from, kSectionDeClickSamples);
+
+    handBack_.reset();
+    wasHeld_ = true;
+}
+
+void ClipAudioSource::renderMaterial(const BlockInfo& block, juce::dsp::AudioBlock<float> out) {
     out.clear();
 
     const ClipStreamFeed::Reader streams(streams_);
@@ -195,10 +283,10 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
     std::array<Sounding, kMaxVoicesPerTrack> sounding;
     auto soundingCount = 0;
 
-    if (handles_ == nullptr)
-        gather(track->audio, lane, 0, table, sounding, soundingCount);
-    else
+    if (section_ == Section::Session)
         gatherSession(*track, lane, table, sounding, soundingCount);
+    else
+        gather(track->audio, lane, 0, table, sounding, soundingCount);
 
     for (auto& voice : voices_) {
         if (voice.clipId() == INVALID_CLIP_ID)
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -161,16 +161,15 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
 }
 
 void ClipAudioSource::applySectionHold(juce::dsp::AudioBlock<float> out) {
-    const LaunchHandleFeed::Reader handles(*handles_);
-    const auto hold = handles ? sectionHold(*handles.get(), trackId_) : SectionHold{};
-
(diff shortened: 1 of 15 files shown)


magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -114,20 +119,103 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
...
+
+        stillSounding = stillSounding || status.playingAtEnd();
     };
 
     forEachSlot(*handles.get(), track, renderSlot);
+
+    // The ramp is bound to the same grain as the signal it corrects, which is
+    // the track: the op is per track because a track sounds one session clip at
+    // a time, so a sum that did not step needs no correction.
+    if (stillSounding)
```
lucaromagnoli: **P1 — Preserve the outgoing ramp during a session-slot switch**

When one slot stops and another starts in the same block, `stillSounding` clears `sessionSilentFrom_`. The incoming clip’s start ramp only corrects its own edge; it does not carry the outgoing clip’s last sample down, so the track sum still jumps from the old signal to the new one and can click.

Please retain/apply the outgoing `StopDeClick` when a slot stops even if another slot is playing at the end of the block, and cover a same-sample old-slot stop/new-slot launch with different signal levels.

— Codex

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -586,14 +587,54 @@ void ClipMidiSource::render(const BlockInfo& block, juce::MidiBuffer& out) {
...
+            }
+
+            wasHeld_ = true;
+            lastSnapshot_ = live;
+            return;
+        }
+
+        // Taken back. Nothing sounded while the session held it, so there is
+        // nothing to chase and the lane simply resumes where the timeline is.
+        wasHeld_ = false;
```
lucaromagnoli: **P1 — Chase arrangement MIDI when taking the section back**

The hand-over calls `endAll`, so the arrangement source has no active notes while the session holds the track. On release this only clears `wasHeld_`; `playLane` then sees the transport’s continuous block and skips its chase path. If the timeline is already inside a sustained arrangement note, that note—and the current controller state—remains absent until a later event happens naturally.

Please treat the hand-back as a discontinuity or explicitly chase the arrangement lane at the block start. A regression test should release while the timeline is already inside a sustained note, rather than before its onset.

— Codex

magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -114,20 +122,101 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
...
+
+        case Section::Session:
+            if (sessionSilentFrom_) {
+                sessionStop_.begin(out, *sessionSilentFrom_, kSectionDeClickSamples);
+            } else {
+                sessionStop_.advance(out);
+
+                // What sounded, remembered for a stop that lands on the first
+                // sample of some later block.
+                sessionStop_.push(out);
```
lucaromagnoli: **P1 — Do not overwrite an unfinished stop ramp’s held sample**

When a stop ramp spans multiple continuation blocks, this `push` replaces `StopDeClick::held_` with the last sample of each partially rendered ramp. The following block then multiplies the remaining cosine by that already-decayed value instead of by the original discontinuity, producing another step at the block seam.

For example, with the 32-sample section ramp and 8-sample blocks, `advance` renders eight samples correctly, then `push` stores the eighth ramp sample; the next eight samples decay from that reduced value. `RenderContext` explicitly permits blocks shorter than `maxBlockSize`, so this makes output depend on callback partitioning. Please only refresh the remembered sample once `sessionStop_` is inactive (or otherwise keep the original held value until the ramp completes), and add a source-level regression using blocks shorter than 32 samples.

— Codex

magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -114,20 +122,116 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
...
+            break;
+
+        case Section::Session:
+            // A stop landing on this block's first sample. The handle reports
+            // it as a block that was never playing, which is the same
+            // SplitStatus as a slot that stopped long ago, so the edge is only
+            // visible from here: something sounded last block and nothing does
+            // now. The value it steps down from is the one push remembered,
+            // which is what push is for.
+            if (!sessionSilentFrom_ && soundedLastBlock && !sessionSounded_)
```
lucaromagnoli: **P1 — Preserve a boundary stop when another slot starts there**

This detects a stop on sample zero only when the whole session becomes silent. If one slot stops and its replacement starts on that same block boundary, the old handle reports no playing `beforeEvent`, while the new handle sets `sessionSounded_` true. The old stop is therefore invisible both here and to the earlier `sessionSilentFrom_` check, so the output jumps directly from the outgoing slot’s last sample to the incoming slot’s first sample without carrying the outgoing edge down.

This is the boundary-aligned version of the slot-switch case fixed in the previous commit. Please track whether the same slot continued (or otherwise retain the boundary stop independently of whether another slot now sounds), and add a regression with old `stop(std::nullopt)` plus new `play(std::nullopt)` before the same advance.

— Codex

magda/engine/clip/FadeCurves.cpp
```diff
@@ -120,4 +120,102 @@ void StartDeClick::applyFrom(juce::dsp::AudioBlock<float> audio, int alreadyDone
...
+    // signal stopped at, and the buffer it was just written into now holds that
+    // value part way down. Taking the remembered sample from there would
+    // multiply the rest of the cosine by an already decayed number, so the ramp
+    // would bend at every block seam and a stop would come out differently at
+    // 8 samples a block and at 1024.
+    //
+    // Refusing here rather than at the callers, because every caller pushes
+    // unconditionally every block: which of those blocks a ramp happens to be
+    // running through is exactly what they should not have to know.
+    if (active())
```
lucaromagnoli: **P1 — Keep the last output sample separately from the active ramp anchor**

Skipping `push` while active preserves the current ramp’s anchor, but it also discards the actual last sample of the rendered block. If another stop lands on sample zero before this ramp completes, `begin` calls `hold(audio, 0)`, which cannot refresh anything, and starts the new ramp from the first stop’s original `held_` value rather than from the sample immediately before the new stop.

This occurs when slot A hands over to B near a block end, the outgoing-A correction continues into a short next block, and B stops on the following block boundary. The previous sample is B plus the partially decayed A correction, but the new tail starts from A’s original level, creating another discontinuity. The existing second-stop test only reanchors at a nonzero offset inside the same block.

Please retain an always-updated last-output sample separately from the value anchoring the active ramp, so `begin(..., 0, ...)` can reanchor correctly, and cover a second stop on a boundary while the first ramp is still active.

— Codex

magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -114,20 +116,102 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
...
+        wasHeld_ = false;
+        return;
+    }
+
+    // The session has the track. What the arrangement rendered still advanced
+    // every voice, stream and stretcher, which is what makes handing the track
+    // back land where the timeline says rather than where the arrangement was
+    // when it lost it. Only the output is dropped.
+    const auto numSamples = static_cast<int>(out.getNumSamples());
+    const auto from = wasHeld_ ? 0 : std::clamp(hold.takenAt.value, 0, numSamples);
```
lucaromagnoli: **P1 — Represent a release followed by reacquisition in the same block**

`SectionHold` reports only who owns the track at the end of the block and one `takenAt` sample. If the session held the track last block, Back to Arrangement releases it at sample zero, and another slot launches later in this block, the fold reports `session == true` with the later launch sample. This branch nevertheless uses zero because `wasHeld_` was true, so the arrangement is muted for the whole block instead of sounding from sample zero until the relaunch. The MIDI source makes the analogous early return and loses the same interval.

This can occur when a pending quantized launch remains while the current slot is released, or when release and launch requests arrive before the same advance. Please represent the block’s ownership intervals/transitions explicitly rather than combining final state with source-local previous state, and add a release-at-zero/relaunch-mid-block regression for both audio and MIDI.

— Codex

magda/engine/launch/LaunchHandle.cpp
```diff
@@ -170,12 +170,28 @@ std::optional<BeatRange> LaunchHandle::lastPlayedRange() const {
...
 
+void LaunchHandle::releaseSection() {
+    // A request rather than a state change, like every other way a slot stops.
+    // The advance is what turns a stop into an edge a source can see and ramp;
+    // a run ended behind their backs is a step nobody is told about, and the
+    // slot would cut off rather than decay (#2344 review). It also replaces a
+    // queued launch, so nothing fires after the track has been handed back.
+    stop(std::nullopt);
+
+    releasePending_ = true;
```
lucaromagnoli: **P1 — Keep the release stop atomic with the ownership release**

`releaseSection()` represents one action in two independently replaceable fields: the stop is stored in `pending_`, while the ownership change is stored in `releasePending_`. If this playing handle receives `play(futureBeat)` before the next `advance()`, `play()` replaces the queued stop but leaves `releasePending_` set. The advance then hands ownership to the arrangement at sample 0 while the old session run remains in `beforeEvent` until the future relaunch, so both sections render during the supposed arrangement gap.

Please make the stop and ownership release one atomic pending transition, or ensure replacement updates both halves consistently. The regression should use release and relaunch on the same handle; the current tests use two handles and do not exercise this split state.

— Codex

magda/engine/clip/ClipAudioSource.cpp
```diff
@@ -114,20 +116,87 @@ void ClipAudioSource::gatherSession(const TrackClipPlayback& track, const BlockI
...
+    // it to. One built without the feed is the whole track.
+    if (section_ == Section::Arrangement && handles_ != nullptr)
+        applySectionHold(out);
+}
+
+void ClipAudioSource::applySectionHold(juce::dsp::AudioBlock<float> out) {
+    const auto numSamples = static_cast<int>(out.getNumSamples());
+
+    const LaunchHandleFeed::Reader handles(*handles_);
+    const auto hold = handles ? sectionHold(*handles.get(), trackId_, numSamples) : SectionHold{};
```
lucaromagnoli: **P2 — Default a missing handle snapshot to arrangement ownership**

`SectionHold{}` now has `until == 0`, so this fallback treats a null `LaunchHandleFeed::Reader` as though the session owned the entire block. The feed contract explicitly permits null until its first publication; with no handle table, no session handle can own the track, so the arrangement must keep sounding. As written, the audio path clears the whole block, and the identical fallback in `ClipMidiSource` suppresses all arrangement MIDI.

Please make the fallback span the complete block (preferably through an explicit arrangement-owned factory rather than a default whose length cannot be known), and cover both arrangement audio and MIDI with a clip snapshot present but no handle table published.

— Codex

PR #2349, opened 2026-09-03
Make the plan the only implementation of MIDI in thru (#2345)

### LINKED ISSUE #2345: MIDI thru is implemented twice on the native leg, so a thru device plays every note twice
## Problem

`DeviceInfo::midiInThru` has two implementations, and on the native leg both fire for the same device.

**The plan.** `PlanCompiler::emitDevice` wraps a MIDI-producing device whose node `passesRawMidiInput()` in a `MergeMidi{signal.midi, deviceMidi}` op with role `ChainMidiMerge` (`PlanCompiler.cpp:936-943`), from `MidiOutputPolicy::MergeRawInputAndPluginOutput` (`ChainRoutingModel.hpp:90-92`).

**The device.** The SDK hands a device one buffer for both directions. `EngineMagdaDevice::process` decodes `block.midiIn` into its scratch (`EngineMagdaDevice.cpp:360-375`) and writes the whole scratch back to `block.midiOut` (`:421-431`). `StepSequencerPlugin::process` clears that buffer only when its own `midiThru` flag is false (`StepSequencerPlugin.cpp:345`), and `PolyStepSequencerPlugin` likewise (`:362`). So with the flag on, the device's output port carries **input + its own notes**.

The flag is not a live copy of the model's: nothing on the native leg syncs it. `createEngineDevice` calls `restoreSavedState` (`EngineDeviceFactory.cpp:177`) and both sequencers restore `seqMidiThru` from saved plugin state (`StepSequencerPlugin.cpp:223-224`), so the native device runs on whatever the TE leg last saved — a stale second copy of a fact the model owns. The only writer is `syncPluginMidiInThru` (`MidiInThruSync.hpp`), which takes a `te::Plugin*` and is reachable only from TE-leg code.

## Measured

A media track (armed, `midiInputDevice = "midi-hardware"`) with a step sequencer followed by an instrument. One note-on fed from the live MIDI input, one block, counting note-ons reaching the instrument:

| model `midiInThru` | device flag | note-ons downstream |
|---|---|---|
| on | on | **2** |
| on | off | 1 — correct |
| off | on | 1 — thru happens although the toggle is off |
| off | off | 0 — correct |

So the toggle does not turn thru on; it adds a second copy. Thru happens either way.

The plan dump confirms the shape: `Device role deviceProcess dev 10` followed by `MergeMidi role chainMidiMerge dev 10`.

## Why the plan should own it, not the device

- **A host cannot make a third-party plugin pass its input through.** Every JUCE format replaces the host buffer with the plugin's declared MIDI output — VST3 does `midiMessages.clear(); MidiEventList::toMidiBuffer(...)` unconditionally (`juce_VST3PluginFormatImpl.h:2512`), VST2 likewise, AU when it wants or produces MIDI. The host's only way to offer thru is to add the raw stream itself, which is the plan's merge. Device-owned thru would therefore always need the plan's merge as well for externals — two mechanisms, which is exactly the arrangement that broke here.
- **The flag is already decorative for most MIDI devices.** `ArpeggiatorPlugin.cpp:414`, `MidiStrumPlugin.cpp:353` and `MidiChordEnginePlugin.cpp:49` clear unconditionally. Only the two sequencers honour it, so one model flag means two different things across the fleet.
- **The plan is blind to a device that passes through.** Everything downstream reasons from "a Device op's MIDI output port carries what the device produced". With internal thru that is false, and every computation over it is wrong — see #2341, where the byte-budget reservation compounds to `2B + 4096` per stage because each port secretly contains its input.
- **The MIDI-sidechain case is already plan-owned** (`ChainRoutingModel.hpp:88-89`, `PlanCompiler.cpp:777-781`), so this makes the two consistent.

## Direction

- Both sequencers clear their MIDI buffer unconditionally, as arp/strum/chord already do: `StepSequencerPlugin.cpp:345` and `PolyStepSequencerPlugin.cpp:362`.
- Delete the `midiThru` atomic from both, drop `seqMidiThru` from save (`StepSequencerPlugin.cpp:187`, `PolyStepSequencerPlugin.cpp:172`) while still tolerating it on load, and remove `MidiInThruSync.hpp` with its call sites (`AudioBridge.cpp:610`, `PluginManagerSync.cpp:2403`, `RackSyncManager.cpp:753`, `:1063`).
- Thru is then the plan's `ChainMidiMerge` alone, on the toggle the UI already writes.

**The feature is unchanged**: with the toggle on, a keyboard still plays the instrument after the sequencer — via the merge rather than via the device. It just stops arriving twice.

**This removes thru on the TE leg**, where the device flag is the only implementation. Accepted deliberately given the migration to the native engine; TE will not get a plan equivalent.

## Acceptance criteria

- On the native leg, one note into a track with a thru-enabled sequencer reaches the next device exactly once.
- With the toggle off, it does not reach the next device at all.
- No device-level MIDI thru state remains in the model, the plugin state, or the device.
- A project saved with `seqMidiThru` still loads.
- The null-diff corpus gains a `midiInThru = true` case with a MIDI-producing device; it currently has none, and the one MIDI-out case sets `relay.midiInThru = false` (`tests/NullDiffCorpus.cpp:2378`).

## Fallout

- `tests/midi/test_step_sequencer_midi_thru.cpp` asserts the device-level behaviour and is replaced by a plan-level equivalent.
- The `seq->midiThru` expectations in `tests/midi/test_midi_signal_routing_juce.cpp:1203-1347` go.
- #2341 / PR #2343 shrinks: `OpKind::Device` comes out of the forward pass's conduit set, since a device port becomes one producer's worth again. Per-stage reservation goes from `2B + 4096` to `B + 4096`, which resolves the overflow raised in review there.

## Background

Found reviewing PR #2343 (#2341). The exponential reservation flagged there is an honest count of this duplication compounding down a chain: 12 chained thru devices reserve 16 MB on the last port because they genuinely emit 2^12 copies of each input event.

lucaromagnoli:
Closes #2345.

## What was wrong

`DeviceInfo::midiInThru` had two implementations and on the native leg both fired for the same device.

The plan wraps a MIDI-producing device whose node `passesRawMidiInput()` in a `MergeMidi` op behind it (`OpRole::ChainMidiMerge`). The device did it too: the SDK hands a device one buffer for both directions, and both step sequencers left their input in it when their own `midiThru` flag was on. That flag was not a live copy of the model's — nothing on the native leg wrote it, so the device ran on whatever the TE leg last saved.

| model `midiInThru` | device flag | note-ons downstream |
|---|---|---|
| on | on | **2** |
| on | off | 1 — correct |
| off | on | 1 — thru although the toggle is off |
| off | off | 0 — correct |

The toggle did not turn thru on; it added a second copy.

## Why the plan owns it

- **A host cannot make a plugin pass its input through.** Every JUCE format replaces the host's buffer with the plugin's declared MIDI output. The host's only way to offer thru is to add the raw stream itself, which is the merge — so device-owned thru would always need the merge as well for externals. Two mechanisms, which is exactly what broke here.
- **The flag was already decorative for most MIDI devices.** The arpeggiator, strum and chord engine clear unconditionally; only the two sequencers honoured it, so one model flag meant two different things across the fleet.
- **The plan was blind to a device that passes through.** Everything downstream reasons from "a Device op's MIDI output port carries what the device produced". With internal thru that is false, and every computation over it is wrong.

## Changes

- Both sequencers clear their MIDI buffer unconditionally. The `midiThru` atomic, `SettingIDs::midiThru` and the `seqMidiThru` save are gone; the property is ignored on load, so a project carrying one still opens.
- `MidiInThruSync.hpp` deleted with its four call sites (`AudioBridge`, `PluginManagerSync`, `RackSyncManager` ×2).
- `OpKind::Device` comes out of the forward pass's conduit set in `PlanExecutor::prepare` — a device port is one producer's worth again. Per-stage reservation goes from `2B + 4096` to `B + 4096`, which resolves the overflow raised in review on #2343: twelve chained thru devices no longer reserve 16 MB on the last port, because they no longer emit 2^12 copies of each input event.
- Contract comments rewritten where they described device-owned thru: `EngineDevice.hpp`, both engine adapters, `PlanExecutor`, and `ChainRoutingModel::makeRoutingNode`, where the invariant now reads.

`setMidiOutputBoundBytes` stays. The input and output figures are now independent for a different reason than before — a device may read a merged stream far larger than one producer's budget and still emit only its own — and `EngineMagdaDevice` sizes its scratch from the wider of the two.

## Tests

- `tests/midi/test_step_sequencer_midi_thru.cpp` deleted, replaced by a plan-level test in `test_render_plan_executor.cpp`: with thru on the chain's note reaches the next device exactly once beside the device's own; with it off, not at all.
- `test_engine_device_layer.cpp`: the two thru-based tests reworked into "a step sequencer's MIDI output is its own" (mono and poly, densest legal block in, none of it out) and "a device nothing feeds still emits its whole producer's worth". The `outBound` expectation is now `kMaxMidiBytesPerPort`.
- `test_midi_signal_routing_juce.cpp`: the three device-thru tests replaced by one that loads a project carrying `seqMidiThru` and checks it neither blocks the settings saved beside it nor reappears on the way out.
- The null-diff corpus gains `plugin.instrument.midiout.thru` — a `midiInThru = true` case with a MIDI-producing device, which it had none of. Both legs agree.

Each of the three was verified negatively: re-adding the device's conditional clear fails the device test, putting `OpKind::Device` back in the conduit set fails the `outBound` test, and removing the compiler's merge fails the plan test.

## Behaviour

The feature is unchanged. With the toggle on, a keyboard still plays the instrument after the sequencer — via the merge rather than via the device. It just stops arriving twice.

**This removes thru on the TE leg for a device that is not wrapped in an instrument rack**, which is where the device flag was the only implementation — in practice the two step sequencers. Accepted deliberately given the migration to the native engine; racked devices keep it through `InstrumentRackManager`, which is what the new corpus case covers, and TE will not get a plan equivalent.

## Results

Full Catch2 suite green (3421 passed, 13 skipped). JUCE suites MIDI Signal Routing, Null Diff Maps, Null Diff Suite Rules and Legacy Corpus Device State Migration all green.

Null Diff Corpus and Real Project Corpus each fail their first case (`placement.grid`, `project.v1bounces`) on `tracktion_DeviceManager.cpp:432`. Reproduced on unmodified `origin/dev/0.20.0` — pre-existing and unrelated.


lucaromagnoli:
**[P3] Remove `seqMidiThru` from the reused plugin state** (`StepSequencerPlugin.cpp:185`, and the poly equivalent)

Dropping `setProperty` only prevents the field appearing in a fresh tree. A plugin restored through `devicePluginTreeFromState()` is constructed with `seqMidiThru` already on its live `state`; `TracktionMagdaDevicePlugin::flushPluginStateToValueTree()` passes that same tree to `flushState()`. Since neither sequencer removes the property, `captureInternalDeviceState()` captures it again and every subsequent save preserves the retired device-level state.

The new test misses this lifecycle because it restores from one tree and then calls `seq->flushState()` on a separate fresh tree. Please either remove `seqMidiThru` explicitly in both `flushState()` implementations or filter it during capture, and exercise construction from saved state followed by `captureInternalDeviceState()` in the test.

lucaromagnoli:
Confirmed and fixed in 6e4cf59f — thanks, this was real.

`PluginManager::createInternalPlugin` hands the saved tree straight to `edit.getPluginCache().createNewPlugin(savedState)`, so for a loaded project that tree *is* `plugin.state`. `flushPluginStateToValueTree()` then passes the same one to `flushState()`, and `captureInternalDeviceState()` copies every property on it except the parameter IDs. So `seqMidiThru` survived every save, and not writing it was never going to be enough.

Both sequencers now `removeProperty` it in `flushState()`, against a file-local `kRetiredMidiThru` with the reasoning beside it.

The test loads the way a project does — `createNewPlugin(saved)` rather than a restore onto a fresh plugin — then captures through `captureInternalDeviceState()` and reads back with `devicePluginTreeFromState()`, asserting `seqMidiThru` is gone and `seqRampCycles` survived the same round trip. Verified negatively: with the `removeProperty` taken out, the test fails on exactly that assertion.

Full Catch2 suite green; MIDI Signal Routing and Legacy Corpus Device State Migration green.

PR #2350, opened 2026-09-03
Port the Faust Instrument to MagdaDevice (#2315)

### LINKED ISSUE #2315: Port the Faust Instrument to MagdaDevice

lucaromagnoli:
Closes #2315. Part of #2299.

## The port

`FaustInstrumentPlugin` was `te::Plugin` only, so a project using it rendered as passthrough on the native leg — `createSdkDevice()` returned null and the executor bound a `Passthrough` for the Device op. It is a `MagdaDevice` now, following the runtime Faust effect that crossed in #2196.

- `properties` / `prepare` / `release` / `reset` / `process` replace `initialise` / `deinitialise` / `applyToBuffer`.
- The 64 pool slots become relaxed atomics read on the audio thread instead of `te::AutomatableParameter` + `CachedValue`. The three host-owned controls (voice mode, glide, bend range) keep their indices past the end of the pool, so a patch still has all 64 of its own and nothing a macro, mod or automation lane pinned to moves.
- State goes through `flushState` / `restoreState` under the spellings the retired plugin registered, so a saved project reads back onto the same controls.
- Voice allocation, glide, bend, the mono note stack and the retire timer are unchanged. This is a port, not a rewrite, and the corpus case below is what says so.

The compiled factory and its voices now live for as long as the device does — the runtime store's lifetime rather than the plan's — so recompiling a plan because a fader moved never rebuilds a voice mid-note.

## The editor was unreachable through the wrapper

`dynamic_cast<IFaustEditorModel*>` on a `te::Plugin` has returned null since the Faust **effect** became a `MagdaDevice`: `TracktionMagdaDevicePlugin` is `final` and is not that interface. The instrument needs the unwrap, and the same `deviceFromPlugin<IFaustEditorModel>` call repairs the effect's inline editor and `PluginApiLive::applyFaustSource` at the same time (`DeviceSlotInlineUiFactory.cpp` ×2, `plugin_api_live.cpp`).

## Null-diff, and what it found

The issue asks for this to be null-diffed against the TE leg, and the corpus had no Faust case at all. `plugin.faust.instrument` renders overlapping chords through a deliberately trivial patch — a gated DC per voice, no oscillator, no envelope — so the output is a staircase and any difference can only come from *when* a gate flips.

It failed on the first run: `peak=-36.2 rms=-74.6 shift=0.000`. The cause is not the port.

**The fork ends every note 0.0001 s early.** `MidiNote::getPlaybackTime` subtracts it — *"nudge the note-up backwards just a bit to make sure the ordering is correct"* (`tracktion_MidiList.cpp:829-831`) — which at 44.1 kHz is 4.41 samples, so after rounding the fork releases four samples before the engine does. The engine orders its events when it compiles them and keeps the length the note was drawn at.

The corpus has always known the figure: `Case::incumbentNoteEndEarlySeconds` (`NullDiffCase.hpp:208`) declares it, with that explanation. Its only consumer was the MIDI comparison (`NullDiffRunner.cpp:421`). The audio comparison could not see it, and until now never needed to — every corpus instrument answers a note-on with an impulse and ignores the off (`NullDiffHostedPlugin.cpp` `processVoice`), so no rendered audio depended on when a note ended. A sustaining synth is the first case that hears it.

Every divergence was a contiguous four-sample window ending exactly on an authored note-off, with a magnitude equal to the sum of the released notes' DC levels, and **no divergence at any note-on**.

So the audio comparison learns the fact the corpus already wrote down:

- `AudioCompareOptions::excludedRanges` — half-open sample ranges the residual is not taken over, counted out of `comparedSamples`.
- `noteEndNudgeRanges()` derives them from the authored notes and the declared constant: `[end - nudge, end)` per note, clipped by the clip end, sorted and merged.
- `Case::audioChangesAtNoteEnds` opts a case in, with the mechanism recorded beside it.

This is a change to the shape of the comparison, not a widened tolerance. The samples are named per note rather than found in the residual, and every other sample of the case — note-ons included — is still held to bit identity. Result: `peak=-inf`.

## Results

`plugin.faust.instrument` reports `peak=-inf` on both legs. Verified negatively: with `audioChangesAtNoteEnds` off, `peak=-36.2 rms=-74.6` comes back exactly.

Full Catch2 suite green (3421 passed, 13 skipped). JUCE suites **Null Diff Corpus**, **Faust Instrument Project Tempo**, **Faust Instrument Voice Mode** and **MIDI Signal Routing** all clean. `placement.grid` still fails on a `tracktion_DeviceManager.cpp:432` assertion — reproduced on unmodified `dev/0.20.0`, pre-existing and unrelated.

## Follow-ups filed

- #2351 - the mono held-note stack has no dedup, where the compiled poly instrument erases a matching pitch first. Pre-existing, not introduced here.
- #2352 - the corpus has no case for `MagdaCompiledPolyInstrument`, which would have hit the same upstream divergence. The harness change above is what it needs; it is now waiting on the case.

### DIFF
--- a/magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
+++ b/magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
@@ -959,6 +959,10 @@ void FaustInstrumentPlugin::restoreState(const juce::ValueTree& v) {
             stageSourceForEditing(savedName.isNotEmpty() ? savedName : juce::String("Loaded"),
                                   savedSource);
         }
+    } else if (savedName.isNotEmpty()) {
+        // The name is state of its own: a patch renamed without being edited
+        // needs no recompile and would otherwise keep the default's name.
+        dspName_ = savedName;
     }
 
     // Stable ids (param_01 ... param_64), so a macro or automation lane keeps
--- a/magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
+++ b/magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
@@ -431,10 +431,8 @@ std::shared_ptr<FaustInstrumentPlugin::FaustState> FaustInstrumentPlugin::compil
 }
 
 int FaustInstrumentPlugin::readVoiceMode() const {
-    if (!voiceModeParam_)
-        return Poly;
     // The parameter is normalised 0..1 over three modes.
-    const float real = voiceModeParam_->getCurrentValue() * 2.0f;
+    const float real = hostValue(kVoiceModeParamIndex) * 2.0f;
     return juce::jlimit(0, 2, static_cast<int>(std::lround(real)));
 }
 
@@ -463,8 +461,7 @@ void FaustInstrumentPlugin::resetAllVoices(const std::shared_ptr<FaustState>& st
 float FaustInstrumentPlugin::readBendRatio() const {
     if (bendNormalised_ == 0.0f)
         return 1.0f;
-    const float semitones =
-        (bendRangeParam_ ? bendRangeParam_->getCurrentValue() : 0.0f) * kMaxBendSemitones;
+    const float semitones = hostValue(kBendRangeParamIndex) * kMaxBendSemitones;
     return std::pow(2.0f, bendNormalised_ * semitones / 12.0f);
 }
 
@@ -554,7 +551,6 @@ void FaustInstrumentPlugin::handleMonoNoteOff(const std::shared_ptr<FaustState>&
 void FaustInstrumentPlugin::initialiseUnsetPoolValues(
     const std::vector<FaustParamPool::ActiveBindingDescriptor>& bindings,
     const std::array<FaustParamSlot, FaustParamPool::kSize>& previousSlots) {
-    auto* um = getUndoManager();
     for (const auto& binding : bindings) {
         const int slotIndex = binding.slotIndex;
         if (slotIndex < 0 || slotIndex >= FaustParamPool::kSize)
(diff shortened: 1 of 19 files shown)


magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
```diff
@@ -567,74 +563,38 @@ void FaustInstrumentPlugin::initialiseUnsetPoolValues(
...
             continue;
 
-        const float normalisedDefault = normaliseDefaultForSlot(pool_.slot(slotIndex));
-        auto& cached = poolCached_[static_cast<size_t>(slotIndex)];
-        cached.setValue(normalisedDefault, um);
-
-        auto& param = poolParams_[static_cast<size_t>(slotIndex)];
-        if (param)
-            param->updateFromAttachedValue();
+        poolValues_[static_cast<size_t>(slotIndex)].store(
```
lucaromagnoli, line 566: [P1] This now resets only the device-side atomic. The Tracktion wrapper's AutomatableParameter/CachedValue for the same slot remains at the prior patch value, and syncParametersToDevice() writes that stale value back before the next process call. refreshDeviceParameters() also reads the host value before capture, so a newly bound control can inherit the old patch's normalized value instead of its own default. The retired implementation updated both the CachedValue and AutomatableParameter; this port needs equivalent device-to-host synchronization when the control identity changes.

magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
```diff
@@ -971,20 +953,71 @@ void FaustInstrumentPlugin::restorePluginStateFromValueTree(const juce::ValueTre
...
+
+void FaustInstrumentPlugin::refreshPoolDomains() {
+    for (int i = 0; i < FaustParamPool::kSize; ++i)
+        poolDomains_[static_cast<size_t>(i)] =
+            ParameterUtils::domainOf(paramInfoFromSlot(pool_.slot(i)));
+}
+
+ParameterInfo FaustInstrumentPlugin::parameterInfo(int index) const {
+    if (index >= FaustParamPool::kSize && index < parameterCount())
+        return faustInstrumentHostParamInfo(index - FaustParamPool::kSize);
```
lucaromagnoli, line 997: [P1] These host ParameterInfos have no stableId, so the generic wrapper registers faust-instrument_param_64 through _66 instead of the released voiceMode, glide, and bendRange IDs. restoreState() reads the legacy properties into the atomics, but the wrapper parameters remain at defaults and syncParametersToDevice() overwrites the restored values. Please assign the legacy IDs as the stable IDs for these three parameters.

magda/daw/audio/plugins/FaustInstrumentPlugin.cpp
```diff
@@ -567,74 +563,38 @@ void FaustInstrumentPlugin::initialiseUnsetPoolValues(
...
-    if (!compiled) {
-        DBG("FaustInstrumentPlugin: failed to compile saved source: " << err << " - using default");
-        compiled = compileAndRebind(kDefaultDspSource, err);
-    }
+    auto compiled = compileAndRebind(kDefaultDspSource, err);
 
-    dspSource_ = savedSource.isNotEmpty() ? savedSource : juce::String(kDefaultDspSource);
-    dspName_ = savedName.isNotEmpty() ? savedName : juce::String("Simple Synth");
+    dspSource_ = kDefaultDspSource;
+    dspName_ = "Simple Synth";
```
lucaromagnoli, line 591: [P2] The constructor now always seeds the display name as Simple Synth, while restoreState() only applies savedName when savedSource differs from dspSource_. A state containing this same DSP source under a renamed display name therefore never restores that name. Please restore a non-empty savedName independently of whether recompilation is needed.

lucaromagnoli:
All three fixed in `90760dd1`. The stableId one led to two more.

**[P1] stableId** — right, and worse than stated: the wrapper keyed the three on `faust-instrument_param_64..66` while `flushState`/`restoreState` used `voiceMode`/`glide`/`bendRange`, so a saved value was never read back *and* the two legs disagreed — the native leg restored it, the fork did not. They carry the released ids now.

**[P1] identity change** — added `TracktionMagdaDevicePlugin::pullParametersFromDevice()`, called from `refreshDeviceParameters()`, which is already what runs after a compile. Message thread only: `syncParametersToDevice()` runs from `applyToBuffer`, so the pull could not live there.

**[P2] display name** — `restoreState` now takes a non-empty `savedName` whether or not a recompile is needed.

The test I wrote for the first one then failed on two more, both in the wrapper and both affecting **every** ported device, not just this one:

- `flushPluginStateToValueTree()` asked the device to flush values the host had never pushed into it — stale since construction if the plugin has not rendered — so a save wrote those over what the parameters held. It calls `syncParametersToDevice()` first now.
- `restorePluginStateFromValueTree()` set the `CachedValue` and left the parameter caching its old one, so a restore reached the tree and nothing else, and the next sync pushed the stale value back. It calls `updateFromAttachedValue()`, which the retired implementation did and the wrapper had dropped.

New suite **Faust Instrument Host Parameter State Tests** covers it: the ids are the released ones, a non-default value survives a flush and reload, and it arrives at both the host parameter and the device. 18 assertions, and it failed on each of the three bugs in turn while I was fixing them.

Catch2 3421 passed. JUCE: Null Diff Corpus (`plugin.faust.instrument` still `peak=-inf`), both Faust Instrument suites, MIDI Signal Routing, Legacy Corpus Device State Migration all clean. Only `placement.grid` fails, pre-existing on `dev`.

PR #2355, opened 2026-09-04
Port the sampler to MagdaDevice (#2271)

### LINKED ISSUE #2271: The native engine has no sampler, so no Drum Grid project renders through it

lucaromagnoli:
Closes #2271. Part of #1882.

## The port

`MagdaSamplerPlugin` was `te::Plugin` only, so `createSdkDevice()` returned null and the executor bound a `Passthrough` for the Device op. Every Drum Grid pad is one (`DrumGridPads.cpp`), so a drum kit built the ordinary way rendered as passthrough on the native leg.

- `properties` / `prepare` / `release` / `reset` / `process` replace `initialise` / `deinitialise` / `applyToBuffer`.
- The fourteen parameters become normalised atomics behind a frozen slot table. The ids, order and curves are the ones the retired plugin registered, so saved automation, macro links and mod links keep pointing at the same controls. The JUCE skews map onto `ParameterScale::Exponential` by their **reciprocal** — `NormalisableRange::convertFrom0to1` computes `min + span * norm^(1/skew)` and MAGDA's Exponential is `norm^k * span + min`.
- Parameter values are the model's (#2317), so `flushState`/`restoreState` carry only what the device owns: the sample path, the root note and the loop switch, under the spellings the retired plugin used.
- Voice allocation, glide, the mono/legato note stack and the ADSR are untouched. This is a port, not a rewrite, and the corpus case below is what says so.

The sample stays a **path** rather than a blob in device state. Unlike the convolution's impulse response, a sample is arbitrarily large and the project media tree already owns relocating it — which is what `relocateSample` exists for.

The processor becomes a `MagdaDeviceProcessor`, and the registry entry moves to `SavedStateOrFresh` + `createValueTreePlugin` for the same reason the convolution's did: the sample path is device state, so a restore has to rebuild the plugin from it rather than from a fresh tree.

## Both sides of the wrapper have to be written

Eighteen `dynamic_cast<MagdaSamplerPlugin*>` sites become `deviceFromPlugin`, which is the mechanical half. The half that is not mechanical is `SamplerHostBinding.hpp`.

The faceplate, the drum-grid commands and the media relocator all work in display units — seconds, dB, semitones — while the wrapper's parameters are normalised. Neither side can be written on its own:

- The host is the authority every block: `syncParametersToDevice()` pushes on every `applyToBuffer`, so a write straight to the device is undone at the next render.
- The device is what actually holds a value between renders, so a write to the parameter alone is invisible until something renders.

So a write goes to both. And a **device-driven** change travels the other way: loading a sample makes the device derive the trim and loop markers from the audio, and without `pullParametersFromDevice()` the host's stale markers overwrite them at the next render — the pad then plays a region belonging to whatever was loaded before it. `Sampler Relocate Tests` asserts the host parameter carries what the load derived; reverting the pull fails all four markers.

## Null-diff

`plugin.sampler` renders overlapping notes through a sampler whose envelope is flattened to a gate, over staircase material read back at its own rate at root note, so a read head in the wrong place shows as a step landing on the wrong sample rather than as a level change.

**It found something the Faust case could not.** #2350 taught the audio comparison about the fork's note-end nudge — `MidiNote::getPlaybackTime` subtracts 0.0001 s, four samples at 44.1 kHz — and excluded a four-sample window per note. That was sufficient there because the patch has no envelope: the gate flips and the difference ends. An instrument with a **release** runs the same ramp four samples apart, so the whole ramp differs, not just the gap between the two releases. The first run said `peak=-68.9 rms=-78.9`.

- `Case::noteEndReleaseSeconds` declares how long the instrument carries a note-off difference forward. It is the device's number, not the fork's.
- `noteEndNudgeRanges()` widens each window to `[end - nudge, end + release)`.
- The case sets attack, decay and release to the device's minimum and sustain wide open, so that window is 48 samples rather than the 4410 the default release would ask for. The comment says why: the longer it is, the less of the render is held to bit identity.

Result: `peak=-inf`. Verified negatively in **both** directions — with `audioChangesAtNoteEnds` off, and with `noteEndReleaseSeconds` left at zero, `peak=-28.9` comes back each time. Both halves are load-bearing.

`plugin.sampler` also declares the `internalDevices()` DAWproject loss the other internal-device cases do.

## The marker #2271 named for itself

The Drum Grid fixture's `expectedDiagnostics` line is gone from `MgdFixtures.cpp`. That declaration was written to fail in the direction it was asserted in the day the sampler ran natively, and it did.

## Results

Full Catch2 suite: 3512 passed, 13 skipped, **1 failed** — `test_engine_device_layer.cpp:655` (`synthProbe.bound == 2 * kMaxMidiBytesPerPort`, 12288 vs 8192), tagged `[2341][2345]`. Confirmed pre-existing: it fails identically on unmodified `dev/0.20.0` with this branch stashed. Filed separately.

All JUCE suites green, including **Null Diff Corpus**, **Sampler Relocate Tests**, **DrumGrid Pad Chain Serialization**, **Device State Schema** and **MIDI Signal Routing**.

Not verified in the running app — I can't drive MAGDA's UI. The sampler faceplate and the Drum Grid pads are worth a click-through: loading a sample, dragging the trim and loop markers, and switching a pad's voice mode.

### DIFF
--- a/magda/daw/audio/ClipCommands.cpp
+++ b/magda/daw/audio/ClipCommands.cpp
@@ -14,6 +14,7 @@
 #include "audio/plugins/DrumGridPlugin.hpp"
 #include "audio/plugins/InsertCapturePlugin.hpp"
 #include "audio/plugins/MagdaSamplerPlugin.hpp"
+#include "audio/plugins/tracktion/TracktionMagdaDevicePlugin.hpp"
 #include "audio/racks/InstrumentRackManager.hpp"
 #include "core/ClipOcclusion.hpp"
 #include "core/ClipOperations.hpp"
@@ -21,6 +22,7 @@
 #include "core/Config.hpp"
 #include "core/ControlTarget.hpp"
 #include "core/DrumGridPads.hpp"
+#include "core/ParameterUtils.hpp"
 #include "core/TrackManager.hpp"
 
 namespace magda {
@@ -2858,20 +2860,38 @@ void prepareDrumGridAdsrMacros(DeviceInfo& drumGridDevice) {
     }
 }
 
-void zeroSamplerAdsrBase(daw::audio::MagdaSamplerPlugin& sampler) {
-    const float attackMin = sampler.attackParam->getValueRange().getStart();
-    const float decayMin = sampler.decayParam->getValueRange().getStart();
-    const float releaseMin = sampler.releaseParam->getValueRange().getStart();
+/// Write one sampler slot in its display units, through the HOST WRAPPER's
+/// parameter rather than onto the device. The wrapper mirrors its parameters
+/// into the device on every sync, so a write straight to the device is undone
+/// at the next one. setParameterFromHost, not setParameter: the latter is
+/// silently dropped once a macro or mod is attached to the parameter.
+void setSamplerSlot(te::Plugin& plugin, int index, float displayValue) {
+    auto* sampler =
+        daw::audio::tracktion_adapter::deviceFromPlugin<daw::audio::MagdaSamplerPlugin>(&plugin);
+    if (sampler == nullptr)
+        return;
 
-    sampler.attackParam->setParameterFromHost(attackMin, juce::dontSendNotification);
-    sampler.decayParam->setParameterFromHost(decayMin, juce::dontSendNotification);
-    sampler.sustainParam->setParameterFromHost(0.0f, juce::dontSendNotification);
-    sampler.releaseParam->setParameterFromHost(releaseMin, juce::dontSendNotification);
+    auto params = plugin.getAutomatableParameters();
+    if (index < 0 || index >= params.size())
+        return;
 
-    sampler.attackValue = attackMin;
-    sampler.decayValue = decayMin;
-    sampler.sustainValue = 0.0f;
-    sampler.releaseValue = releaseMin;
+    const float normalized =
+        ParameterUtils::realToNormalized(displayValue, sampler->parameterInfo(index));
+    params[index]->setParameterFromHost(normalized, juce::dontSendNotification);
+}
+
+/// Flatten the pad sampler's envelope so the DrumGrid ADSR macros, which are
+/// added next, are the only thing shaping it.
+void zeroSamplerAdsrBase(te::Plugin& plugin) {
+    using Sampler = daw::audio::MagdaSamplerPlugin;
+    auto* sampler = daw::audio::tracktion_adapter::deviceFromPlugin<Sampler>(&plugin);
+    if (sampler == nullptr)
+        return;
+
+    setSamplerSlot(plugin, Sampler::kAttack, sampler->parameterInfo(Sampler::kAttack).minValue);
+    setSamplerSlot(plugin, Sampler::kDecay, sampler->parameterInfo(Sampler::kDecay).minValue);
+    setSamplerSlot(plugin, Sampler::kSustain, 0.0f);
+    setSamplerSlot(plugin, Sampler::kRelease, sampler->parameterInfo(Sampler::kRelease).minValue);
 }
 
 void addSamplerAdsrMacroLinks(DeviceInfo& drumGridDevice, TrackId trackId, int chainIndex,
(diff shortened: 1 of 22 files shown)


lucaromagnoli:
Related: #2356 — the one Catch2 failure on this branch, confirmed pre-existing on `dev/0.20.0`.

lucaromagnoli:
Pushed `03290180`, which fixes #2356 — the failure that was making this PR red.

It was **not** a macOS problem and not this branch: `dev/0.20.0` has been red on macOS, Linux and Windows alike since `931925f7`. #2318 gave a track that carries clips a `SessionMidi` source beside its `ClipMidi` one, so a record-armed track's `TrackMidiInput` merge takes three inputs (clips, session, hardware input) where the test still expected two. Confirmed by dumping the merge's inputs, not by adjusting the number until it passed. Three is correct — a section switch has the arranger and a launched session clip live at once (#2302) — so only the expectation moves; the plan and the executor's forward pass are untouched.

Full Catch2 suite is now green locally: 3513 passed, 0 failed, 13 skipped.

Folded in here rather than sent as its own PR so macOS CI runs once instead of twice. Happy to split it out if you would rather keep this PR to the sampler alone.

lucaromagnoli:
Correction on the previous comment: the #2356 fix has been moved to its own PR against `dev/0.20.0` — **#2359** — and taken back off this branch. Folding it in here was the wrong call: it left `dev` red for every other branch until this PR merged.

This branch is back to `39d74240`, the sampler port alone. It stays red on the one pre-existing assertion until #2359 lands on `dev`; after that I will rebase this onto it and CI should be clean.

lucaromagnoli:
Codex review, triaged against \`c43da096^\`:

- **Regression from this PR** — `restoreState` kept the previous `loopEnabled` when the state did not name it, where `copyPropertiesToCachedValues` used to reset it. Fixed in #2377.
- **Pre-existing** — the audio thread reads `currentSound` while a load can free it. Same raw reads at lines 722/799 before the port. Filed as #2378.
- **Pre-existing** — pad sampler edits in `ClipCommands`, `PadDeviceSlot` and `DeviceCustomUIManager` reach the runtime and not the model. Before the port these wrote `setParameterFromHost` plus a CachedValue; the port changed the API, not the authority. Filed as #2379 under #2317.

No findings on #2376.

PR #2357, opened 2026-09-04
Trim comment bloat

### LINKED ISSUE #2353: 73k lines of comment against 309k of code, and the worst headers are 4:1
## The number

Across `magda/`, 1581 C++ files: **73,265 comment lines against 308,617 code lines — 24%.** Nearly a quarter of the tree is prose.

The distribution is worse than the average. Headers with more than 60 lines of code, by ratio of comment to code:

| ratio | comment | code | file |
|---|---|---|---|
| 433% | 273 | 63 | `magda/engine/param/ModLfo.hpp` |
| 384% | 242 | 63 | `magda/daw/audio/plugins/engine/EngineDeviceFactory.hpp` |
| 380% | 232 | 61 | `magda/daw/api/remote_api_host.hpp` |
| 323% | 213 | 66 | `magda/daw/audio/plugins/engine/DeviceControl.hpp` |
| 314% | 239 | 76 | `magda/daw/audio/session/ClipSynchronizer.hpp` |
| 311% | 196 | 63 | `magda/daw/audio/plugins/engine/EngineExternalDevice.hpp` |
| 284% | 176 | 62 | `magda/daw/core/controllers/BindingRegistry.hpp` |
| 264% | 185 | 70 | `magda/engine/param/ModRuntime.hpp` |
| 263% | 329 | 125 | `magda/daw/audio/osc/OscRouter.hpp` |
| 251% | 206 | 82 | `magda/engine/exec/EngineSession.hpp` |

A header where four lines in five are prose is not documented, it is buried.

## Why it costs

- **Reading.** Finding the signature of a five-method interface means paging past three screens of rationale.
- **Drift.** Prose is not compiled. A comment explaining why a field exists outlives the reason, and nothing fails when it does. The longer it is the more of it can quietly go stale.
- **Review.** A change to a 20-line comment block reads in a diff like a change to logic.
- **It compounds.** The style teaches itself: everyone writing new code matches the density of what is around it, including me. That is how it reached a quarter of the tree.

## What is worth keeping

Not all of it. The comments that earn their place say something the code cannot: an issue number, a constraint from outside (a TE quirk, a JUCE contract, a format's rules), a measured figure, or a decision whose alternative looks more obvious than it is. Those are usually one or two lines.

What does not earn its place, and is most of the bulk:

- Restating the code below it in prose.
- Narrating alternatives that were rejected, at length.
- Worked examples and tables that belong in `docs/` or in a test.
- The same explanation repeated on the declaration, the definition and the call site.
- Essays on a struct field.

## Direction

1. **Set a rule and write it down** — in `CLAUDE.md`, so it binds contributors and agents alike. Suggested: a comment is a line or two; anything longer needs a reason, and a link to an issue or a doc beats a paragraph.
2. **Measure it.** The script above is a few lines; a CI check could report the ratio per PR, or fail a file that crosses a threshold, so the number stops growing while the backlog is worked.
3. **Cut the worst by hand.** The ten files above are ~2,300 comment lines between them. They are the ones people read most — interfaces — and cutting them sets the example for everything else.
4. **Move what deserves keeping.** Several of these blocks are genuinely good architectural writing in the wrong place. `docs/architecture/` already exists.

Not a rewrite of every comment in the tree, and not a licence to strip context that is load-bearing. The target is the ratio, taken worst-first.

## Acceptance criteria

- `CLAUDE.md` states the rule.
- The ten files above are under 60% comment-to-code, with nothing load-bearing lost.
- A way to see the ratio move, so this does not have to be rediscovered by counting again.

lucaromagnoli:
## Summary
- Cuts restated-code prose, rejected-alternatives narration, and worked examples from the ten headers named in #2353, moving genuinely useful architectural narrative into `docs/architecture/` for three of them.
- Adds `CLAUDE.md` at the repo root with the comment-density rule.
- Adds `scripts/comment_ratio.py` to measure the comment-to-code ratio going forward (`--top`, `--check PERCENT`, or specific files).

Comment-only changes: verified by diff that no code/declaration lines changed in any of the ten headers.

Closes #2353

## Test plan
- [x] CI build passes

https://claude.ai/code/session_01SdpaazGnnZqbN2sgJ9MEYV

### DIFF
--- a/magda/daw/api/remote_api_host.hpp
+++ b/magda/daw/api/remote_api_host.hpp
@@ -23,55 +23,21 @@ class SubscriptionHub;
 /**
  * @brief Owns the remote API for the lifetime of a running MAGDA (#1856, #1858).
  *
- * The dispatcher, the model bridge that feeds it local edits, and the two
- * transports are objects with one lifetime and a required construction order, so
- * something has to own them together. This is that something, and it is the only
- * place in the app that knows the remote API exists at all.
- *
- * Both transports share one dispatcher and one subscription hub. That is the
- * point of the layering rather than an optimisation: two dispatchers would be
- * two revision counters and two undo groupings over one project, and two hubs
- * would be two projections of the same model drifting apart.
+ * Owns the dispatcher, the model bridge that feeds it local edits, and the two
+ * transports, which share one required construction order. Both transports
+ * share one dispatcher and one subscription hub, not as an optimization: two
+ * dispatchers would be two revision counters and two undo groupings over one
+ * project, and two hubs would be two projections of the model drifting apart.
  *
  * Construct and destroy on the message thread: `ModelChangeBridge` attaches to
  * the model singletons, which notify there.
  *
- * ## The token
- *
- * Generated per run, never persisted to config. A config file gets copied
- * between machines, committed by accident, and pasted into bug reports; a
- * credential in one is a credential that leaks eventually. Instead the token
- * lives in a file beside the other app data, written with owner-only
- * permissions and deleted on shutdown, carrying the port alongside it:
- *
- *     remote-api-<pid>.json
- *     {"port":51734,"token":"…","url":"ws://127.0.0.1:51734/rpc",
- *      "mcpPort":51735,"mcpUrl":"http://127.0.0.1:51735/mcp","pid":4021}
- *
- * A client reads that file to learn where to connect and what to present. This
- * is the same shape Jupyter uses for its local server, and it means a client on
- * the machine needs no configuration while a process elsewhere gets nothing.
- *
- * Both transports are named because a client picks one: an MCP host wants
- * `mcpUrl`, a script that needs pushed state wants `url`. The token is the same
- * for both — it authenticates the user at the keyboard, not a protocol.
- *
- * The record is named after the process because MAGDA allows more than one
- * instance. One shared file would mean the second instance to start hides the
- * first, and the first to stop deletes the record of the one still running —
- * and with a fixed port, where SO_REUSEPORT lets both listeners bind, a client
- * could hold one instance's token and be routed to the other. A client that
- * finds several records is looking at several running instances and has to
- * choose; each is independently valid for the port it names.
- *
- * A record whose process is gone is debris from a crash, and any instance's
- * `start()` will collect it. An instance only ever deletes its own live record.
- *
- * ## Disabled means disabled
- *
- * `start()` returns false and opens no socket when the feature is off in
- * config, when no token could be generated, or when the port is taken. There is
- * no partial state where MAGDA is listening but unauthenticated.
+ * The auth token is generated per run, never persisted to config, and lives in
+ * an owner-only, per-process discovery file deleted on shutdown (see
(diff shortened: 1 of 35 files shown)


magda/daw/audio/plugins/engine/DeviceControl.hpp
```diff
@@ -13,64 +13,39 @@
...
+ * Nothing here writes a model directly: a capture comes back as data, the
+ * caller checks the assignment it was read from still holds, then writes it
+ * -- the same boundary completeExternalPluginLoad() applies on the load path,
+ * for the same reason (a slot can be given a different plugin, or none,
+ * between asking and answering).
+ *
+ * Not here yet: the editor window, MAGDA's other caller that reaches for an
+ * instance. It belongs here when it's ported off the fork's plugin windows --
+ * an editor is a control-plane request like any other.
  */
```
lucaromagnoli: Shorten, also don't add content that could rot easily. e.g " Not here yet". State what is here not what isn't. Also don't describe a feature by the source ticket number. Humans do not remember ticket numbers. You may reference the number for completeness but you can't lead with that

lucaromagnoli: Fixed: reworded to not lead with the ticket number, and dropped the "Not here yet" paragraph about the editor window (unrelated, absent feature, would've rotted once it shipped). Also caught and fixed the same ticket-number pattern in remote_api_host.hpp, and reverted ClipSynchronizer.hpp entirely — its @brief/@param/@return docs were already at the right density and I'd wrongly stripped the @param/@return tags there trying to shorten it further.

magda/daw/api/remote_api_host.hpp
```diff
@@ -23,55 +23,21 @@ class SubscriptionHub;
...
  * the model singletons, which notify there.
  *
- * ## The token
- *
- * Generated per run, never persisted to config. A config file gets copied
- * between machines, committed by accident, and pasted into bug reports; a
- * credential in one is a credential that leaks eventually. Instead the token
- * lives in a file beside the other app data, written with owner-only
- * permissions and deleted on shutdown, carrying the port alongside it:
- *
```
lucaromagnoli, line 46: leave an example payload

lucaromagnoli: Restored the discovery-record JSON payload inline, with a pointer to docs/architecture/remote-api-host.md for the why (crash recovery, multi-instance, why not config).

magda/daw/core/LinkModeManager.hpp
```diff
@@ -7,48 +7,12 @@
...
-//
-//   class MyComponent : public LinkModeManagerListener {
-//       void modLinkModeChanged(bool active, const ModSelection& sel) override;
-//       void macroLinkModeChanged(bool active, const MacroSelection& sel) override;
-//   };
-//
-// Register in constructor: LinkModeManager::getInstance().addListener(this);
-// Unregister in destructor: LinkModeManager::getInstance().removeListener(this);
-//
-// ============================================================================
```
lucaromagnoli, line 51: leave this example

lucaromagnoli: Restored the LinkModeManagerListener implementation snippet inline.

Issue #2303, opened 2026-09-01
Session launcher slice 4: played ranges reported without a poll


lucaromagnoli:
Slice 4 of #1894. Depends on slice 1.

## What the fork does, and why it should not be copied

`SessionClipScheduler::processStateEvents()` says it: "Must be called periodically from the message thread (e.g. from PlaybackPositionTimer)." The UI learns what the launcher did by polling a timer. The cost is visible in the same header, where `relaunchActiveClips()` exists to "call synchronously when transport starts to avoid the 33ms polling delay" -- a second entry point whose only reason to exist is that the first one is a poll.

A launcher that reports through a timer has a UI that is up to a frame behind the audio on every launch, and a workaround per case where that mattered enough to notice.

## What to build instead

The read-back path already exists. Value taps landed in #2122 and are how the parameter system reports what the audio thread resolved without the message thread asking. A launch handle's play state, queued state and played range are the same kind of value: audio-written, message-read, wanted at frame rate by the UI and never on the audio thread's critical path.

- Play state and queued state per handle, so a slot button knows it is queued without asking.
- The played range, which is what draws a session playhead. `ClipInfo::sessionPlayheadPos` exists and `getActiveClipPlayheadPositions()` is how the fork answers it; the native path reports the range and lets the UI derive the position rather than converting beats to looped seconds inside the engine, which is what `elapsedBeatsToSeconds` does today.
- Enough for the UI to blink a pending stop, which `isSessionTrackStopPending` currently answers by comparing intent against a polled state.

## Why it is its own slice

Because the alternative is that it never happens. If the launcher is built with the fork's polling shape and the UI is wired to it, the poll is load-bearing by the time anybody revisits it, and every case like `relaunchActiveClips` gets rebuilt on the new engine rather than deleted with the old one.

## Verification

The reported range after a known sequence of launches matches what the handle advanced through, and it is correct on the block the launch happened in rather than on the next one a timer would have seen.

lucaromagnoli:
Merged in #2372.

PR #2358, opened 2026-09-04
Carry launch requests to the audio thread on a queue (#2305)

### LINKED ISSUE #2305: Session launcher slice 6: handle lifetime on the plan-swap epoch

lucaromagnoli:
Closes #2305. Slice 6 of #1894.

Handle lifetime was already on the epoch: #2301 put allocation and retirement in `RuntimeStateStore::publishHandles`, where the swap waits for the block the callback is in, and `test_launch_handle_lifetime.cpp` covers it. What was missing is the other half of the issue, the lane a launch travels down. `SessionLauncher.hpp` said so: *"until then a handle is driven directly, which is safe only when nothing is rendering"*, and `EngineSession::launchHandle()` handed out a mutable pointer to say it.

## A launch is an event, so it travels on a queue

The engine's one lane publishes a value whole. That is right for what it carries, and wrong here for three reasons that only show up once you try:

- **A request put through it is applied again every block** until something overwrites it, so a launch would retrigger the clip for ever. Closing that needs a serial per slot and somewhere to keep the last one applied.
- **That somewhere is the handle**, and a handle is per slot. A slot emptied and refilled gets a *new* handle whose last-applied is zero, while the published table still names the old request — so the refilled slot comes up already playing. That is exactly the case `test_launch_handle_lifetime.cpp` was written to stop, reintroduced one layer up.
- **A published table has no order.** `setLooping` before `play` is an order: a length set for a clip about to start has to arrive first.

A queue has none of those. A request is written once and read once, so it cannot be re-applied, cannot outlive the block that consumed it, and keeps its order for free.

**A request names a slot and never a handle**, which is what keeps the lane out of the lifetime question entirely rather than making it a second answer to it. Nothing in the ring points at anything, so nothing in it can outlive anything; a request for a slot whose handle has gone finds nothing and does nothing. `playSynced` carries the leader's `SlotKey` too, and is resolved against the table that is live when it is applied — a scene launch decided on the message thread would otherwise join where the leader was a block ago.

## A gesture is what makes a scene one event

`LaunchRequestQueue::Gesture` is RAII over the ring: it fills slots at a private cursor and publishes that cursor once, on destruction. So the audio thread sees every request of a gesture or none of them, and eight clips launched together start on one sample rather than over eight consecutive blocks. It states where a gesture ends in the type rather than leaving it to whoever remembers to call something.

A gesture too large for the ring is dropped whole and counted (`overflows()`), not truncated. Half a scene launching is worse than none: the tracks that made it would be a bar ahead of the ones that did not for as long as they played. Counted rather than left silent for the reason the clock counts its loop wraps — a launch that quietly did not happen is otherwise blamed on the audio device.

The capacity check is at each push rather than at the commit. The slots a full ring would write over are the ones the reader still owns, and rolling a cursor back afterwards does not put their contents back.

## Applied in the pass that advances

`advanceLaunchHandles` takes the queue and drains it whole before advancing anything, which is why it is one call and not two. A request applied after a handle had already been advanced would take effect a block late; one applied to some handles and not others would split a scene down the middle. The queue is drained even when no table is published, and `EngineSession::process` drains it when no plan is — a queue filling while nothing exists to launch would deliver a launch made minutes ago at whatever moment a session appeared.

`launchHandle()` returns `const LaunchHandle*` now, so nothing can drive a handle from the publishing thread. Reading is still only nearly safe and is what there is until the read-back path (#2303).

## The plan swap

Nothing had to change for it, which is the point: handles live in the store, not in the plan, so a swap has nothing of theirs to take. The new end-to-end case asserts it — a track added while a slot plays compiles a new plan and swaps it in, and the clip is not dropped, not restarted (the run's origin is unchanged) and not doubled (three more blocks advance the run by exactly three blocks, where a handle advanced twice over a swap would be half a bar ahead of the transport).

## Tests

New `tests/engine/test_launch_requests.cpp`, 13 cases: the block a request lands in, that an open gesture is invisible, that a request cannot be applied twice, that order is kept, a scene launch in phase, a synced launch whose leader has gone, a request for a slot that is not there, the refilled-slot case above, back-to-arrangement and stop-still-holds over the lane, a request made with nothing published, and the ring's overflow / exact-fit / reuse behaviour. Plus the structural-edit case in `test_engine_device_layer`'s neighbour `test_engine_session.cpp`.

`[launch]` 45 cases and `[session]` 138 cases pass.

### DIFF
--- a/magda/engine/launch/SessionLauncher.cpp
+++ b/magda/engine/launch/SessionLauncher.cpp
@@ -35,16 +35,15 @@ LaunchHandle* LaunchHandleTable::find(const SlotKey& key) const {
 
 namespace {
 
-/// One request against the table that is live now. A slot the table does not
-/// name is one emptied between the ask and this block, and there is nothing
-/// left to ask.
-///
-/// The incarnation is the rest of that question. A slot emptied and refilled
-/// between the ask and this block is the same key on a different clip, and
-/// applying the launch to it would start something the user never clicked
-/// (#2305 review). The key alone cannot tell those apart; the incarnation can,
-/// and it is a number rather than a pointer, so nothing here holds a handle
-/// that could have been retired.
+/**
+ * @brief Apply @p request against the table that is live now.
+ *
+ * Dropped when the slot has gone, and equally when it was emptied and refilled:
+ * that is the same key on a different clip, and launching it would start
+ * something the user never clicked. The incarnation is what tells those apart,
+ * and it is a number rather than a pointer, so nothing here can name a retired
+ * handle.
+ */
 void apply(const LaunchRequest& request, const LaunchHandleTable& table) {
     const auto* entry = table.findEntry(request.key);
     if (entry == nullptr || entry->handle == nullptr || entry->incarnation != request.incarnation)
@@ -55,21 +54,15 @@ void apply(const LaunchRequest& request, const LaunchHandleTable& table) {
     switch (request.kind) {
         case LaunchRequest::Kind::play:
             if (request.syncTo) {
-                // The run to join is the one that handle holds now, which is
-                // why the request carries a slot and not a run: a scene launch
-                // decided on the message thread would join where the leader was
-                // a block ago.
+                // A slot and not a run, so the run joined is the one the leader
+                // holds now rather than where it was when the gesture was made.
                 if (const auto* with = table.find(*request.syncTo); with != nullptr) {
-                    // Unless the leader has a launch of its own still queued, in
-                    // which case it has no run to join yet: the one it is about
-                    // to begin is the one the follower means. Joining what it is
-                    // playing now would put the follower on the run the leader
-                    // is leaving, and a scene relaunched while it was already
-                    // playing would come back out of phase with itself.
-                    //
-                    // Both then begin fresh on the same instant, which is the
-                    // same origin, because a run's origin is the sample it
-                    // started on and nothing else (#2336).
+                    // A leader with a launch of its own still queued has no run
+                    // to join yet, and the one it is about to begin is the one
+                    // the follower means. Both then begin on the same instant,
+                    // which is the same origin (#2336); joining what it is
+                    // playing would leave a relaunched scene out of phase with
+                    // itself.
                     if (with->queuedState() == LaunchHandle::QueueState::playQueued) {
                         handle->play(with->queuedPosition());
                         return;
@@ -79,9 +72,8 @@ void apply(const LaunchRequest& request, const LaunchHandleTable& table) {
                     return;
                 }
 
-                // The leader's slot was emptied in between. Starting alone is
-                // the honest answer: the clip was asked to play, and there is
-                // no longer anything for it to be in phase with.
+                // The leader's slot was emptied in between, so there is nothing
+                // left to be in phase with.
             }
 
             handle->play(request.position);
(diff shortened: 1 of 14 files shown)


magda/engine/launch/SessionLauncher.cpp
```diff
@@ -28,10 +28,67 @@ LaunchHandle* LaunchHandleTable::find(const SlotKey& key) const {
...
 }
 
-void advanceLaunchHandles(LaunchHandleFeed& handles, const BlockInfo& block) {
+namespace {
+
+/// One request against the table that is live now. A slot the table does not
+/// name is one emptied between the ask and this block, and there is nothing
+/// left to ask.
+void apply(const LaunchRequest& request, const LaunchHandleTable& table) {
+    auto* handle = table.find(request.key);
```
lucaromagnoli: **P1 — Bind requests to a slot incarnation, not only its key**

A committed request can remain in the queue while the publishing thread empties this slot and refills the same `SlotKey` before the next callback. `publishHandles()` then retires the old handle and creates a new one, but this lookup finds that replacement and applies the old launch to it. The new refill test does not cover this ordering because it drains the request before replacing the handle. Please carry a non-pointer incarnation/generation (or otherwise invalidate stale requests across handle replacement) so a request cannot launch a different clip that later occupies the same slot.

magda/engine/launch/SessionLauncher.cpp
```diff
@@ -28,10 +28,67 @@ LaunchHandle* LaunchHandleTable::find(const SlotKey& key) const {
...
+
+    switch (request.kind) {
+        case LaunchRequest::Kind::play:
+            if (request.syncTo) {
+                // The run to join is the one that handle holds now, which is
+                // why the request carries a slot and not a run: a scene launch
+                // decided on the message thread would join where the leader was
+                // a block ago.
+                if (const auto* with = table.find(*request.syncTo); with != nullptr) {
+                    handle->playSynced(*with, request.position);
```
lucaromagnoli, line 71: **P1 — Sync followers to the leader’s pending relaunch**

When a gesture uses the exact scene pattern in the new test—`play(leader)` followed by `playSynced(follower, leader)`—and the leader is already playing, `play()` only replaces the leader’s pending request; it leaves its current `run_` unchanged. This call therefore snapshots the old run. During the following advance the leader starts a new run while the follower joins the old one, so their origins diverge. Please resolve the leader’s pending launch (or represent the gesture dependency explicitly) so relaunching a scene keeps every requested slot in phase, and add the already-playing-leader case.

PR #2372, opened 2026-09-04
Publish a slot's state from the block that decided it (#2303)

### LINKED ISSUE #2303: Session launcher slice 4: played ranges reported without a poll

lucaromagnoli:
Closes #2303. Slice 4 of #1894.

The fork's UI learns what the launcher did by polling: `SessionClipScheduler::processStateEvents` off a 33 ms timer, plus `relaunchActiveClips` as a second entry point that exists only because the first one is a poll. `LaunchTap` is the read-back the value taps already established for parameters (#2122), applied to a slot: the block that decides the state publishes it, in the pass that advanced the handle, so the reading and the audio it describes are the same block.

## What a slot publishes

`play state`, `queued state`, the section hold, and how far the run has got. The issue asks for the played range rather than a position, and what a playhead actually needs off it is its length -- elapsed beats since the run began, unlooped. Wrapping that against the clip's own length stays in the UI, because the length belongs to the model and `elapsedBeatsToSeconds` inside the engine is exactly the conversion the issue objects to.

The section hold rides along because #2302 made it a separate question from playing: silence after a stop is the session still holding the track, and a UI drawing the arrangement as live there would be drawing the opposite of what is sounding.

## One word

A reading pairs the position with the state it was in. Split across two atomics, a slot could draw a playhead from before the launch that started it, or a queued badge beside a position that has already begun moving. A float and a 32-bit status fit in one 64-bit word: one store on the audio thread, one load off it, nothing to retry -- the same argument `ValueTap` makes for its value-and-count pair.

## No write count, unlike ValueTap

`ValueTap` counts blocks because a modifier holding while the engine is stopped is indistinguishable from one that is not moving, and because a value nothing publishes has to fall back to the model's. Neither applies here. A slot that is not playing is not playing whether or not a block has ever said so, so the default reading is already the true one and a count would have nothing to disambiguate. That also means a tap holds correctly across a stopped transport rather than needing to be told the engine went away.

## Lifetime

The tap is made and retired with the handle beside it, in `RuntimeStateStore::publishHandles`, so `EngineSession::launchTap()` hands out the same lifetime `valueTap()` already does: good until a `publishClips()` stops naming the slot. It lives beside the handle rather than in the published table so the address a host is reading stays put across a publish that did not retire the slot -- there is a case for that.

`launchHandle()` is now documented as being for tests and diagnostics: what the state is belongs to the tap, and what changes it belongs to `launchRequests` (#2305).

## Tests

New `tests/engine/test_launch_tap.cpp`, 9 cases: an unrendered slot reads as stopped, a launch is visible in the block that made it, a queued launch is visible before it sounds, a pending stop is visible while the slot still sounds, the position is what the run covered, the stop-still-holds distinction, a reading is one block's own, a scene reports one position across its slots, and a stopped transport holds. Plus an end-to-end case on `EngineSession` for the store plumbing and the tap's identity across a republish.

`[tap]` 43, `[launch]` 58 and `[session]` 151 cases pass.

## Also here

`.claude/skills/comments-not-essays/SKILL.md`. `CLAUDE.md` states the comment rule; this is the working form of it, naming the failure modes that keep recurring -- markdown headings inside a doc block, rejected-alternatives narration, argument connectives, worked arithmetic. Bundled rather than given its own PR so it does not cost a second CI run.

The tap's own doc blocks were an example of the problem before the last two commits, and are now the size the rule asks for.

### DIFF
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -173,9 +173,12 @@ class EngineSession {
     /**
      * @brief Where @p key's state is published for the UI, or null (#2303).
      *
-     * Read from any thread, as often as a repaint likes. The pointer is good
-     * until a publishClips() stops naming the slot, the same lifetime
-     * valueTap() hands out.
+     * Both the lookup and the pointer belong to the publishing thread, exactly
+     * as valueTap() does: the next publishClips() that stops naming the slot
+     * destroys the tap on that thread, with nothing deferring it, so a pointer
+     * cached and read from a paint or animation thread is a use-after-free
+     * waiting for a clip edit. Fetch it here, use it before returning to the
+     * message loop, and ask again after every publish.
      */
     const LaunchTap* launchTap(const SlotKey& key) const {
         return store_.launchTap(key);
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -171,12 +171,26 @@ class EngineSession {
     }
 
     /**
+     * @brief Where @p key's state is published for the UI, or null (#2303).
+     *
+     * What a slot button and a session playhead draw from. Read from any
+     * thread, as often as a repaint likes; the block that decided the state is
+     * what published it, so nothing polls and nothing is a frame behind.
+     *
+     * The pointer belongs to the publishing thread and is good until a
+     * publishClips() stops naming the slot, which is the same lifetime
+     * valueTap() hands out. Null until publishClips() has named it.
+     */
+    const LaunchTap* launchTap(const SlotKey& key) const {
+        return store_.launchTap(key);
+    }
+
+    /**
      * @brief The handle for @p key, or null. On the publishing thread.
      *
-     * To read, and only to read: a handle is advanced on the audio thread, so
-     * changing one from here races the callback. Every change goes through
-     * @ref launchRequests. Reading is itself only nearly safe, and is what
-     * there is until the read-back path is built (#2303).
+     * For tests and diagnostics. What the state is belongs to @ref launchTap
+     * and what changes it belongs to @ref launchRequests; a handle is advanced
+     * on the audio thread, so reading one from here races the callback.
      */
     const LaunchHandle* launchHandle(const SlotKey& key) const {
         return store_.findHandle(key);
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -173,24 +173,20 @@ class EngineSession {
     /**
      * @brief Where @p key's state is published for the UI, or null (#2303).
      *
-     * What a slot button and a session playhead draw from. Read from any
-     * thread, as often as a repaint likes; the block that decided the state is
-     * what published it, so nothing polls and nothing is a frame behind.
-     *
-     * The pointer belongs to the publishing thread and is good until a
-     * publishClips() stops naming the slot, which is the same lifetime
-     * valueTap() hands out. Null until publishClips() has named it.
+     * Read from any thread, as often as a repaint likes. The pointer is good
+     * until a publishClips() stops naming the slot, the same lifetime
+     * valueTap() hands out.
      */
     const LaunchTap* launchTap(const SlotKey& key) const {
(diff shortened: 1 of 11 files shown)


magda/engine/exec/EngineSession.hpp
```diff
@@ -171,12 +171,22 @@ class EngineSession {
...
      *
-     * To read, and only to read: a handle is advanced on the audio thread, so
-     * changing one from here races the callback. Every change goes through
-     * @ref launchRequests. Reading is itself only nearly safe, and is what
-     * there is until the read-back path is built (#2303).
+     * Read from any thread, as often as a repaint likes. The pointer is good
+     * until a publishClips() stops naming the slot, the same lifetime
+     * valueTap() hands out.
+     */
+    const LaunchTap* launchTap(const SlotKey& key) const {
```
lucaromagnoli, line 183: **P1 — Make the advertised cross-thread lifetime safe**

This returns a raw pointer that the preceding comment says may be read from any thread, but publishHandles() destroys taps for removed slots immediately after synchronizing only with the audio thread. A repaint or other reader can therefore still be inside LaunchTap::read() when the atomic itself is freed, making an otherwise valid advertised use a use-after-free.

Please return a lifetime-bearing lease/shared pointer, defer reclamation until readers quiesce, or restrict lookup and use to the publishing thread as valueTap() does.

magda/engine/tap/LaunchTap.hpp
```diff
@@ -0,0 +1,106 @@
...
+        Reading reading;
+        reading.playing = handle.playState() == LaunchHandle::PlayState::playing;
+        reading.holdsSection = handle.holdsSection();
+
+        if (const auto queued = handle.queuedState())
+            reading.queued =
+                *queued == LaunchHandle::QueueState::playQueued ? Queued::play : Queued::stop;
+
+        if (const auto played = handle.playedRange())
+            reading.elapsedBeats = static_cast<float>(played->length());
```
lucaromagnoli: **P2 — Keep precision for the unbounded run length**

playedRange() accumulates for the entire run, including across ordinary clip loops, so narrowing its length to float eventually makes the displayed playhead stall and jump. At 120 BPM the float step reaches about 31 ms after three days, 125 ms after twelve days, and 250 ms after thirty-three days. Long-running installations will therefore lose the frame-rate readback this tap is meant to provide.

Please preserve more precision, or publish a bounded phase separately while retaining whatever unbounded value the UI needs.

PR #2375, opened 2026-09-04
Act on a slot's follow action in the block it falls in (#2304)

### LINKED ISSUE #2304: Session launcher slice 5: follow actions, and an engine-initiated launch reaching the app

lucaromagnoli:
Closes #2304. Slice 5 of #1894.

The engine's first launch of its own. Every other arrives down the queue in `LaunchRequests.hpp` (#2305); this one is a beat the clip named when it was written, so the audio thread works it out and acts on it.

## Where it runs

In `advanceLaunchHandles`, after the request drain and before the advance. That ordering is the whole of the sample accuracy: a launch decided after the handles had moved would take effect on the next callback boundary, and both halves -- the slot that stops and the slot that starts -- are handed the same beat, so the advance cuts both blocks at the same sample.

A slot with anything queued is left alone. A handle holds one pending request and the last write wins, so a follow action written over a queued launch would drop what the user asked for. The incumbent only defends against a queued stop; this is the same rule taken to its end.

## Counting the passes

The due beat counts from `LaunchHandle::scheduleBeat`, which is carried across every re-trigger, not from the run's origin, which is not.

That difference is a bug in the incumbent rather than a preference. `SlotControlNode` computes its stop point as `playedMonotonicRange().start + stopDuration`, and TE's own loop wrap resets that start. With `FollowActionDurationType::loops` and a count above one, the stop point moves forward by a loop length every time the loop wraps, so it is never reached and the action never fires. Counting from the schedule beat is exact and stable across a thousand wraps, which is what #2336 put it there for.

`loopCount` multiplies only while the handle is re-triggering, matching `syncFollowActionToTracktionClip`: a slot with one pass in it has one to count.

## A run that ends

A non-looping slot now stops at the end of its material. The engine did not do this at all before -- a launched slot ran until something stopped it -- and it is the same question a follow action is asked at, so it falls out of the same calculation rather than needing its own.

## Random

Hashed from the slot and the due beat, not carried in a generator. Two renders of one project have to produce one file, and a seed carried block to block would put every random follow action beyond the null-diff harness (#1896). splitmix64's finaliser, so two adjacent beats pick unrelated slots.

## A target that has gone

Nothing stores one. `next`, `previous` and `random` are resolved against the handle table as it stands at the instant the action fires, and that table is built from the slots the snapshot named, so a slot deleted since the project loaded is simply not in the list searched and a moved one is at its new scene index. An action with nothing left to name does nothing beyond the stop, which is also what the incumbent does at the ends of a track: `trackNext` at the last slot returns nothing rather than wrapping.

## The read-back, and what is not here

The action reaches the handle through `play` and `stop` -- what a user launch reaches -- so the slot it starts publishes through its own tap (#2303) exactly as a clicked one does. `test_follow_actions.cpp` asserts that field by field against a control rig where the user clicked on the same beat instead.

So there is nothing to adopt, which is what #1894 asked for. What there is not is an app-side subscription: nothing in the app drives the native launcher yet, and the fork's `SessionClipScheduler::processStateEvents` still owns `activeSessionClipId` and its adoption loop. That is the issue's "or a stated reason why it cannot" -- the reporting is in place and the model reading it is wiring that belongs with whatever turns the native launcher on.

## Shape

`SlotAction` and `SlotFollow` are named apart from the model's `FollowAction` on purpose: a slot rather than a clip, and no track-group actions, because the model does not offer them. The config rides on `LaunchHandleTable::Entry` beside the handle and the tap, so the audio thread reads one table rather than merging two lists that publish separately, and a follow-action edit republishes it through the existing `entries ==` comparison.

## Tests

New `tests/engine/test_follow_actions.cpp`, 16 cases: a non-looping slot stopping at its end, a looping one with nothing to follow it playing on, the stop landing on its beat, next and previous handing the track over on one instant and doing nothing at the ends, again restarting rather than ending, a loop count holding the action off until the third pass, a loop count being one pass for a slot that does not loop, the delay, a user request outranking the action, a target deleted mid-run, random being deterministic per beat and not per set, an engine launch reading out identically to a clicked one, and a three-slot chain producing a known sequence of played ranges.

Full engine suite green: 554,643 assertions in 875 cases.

## Also here

Every `LaunchHandleTable::Entry` construction site now names its fields. The new member would otherwise have left `-Wmissing-field-initializers` across four test files, and a positional aggregate is what made adding one noisy in the first place.

### DIFF
--- a/magda/engine/launch/FollowActions.cpp
+++ b/magda/engine/launch/FollowActions.cpp
@@ -35,15 +35,15 @@ std::size_t randomIndex(const SlotKey& key, double dueBeat, std::size_t count) {
     return static_cast<std::size_t>(seed % count);
 }
 
-/// The slot @p follow names, or null for an action that starts nothing: a stop,
-/// and a neighbour at the end of the track.
-const LaunchHandleTable::Entry* followTarget(const LaunchHandleTable& table, const SlotKey& key,
-                                             const SlotFollow& follow, double dueBeat) {
+}  // namespace
+
+LaunchHandle* followTarget(const LaunchHandleTable& table, const SlotKey& key,
+                           const SlotFollow& follow, double dueBeat) {
     if (follow.action == SlotAction::none || follow.action == SlotAction::stop)
         return nullptr;
 
     if (follow.action == SlotAction::again)
-        return table.findEntry(key);
+        return table.find(key);
 
     // Every slot of the track that has a handle, in scene order, which is every
     // slot holding a clip: the table is built from the slots the snapshot named
@@ -61,13 +61,13 @@ const LaunchHandleTable::Entry* followTarget(const LaunchHandleTable& table, con
         case SlotAction::next:
             // No wrap at the ends, which is the incumbent's: the last slot of a
             // track stops rather than starting the first again.
-            return at + 1 < count ? first + at + 1 : nullptr;
+            return at + 1 < count ? first[at + 1].handle : nullptr;
 
         case SlotAction::previous:
-            return at > 0 ? first + at - 1 : nullptr;
+            return at > 0 ? first[at - 1].handle : nullptr;
 
         case SlotAction::random:
-            return first + randomIndex(key, dueBeat, count);
+            return first[randomIndex(key, dueBeat, count)].handle;
 
         case SlotAction::none:
         case SlotAction::stop:
@@ -77,8 +77,6 @@ const LaunchHandleTable::Entry* followTarget(const LaunchHandleTable& table, con
 
     return nullptr;
 }
-
-}  // namespace
 
 std::optional<double> followDueBeat(const LaunchHandle& handle, const SlotFollow& follow) {
     const auto schedule = handle.scheduleBeat();
@@ -100,14 +98,6 @@ std::optional<double> followDueBeat(const LaunchHandle& handle, const SlotFollow
     const auto passes = loop ? std::max(1, follow.loopCount) : 1;
 
     return *schedule + (pass * passes) + std::max(0.0, follow.delayBeats);
-}
-
-void applyFollowAction(const LaunchHandleTable& table, const SlotKey& key, const SlotFollow& follow,
-                       double dueBeat) {
-    const auto* target = followTarget(table, key, follow, dueBeat);
-
-    if (target != nullptr && target->handle != nullptr)
-        target->handle->play(dueBeat);
 }
 
 }  // namespace magda::engine
--- a/magda/engine/launch/FollowActions.cpp
+++ b/magda/engine/launch/FollowActions.cpp
@@ -9,13 +9,8 @@ namespace magda::engine {
 
 namespace {
 
-/**
- * @brief Which of @p count slots a random action picks, at @p dueBeat.
- *
- * A hash rather than a generator with state, so two renders of one project
- * produce one file (#1896). splitmix64's finaliser: its avalanche is what makes
- * two adjacent beats pick unrelated slots.
- */
+/// Scrambles @p value for @ref randomIndex below, so that inputs one bit apart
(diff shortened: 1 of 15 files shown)


magda/engine/launch/SessionLauncher.cpp
```diff
@@ -93,6 +95,37 @@ void apply(const LaunchRequest& request, const LaunchHandleTable& table) {
...
+ *
+ * Before the advance, so the stop and the launch it triggers land on one beat
+ * inside this block rather than on the next callback boundary.
+ *
+ * A slot with anything queued is left alone: a handle holds one pending request,
+ * so writing a stop over it would drop what the user asked for.
+ */
+void applyDueFollowActions(const LaunchHandleTable& table, const SyncRange& range) {
+    for (const auto& entry : table.entries) {
+        if (entry.handle == nullptr ||
```
lucaromagnoli, line 110: **P1 — Handle follow deadlines for slots launched in this callback**

A play request drained above only sets the handle pending; its playState remains stopped until advance() below. This guard therefore skips every newly launched slot. If the slot length fits inside this callback, its deadline also falls inside this callback, but it is not noticed until the next callback and the overdue stop/follow is clamped to that callback start. That makes short clips and same-block follow chains transition late, despite the intended in-block sample accuracy. Please include pending launches when scheduling follow events, including any targets launched by an earlier follow action in the same block.

magda/engine/launch/FollowActions.cpp
```diff
@@ -0,0 +1,112 @@
...
+    const auto passes = loop ? std::max(1, follow.loopCount) : 1;
+
+    return *schedule + (pass * passes) + std::max(0.0, follow.delayBeats);
+}
+
+void applyFollowAction(const LaunchHandleTable& table, const SlotKey& key, const SlotFollow& follow,
+                       double dueBeat) {
+    const auto* target = followTarget(table, key, follow, dueBeat);
+
+    if (target != nullptr && target->handle != nullptr)
```
lucaromagnoli: **P1 — Do not overwrite a queued request on the follow target**

The caller only checks queuedState() on the source. A next/previous/random action can select another handle that already has a user play or stop queued, and this unconditional play() replaces that handle’s sole pending request. The user-selected beat or stop is then lost, contrary to the PR’s rule that user requests outrank follow actions. Please preserve a queued request on the selected target as well.

PR #2382, opened 2026-09-04
Route a sampler's faceplate edits through the model (#2379)

### LINKED ISSUE #2379: Pad sampler edits reach the runtime and not the model, so a rebuild loses them
Part of #2317. Found by a Codex review of #2355. **Pre-existing** — the port swapped which runtime API these call, not where the value lands. Before it they wrote `setParameterFromHost` plus a `CachedValue`; after it they write the host parameter and the device. Neither is the model.

## The sites

| site | what it writes | what it should write |
| --- | --- | --- |
| `ClipCommands.cpp` (`buildDrumGridFromSlices`, `zeroSamplerAdsrBase`) | slice start/end and the flattened ADSR onto the pad's host parameters | the pad's model parameters |
| `PadDeviceSlot.cpp` (`setupForSampler`) | faceplate parameter edits, the loop switch and the root note onto the device | the pad's `DeviceInfo` via its `ChainNodePath` |
| `DeviceCustomUIManager.cpp` (`executeSamplerCommand`) | the loop switch, the root note and a sample load onto the device | model authored state, projected to the engine |

## What it costs

#2317 makes the model the authority for internal-device parameters and authored state: the engine is a projection of it, and a rebuild or a save reads the model. A value that only ever reached the runtime is therefore lost at the next rebuild, missing from a save, and invisible to undo — and on the native leg, absent from the render, which is the half that matters now that the sampler runs there (#2271).

The slice case is the sharpest: `buildDrumGridFromSlices` is how a sliced loop becomes a kit, and every pad's start and end marker is written this way. The model keeps the defaults, so a rebuilt or reopened project has sixteen pads all playing the whole file.

## Direction

Route each through the model rather than the plugin: `TrackManager::setDeviceParameterValue` for the parameters, and the authored-state path for the loop switch, root note and sample. The pad's `ChainNodePath` is what addresses a nested device, and `linkAssignedDrumGridSamplerAdsrMacros` in the same file already builds one for the macro links, so the address is available at every one of these sites.

The loop switch and root note are not parameters, so they need the authored-state route rather than the parameter one — the same shape the convolution's IR load takes since #2319.

lucaromagnoli:
Closes #2379. Part of #2317.

Three sites wrote a sampler's values onto the runtime and never onto the model. Pre-existing — the port to `MagdaDevice` swapped which runtime API they call, not where the value lands. Since #2317 the model is the authority and the engine is a projection, so a value that only reached the runtime is gone at the next rebuild, missing from a save, and absent from the native leg's render.

| site | now writes |
| --- | --- |
| `ClipCommands.cpp` — `buildDrumGridFromSlices` | the pad's model parameters for the slice markers, at its `PadRack > PadChain > Device` path, taken from `setPadDevice`'s returned id rather than by hunting the plugin |
| `ClipCommands.cpp` — `zeroSamplerAdsrBase` | the same path for the flattened ADSR; the local `setSamplerSlot` host-parameter helper is gone |
| `PadDeviceSlot.cpp` — `setupForSampler` | the model at the slot's own `devicePath_` for parameters, and the authored-state route for the loop switch and root note |
| `DeviceCustomUIManager.cpp` — `executeSamplerCommand` | the model's state document for the loop switch, the root note and a sample load |

The slice case was the sharpest: the model kept the defaults, so a rebuilt or reopened kit had sixteen pads playing the whole file.

The three authored-state edits share a new `audio/sampling/SamplerModelEdits`, which is `updateDeviceAuthoredState` plus `markDirty` — the shape `writeDeviceSettings` already gives the arpeggiator and the sequencers.

## Two consequences the routing forces

**A sample load has to author its markers.** The projection preserves whatever markers the model holds, so a newly chosen sample would otherwise inherit the previous one's trim. `MagdaSamplerPlugin::readSampleChoice()` reads the root note and the length off the file header, and the helper writes the four marker parameters after the state patch.

**`restoreState` skips the reload when the document names the source it already holds.** An authored-state edit is projected as a whole document, so a loop-switch toggle was re-reading the audio and `clearSounds()`-ing every sounding voice.

`AudioBridge::loadSamplerSample` and `SamplerFileLoader` go with them: dead once the load travels through the model, and a runtime-only route worth not leaving reachable.

## Not in scope

Undo. These values now ride every model snapshot, but none of them gets its own undo step — the same as the arpeggiator's and the sequencers' settings.

Adjacent, left alone: `PadDeviceSlot::setupForSharedDeviceUi` has the same defect for non-sampler pad devices — it writes `setParameterFromHost` plus the slot's local `DeviceInfo` copy, never the model.

## Tests

`tests/devices/test_sampler_model_edits.cpp` (4 Catch2 cases) covers the authored switches, a load's path and markers, a second load not inheriting the first's trim, and an unreadable file being refused before anything is written. One JUCE case in `test_sampler_relocate_juce.cpp` pins the same-source restore.

Both verified negatively: disabling the restore guard fails the JUCE case, disabling the marker writes fails two of the Catch2 cases.

Full Catch2 suite green (3579 cases), `Sampler Relocate Tests` 5/5, app builds.

Found by a Codex review of #2355.


magda/daw/audio/plugins/MagdaSamplerPlugin.cpp
```diff
@@ -662,6 +672,14 @@ void MagdaSamplerPlugin::restoreState(const juce::ValueTree& state) {
         return;
     }
 
+    // Already holding this audio is not a reload. An authored-state edit is
+    // projected as a whole document, so without this a loop-switch or root-note
+    // write would re-read the file and cut every sounding voice (#2379).
+    if (savedPath == samplePath_ && getSampleLengthSeconds() > 0.0) {
```
lucaromagnoli: Explicitly loading a sample that is already at `samplePath_` now skips rereading the file. `sampler_edits::loadSample()` only authors the source and markers into the model and relies on projection to reach the plugin, so if a file is replaced or edited in place and the same path is selected again, this branch returns before `loadSample()`: playback stays on the old audio buffer while markers may be derived from the new file. Previously the chooser unconditionally invoked the plugin load path. Could the explicit load path carry a force-reload signal or otherwise bypass this guard, with a regression test that replaces the contents at the same pathname?

lucaromagnoli: Confirmed and fixed in 2c49d453.

Worse than stale playback: `sampler_edits::loadSample()` reads the **new** file for the root note and the marker span and authors those on the model, so the model described audio the device was not holding.

I went for content identity rather than a force-reload flag. The projection is the only channel from model to device and it carries a whole document, so a force signal would need a side channel that only the explicit-load path could use — and it would still leave the case where the file changes underneath with no UI action at all. So the shortcut now asks the right question: is the audio I hold the file at that path, unchanged since I read it?

```cpp
bool MagdaSamplerPlugin::holdsAudioFrom(const juce::String& path) const {
    if (path.isEmpty() || path != samplePath_ || getSampleLengthSeconds() <= 0.0)
        return false;

    // Size and modification time, not the bytes: a sample is arbitrarily large
    // and this runs on every projection of the device's document.
    const juce::File file(path);
    return file.getSize() == sampleFileSize_ &&
           file.getLastModificationTime().toMilliseconds() == sampleFileModifiedMs_;
}
```

`loadSample()` stamps size and mtime, `unloadSample()` clears them. A loop-switch projection still skips the reload; anything that changed the file falls through, whichever route the projection arrived by.

Regression test as asked — `A file replaced under the same name is read again` in `test_sampler_relocate_juce.cpp`: write 1s, load, overwrite the same pathname with 2s, restore, assert the device holds 2s. Verified negatively: with the old path-only guard it reports 1s.

The one gap left is a replacement with byte-identical size *and* modification time, which I did not try to close.

PR #2383, opened 2026-09-04
Take the sampler's audio thread off the sound a load can free (#2378)

### LINKED ISSUE #2378: The sampler's audio thread reads currentSound while a load can free it
Found by a Codex review of #2355. **Pre-existing, not introduced by the port** — the retired `te::Plugin` had the same raw pointer and the same unlocked reads (`MagdaSamplerPlugin.cpp` at `c43da096^`, lines 722-724 and 799).

## The race

`currentSound` is a raw `SamplerSound*` owned by the synthesiser. The audio thread dereferences it outside the synthesiser's lock:

- `updateVoiceParameters()` reads `currentSound->sourceSampleRate` and `currentSound->audioData.getNumSamples()` at the top of every `process()`.
- `process()` reads `currentSound->sourceSampleRate` again for the playhead.

The message thread replaces it in `loadSample()`:

```cpp
synthesiser.clearSounds();      // destroys the old SamplerSound
synthesiser.addSound(newSound);
currentSound = newSound;
```

`clearSounds()` takes the synthesiser's lock and **destroys** the old sound. The audio thread's reads take no lock, so a `loadSample()` landing between the two can have the audio thread reading an `AudioBuffer` that has just been freed. `unloadSample()` and `setRootNote()` have the same shape.

The window is small but the consequence is a use-after-free on the audio thread, and every Drum Grid pad is one of these — loading a sample onto a pad while the kit is playing is the ordinary way to hit it.

## Direction

Codex's suggestion, which looks right: publish the sound as an immutable reference-counted snapshot the audio thread takes atomically, so a swap never frees anything a block is still reading. The runtime Faust devices already do this shape with their atomic state swap and retire timer (`FaustPlugin`/`FaustInstrumentPlugin`), so there is a pattern in the tree to follow rather than invent.

The alternative — taking the synthesiser lock around the audio thread's reads — is not one: that is a lock on the audio thread.

## Not urgent enough to rush, but real

It has been there since the sampler was written, so nothing is newly broken. It is worth doing before the native engine is the only engine, because the pad case makes it reachable by ordinary use rather than by a race in an edge path.

lucaromagnoli:
Closes #2378.

`updateVoiceParameters()` and the playhead both dereferenced `currentSound`, a raw `SamplerSound*` owned by the synthesiser, with no lock. `loadSample()` replaces it through `clearSounds()`, which destroys the old sound under the synthesiser's lock the audio thread never takes, so a load landing inside a block could have `applyToBuffer` reading a freed `AudioBuffer`. The constructor and `unloadSample()` install sounds the same way.

Every Drum Grid pad is one of these, so dropping a sample onto a pad while the kit plays is the ordinary way in rather than a race in an edge path.

## Not the snapshot the issue proposed

The issue suggested publishing the sound as an immutable reference-counted snapshot with a retire timer, following `FaustPlugin`. Reading the call sites, that is more machinery than the problem needs.

The audio thread only ever wanted two numbers off that pointer: `sourceSampleRate`, and `audioData.getNumSamples()`. It never reaches the buffer through it. The buffer the voices play is held by the synthesiser through JUCE's own reference counting, under its own lock, and that part was never in question.

So rather than making the object outlive its reader, this removes the dereference. `publishSoundFacts()` stores the source rate and the sample's length as atomics on the message thread, at all three sites that install a sound. There is no snapshot to retire, no reference count touched on the audio thread, and nothing new to keep in step.

`currentSound` stays exactly where it was for the message thread, which is where the buffer accessors and `setRootNote` already run.

## Two words, not one

They are read to clamp marker ranges and to scale a playhead. A pair torn across a load costs one block's clamp computed against a mismatched length, and never a dereference. Packing them into a single word the way `LaunchTap` does would buy nothing that matters here.

## No test, and why

`[sampler]` matches nothing, so there is no suite to extend, and the defect is a race whose window is a few instructions between `clearSounds()` and the next block. A test that provoked it would have to hammer `loadSample()` against a running `applyToBuffer` and would still be probabilistic, so a green run would say very little.

What is checkable instead is the property: no `currentSound` dereference remains on the audio thread. `grep currentSound MagdaSamplerPlugin.cpp` now shows only the three assignments, `publishSoundFacts()`, and the message-thread accessors.

## Note on overlap

`fix/2379-sampler-model-writes` is in flight and touches `MagdaSamplerPlugin.{cpp,hpp}` as well. This change is confined to the two audio-thread reads, the three install sites and two new members, so it should rebase cleanly, but whichever lands second is worth a look.

### DIFF
--- a/magda/daw/audio/plugins/MagdaSamplerPlugin.hpp
+++ b/magda/daw/audio/plugins/MagdaSamplerPlugin.hpp
@@ -18,7 +18,11 @@ namespace magda::daw::audio {
 struct SamplerSound : public juce::SynthesiserSound {
     juce::AudioBuffer<float> audioData;
     double sourceSampleRate = 44100.0;
-    int rootNote = 60;
+
+    /// Read on the audio thread, in SamplerVoice::startNote, and written by
+    /// setRootNote off it. Atomic because a note starting during that write is
+    /// otherwise a data race, whatever the hot path does (#2383 review).
+    std::atomic<int> rootNote{60};
 
     bool appliesToNote(int) override {
         return true;
@@ -301,9 +305,9 @@ class MagdaSamplerPlugin : public MagdaDevice {
     /// applyToBuffer (#2378).
     SamplerSound* currentSound = nullptr;
 
-    /// What the audio thread needs of the loaded sound, which is only these two
-    /// numbers. Published by the message thread once the synthesiser holds the
-    /// sound they describe.
+    /// What the audio thread needs of the loaded sound besides its root note,
+    /// which the sound carries atomically. Published by the message thread once
+    /// the synthesiser holds the sound they describe.
     ///
     /// Two words rather than one: they are read to clamp marker ranges and to
     /// scale a playhead, so a pair torn across a load costs one block's clamp
--- a/magda/daw/audio/plugins/MagdaSamplerPlugin.hpp
+++ b/magda/daw/audio/plugins/MagdaSamplerPlugin.hpp
@@ -280,6 +280,10 @@ class MagdaSamplerPlugin : public MagdaDevice {
   private:
     //==============================================================================
     void updateVoiceParameters();
+
+    /// Republish what the audio thread reads of the loaded sound. Message
+    /// thread, after the synthesiser holds it (#2378).
+    void publishSoundFacts();
     /// Put the synthesiser back to holding no audio. What an authored document
     /// with no source means (see restoreState).
     void unloadSample();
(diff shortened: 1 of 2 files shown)


magda/daw/audio/plugins/MagdaSamplerPlugin.hpp
```diff
@@ -269,12 +269,31 @@ class MagdaSamplerPlugin : public MagdaDevice {
...
     SamplerSynth synthesiser;
-    SamplerSound* currentSound = nullptr;  // owned by synthesiser
+
+    /// Owned by the synthesiser, and only ever read on the message thread. The
+    /// audio thread reads @ref soundSourceRate_ and @ref soundLengthSeconds_
+    /// instead: a load frees this one while a block may still be inside
+    /// applyToBuffer (#2378).
+    SamplerSound* currentSound = nullptr;
+
+    /// What the audio thread needs of the loaded sound, which is only these two
```
lucaromagnoli: **P2 — Publish the root note too**

The audio thread also reads `SamplerSound::rootNote`: `SamplerVoice::pitchRatioForNote()` reaches it from `startNote()` while `setRootNote()` writes it from the message thread without the synthesiser lock. Changing the root note as a MIDI note starts therefore remains a C++ data race, despite this comment claiming the two atomics are the complete cross-thread state. Please make `SamplerSound::rootNote` atomic (and load/store it explicitly), or include it in equivalent published state.

— Codex

PR #2386, opened 2026-09-04
Clock Nimbus's DSP at 32 kHz and declare its priming latency

### LINKED ISSUE #2365: Set Clouds' resampler steps the way its contract says, and declare its priming latency
## Problem

`MutableCloudsPlugin::Resampler` documents `step = Fin/Fout` and implements it (magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp:72-96: `pos += step` per pulled output sample, one input sample popped per integer crossing). `prepare` sets both steps the other way round (:182-183): the input resampler, which reads host-rate data and produces 32 kHz, gets `32000/host`, and the output resampler gets `host/32000`. Elements has the same resampler the right way (MutableElementsPlugin.cpp:129, :201).

Throughput balances, so nothing starves and dry pitch survives, which is why it was never heard as wrong. What is wrong is the engine's clock: it is fed `(host/32000)^2` more samples per real second than it assumes, 2.25x at 48 kHz and 1.9x at 44.1 kHz. Grain size and density, reverb and diffuser decay, the looping delay length, the buffer's memory span and the feedback filter corner are all off by that factor, and only right on a 32 kHz host.

A second defect in the same code: the device declares `latencySeconds` 0 (MutableCloudsPlugin.hpp:58-65) while its output is delayed by one 32-sample internal block plus the eight resampler taps (:82-84, :210-228; :226 calls it priming latency). The delay applies to the dry path too, since Clouds mixes dry and wet internally, so both hosts leave it uncompensated.

## Direction

Swap the two steps and declare the priming delay as latency (about 40 samples at 32 kHz, scaled to the host rate). Existing projects will change sound, because they were tuned against the wrong clock; that is the correct outcome.

Pre-existing, carried through the port unchanged.

lucaromagnoli:
Closes #2365.

## The clock

`Resampler` documents `step = Fin/Fout` and implements it, but `prepare` set both steps the other way round: the input resampler, which reads host-rate data and produces 32 kHz, got `32000/host`, and the output one got `host/32000`. Elements has the same resampler the right way.

Throughput balanced either way, so nothing starved and dry pitch survived — which is why it was never heard as wrong. The engine's clock was not: it was fed `(host/32000)^2` more samples per real second than it assumes, 2.25x at 48 kHz and 1.9x at 44.1 kHz. Grain size and density, the reverb and diffuser decay, the looping delay length, the buffer's memory span and the feedback filter corner were all off by that factor, and only right on a 32 kHz host.

**Existing projects will change sound.** They were tuned against the wrong clock; this is the correct outcome.

## The latency

The device declared `latencySeconds` 0 while its output lags its input by a 32-sample grain block plus the resampler windows — 33 samples at 32 kHz, measured. Clouds mixes dry and wet internally, so the dry path is delayed too, and neither host was compensating for any of it. Declared in seconds because `properties()` is read once, at construction, before the host rate is known.

## Tests

`tests/devices/test_mutable_clouds.cpp` (new, `[nimbus]`), driving the device directly:

- **Rate.** The reverb tail's 90%-energy span has to be the same number of real seconds at 44.1 and 48 kHz as at 32 kHz — the one host rate whose steps are 1 either way round, so the one the engine was always fed correctly at. A cumulative integral rather than a level crossing, because the reverb's envelope ripples by several dB. Measured: 0.83597 s / 0.83603 s / 0.83594 s.
- **Latency.** The energy centroid of an impulse through the dry path has to land where `properties().latencySeconds` says. Measured 50.0 samples at 48 kHz against 50 declared.

Both verified negatively. Restoring the swapped steps: the spans become 0.836 / 0.372 / 0.199 s, and the impulse lands at 21.2 samples. Dropping the `latencySeconds` line: declared 0, measured 50.

## Test runs

- `magda_tests`: 3584 cases, all passed.
- `magda_juce_tests` "Real Project Corpus": `project.fmchain` — the corpus project that hosts a Nimbus — passes. `project.v1bounces` fails, and fails identically on `origin/dev/0.20.0` with these changes reverted.
- `magda_juce_tests` "Null Diff Corpus": `placement.grid` fails, also identically at base. Neither project contains a Nimbus.

Both of those are a red base, not this change.

### DIFF
--- a/magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
+++ b/magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
@@ -1,9 +1,11 @@
 #include "plugins/mutable/MutableCloudsPlugin.hpp"
 
+#include <algorithm>
 #include <array>
 #include <cmath>
 #include <cstdint>
 #include <cstring>
+#include <iterator>
 
 // Upstream Mutable Instruments DSP (third_party/eurorack, magda::mutable),
 // compiled with -DTEST. Walled behind the pimpl so this is the only TU in the
@@ -27,6 +29,80 @@ inline float cubic(float x0, float x1, float x2, float x3, float t) {
     const float c = -0.5f * x0 + 0.5f * x2;
     return ((a * t + b) * t + c) * t + x1;
 }
+
+// One biquad section, transposed direct form II.
+struct Biquad {
+    float b0 = 1.0f, b1 = 0.0f, b2 = 0.0f, a1 = 0.0f, a2 = 0.0f;
+    float z1 = 0.0f, z2 = 0.0f;
+
+    void clear() {
+        z1 = z2 = 0.0f;
+    }
+    void setLowpass(double fc, double fs, double q) {
+        const double w0 = 2.0 * juce::MathConstants<double>::pi * fc / fs;
+        const double cw = std::cos(w0);
+        const double alpha = std::sin(w0) / (2.0 * q);
+        const double a0 = 1.0 + alpha;
+        b0 = static_cast<float>((1.0 - cw) * 0.5 / a0);
+        b1 = static_cast<float>((1.0 - cw) / a0);
+        b2 = b0;
+        a1 = static_cast<float>(-2.0 * cw / a0);
+        a2 = static_cast<float>((1.0 - alpha) / a0);
+    }
+    float process(float x) {
+        const float y = b0 * x + z1;
+        z1 = b1 * x - a1 * y + z2;
+        z2 = b2 * x - a2 * y;
+        return y;
+    }
+};
+
+// Pole Qs of an 8th-order Butterworth, 1/(2 cos(pi (2k+1)/16)).
+constexpr double kButterworthQ[4] = {0.50979558, 0.60134489, 0.89997622, 2.56291545};
+constexpr double kBandLimitHz = MutableCloudsPlugin::kBandLimitHz;
+
+// A section's DC group delay is 1/(Q w0), so the filter's is sum(1/Q)/w0. The
+// header declares the latency from it and cannot see this table, so tie them.
+constexpr double kQSum = 1.0 / kButterworthQ[0] + 1.0 / kButterworthQ[1] + 1.0 / kButterworthQ[2] +
+                         1.0 / kButterworthQ[3];
+static_assert(kQSum > MutableCloudsPlugin::kBandLimitQSum - 1e-6 &&
+                  kQSum < MutableCloudsPlugin::kBandLimitQSum + 1e-6,
+              "the declared latency is derived from these pole Qs");
+static_assert(kBlock == 32, "kLatencySeconds counts a 32-sample grain block");
+
+// The 32 kHz DSP is band-limited to 16 kHz and the cubic interpolators do not
+// enforce it: unfiltered, a 20 kHz tone folds to 12 kHz louder than it left,
+// and a 10 kHz tone's image sits 17 dB down coming back up. One lowpass per
+// crossing, both at host rate, both idle at 32 kHz where neither leg resamples.
+struct BandLimit {
+    Biquad l[4], r[4];
+    bool active = false;
+
+    void prepare(double hostRate) {
+        active = hostRate > static_cast<double>(kInternalRate) + 1.0;
+        if (!active)
+            return;
+        for (int i = 0; i < 4; ++i) {
+            l[i].setLowpass(kBandLimitHz, hostRate, kButterworthQ[i]);
+            r[i].setLowpass(kBandLimitHz, hostRate, kButterworthQ[i]);
+        }
+        clear();
+    }
+    void clear() {
+        for (int i = 0; i < 4; ++i) {
+            l[i].clear();
+            r[i].clear();
+        }
+    }
+    void process(float& L, float& R) {
+        if (!active)
+            return;
+        for (int i = 0; i < 4; ++i) {
+            L = l[i].process(L);
+            R = r[i].process(R);
+        }
+    }
+};
 
 inline short floatToShort(float x) {
     return static_cast<short>(juce::jlimit(-32768, 32767, juce::roundToInt(x * 32767.0f)));
@@ -80,7 +156,10 @@ struct Resampler {
         pos = 1.0;
     }
     bool pull(StereoRing& ring, float& oL, float& oR) {
-        if (ring.count < 4)
+        // The 4-point window plus whatever this call pops: step can exceed 1,
+        // and popping past count leaves it negative, which puts the next push
+        // behind head (#2365 review).
+        if (ring.count < 3 + static_cast<int>(std::ceil(step)))
             return false;
         const auto t = static_cast<float>(pos - 1.0);
         oL = cubic(ring.L(0), ring.L(1), ring.L(2), ring.L(3), t);
@@ -184,6 +263,8 @@ struct MutableCloudsPlugin::Impl {
         // the output one reads 32 kHz.
         inResampler_.step = hostRate / kInternalRate;   // host -> 32k (downsample)
         outResampler_.step = kInternalRate / hostRate;  // 32k -> host (upsample)
+        inLimit_.prepare(hostRate);
+        outLimit_.prepare(hostRate);
         reset();
     }
 
@@ -192,6 +273,8 @@ struct MutableCloudsPlugin::Impl {
         wet32k_.clear();
         inResampler_.reset();
         outResampler_.reset();
+        inLimit_.clear();
+        outLimit_.clear();
         grainFill_ = 0;
     }
 
@@ -202,33 +285,48 @@ struct MutableCloudsPlugin::Impl {
         processor_.set_playback_mode(static_cast<clouds::PlaybackMode>(mode));
     }
 
-    // Push one host input frame, drain whatever full grain blocks it enables,
-    // and pull host output frames back out. Output lags input by the resampler
-    // + grain-block latency; the rings carry the cushion.
+    // Push host input, drain whatever full grain blocks it enables, and pull
+    // host output back out. Output lags input by the resampler + grain-block
+    // latency; the rings carry the cushion.
+    //
+    // In chunks, because pushing a whole call before draining any of it makes
+    // StereoRing::kCap a limit on the host's buffer size: at 8192 the ring
+    // overruns and drops input continuously (#2365 review). A chunk bounds
+    // occupancy at roughly kChunk * max(1, 32000/hostRate) whatever the host
+    // hands over, so prepare() does not have to be told the block size.
     void process(float* L, float* R, int n) {
-        for (int i = 0; i < n; ++i)
-            hostIn_.push(L[i], R != nullptr ? R[i] : L[i]);
+        constexpr int kChunk = 256;
 
-        float sL, sR;
-        while (inResampler_.pull(hostIn_, sL, sR)) {
-            grainIn_[grainFill_].l = sL;
-            grainIn_[grainFill_].r = sR;
-            if (++grainFill_ == kBlock) {
-                runBlock();
-                grainFill_ = 0;
+        for (int base = 0; base < n; base += kChunk) {
+            const int m = std::min(kChunk, n - base);
+
+            for (int i = 0; i < m; ++i) {
+                float inL = L[base + i];
+                float inR = R != nullptr ? R[base + i] : inL;
+                inLimit_.process(inL, inR);
+                hostIn_.push(inL, inR);
             }
-        }
 
-        for (int i = 0; i < n; ++i) {
-            float oL, oR;
-            if (outResampler_.pull(wet32k_, oL, oR)) {
-                L[i] = oL;
+            float sL, sR;
+            while (inResampler_.pull(hostIn_, sL, sR)) {
+                grainIn_[grainFill_].l = sL;
+                grainIn_[grainFill_].r = sR;
+                if (++grainFill_ == kBlock) {
+                    runBlock();
+                    grainFill_ = 0;
+                }
+            }
+
+            for (int i = 0; i < m; ++i) {
+                // Zero unless the ring has a window yet: priming latency only.
+                // The filter still sees those samples, so its state stays
+                // continuous across the transition into real output.
+                float oL = 0.0f, oR = 0.0f;
+                outResampler_.pull(wet32k_, oL, oR);
+                outLimit_.process(oL, oR);
+                L[base + i] = oL;
                 if (R != nullptr)
-                    R[i] = oR;
-            } else {
-                L[i] = 0.0f;  // priming latency only
-                if (R != nullptr)
-                    R[i] = 0.0f;
+                    R[base + i] = oR;
             }
         }
     }
@@ -255,6 +353,8 @@ struct MutableCloudsPlugin::Impl {
     StereoRing wet32k_;  // 32 kHz processed output awaiting upsample
     Resampler inResampler_;
     Resampler outResampler_;
+    BandLimit inLimit_;   // before the decimation, against aliasing
+    BandLimit outLimit_;  // after the interpolation, against imaging
 
     clouds::FloatFrame grainIn_[kBlock];
     int grainFill_ = 0;
@@ -284,10 +384,47 @@ float MutableCloudsPlugin::parameterValue(int index) const {
     return values_[static_cast<size_t>(index)];
 }
 
+double MutableCloudsPlugin::tailSecondsFor(float reverb, float feedback, bool freeze) {
+    // Upstream's own reverb_amount (granular_processor.cc): feedback only joins
+    // it once frozen, and reverb alone tops out at 0.95.
+    const float amount =
+        juce::jlimit(0.0f, 1.0f, reverb * 0.95f + (freeze ? feedback * (2.0f - feedback) : 0.0f));
+
+    // Measured 90%-decay against reverb, +25% headroom (see the test). A curve
+    // rather than a formula: the decay is not a single exponential, so a fitted
+    // one is either wrong in the middle or absurd at the top.
+    constexpr float kAmount[] = {0.0f,   0.095f,  0.2375f, 0.38f, 0.5225f,
+                                 0.665f, 0.8075f, 0.95f,   1.0f};
+    constexpr double kSeconds[] = {0.1, 1.3, 2.3, 3.2, 4.2, 6.2, 10.7, 20.0, 20.0};
+    constexpr int kPoints = static_cast<int>(std::size(kAmount));
+
+    double seconds = kSeconds[kPoints - 1];
+    for (int i = 1; i < kPoints; ++i) {
+        if (amount <= kAmount[i]) {
+            const double t = (amount - kAmount[i - 1]) / (kAmount[i] - kAmount[i - 1]);
+            seconds = kSeconds[i - 1] + t * (kSeconds[i] - kSeconds[i - 1]);
+            break;
+        }
+    }
+
+    // Unfrozen feedback recirculates on its own: below 0.7 it dies with the
+    // grain, from 0.85 it never does. Declare the ceiling rather than a decay
+    // the device is not performing.
+    if (feedback >= 0.8f)
+        seconds = kMaxTailSeconds;
+
+    return juce::jlimit(0.25, kMaxTailSeconds, seconds);
+}
+
 void MutableCloudsPlugin::setParameterValue(int index, float value) {
     if (index < 0 || index >= kNumParams)
         return;
     values_[static_cast<size_t>(index)] = juce::jlimit(0.0f, 1.0f, value);
+
+    if (index == kReverb || index == kFeedback || index == kFreeze)
+        tailSeconds_.store(tailSecondsFor(displayValue(kReverb), displayValue(kFeedback),
+                                          displayValue(kFreeze) > 0.5f),
+                           std::memory_order_relaxed);
 }
 
 float MutableCloudsPlugin::displayValue(int index) const {
--- a/magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
+++ b/magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
@@ -50,9 +50,9 @@ struct Biquad {
         a2 = static_cast<float>((1.0 - alpha) / a0);
     }
     float process(float x) {
-        const float y = b0 * x + z1;
-        z1 = b1 * x - a1 * y + z2;
-        z2 = b2 * x - a2 * y;
+        const float y = (b0 * x) + z1;
+        z1 = (b1 * x) - (a1 * y) + z2;
+        z2 = (b2 * x) - (a2 * y);
         return y;
     }
 };
@@ -63,8 +63,8 @@ constexpr double kBandLimitHz = MutableCloudsPlugin::kBandLimitHz;
 
 // A section's DC group delay is 1/(Q w0), so the filter's is sum(1/Q)/w0. The
 // header declares the latency from it and cannot see this table, so tie them.
-constexpr double kQSum = 1.0 / kButterworthQ[0] + 1.0 / kButterworthQ[1] + 1.0 / kButterworthQ[2] +
-                         1.0 / kButterworthQ[3];
+constexpr double kQSum = (1.0 / kButterworthQ[0]) + (1.0 / kButterworthQ[1]) +
+                         (1.0 / kButterworthQ[2]) + (1.0 / kButterworthQ[3]);
 static_assert(kQSum > MutableCloudsPlugin::kBandLimitQSum - 1e-6 &&
                   kQSum < MutableCloudsPlugin::kBandLimitQSum + 1e-6,
               "the declared latency is derived from these pole Qs");
@@ -73,15 +73,14 @@ static_assert(kBlock == 32, "kLatencySeconds counts a 32-sample grain block");
 // The 32 kHz DSP is band-limited to 16 kHz and the cubic interpolators do not
 // enforce it: unfiltered, a 20 kHz tone folds to 12 kHz louder than it left,
 // and a 10 kHz tone's image sits 17 dB down coming back up. One lowpass per
-// crossing, both at host rate, both idle at 32 kHz where neither leg resamples.
+// crossing, both at host rate.
 struct BandLimit {
     Biquad l[4], r[4];
-    bool active = false;
 
     void prepare(double hostRate) {
-        active = hostRate > static_cast<double>(kInternalRate) + 1.0;
-        if (!active)
-            return;
+        // At every rate, 32 kHz included. Bypassing where nothing resamples
+        // would save a filter nobody hears and put the declared latency 3
+        // samples out at that rate (#2365 review).
         for (int i = 0; i < 4; ++i) {
             l[i].setLowpass(kBandLimitHz, hostRate, kButterworthQ[i]);
             r[i].setLowpass(kBandLimitHz, hostRate, kButterworthQ[i]);
@@ -95,8 +94,6 @@ struct BandLimit {
         }
     }
     void process(float& L, float& R) {
-        if (!active)
-            return;
         for (int i = 0; i < 4; ++i) {
             L = l[i].process(L);
             R = r[i].process(R);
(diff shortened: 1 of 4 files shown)


magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
```diff
@@ -54,12 +54,19 @@ class MutableCloudsPlugin : public MagdaDevice {
...
+    /// known; the dry path carries it too, since Clouds mixes dry in itself.
+    static constexpr double kLatencySeconds = 33.0 / 32000.0;
+
     //==============================================================================
     DeviceProperties properties() const override {
         return {
             .pluginId = xmlTypeName,
             .name = getPluginName(),
             .shortName = "Nimbus",
+            .latencySeconds = kLatencySeconds,
```
lucaromagnoli, line 87: **Corrected after measuring: this finding was wrong.** The table below was produced by a numerical model of the plumbing rather than by running the code, and the model was off. Built and measured against the PR's own harness, sweeping rate against block size:

| host | block | declared | measured | error |
|---|---|---|---|---|
| 32000 | 16..2048 | 33.00 | 33.00 | 0.00 |
| 44100 | 16..256 | 45.48 | 46.92 | 1.44 |
| 44100 | 512..2048 | 45.48 | 45.92 | 0.44 |
| 48000 | 16..2048 | 49.50 | 50.00 | 0.50 |
| 88200 | 16..2048 | 90.96 | 92.50 | 1.54 |

The declared constant is fine. Worst error anywhere is 1.54 samples, inside the test's own tolerance. There is a block-size dependence, but it is a single sample at 44.1 kHz between 256 and 512, not the 7-sample swing claimed above: 44.1 kHz at 2048 measures 45.92, not 39.92.

What was real in the neighbourhood is that the cushion is discovered rather than placed. The fix was not priming the ring but chunking `process()`, which was needed anyway for the `kCap` problem below, and which makes the latency block-size independent at every rate as a side effect. Sorry for the noise.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
```diff
@@ -54,12 +54,19 @@ class MutableCloudsPlugin : public MagdaDevice {
...
             .tailLengthSeconds = 2.0,  // diffuser/reverb + grain tail
```
lucaromagnoli: Right that `tailLengthSeconds` needs the same correction the steps got, **but the magnitude here is wrong**: it is not 2.25x, it is about 9x.

Measured at 48 kHz, looping delay, no feedback, time for the tail to fall below -60 dBFS:

| reverb | t60 |
|---|---|
| 0.0 | 0.059 s |
| 0.5 | 3.2 s |
| 1.0 | 18.7 s |

So 2.0 s covers reverb up to about 0.35 and truncates everything above it, worst case by 16.7 s.

Two things I got wrong beyond the number. Feedback does not add into `reverb_amount` unless the patch is frozen (`reverb_amount += feedback * (2 - feedback) * freeze_lp_`), so it is reverb that drives the tail. But unfrozen feedback has its own loop: at 0.7 the tail is 0.13 s, at 0.85 and above it never decays inside a 45 s window at all, which is sustain rather than decay and no finite declaration covers it.

A constant is the wrong shape here either way. `TracktionMagdaDevicePlugin::getTailLength()` reads `device_->properties().tailLengthSeconds` live on every call, and both `MagdaConvolutionPlugin` and `MagdaSamplerPlugin` already publish a changing value through it.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -179,8 +179,11 @@ struct MutableCloudsPlugin::Impl {
     }
 
     void prepare(double hostRate) {
-        inResampler_.step = kInternalRate / hostRate;   // host -> 32k (downsample)
-        outResampler_.step = hostRate / kInternalRate;  // 32k -> host (upsample)
+        // step = Fin/Fout, the rate the ring is read at over the rate it is
+        // pulled at (see Resampler). The input resampler reads host-rate data,
+        // the output one reads 32 kHz.
+        inResampler_.step = hostRate / kInternalRate;   // host -> 32k (downsample)
```
lucaromagnoli, line 266: The finding holds and is measurable, but **two details above are wrong**, and it understates the problem by looking at only one leg.

Measured at 48 kHz, dry only, each spur read against a 1 kHz passband tone at the same input level:

| | spur | below reference |
|---|---|---|
| input alias | 18 kHz folds to 14 kHz | 6.1 dB |
| input alias | 20 kHz folds to 12 kHz | 5.5 dB |
| output image | 10 kHz images at 22 kHz | 18.1 dB |
| output image | 5 kHz images at 27 kHz | 38.4 dB |

So the **output leg needs a filter as much as the input one**, which I missed: a 10 kHz tone's image sits 18 dB down, which is worse than most of the input aliasing. Both crossings of the 32 kHz boundary are unfiltered, not just the new one.

The 96 kHz claim is overstated. A lone single-sample impulse does produce zero energy, but continuous audio passes normally: a 220 Hz burst comes through at gain 0.349 at 96 kHz, identical to 48 kHz. The integer-ratio degeneracy is real, it just is not silence.

Fixed on the branch with an 8th-order Butterworth at 15 kHz on each leg, at host rate, bypassed at 32 kHz where neither leg resamples. That puts the four rows above at 39.8, 70.1, 81.2 and 87.1 dB down, and costs 0.9 dB at 14 kHz and nothing below 12 kHz. The rolloff below the corner turned out to be the cubic interpolators' already.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -179,8 +179,11 @@ struct MutableCloudsPlugin::Impl {
...
+        outResampler_.step = kInternalRate / hostRate;  // 32k -> host (upsample)
         reset();
```
lucaromagnoli, line 270: Confirmed by measurement, with the threshold corrected: 4096 is fine in practice, 8192 is not.

Rendering a 1 kHz tone at 48 kHz, dry only, at increasing host buffer sizes:

| block | gain | 1 kHz energy |
|---|---|---|
| 2048 | 0.3507 | 4.42e6 |
| 4096 | 0.3507 | 2.78e6 |
| 8192 | 0.2509 | 4.61e4 |

At 8192 the ring is dropping input continuously and the level collapses by about two orders of magnitude.

Draining incrementally is the better of the two fixes you suggest, since it also removes the dependence on `maximumBlockSize` entirely rather than trading one fixed limit for another.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -179,8 +179,11 @@ struct MutableCloudsPlugin::Impl {
     }
 
     void prepare(double hostRate) {
-        inResampler_.step = kInternalRate / hostRate;   // host -> 32k (downsample)
-        outResampler_.step = hostRate / kInternalRate;  // 32k -> host (upsample)
+        // step = Fin/Fout, the rate the ring is read at over the rate it is
+        // pulled at (see Resampler). The input resampler reads host-rate data,
+        // the output one reads 32 kHz.
+        inResampler_.step = hostRate / kInternalRate;   // host -> 32k (downsample)
+        outResampler_.step = kInternalRate / hostRate;  // 32k -> host (upsample)
```
lucaromagnoli, line 267: This assignment is now the one that can exceed the guard in `Resampler::pull`. The guard is `ring.count < 4`, but the pop loop runs `ceil(step)` times: `pos` enters in [1,2), leaves in [1+step, 2+step), and pops until it is back below 2. That is fine while `step <= 4`, and the input step only crosses 4 above a 128 kHz host, but at 176.4 or 192 kHz the step is 5.51 or 6.0, a pull entered with `count == 4` pops 6, and `count` goes to -2.

Nothing reads out of bounds (every index is masked), but the ring desynchronises: the next `push` writes at `(head + count)` which is behind `head`, so the first two samples of the next block land outside the readable window and are skipped. Cheap fix is to require `count >= static_cast<int>(std::ceil(step)) + 3` rather than a hardcoded 4, or to clamp the pop to `count`.

tests/devices/test_mutable_clouds.cpp
```diff
@@ -0,0 +1,184 @@
...
+    // Steps swapped, the engine runs (host/32000)^2 fast - 1.9x at 44.1 kHz,
+    // 2.25x at 48 kHz. 5% is far inside that and far outside the difference
+    // between two correctly clocked runs.
+    CHECK(at44k == Catch::Approx(at32k).epsilon(0.05));
+    CHECK(at48k == Catch::Approx(at32k).epsilon(0.05));
+}
+
+TEST_CASE("Nimbus declares the priming delay it imposes on the dry path",
+          "[nimbus][clouds][latency]") {
+    constexpr double kRate = 48000.0;
```
lucaromagnoli: **Correcting the numbers here**: they came from the same bad model as my latency comment above. Measured, 44.1 kHz at 2048 is 45.92 against 45.48 declared, not 39.92, and it does not fail. Worst error across 32/44.1/48/88.2 kHz and every block size from 16 to 2048 is 1.54 samples.

The shape of the point still stands, though for a weaker reason: the case pins one rate and one block size, and the rate case above already loops three rates, so covering them costs little.

One thing worth changing on its own merits: the energy centroid counts filter ringing as delay, so it drifts once anything with a tail is in the path. Correlating the output against the input measures the bulk delay a host actually compensates for and stays stable across rates.

tests/devices/test_mutable_clouds.cpp
```diff
@@ -0,0 +1,184 @@
...
+    buffer.setSample(0, impulseAt, 0.5f);
+    buffer.setSample(1, impulseAt, 0.5f);
+    rig.render(buffer);
+
+    // Energy centroid rather than the peak: the cubic interpolators spread the
+    // impulse over a few samples, and which one is highest moves with the phase
+    // the impulse lands on.
+    double weighted = 0.0;
+    double energy = 0.0;
+    for (int i = impulseAt; i < buffer.getNumSamples(); ++i) {
```
lucaromagnoli: The centroid integrates 1536 samples past the impulse while the thing being measured sits in the first ~55, so the assertion's headroom depends on the tail being exactly zero. It is today only because `dry_wet` 0 makes `fade_in` exactly 0 and the wet path contributes nothing. Any residue at all moves it a long way: a -80 dBFS floor over that window carries ~6% of the impulse's energy at a mean offset of ~768 samples, which drags the centroid by roughly 4 samples and blows the `<= 2.0` tolerance.

Bounding the window to something like 128 samples after the impulse keeps the same measurement and stops an unrelated change to the wet path from failing this case for the wrong reason.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -179,8 +179,11 @@ struct MutableCloudsPlugin::Impl {
     }
 
     void prepare(double hostRate) {
-        inResampler_.step = kInternalRate / hostRate;   // host -> 32k (downsample)
-        outResampler_.step = hostRate / kInternalRate;  // 32k -> host (upsample)
+        // step = Fin/Fout, the rate the ring is read at over the rate it is
```
lucaromagnoli, line 263: The comment is doing work that the type should be doing. `step` is a bare ratio and the caller has to compose it in the right order from two rates it holds anyway, which is exactly the mistake #2365 was, and a comment is the only thing stopping the next person from writing it the same way round again.

`void setRates(double inRate, double outRate) { step = inRate / outRate; }` on `Resampler` makes the transposition unwriteable at the call site: `inResampler_.setRates(hostRate, kInternalRate)` and `outResampler_.setRates(kInternalRate, hostRate)` read as the direction each resampler runs, and neither can be silently inverted. The three-line comment then goes away with it.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
```diff
@@ -54,12 +54,19 @@ class MutableCloudsPlugin : public MagdaDevice {
     }
     static const char* xmlTypeName;
 
+    /// Output lag through the resampler pair and the 32-sample grain block: the
+    /// block plus a sample of resampler group delay, at the DSP's own 32 kHz.
+    /// In seconds because properties() is read once, before the host rate is
+    /// known; the dry path carries it too, since Clouds mixes dry in itself.
+    static constexpr double kLatencySeconds = 33.0 / 32000.0;
```
lucaromagnoli: Both magic numbers here already exist in the .cpp, and nothing ties them together: 32 is `kBlock`, which is `clouds::kMaxBlockSize` from the submodule, and 32000 is `kInternalRate`. A TE-side submodule bump that changed `kMaxBlockSize`, or any edit to `kInternalRate`, would leave this constant quietly wrong with nothing failing to build.

The header cannot see the eurorack headers, which is the reason for the literals, but the .cpp can assert the link:

```cpp
static_assert(kBlock == 32 && kInternalRate == 32000.0f,
              "kLatencySeconds hardcodes both");
static_assert(MutableCloudsPlugin::kLatencySeconds == (kBlock + 1) / 32000.0);
```

While there, `kInternalRate` is a `const float` at namespace scope; `constexpr double` would let it into the assert and keep the rate arithmetic in one type.

lucaromagnoli (commented):
Review of the resampler-step fix and the latency declaration. The step swap itself is right: `Resampler::pull` consumes `step` ring samples per output sample, so the ring fed at Fin needs `Fin/Fout`, and Elements' `ratio_ = kInternalRate / hostRate` is the same shape as the new `outResampler_`. Nine findings below, mostly around the latency constant and what else was tuned against the fast clock.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -281,10 +384,47 @@ float MutableCloudsPlugin::parameterValue(int index) const {
...
+        }
+    }
+
+    // Unfrozen feedback recirculates on its own: below 0.7 it dies with the
+    // grain, from 0.85 it never does. Declare the ceiling rather than a decay
+    // the device is not performing.
+    if (feedback >= 0.8f)
+        seconds = kMaxTailSeconds;
+
+    return juce::jlimit(0.25, kMaxTailSeconds, seconds);
```
lucaromagnoli, line 456: **[P2] Include the playback buffer in the declared tail**

With reverb and feedback off, this returns 250 ms regardless of delay position, reducing the previous 2-second declaration. Using Looping Delay at 48 kHz, fully wet, density 0 and texture 0.5, after letting the delay settle, a 250 ms / 220 Hz burst produces output above -60 dBFS for 568 ms after the burst at position 0.5 and 1.078 seconds at position 1. A render trusting getTailLength() therefore truncates delayed audio. Please account for the playback buffer's contribution alongside reverb decay, and cover nonzero delay positions in the tail tests.

— Codex

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
```diff
@@ -54,16 +55,40 @@ class MutableCloudsPlugin : public MagdaDevice {
...
+    static constexpr double kBandLimitHz = 15000.0;
+    static constexpr double kBandLimitQSum = 5.12583089;
+    static constexpr double kBandLimitDelaySeconds =
+        kBandLimitQSum / (2.0 * juce::MathConstants<double>::pi * kBandLimitHz);
+
+    /// The 32-sample grain block and a sample of resampler delay at 32 kHz, plus
+    /// both filters. In seconds because properties() is a construction-time
+    /// snapshot; the dry path carries it too, since Clouds mixes dry itself. A
+    /// 32 kHz host bypasses the filters and lands ~3 samples early against it.
+    static constexpr double kLatencySeconds = 33.0 / 32000.0 + 2.0 * kBandLimitDelaySeconds;
```
lucaromagnoli: **[P2] Match the declared latency across supported sample rates**

The constant does not match the dry path at every rate. Extending the PR's bulkDelaySamples probe gives 33 samples at 32 kHz versus 36.48 declared, and 224 samples at 192 kHz versus 218.88 declared (block sizes 32, 48, 64, 96, 192 and 256). Both host adapters use this declaration for compensation, so parallel paths remain several samples out of alignment even with Nimbus fully dry. Please make the actual delay and declaration agree across supported rates, and include 192 kHz plus the 32 kHz filter-bypass case in that verification.

— Codex

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
lucaromagnoli: Confirmed and fixed in 948385e22. Reproduced first: at position 1 the burst runs 1.078 s past the input against the 250 ms I was declaring, position 0.5 gives 0.568 s, and it is rate independent as you would expect from a buffer measured in 32 kHz samples.

Chasing it turned up a worse miss of my own. I had measured the reverb curve in Looping Delay, and granular is the longest mode by about 60%:

| reverb | granular | looping delay |
|---|---|---|
| 0.5 | 5.30 s | 3.30 s |
| 0.75 | 9.86 s | 6.35 s |
| 1.0 | 32.59 s | 20.26 s |

So the table is re-measured in granular at grain size 1, `position` adds 0.8 s per unit on top, and `kMaxTailSeconds` goes from 20 s to 40 s to clear the 32.6 s worst case rather than truncate it.

Spectral will not give an envelope: three identical renders at reverb 0.75 returned 0 s, 0 s and 20.7 s, since the resynthesis randomises phase. It takes double the curve and the ceiling catches the rest.

The tail test now sweeps all four modes and grain sizes 0/0.5/1 at each position, which is what would have caught this the first time.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
lucaromagnoli: Confirmed and largely fixed in 948385e22. Reproduced both ends: 33 samples at 32 kHz against 36.48 declared, 224 at 192 kHz against 218.88, and 208 against 201.10 at 176.4 kHz, which is the worst of the set at 6.9 samples.

The 32 kHz gap was the band-limit bypass, so the filters now run at every rate. That was the largest single error and it also removes a special case from the code.

The rest is the input guard: it holds `3 + ceil(step)` samples, a fixed count and so a shrinking duration as the rate climbs, 125 us at 32 kHz against 47 us at 192 kHz. That spread lands directly in a delay declared in seconds. I tried padding the hold to a fixed duration so the constant could be exact, and it is worse: it moves the grain-block phase, and 96 kHz went to -265 us and 192 kHz became unstable across block sizes. Reverted.

What is left is declared rather than hidden. The residual is within 80 us from 32 kHz to 192 kHz, `kLatencyToleranceSeconds` names it, and the test now covers 32/44.1/48/88.2/96/176.4/192 kHz at four block sizes each, asserting in seconds rather than samples since that is the unit the declaration is in.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -28,6 +30,77 @@ inline float cubic(float x0, float x1, float x2, float x3, float t) {
...
+// crossing, both at host rate.
+struct BandLimit {
+    Biquad l[4], r[4];
+
+    void prepare(double hostRate) {
+        // At every rate, 32 kHz included. Bypassing where nothing resamples
+        // would save a filter nobody hears and put the declared latency 3
+        // samples out at that rate (#2365 review).
+        for (int i = 0; i < 4; ++i) {
+            l[i].setLowpass(kBandLimitHz, hostRate, kButterworthQ[i]);
```
lucaromagnoli: **[P1] Keep the lowpass cutoff valid at the offline analysis sample rate**

Removing the rate guard now constructs these biquads with a 15 kHz cutoff even at 22.05 kHz, above Nyquist. This is a real call path: OfflineMixAnalysis.cpp renders at 22050 Hz with usePlugins = true. Driving the device directly at that rate, fully dry, with a 100 ms / 1 kHz tone produced 2,143 non-finite output samples out of 2,205; finite output peaked at approximately 2.55e37 before overflowing. The previous revision bypassed these filters at this rate. Please use a valid cutoff/filter configuration for lower rates and add coverage for the analysis render rate.

— Codex

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
```diff
@@ -281,10 +381,65 @@ float MutableCloudsPlugin::parameterValue(int index) const {
...
+    // reverb 0.75 gave 0 s, 0 s and 20.7 s, where the curve above says 11.9.
+    // Nothing measurable to fit, so double it and let the ceiling catch the
+    // rest.
+    if (mode == 3)
+        seconds *= 2.0;
+
+    // Unfrozen feedback recirculates on its own: below 0.7 it dies with the
+    // grain, from 0.85 it never does. Declare the ceiling rather than a decay
+    // the device is not performing.
+    if (feedback >= 0.8f)
```
lucaromagnoli: **[P2] Account for decaying feedback below the sustain threshold**

Feedback below 0.8 contributes nothing to this tail declaration, but it can extend a nonzero delay for many seconds. At 48 kHz in Looping Delay, fully wet, position 1, density 0, texture 0.5, feedback 0.7 and reverb off, I settled the device for 5 seconds and rendered a 250 ms / 220 Hz burst followed by silence. The declaration is 1.5 seconds, while output remains above -60 dBFS until 17.256 seconds after the burst. A render trusting getTailLength() therefore still truncates the repeats. Please account for feedback decay together with delay length, and cover moderate feedback at nonzero positions in the tail tests.

— Codex

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
lucaromagnoli: Rechecked 948385e22 with the PR's bulk-delay probe. This finding remains unresolved: 192 kHz still measures 224 samples against 218.885 declared, and 176.4 kHz measures 208 against 201.1. Enabling the filters at 32 kHz moves that measurement from 33 to 34 samples, still against 36.481 declared. The new 80 us tolerance allows those errors, but neither host adapter uses that tolerance; they continue compensating from latencySeconds. Please keep this open until the actual delay and host compensation agree, rather than treating the wider test tolerance as a compensation fix.

— Codex

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
lucaromagnoli: Right, and this one was mine to start with. Fixed in 2b94f865a.

Reproduced before touching it: at 22.05 kHz a 1 kHz tone came back with 2143 of 2205 samples non-finite, peaking near 1e37 first. Removing the bypass put a 15 kHz corner above an 11.025 kHz Nyquist, and the biquad that produces is unstable.

The corner is `min(15 kHz, 0.45 * hostRate)` now, so it stays under Nyquist everywhere without going back to bypassing, which was worth 3 samples of declared latency at 32 kHz. Measured after: 0 non-finite at 22.05, 24 and 32 kHz, peak 0.18, which matches 44.1 kHz.

Worth saying what this does not fix. Below 32 kHz the roles swap, the output leg is the one decimating, and my output filter runs after that resampler rather than before it. So at the analysis rate the output is band-limited and finite but the aliasing has already happened upstream of the filter. Putting it right means filtering at 32 kHz ahead of the downsample, which is a different structure from the one this PR adds. For a 22.05 kHz measurement render I do not think that is worth the restructure, but flag it if you disagree.

There is a regression test at 22.05, 24 and 32 kHz asserting finite output and a sane peak.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.cpp
lucaromagnoli: Confirmed and fixed in 2b94f865a. Your figure reproduces: feedback 0.7 at position 1 runs 17.256 s against the 1.5 s I was declaring.

My threshold was the wrong shape entirely. Feedback is a delay line, so what matters is the delay length as much as the gain, and I had measured feedback only at position 0 where it contributes nothing. The surface:

| feedback | pos 0 | pos 0.25 | pos 0.5 | pos 1 |
|---|---|---|---|---|
| 0.3 | 0.06 s | 0.79 | 1.55 | 3.08 |
| 0.5 | 0.06 s | 1.57 | 3.10 | 6.16 |
| 0.7 | 0.13 s | 4.47 | 8.59 | 17.26 |
| 0.79 | 1.23 s | >45 | >45 | >45 |

So it is the repeats needed to reach -60 dBFS: delay length times `3 / -log10(gain)`. Analytic rather than fitted, which sits above every measured cell and over-declares by up to 2x at low feedback, which I would rather have than a truncated render.

The sustain threshold moves from 0.8 to 0.75 on the strength of that last row. Tests now sweep feedback 0.3/0.5/0.7 against positions 0.25 and 1.

magda/daw/audio/plugins/mutable/MutableCloudsPlugin.hpp
lucaromagnoli: Fair, and I should not have presented the tolerance as a fix. It is a bound, and the header now says that in those words.

Current numbers on 2b94f865a, bulk delay against the declaration:

| rate | declared | measured | error |
|---|---|---|---|
| 22050 | 25.14 | 27 | +84.5 us |
| 32000 | 36.48 | 34 | -77.5 us |
| 44100 | 50.28 | 51 | +16.4 us |
| 48000 | 54.72 | 53 | -35.9 us |
| 88200 | 100.55 | 103 | +27.8 us |
| 96000 | 109.44 | 110 | +5.8 us |
| 176400 | 201.10 | 208 | +39.1 us |
| 192000 | 218.88 | 224 | +26.6 us |

The constraint is that `TracktionMagdaDevicePlugin::properties_` is `const` and built in the constructor's init list, before `prepare()`, so `latencySeconds` cannot see the rate at all. Nothing I do to the constant fixes more than one rate.

Making them actually agree needs a compensating output delay: measure the natural delay at `prepare()`, then pad the output up to the declared figure. Padding the input is what I tried and reverted, since it moves grain-block phase; padding after the output resampler is inert and would work. The delay is not analytic, so it means probing at `prepare()`, and since the DSP itself always runs at 32 kHz its own contribution is rate independent, so a plumbing-only probe captures all the variation with one calibration constant for the rest.

That is a real change and a magic constant, against a residual of 1.7 samples at 48 kHz and 6.9 at 176.4 kHz. I have left it out of this PR rather than smuggle it in, and I am happy to be argued into it. Leaving this thread open as you asked.

lucaromagnoli:
Windows CI failure here is infrastructure, not the change.

The job dies in **Configure CMake (Debug)**, building JUCE's `juceaide` helper. That step compiles no MAGDA source, so nothing from this branch is involved yet:

```
sccache: error: Timed out waiting for server startup.  Maybe the remote
service is unreachable?

Run with SCCACHE_LOG=debug SCCACHE_NO_DAEMON=1 to get more information
...
ninja: build stopped: subcommand failed.
-- Configuring incomplete, errors occurred!
```

The same error repeats across every parallel compile in that step, which is the shape of the sccache daemon failing to reach the remote cache rather than a compilation problem.

History on this branch:

| commit | Windows |
|---|---|
| 2e908285a | pass |
| f0fb22418 | pass |
| 948385e22 | fail, sccache timeout |
| 8d23ed963 | fail |

Everything else on the current run is green: macOS, code-quality, security validation and the release gate all pass, Linux still running. Locally the full Catch2 suite is 857,642 assertions passing, Nimbus included.

Nothing to fix in the code for this one. It needs a re-run once the cache service is reachable, and if it keeps timing out the Windows job may need `SCCACHE_IGNORE_SERVER_IO_ERROR` or a fallback to a local cache so a remote outage cannot fail configure.

lucaromagnoli:
## On the timing/tail stabilisation plan

Worked through it. It lands one hit that mattered, and I have acted on that in 8d23ed963. I disagree with roughly half the rest, and want to be explicit about which half rather than quietly not doing it.

### Where it is right, and what changed

The sharpest claim is that the DC-delay approximation in the header is not a timing specification for these filters. That is correct and I had it wrong.

`kBandLimitDelaySeconds` was `sum(1/Q)/w0`, which is the **analog prototype's** DC group delay. The bilinear transform collapses the digital filter's delay as its corner approaches Nyquist, so the figure the header asserted was out by a lot, always in the direction of over-declaring:

| host rate | header claimed | actual digital, DC | error over both filters |
|---|---|---|---|
| 192 kHz | 54.4 us | 53.3 us | 2 us |
| 96 kHz | 54.4 us | 50.0 us | 9 us |
| 48 kHz | 54.4 us | 35.7 us | 37 us |
| 32 kHz | 56.7 us | 12.7 us | 88 us |
| 22.05 kHz | 82.2 us | 18.4 us | 128 us |

The plan's second point follows from the same place: an IIR's group delay is not one number. At 48 kHz the pair runs 35.7 us at DC against 88.3 us at 12 kHz, so no single latency figure cancels it across the band. Asking for the declaration and the delay to "agree" is only well posed once you say at which frequency.

So the constant is now measured rather than derived, and the `static_assert` tying the header to the pole table is gone, because it was asserting the wrong model. Measured bulk delay spans 1062 to 1179 us across 32 kHz to 192 kHz, so the constant is the midpoint of that, which is provably the tightest a single construction-time number can be. The bound goes from 90 us to 60 us.

The 22.05 kHz analysis render sits 104 us past it and is now stated as outside the compensation contract: it is a measurement pass with nothing to align against. It keeps its own case pinning finite, bounded output.

### Where I push back

**Sections 3 and 5 are not proportionate.** Sample-time accounting across callbacks, deliberate priming as an invariant, a rate-aware timing query threaded through `MagdaDevice` and both host adapters, plus corpus-level host verification. That is a multi-day restructure of the device framework. What it buys over what is now in the branch is under 60 us, which is 1.7 samples at 48 kHz. This PR started as a two-line step swap; a framework change does not belong inside it, and I would rather it were argued on its own merits in its own issue than smuggled in as review follow-up.

**One factual correction in section 4.** It asks that saved-state restoration and automation not leave a stale tail. `TracktionMagdaDevicePlugin::syncParametersToDevice()` writes through `setParameterValue`, which is what publishes `tailSeconds_`, so the device does track automation as it plays, and Nimbus has no `restoreState` override that bypasses it. The real exposure is narrower than stated: it is *when the host samples the tail*, not whether the device follows. That is host policy and not the device's to fix.

**"Do not enable filters solely to alter the latency measurement"** reads as a fair caution but inverts what happened. The filters run at every rate because bypassing them was worth 3 samples of declared latency at 32 kHz, and because one code path beats two. The Nyquist bug that came with it was mine and is fixed, with a regression case.

### What I think genuinely remains

Three things, which I would take as follow-up issues rather than more commits here:

1. **Filter placement below 32 kHz.** The plan is right that a cutoff clamp is not alias suppression. Below 32 kHz the legs swap, the output leg is the one decimating, and my output filter runs after that resampler rather than before it. Output is finite and band-limited, but the aliasing has already happened upstream. Fixing it means filtering at 32 kHz ahead of the downsample, which is a different structure from the one this PR adds. For a 22.05 kHz measurement render I do not think it earns the restructure, but it should be written down.
2. **The feedback loop-gain assumption.** My tail treats the knob value as the loop gain in `delay * 3 / -log10(gain)`. It sits above every cell of the measured surface, but it is not derived from the actual loop, which has a nonlinear path and filtering in it. Conservative by observation, not by construction.
3. **The ceiling is a render policy.** 40 s is what a render is willing to spend, not a claim the device is silent by then. Freeze and feedback at or above 0.75 genuinely never decay. Worth representing as policy internally rather than as a measurement.

None of those is a correctness bug. The one correctness bug in the set was the 22.05 kHz NaN, and that is fixed and pinned.

Issue #2362, opened 2026-09-04
Emit arpeggiator steps whose swung or quantized beat lands outside the block they were walked in


lucaromagnoli:
## Problem

The step walk in `ArpeggiatorPlugin::process` (magda/daw/audio/plugins/ArpeggiatorPlugin.cpp:521-592) iterates on the raw step beat, applies swing and quantize to get `swungBeat`, emits only when `swungBeat` is inside the block (:540), and advances `currentStep_` regardless (:592). A step whose swung position falls past the block end, or whose quantized position falls before the block start, is consumed without a note and never revisited: the next block's catch-up loop (:521) tests the raw beat, and `currentStep_` has already moved on.

## Scenario

1/16 rate, 120 bpm, swing at 50%: odd steps shift by 31 ms, which is three 512-sample blocks at 48 kHz. Nearly every odd step is dropped, so swing silences half the pattern instead of delaying it. Quantize at a high amount does the same in the other direction.

## Direction

Walk on the warped position, not the raw one: keep a step pending until its swung beat has been emitted or has passed, the way the pending note-off (:507-515) already carries across blocks.

Both hosts, pre-existing, carried through the port unchanged.

lucaromagnoli:
Shares `ArpeggiatorPlugin::process` with #2361. Land #2361 first if they are split: it reorders the input pass and the `midi.clear()` this walk emits after.

lucaromagnoli:
Fixed by #2434.

PR #2413, opened 2026-09-05
Fix arpeggiator input references and lost note-offs

### LINKED ISSUE #2361: Fix the arpeggiator's input pass: a dangling event reference and note-offs lost to the clear
## Problem

Two defects in the input pass of `ArpeggiatorPlugin::process` (magda/daw/audio/plugins/ArpeggiatorPlugin.cpp:371-414).

**A reference that can dangle on the TE leg.** The loop holds `const auto& msg = midi.message(eventIndex)` (:373). On the latch path it calls `sendAllNotesOff(midi)` (:380), which appends to the same buffer, and then reads `msg` again (:389). The TE view's `addEvent` is `te::MidiMessageArray::addMidiMessage`, a `juce::Array` add (TracktionMagdaDevicePlugin.cpp:106), so it can reallocate and the reference is freed storage. The native view cannot reallocate (its scratch is reserved to the port bound and refuses past it), so only the TE leg is exposed. Latch on, all keys released so the latched set is stale, an arp note sounding, then a new key pressed while the array is at capacity: the note number and velocity are read from freed memory.

**Note-offs that never leave.** The note-off appended by that same `sendAllNotesOff` (:380) is wiped by `midi.clear()` at :414, and `lastPlayedNote_` is already -1, so the sounding note never gets its off. The CC123/CC120 path (:407-410) resets the arp without emitting one at all and then drops the CC with the clear. Either way the downstream instrument hangs a note.

## Direction

Read what is needed out of the event before anything appends, and move every note-off the input pass produces to after the clear, where the transport-edge note-off (:422) already lives. All-notes-off from upstream should emit the pending note-off rather than forget it.

The hung notes are pre-existing; the reference became a hazard with the port, when the buffer behind the view started reallocating.

lucaromagnoli:
Latch replacement could append a note-off while holding a reference into the MIDI input buffer, invalidating that reference when the Tracktion buffer reallocates. Clearing the input then discarded the note-off. CC120/CC123 and same-block unlatched chord replacement could also reset the arp while forgetting a sounding note. Host panic delivered through the buffer-level all-notes-off flag was ignored and erased by the clear.

Keep the input scan read-only and capture the sounding note as part of each input-triggered reset. Emit its note-off after clearing the buffer, before generating new output. Process buffer-level panic before fresh input and preserve the flag downstream, including when no arp note is sounding. Repeated resets in one block retain exactly one pending note-off. The step scheduler and transport-stop behavior are unchanged; the misleading loop-boundary comment is corrected.

Closes #2361.

Validation:

- 169 assertions across 13 focused arpeggiator, device SDK, and engine device tests passed.
- Coverage includes latch replacement, pitch/velocity preservation, CC120/CC123, buffer-level panic, fresh input after panic, repeated resets, unlatched chord replacement, release, and transport stop.
- The CC120 regression fails against the original implementation because it produces no note-off. All three buffer-panic regressions fail before the follow-up fix. All pass with the fixes applied.
- Pre-commit checks passed.

Sounding-note cleanup is shared by resets and immediate note-offs. Removed write-only state and extracted shared device MIDI test storage.

Review follow-ups tracked separately:

- #2415: event-accurate input scheduling and note-off ordering.
- #2416: seeks, loops, and held-key/free-run behavior across transport changes.
- #2417: non-note MIDI passthrough with metadata and capacity handling.

### DIFF
--- a/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
+++ b/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
@@ -261,24 +261,25 @@ void ArpeggiatorPlugin::clearHeldNotes() {
 }
 
 void ArpeggiatorPlugin::sendAllNotesOff(DeviceMidiBuffer& midi) {
-    if (lastPlayedNote_ >= 0) {
-        midi.addEvent({juce::MidiMessage::noteOff(1, lastPlayedNote_), 0});
-        lastPlayedNote_ = -1;
-        clearMidiOutDisplay();
-    }
+    sendNoteOff(midi, takeSoundingNote());
 }
 
-void ArpeggiatorPlugin::resetArpState() {
-    currentStep_ = 0;
-    goingUp_ = true;
-    arpOriginBeat_ = -1.0;
+int ArpeggiatorPlugin::takeSoundingNote() {
+    const int note = lastPlayedNote_;
     lastPlayedNote_ = -1;
-    lastPlayedVelocity_ = 0;
     lastNoteOffBeat_ = -1.0;
+    clearMidiOutDisplay();
+    return note;
+}
+
+int ArpeggiatorPlugin::resetArpState() {
+    const int note = takeSoundingNote();
+    currentStep_ = 0;
+    arpOriginBeat_ = -1.0;
     wasPlaying_ = false;
     currentPlayStep_.store(-1, std::memory_order_relaxed);
     currentSeqLength_.store(0, std::memory_order_relaxed);
-    clearMidiOutDisplay();
+    return note;
 }
 
 ArpeggiatorPlugin::ExpandedSequence ArpeggiatorPlugin::buildSequence() const {
@@ -371,9 +372,9 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
     // note can be sounding on entry; this pass does not generate new notes.
     int pendingNoteOff = -1;
     const auto resetFromInput = [&] {
-        if (lastPlayedNote_ >= 0)
-            pendingNoteOff = lastPlayedNote_;
-        resetArpState();
+        const int note = resetArpState();
+        if (note >= 0)
+            pendingNoteOff = note;
         freeRunSamples_ = 0.0;
     };
     // Hosts can signal panic without a CC event (for example on a playhead
@@ -582,7 +583,6 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
                 timeInBlock);
 
             lastPlayedNote_ = note.noteNumber;
-            lastPlayedVelocity_ = vel;
             setMidiOutDisplay(note.noteNumber, vel);
 
             // Schedule note-off
--- a/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
+++ b/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
@@ -370,14 +370,19 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
     // and the clear below would discard any note-offs emitted here. Only one
     // note can be sounding on entry; this pass does not generate new notes.
     int pendingNoteOff = -1;
-    const auto deferNoteOff = [&] {
-        if (lastPlayedNote_ >= 0) {
+    const auto resetFromInput = [&] {
+        if (lastPlayedNote_ >= 0)
             pendingNoteOff = lastPlayedNote_;
-            lastPlayedNote_ = -1;
-            lastNoteOffBeat_ = -1.0;
-            clearMidiOutDisplay();
-        }
+        resetArpState();
+        freeRunSamples_ = 0.0;
     };
+    // Hosts can signal panic without a CC event (for example on a playhead
+    // jump). Reset before consuming fresh input, and forward the flag below.
+    const bool inputPanic = midi.isAllNotesOff();
+    if (inputPanic) {
+        clearHeldNotes();
+        resetFromInput();
+    }
     const int incomingCount = midi.size();
     for (int eventIndex = 0; eventIndex < incomingCount; ++eventIndex) {
         const auto& msg = midi.message(eventIndex);
@@ -386,21 +391,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
 
             // Latch: if old set is stale (all keys were released), clear before adding
             if (isLatched && latchedSetStale_) {
-                deferNoteOff();
                 heldCount_ = 0;
                 nextOrder_ = 0;
                 latchedSetStale_ = false;
-                currentStep_ = 0;
-                goingUp_ = true;
             }
 
             bool wasEmpty = (heldCount_ == 0);
             addHeldNote(msg.getNoteNumber(), msg.getVelocity());
             // Reset free-running clock when first note arrives
             if (wasEmpty && heldCount_ > 0) {
-                freeRunSamples_ = 0.0;
-                deferNoteOff();
-                resetArpState();
+                resetFromInput();
             }
         } else if (msg.isNoteOff()) {
             --physicallyHeldCount_;
@@ -415,21 +415,19 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
                 removeHeldNote(msg.getNoteNumber());
             }
         } else if (msg.isAllNotesOff() || msg.isAllSoundOff()) {
-            deferNoteOff();
             clearHeldNotes();
-            resetArpState();
+            resetFromInput();
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
-    if (pendingNoteOff >= 0)
-        midi.addEvent({juce::MidiMessage::noteOff(1, pendingNoteOff), 0});
+    midi.setAllNotesOff(inputPanic);
+    sendNoteOff(midi, pendingNoteOff);
 
     // --- 3. Handle transport transitions ---
-    // Note: the host briefly toggles isPlaying at loop boundaries, so we only
-    // reset the arp on a genuine stop (isPlaying goes false). On start we
-    // just update the flag — the step counter continues from where it was.
+    // Reset on a playing-to-stopped edge. On start, just update the flag;
+    // the step counter continues from where it was.
     if (context.isPlaying && !wasPlaying_) {
         wasPlaying_ = true;
     } else if (!context.isPlaying && wasPlaying_) {
--- a/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
+++ b/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
@@ -366,8 +366,18 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
     const bool isLatched = displayValue(kLatch) >= 0.5f;
 
     // --- 1. Capture incoming MIDI ---
-    // Over the block's incoming events only: sendAllNotesOff() appends to the
-    // same buffer mid-loop, and the arp must not re-consume its own output.
+    // Keep the input view read-only: appending can invalidate message references,
+    // and the clear below would discard any note-offs emitted here. Only one
+    // note can be sounding on entry; this pass does not generate new notes.
+    int pendingNoteOff = -1;
+    const auto deferNoteOff = [&] {
+        if (lastPlayedNote_ >= 0) {
+            pendingNoteOff = lastPlayedNote_;
+            lastPlayedNote_ = -1;
+            lastNoteOffBeat_ = -1.0;
+            clearMidiOutDisplay();
+        }
+    };
     const int incomingCount = midi.size();
     for (int eventIndex = 0; eventIndex < incomingCount; ++eventIndex) {
         const auto& msg = midi.message(eventIndex);
@@ -376,8 +386,7 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
 
             // Latch: if old set is stale (all keys were released), clear before adding
             if (isLatched && latchedSetStale_) {
-                // Send note-off for any currently playing note before clearing
-                sendAllNotesOff(midi);
+                deferNoteOff();
                 heldCount_ = 0;
                 nextOrder_ = 0;
                 latchedSetStale_ = false;
@@ -390,6 +399,7 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
             // Reset free-running clock when first note arrives
             if (wasEmpty && heldCount_ > 0) {
                 freeRunSamples_ = 0.0;
+                deferNoteOff();
                 resetArpState();
             }
         } else if (msg.isNoteOff()) {
@@ -405,6 +415,7 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
                 removeHeldNote(msg.getNoteNumber());
             }
         } else if (msg.isAllNotesOff() || msg.isAllSoundOff()) {
+            deferNoteOff();
             clearHeldNotes();
             resetArpState();
         }
@@ -412,6 +423,8 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
+    if (pendingNoteOff >= 0)
+        midi.addEvent({juce::MidiMessage::noteOff(1, pendingNoteOff), 0});
 
     // --- 3. Handle transport transitions ---
     // Note: the host briefly toggles isPlaying at loop boundaries, so we only
--- a/tests/devices/test_arpeggiator_input.cpp
+++ b/tests/devices/test_arpeggiator_input.cpp
@@ -81,8 +81,9 @@ struct Rig {
     }
 
     MidiBuffer run(double start, std::initializer_list<juce::MidiMessage> input = {},
-                   bool playing = true) {
+                   bool playing = true, bool panic = false) {
         MidiBuffer midi;
+        midi.setAllNotesOff(panic);
         for (const auto& message : input)
             midi.events.push_back({message, 42});
         midi.events.shrink_to_fit();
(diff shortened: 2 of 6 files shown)


magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
             }
         } else if (msg.isAllNotesOff() || msg.isAllSoundOff()) {
+            deferNoteOff();
             clearHeldNotes();
             resetArpState();
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
```
lucaromagnoli, line 425: `midi.clear()` also wipes the buffer-level all-notes-off flag, so the panic this PR is about never reaches the synth behind the arp.

`te::MidiMessageArray::clear()` sets `isAllNotesOff = false`, and `tracktion_PluginNode.cpp:213-241` sets that flag on the very buffer handed to this plugin whenever `playHeadState.didPlayheadJump()` or the track was just muted. The array is then swapped straight into `outputBuffers.midi`, so after the arp clears it every downstream plugin sees `isAllNotesOff == false`.

`DrumGridPlugin.cpp:246` already does the right thing (`chainMidi_.isAllNotesOff = inputMidi.isAllNotesOff;`). Capture `midi.isAllNotesOff()` before the clear and re-raise it with `midi.setAllNotesOff(...)` after.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
                 removeHeldNote(msg.getNoteNumber());
             }
         } else if (msg.isAllNotesOff() || msg.isAllSoundOff()) {
```
lucaromagnoli, line 418: The arp only reacts to CC120/CC123 as *messages*; the host's real panic arrives as the buffer flag, which `process()` never reads.

`DeviceMidiBuffer::isAllNotesOff()` exists and both adapters implement it, but nothing in this function calls it. On a playhead jump or a just-muted track TE raises the flag without putting a CC in the buffer, so `lastPlayedNote_` stays set and the note hangs downstream until the next step boundary or transport stop.

Add `if (midi.isAllNotesOff()) deferNoteOff();` (plus the held-note/state reset) alongside the CC branch.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
+    if (pendingNoteOff >= 0)
+        midi.addEvent({juce::MidiMessage::noteOff(1, pendingNoteOff), 0});
 
     // --- 3. Handle transport transitions ---
     // Note: the host briefly toggles isPlaying at loop boundaries, so we only
```
lucaromagnoli: This comment contradicts the code directly below it.

It claims the arp resets "only on a genuine stop", but line 435 is `else if (!context.isPlaying && wasPlaying_)`, which fires on *any* `isPlaying` -> false edge, the brief loop-boundary toggle the comment describes included. When it fires it runs `clearHeldNotes()`, dropping a chord the player is still physically holding, so a live-played arp dies at every loop point and only comes back when the keys are re-pressed.

Either the comment is stale and needs deleting, or the loop toggle needs the guard the comment promises.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
             resetArpState();
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
+    if (pendingNoteOff >= 0)
+        midi.addEvent({juce::MidiMessage::noteOff(1, pendingNoteOff), 0});
 
     // --- 3. Handle transport transitions ---
```
lucaromagnoli, line 429: A backward playhead jump strands `currentStep_` ahead of the transport: the arp goes silent and leaves a hung note.

`currentStep_` and `arpOriginBeat_` are only re-anchored by `resetArpState()`. If the playhead moves backwards while `isPlaying` stays true (dragging the cursor during playback, a loop that does not toggle `isPlaying`), `computeStepBeat(currentStep_)` is far past `blockEndBeat`, so the step loop emits nothing, and the step-7 guard `lastNoteOffBeat_ >= blockStartBeat && lastNoteOffBeat_ < blockEndBeat` never matches either, so the sounding note is never released. Playback resumes only once the transport reaches the old position again.

The mirror case is a large *forward* jump: the catch-up loop `while (computeStepBeat(currentStep_) < blockStartBeat) ++currentStep_;` is unbounded and runs the bezier per iteration on the audio thread (a 10-minute jump at 1/32 is ~9600 iterations in one callback). Re-anchoring `arpOriginBeat_`/`currentStep_` from `blockStartBeat` when the jump flag is set fixes both.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
+            deferNoteOff();
             clearHeldNotes();
             resetArpState();
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
+    if (pendingNoteOff >= 0)
+        midi.addEvent({juce::MidiMessage::noteOff(1, pendingNoteOff), 0});
```
lucaromagnoli: The recovered note-off is stamped at the block start rather than at the event that caused it, so the previous note is cut up to a full block early.

`juce::MidiMessage::noteOff(1, pendingNoteOff)` carries timestamp 0. If the latch replacement / CC123 arrived at 90% through the block, the old note is silenced one block ahead of when it should be. `msg.getTimeStamp()` is right there in the input scan (the same units the arp writes for its own events: seconds within the block), so `deferNoteOff` can capture it alongside the note number.

`checkOff()` in the new test asserts `getTimeStamp() == 0.0`, which cements the approximation.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
         } else if (msg.isAllNotesOff() || msg.isAllSoundOff()) {
+            deferNoteOff();
             clearHeldNotes();
             resetArpState();
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
     midi.clear();
+    if (pendingNoteOff >= 0)
```
lucaromagnoli: `MidiMagdaDevice::sendNoteOff(midi, noteNumber)` already exists for exactly this, and it carries the `>= 0` guard too.

`midi.addEvent({juce::MidiMessage::noteOff(1, pendingNoteOff), 0});` is the third copy of that one-liner in this file (the base helper, `sendAllNotesOff` at line 265, and now here). `sendNoteOff(midi, pendingNoteOff);` replaces both this line and the guard above it.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -366,8 +366,18 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
 
     // --- 1. Capture incoming MIDI ---
-    // Over the block's incoming events only: sendAllNotesOff() appends to the
-    // same buffer mid-loop, and the arp must not re-consume its own output.
+    // Keep the input view read-only: appending can invalidate message references,
+    // and the clear below would discard any note-offs emitted here. Only one
+    // note can be sounding on entry; this pass does not generate new notes.
+    int pendingNoteOff = -1;
+    const auto deferNoteOff = [&] {
+        if (lastPlayedNote_ >= 0) {
```
lucaromagnoli: `deferNoteOff`'s state clearing is dead at all three call sites, and it silently duplicates `sendAllNotesOff`.

Every caller runs `resetArpState()` immediately afterwards - line 403 and 420 explicitly, and the latch-stale site at 389 because `wasEmpty` is then necessarily true, so line 403 runs for it as well. `resetArpState()` already sets `lastPlayedNote_ = -1`, `lastNoteOffBeat_ = -1.0` and calls `clearMidiOutDisplay()`, so lines 376-378 never do anything a caller depends on.

The lambda can shrink to `if (lastPlayedNote_ >= 0) pendingNoteOff = lastPlayedNote_;`.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -366,8 +366,18 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
+            pendingNoteOff = lastPlayedNote_;
+            lastPlayedNote_ = -1;
+            lastNoteOffBeat_ = -1.0;
```
lucaromagnoli: Two functions now stop the sounding note with different state effects.

`deferNoteOff` clears `lastNoteOffBeat_`; `sendAllNotesOff` (line 263) clears `lastPlayedNote_` and the display but leaves `lastNoteOffBeat_` pointing at a beat in the future. Today the divergence is masked (step 7 also tests `lastPlayedNote_ >= 0`), but the two helpers doing the same job with different post-conditions is exactly the shape of the bug this PR is fixing.

One helper that returns the stopped note and clears the whole set, used by both, removes the asymmetry.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -390,6 +399,7 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
             // Reset free-running clock when first note arrives
             if (wasEmpty && heldCount_ > 0) {
                 freeRunSamples_ = 0.0;
+                deferNoteOff();
```
lucaromagnoli: The special case here is a symptom of `resetArpState()` forgetting the sounding note.

Three separate sites now have to remember to pair `deferNoteOff()` with `resetArpState()`, and any future reset path added to this function has to remember it too - which is how #2361 happened in the first place (line 403's `resetArpState()` used to drop `lastPlayedNote_` on the floor with no note-off at all).

Making the capture part of `resetArpState()` itself - have it return the note that was sounding, or take the `int&` to fill - makes the invariant impossible to forget, and the three `deferNoteOff()` lines disappear.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -405,13 +415,16 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
                 removeHeldNote(msg.getNoteNumber());
             }
         } else if (msg.isAllNotesOff() || msg.isAllSoundOff()) {
+            deferNoteOff();
             clearHeldNotes();
             resetArpState();
         }
     }
 
     // --- 2. Clear MIDI buffer (arp replaces input) ---
```
lucaromagnoli, line 424: "The arp replaces input" also destroys every non-note message: CC, pitch bend, sustain pedal, aftertouch, program change.

`midi.clear()` is unconditional and step 8 only ever writes note-on/note-off, so a mod wheel, expression pedal or pitch-bend upstream of the arp never reaches the synth behind it. Note messages are the ones the arp actually consumes; forwarding the rest (or at least channel-voice non-note messages) costs one branch in the input scan.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -390,6 +399,7 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
             // Reset free-running clock when first note arrives
             if (wasEmpty && heldCount_ > 0) {
                 freeRunSamples_ = 0.0;
+                deferNoteOff();
                 resetArpState();
```
lucaromagnoli: `goingUp_` and `lastPlayedVelocity_` are write-only.

`goingUp_` is assigned at line 273 and 394 and read nowhere; `lastPlayedVelocity_` is assigned at 276 and 587 and read nowhere (`setMidiOutDisplay` gets the velocity passed directly). `StepClock` has a live `goingUp_` for ping-pong, but this class's Up/Down handling is entirely in `buildSequence()`. Both fields and their four assignments can go.

tests/devices/test_arpeggiator_input.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include "magda/daw/audio/plugins/ArpeggiatorPlugin.hpp"
+
+namespace {
+namespace audio = magda::daw::audio;
+using Arp = audio::ArpeggiatorPlugin;
+
+// Like the Tracktion view, additions may invalidate references. Also detect
+// writes before clear deterministically, without depending on a freed read
+// happening to return the wrong note on a particular allocator.
+class MidiBuffer final : public audio::DeviceMidiBuffer {
```
lucaromagnoli: This `MidiBuffer` stub duplicates the one in `tests/devices/test_device_sdk_handles.cpp:149`.

The two are near-identical implementations of the same ten-method interface; this one adds `cleared` / `additionsBeforeClear` and an `std::stable_sort`. A third device test will copy it again. Lifting it to a shared fixture header at `tests/` root (already on the include path per `tests/CMakeLists.txt:517`) and letting this file subclass it for the write-before-clear counter keeps one copy.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli, line 429: The transport-stop branch calls `clearHeldNotes()`, which zeroes `physicallyHeldCount_` and drops keys that are still physically down, killing free-run mode.

`freeRunSamples_` exists precisely so the arp keeps running from held keys while the transport is stopped, but the stop edge wipes `heldNotes_` and `physicallyHeldCount_`, so step 4's `!context.isPlaying && physicallyHeldCount_ <= 0` immediately returns. Hitting stop with a chord held silences the arp until every key is released and re-pressed.

The new test locks this in at line 186 (`CHECK(rig.run(0.1, {}, false).size() == 0);`). Releasing the sounding note on stop is right; discarding the physically-held set is not.

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Fixed in 01eff7d13. The incoming buffer-level panic flag is captured before clear and restored afterward, including when no arp note is sounding. Regression coverage checks propagation and that the flag does not persist into the next block.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Fixed in 01eff7d13. Buffer-level panic resets held notes and arp state before fresh input is consumed, captures the sounding note for release after clear, and forwards the flag. Tests cover a stale latched chord and fresh input following panic without duplicate note-offs.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Corrected the misleading comment in 01eff7d13: it now describes the actual playing-to-stopped edge behavior without claiming special loop handling. The behavior and verification of loop/seek/stop transitions are tracked in #2416: https://github.com/Conceptual-Machines/magda-core/issues/2416. That issue explicitly calls for tracing both hosts rather than assuming every loop toggles isPlaying.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Tracked the broader transport work in #2416: https://github.com/Conceptual-Machines/magda-core/issues/2416. The buffer-panic fix already resets the scheduler before fresh input on flagged discontinuities. Unflagged seeks, bounded forward catch-up and preserving live keys across loops need explicit transport/note-ownership semantics, so they are outside this input lifetime PR. The issue includes backward/forward seek reproductions, bounded-work requirements and both-host verification, coordinated with #2362.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Agreed that block-start handling is an approximation. Opened #2415: https://github.com/Conceptual-Machines/magda-core/issues/2415. Merely retaining the input timestamp here can emit the old off after a new arp on: the scheduler currently consumes the whole input block before generating output. The follow-up covers chronological input/step/gate processing, same-pitch retriggers, equal-time ordering and partition-invariant output. This PR keeps its existing block-based scheduling while fixing the lost note and invalid reference.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Fixed in 01eff7d13 for deferred output, and completed in c53b040ee: sendAllNotesOff now uses the existing sendNoteOff helper as well.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Addressed in 01eff7d13 and c53b040ee. Removed the duplicate deferNoteOff state clearing. Input reset captures the note returned by resetArpState(), which performs the reset itself; repeated resets retain the first pending release.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Fixed in c53b040ee. takeSoundingNote() returns the old note and clears lastPlayedNote_, lastNoteOffBeat_ and the MIDI output display. Both sendAllNotesOff() and resetArpState() use it, so their sounding-note state has the same postconditions. All 169 assertions in the 13 focused tests pass.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Fixed in c53b040ee. resetArpState() now returns the captured sounding note, and the single resetFromInput helper retains it for emission after clear. Input callers no longer pair a separate capture with reset. Lifecycle resets can discard the returned note when there is no processing buffer. Repeated-reset and chord-replacement regressions still pass.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Tracked in #2417: https://github.com/Conceptual-Machines/magda-core/issues/2417. This is a valid pre-existing output-policy gap, but an append branch in the scan would be erased by clear and revive the reference hazard. The issue covers bounded forwarding/compaction, timestamps, source IDs, channel policy, panic interaction and both-host capacity behavior, with controller/pitch-bend/aftertouch/program-change tests.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Fixed in c53b040ee. Removed both write-only fields and their remaining assignments; direction is represented by the expanded sequence, and velocity goes directly to the output display.

— Codex

tests/devices/test_arpeggiator_input.cpp
lucaromagnoli: Fixed in c53b040ee. Added tests/TestDeviceMidiBuffer.hpp and reused it in both test files. The arpeggiator fixture subclasses it only to detect additions before clear. Shared storage uses stable timestamp ordering and clears the panic flag like Tracktion. The device SDK and arpeggiator tests pass.

— Codex

magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
lucaromagnoli: Tracked in #2416: https://github.com/Conceptual-Machines/magda-core/issues/2416. Agreed that the current stop test preserves an existing limitation rather than defining the desired live free-run behavior. Retaining all held notes blindly can retain clip notes whose future off will never arrive, so the follow-up addresses live versus clip/latched ownership alongside stop, seek, loop, mute and restart behavior. It includes a held-live-chord regression and a no-hung-clip-notes requirement.

— Codex

PR #2420, opened 2026-09-05
Stop noexcept functions and destructors from throwing

### LINKED ISSUE #2395: Stop noexcept functions and destructors from throwing where clang-tidy flags them
`bugprone-exception-escape` reports 61 sites. Twenty are named functions that must not throw, and those split into three groups by what actually happens if they do.

**Five `noexcept` functions in `MediaDbContext.cpp` call `std::filesystem::exists(path)`** (lines 227, 232, 276, 292, 308). That overload throws `filesystem_error` on any OS error other than not-found, and a throw out of `noexcept` is `std::terminate`. Switch to the `error_code` overload.

**Fourteen destructors do work that allocates or runs arbitrary code**, so a `bad_alloc` or a listener throw during unwinding terminates the process instead of propagating:

- `magda/daw/core/SqliteUtils.hpp:147` `~Transaction` builds the ROLLBACK string
- `magda/daw/core/SelectionManager.hpp:1076` `~NotifyGuard` erases and pushes on the listener vector
- `magda/daw/core/AutomationManager.hpp:452` and `magda/daw/core/ClipManager.hpp:861` `~BatchScope`
- `magda/daw/core/UndoManager.cpp:230` `~CompoundOperationScope`
- `magda/daw/api/remote_mcp_server.cpp:1460` `~RemoteMcpServer`
- `magda/daw/audio/analysis/OfflineMixAnalysis.cpp:87` `~MessageThreadRenderSession`
- `magda/daw/audio/insert_capture/InsertRenderCaptureService.cpp:97` `~InsertRenderCaptureService`
- `magda/daw/engine/PluginWindowManager.cpp:14` `~PluginWindowManager`
- `magda/daw/profiling/PerformanceProfiler.hpp:232` `~MonitoredProfiler`
- `magda/daw/ui/components/chain/params/ParamSlotComponent.cpp:174` `~ParamSlotComponent`
- `magda/daw/ui/components/pianoroll/PianoRollGridComponent.cpp:89` `~PianoRollGridComponent`
- `magda/engine/clip/ClipVoicePool.cpp:192` `~ClipVoicePool`

Most want the work moved to an explicit `finish()` or wrapped in a catch that logs, per site.

**Two `main` functions** (`magda/daw/cli/magda_cli_main.cpp:1011`, `magda/mcp_bridge/main.cpp:685`) have no top-level catch, so an uncaught exception dies without a message. A catch-all that prints `what()` and returns non-zero.

The other 41 sites are plain, non-`noexcept` lambdas in UI and agent code, flagged with a trace through `std::string` allocation. That is not this check's documented scope, so establish why LLVM 23 reports them before acting on any; they may be a new false-positive class to exclude in `.clang-tidy`.

Regenerate with Homebrew LLVM 23:

```
run-clang-tidy -clang-tidy-binary /opt/homebrew/opt/llvm/bin/clang-tidy -p cmake-build-debug -j 4 -extra-arg=-isysroot -extra-arg=$(xcrun --show-sdk-path) -checks='-*,bugprone-exception-escape' '^/path/to/magda-core/magda/'
```

lucaromagnoli:
## Summary
- Closes #2395. Switches five `MediaDbContext` accessors to `std::filesystem::exists`'s `error_code` overload so an OS error other than not-found can't throw out of a `noexcept` function.
- Wraps 14 destructors' teardown work (allocation and/or listener callbacks) in `try/catch(...)` so a throw during unwind can't escape an implicitly-`noexcept` destructor into `std::terminate`.
- Adds a top-level `catch` to `magda_cli`'s and `magda-mcp`'s `main()` that prints `what()` and returns non-zero.
- Documents the other 41 flagged sites (plain, non-`noexcept` UI/agent lambdas passed to `std::function`-typed callback setters) as a known LLVM 23 false-positive class in `.clang-tidy`, rather than working them one by one — they're outside this check's documented scope.
- Adds regression tests for three of the destructor fixes (`AutomationManager::BatchScope`, `ClipManager::BatchScope`, `CompoundOperationScope`): each registers a listener that throws during the scope's flush and asserts the process survives.

## Test plan
- [x] `cmake --build cmake-build-debug --target magda_daw magda_cli magda-mcp magda_tests` — clean build
- [x] Full `magda_tests` suite: 3592 passed, 13 skipped, 0 failed (857233 assertions)
- [x] Re-ran `bugprone-exception-escape` against all 20 fixed sites — zero warnings
- [x] Verified negatively: reverted each of the 3 new destructor tests' underlying fix and confirmed `SIGABRT` (uncaught exception escaping an implicitly-`noexcept` destructor), then restored and confirmed green

https://claude.ai/code/session_0149ko34fXR5PeMBqVL1tDWt

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -74,12 +74,23 @@ Checks: >-
 #
 # bugprone-exception-escape stays enabled — it correctly found 20 real
 # noexcept/destructor/main sites (#2395) — but on LLVM 23 it also flags ~41
-# plain, non-noexcept UI/agent lambdas that only reach a throwing std::string
-# or std::vector copy. The check's documented scope is noexcept functions,
-# destructors, main, and other implicitly-noexcept special members; a lambda
-# handed to a std::function-typed callback setter (juce::Button::onClick,
-# PopupMenu::showMenuAsync, MessageManager::callAsync, ...) has none of that
-# and is free to throw. Treat those as false positives, not a to-do list.
+# plain, non-noexcept UI/agent lambdas. The reported trace is a genuine false
+# positive at the diagnostic site itself: the throwing std::string/std::vector
+# copy happens while constructing the closure (a by-value capture), not inside
+# the lambda's non-noexcept call operator, and a lambda handed to a
+# std::function-typed callback setter (juce::Button::onClick,
+# PopupMenu::showMenuAsync, MessageManager::callAsync, ...) is under none of
+# this check's documented scope (noexcept functions, destructors, main) and is
+# free to throw. Don't chase these as a to-do list.
+#
+# That said, "free to throw" doesn't mean the throw is harmless: a handful of
+# these callAsync lambdas (magda/agents/coder_agent.cpp,
+# magda/agents/sound_design_agent.cpp x3) ran an apply step and then
+# unconditionally signalled a WaitableEvent the caller was blocked on. An
+# exception mid-apply skipped the signal (the caller then just timed out) and,
+# uncaught on the message thread, would have propagated into JUCE's dispatch
+# loop instead of failing only that one apply. Those four now catch and report
+# the error through the same `status`/signal path instead of skipping it.
 
 # Warnings as errors (can be adjusted based on team preference)
 WarningsAsErrors: ''
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -71,6 +71,15 @@ Checks: >-
 # whole run down — so with these enabled `make lint` and `make lint-changed`
 # exited 132/139 on every file and reported nothing at all. Worth re-testing
 # when the toolchain moves.
+#
+# bugprone-exception-escape stays enabled — it correctly found 20 real
+# noexcept/destructor/main sites (#2395) — but on LLVM 23 it also flags ~41
+# plain, non-noexcept UI/agent lambdas that only reach a throwing std::string
+# or std::vector copy. The check's documented scope is noexcept functions,
+# destructors, main, and other implicitly-noexcept special members; a lambda
+# handed to a std::function-typed callback setter (juce::Button::onClick,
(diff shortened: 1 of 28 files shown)


lucaromagnoli:
Pushed a follow-up commit addressing a Codex review pass:

- Every `catch(...)` from the first commit now logs (JUCE files use `juce::Logger`, the JUCE-free `SqliteUtils.hpp`/`ClipVoicePool.cpp` use `fprintf(stderr, ...)`).
- `InsertRenderCaptureService` and `OfflineMixAnalysis` surface a user-facing notification on the teardown-exception path via a new `magda::notifyUserAlert()` seam (`magda/daw/core/UserAlert.{hpp,cpp}`), bridged to the existing `Toast` component by `MainWindow` at startup — kept out of the audio/core layer directly since those files also compile into the headless `magda_cli` target, which doesn't link `Toast.cpp`.
- Codex found several real correctness gaps beyond "don't terminate": `MediaDbContext`'s accessors still funneled through a throwing `Config` getter, `RemoteMcpServer::stop()` could leave a joinable `std::thread` behind (its destructor terminates unconditionally), `ParamSlotComponent`/`ClipVoicePool`/`InsertRenderCaptureService` could skip mandatory cleanup on an earlier throw, `SqliteUtils::Transaction`'s nested rollback could leave a savepoint half-released, and `OfflineMixAnalysis::runOnMessageThreadBlocking` plus four `callAsync` lambdas (`coder_agent.cpp`, `sound_design_agent.cpp` x3) could skip a completion signal / leak an exception onto the message thread. All fixed; `.clang-tidy`'s comment narrowed to reflect it.

Full `magda_tests` suite green (3592 passed, 13 skipped, 0 failed); `magda_cli`/`magda-mcp` still link.

PR #2423, opened 2026-09-05
Run a rack nested inside another rack's chain

### LINKED ISSUE #2171: Nested racks are silent: a rack inside a rack's chain passes audio through unprocessed
A rack nested inside another rack's chain passes audio through unprocessed. Its chains, devices and connections are all built correctly - the devices are just never run.

Found by the rack parity cases in #2139: the `rack.nested` case renders 0.5 through the native engine and 0.6 through the incumbent, and the 0.1 difference is exactly the nested rack doing nothing.

### Reproduction, as measured

An outer rack with one chain holding only a nested rack. The nested rack has two chains, one gain at 0.5 and one at 0.3.

- native engine: peak `0.8` - the nested rack sums its two chains.
- incumbent: peak `1.0` - the input, unchanged.

### Root cause

`RackInstance::applyToBuffer` is empty (`tracktion_RackInstance.cpp:320`). A rack instance does not process anything itself; the processing lives in `RackInstanceNode`, and only `EditNodeBuilder::createNodeForPlugin` substitutes one - it special-cases `RackInstance` at `tracktion_EditNodeBuilder.cpp:1323` and calls `createNodeForRackInstance`.

`RackNode.cpp`, which builds the graph *inside* a rack, has no such special case. Its plugin loop wraps every plugin the rack holds in a plain `PluginNode`:

```cpp
for (auto plugin : rack.getPlugins())
    itemNodes[plugin->itemID] = makeNode<PluginNode> (makeNode<ConnectedNode> (...), plugin, ...);
```

So a nested rack instance becomes a `PluginNode` over a plugin whose `applyToBuffer` does nothing, and the connection matrix faithfully routes signal into it and back out again unchanged.

Everything on the MAGDA side is correct and I verified each step: `RackSyncManager` creates the nested `RackType`, wires its 16 internal connections, adds the instance to the outer rack (`addPlugin` succeeds), records it under `Track[1] > Rack[707] > Chain[1] > Rack[708]`, finds it again when wiring the outer chain, and leaves it at `dry=0 wet=1` with pins `1,2 -> 1,2`.

`RackInstance::canBeAddedToRack()` was already patched to `true` for MAGDA ("MAGDA builds distinct RackTypes for nested rack nodes"), which is what lets the instance be added at all. The node builder never got the matching half.

### Fix

`RackNode.cpp`'s plugin loop needs the same `RackInstance` substitution `EditNodeBuilder` does, so a nested instance becomes a `RackInstanceNode` over the nested `RackType`'s own graph. Worth checking recursion depth and latency reporting at the same time, since a nested rack's latency currently never reaches the outer rack's PDC either.

### Meanwhile

`rack.nested` is on the null-diff calibration list with this mechanism written beside it, so the case runs and prints its number every build. It comes off that list when this is fixed - the corpus asserts membership in both directions, so the fix cannot land unnoticed.

lucaromagnoli:
Closes #2171.

**Depends on [Conceptual-Machines/tracktion_engine#34](https://github.com/Conceptual-Machines/tracktion_engine/pull/34).** The submodule currently points at that PR's branch commit; once it merges I re-point it at the resulting commit on `magda_minimal` before this lands.

## The bug

A rack nested inside another rack's chain passed audio through unprocessed. Its chains, devices and connections were all built correctly, per the tracing in the issue - the devices were just never run.

`RackInstance::applyToBuffer` is empty. An instance processes nothing itself; the work lives in `RackInstanceNode`, and only `EditNodeBuilder` substituted one. `RackNode.cpp`, which builds the graph *inside* a rack, wrapped every plugin it held in a plain `PluginNode`, so the connection matrix faithfully routed signal into a nested instance and back out unchanged. `canBeAddedToRack()` was already patched to `true` for MAGDA, which is what let the instance be added at all; the node builder never got the matching half.

## What the fork change does

`createNodeForRackInstance` and the two rack bus IDs move out of `tracktion_EditNodeBuilder.cpp` into `RackInstanceNode`, which the unity build reaches before both callers, and `RackNode`'s plugin loop makes the same substitution.

The connection matrix needed loosening to accept it: three helpers assumed a `PluginNode` or `ModifierNode` sitting one level above its `ConnectedNode`, and a nested instance is a six-deep send/return chain. `getConnectedNode` searches for the `ConnectedNode` instead, and `getDestChannelIndex` treats anything that is not a Modifier as plugin-shaped rather than asserting.

Two of the three things the issue left open needed nothing, both checked rather than assumed:

- **Building the nested rack's graph.** `getAllPlugins` already iterates `edit.getRackList().getTypes()` and adds each rack's own plugins, so an instance living inside another rack was always found by `getEnabledInstancesForRack`, and its RackType always got its graph and both buses. Only the routing in was missing.
- **Latency and self-nesting.** After `ReturnNode::findSendNodes` runs, the return holds the rack's `SendNode` as a direct input and `getNodeProperties` maxes `latencyNumSamples` over those, so a nested rack's latency reaches the outer PDC by the same path a top-level instance's does. And `findSendNodes` already drops sends that feed back into their own return, so a rack containing itself goes silent on that path rather than deadlocking.

## Measured

`rack.nested` is an outer rack whose one chain holds a nested rack with gains at 0.5 and 0.3.

| | peak | rms |
| --- | --- | --- |
| before | -18.1 dB | -61.5 dB |
| after | -inf | -inf |

An exact null, so the case comes off the null-diff calibration list. The corpus asserts membership in both directions, so this could not have landed unnoticed and cannot silently regress. The other ten `rack.*` cases - parallel, chain fader, fader, chain mute, chain solo, both delta solos, aux, mono, instrument - still null exactly.

Full JUCE suite green. Catch2: 3610 passed, 13 skipped.

## Unrelated, seen while measuring

`placement.grid` fails when the Null Diff Corpus test is run on its own (`JUCE_TEST="Null Diff Corpus"`), with `jassert (activeContexts.isEmpty())` in `DeviceManager::closeDevices`, and passes in the full-suite run. It fails identically on a clean tree, so it is the known JUCE-test async-teardown flakiness rather than anything here. No issue open for it.


lucaromagnoli:
## The dry-path latency finding

Real, and now fixed, but the mechanism turned out to be a layer below where the finding put it. Pushed as two more commits.

### Confirming it

`rack.nested.latency` is now a corpus case: 333 samples of latency both before a nested rack and inside it. It fails at `peak=0.0 dB`.

Two things fell out of bisecting the fixture. Delta solo is not needed to reproduce it, the numbers are identical without it. And latency inside the nested rack alone nulls fine, so what matters is latency *ahead of* the instance.

`rack.latency` is the second case: the same shape with nothing nested. It nulls either way, which is what says the divergence is about nesting rather than about racks.

### The suggested fix did not work

I implemented "defer compensation until bus connections stabilize, or recompute it when latency changes" in `RackReturnNode::transform`: dropped the `hasTransformed` latch, kept the uncompensated dry pointer, resized the `LatencyNode` on any pass where the requirement changed.

It did not move the result by a single sample. Tracing showed why: the nested instance read `wet=333 dry=333 required=0` on every pass. There was nothing to recompute against, because the wet number was permanently wrong rather than merely early.

### What was actually wrong

Dumping the nested instance's wet chain at the last transform pass:

```
ReturnNode    latency=333     <- inner rack's output return
  SummingNode latency=333     <- reports 333 with inputs of 666 and 333
    SendNode  latency=666     <- the inner rack's graph, correct
    SinkNode  latency=333     <- the send-side path
```

`SummingNode::getNodeProperties()` caches its result on the first call and clears the cache only when it changes its own topology. Nothing invalidates it when an input's latency grows underneath it, which is what a `ReturnNode` does when it finds its send on a later pass.

A nested rack walks straight into that. The two rack graphs are siblings in the list `createNodesForRacks` builds, so the inner rack's output return can connect before its input return has. The `SummingNode` that `findSendNodes` puts in to carry the connection then caches the inner graph's latency without whatever precedes the instance, and the number never rises again. The instance compensated its dry path against a wet path it believed was 333 samples shorter than it was.

A top-level instance was never exposed because its rack graph is added to the top summing node ahead of the track chain, so its input return has always connected by the time anything asks.

### The fix

One line in `SummingNode::transform`: drop the cached properties at the top of every pass. The transform loop already runs until nothing changes, so the last pass holds the settled graph.

That is sufficient on its own, so I reverted the `RackReturnNode` rewrite completely. The existing `assert (dryProps.latencyNumSamples <= wetProps.latencyNumSamples)` stays in place and now actually holds.

### Measured

All thirteen `rack.*` cases null exactly, including both new ones. Full JUCE suite green. Catch2: 3610 passed, 13 skipped. The corpus size guard goes from 67 to 69.

### Not addressed

The Linux `test_remote_mcp_server.cpp:1134` failure on the previous run is unrelated to anything here. It passes 5 times out of 5 locally, and it looks like a timing casualty of the Catch2 sharding that landed in #2421 rather than a real regression.

tests/NullDiffCorpus.cpp
```diff
@@ -1835,6 +1835,84 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
...
+        // which is silence while the two are aligned and a pair of
+        // opposite-polarity impulses one compensation apart when they are not.
+        ChainInfo innerChain;
+        innerChain.id = 1;
+        innerChain.name = "Delayed";
+        innerChain.elements.emplace_back(hostedDevice(983, HostedRole::Latency));
+
+        std::vector<ChainInfo> innerChains;
+        innerChains.push_back(std::move(innerChain));
+        auto inner = rackOf(713, "Inner Latency", std::move(innerChains));
```
lucaromagnoli, line 1890: **[P2] Enable delta solo so this case exercises the remaining dry-path bug**

The comments and case description say the inner rack is delta soloed, but `inner.deltaSolo` remains false. On this head, adding `inner.deltaSolo = true` here makes `rack.nested.latency` fail with a 0.0 dB peak residual: the native render is silent, while Tracktion emits -1/+1 impulses 333 samples apart. The committed test therefore passes without exercising the dry path from the original finding.

The stale `SummingNode` cache diagnosis in the follow-up comment is correct, and my earlier explanation missed that layer. However, invalidating that cache alone is insufficient: `RackReturnNode::transform` still latches `hasTransformed` and never revisits the dry compensation after the wet latency changes. Please enable delta solo in the regression coverage and address that remaining compensation issue before considering the finding resolved. If keeping this in the corpus, use a non-silent reference/sibling path or a dedicated cancellation assertion, since the corpus rejects two completely silent renders.

— Codex

lucaromagnoli: Both correct, fixed in 7b1ef2eaa.

I took delta solo out while bisecting and never put it back, and the comment kept claiming it. With it on, the case fails at 0.0 dB peak exactly as described.

`RackReturnNode::transform` no longer latches: it keeps the dry input as given and sizes the LatencyNode over it on any pass where the requirement changes. I had measured that change as unnecessary and reverted it, but that measurement was taken with delta solo missing from the case, so it was measuring nothing.

On the silent-render point, you were right about that too: with the compensation correct both renders go silent and the corpus rejects it. The outer rack now has a parallel sibling chain carrying the material, and the delta-soloed nested rack adds to it only when the alignment is wrong.

Each fix is independently necessary. The case fails at 0.0 dB peak with either the RackReturnNode change or the SummingNode cache invalidation removed.

Issue #2418, opened 2026-09-05
Carry the buffer panic flag through the native engine's MIDI port


lucaromagnoli:
`DeviceMidiBuffer::isAllNotesOff()` is part of the device SDK, but `EngineMidiBufferView` (`magda/daw/audio/plugins/engine/EngineMagdaDevice.cpp:95`) holds it in a per-block member (`allNotesOff_`, line 193) that is never seeded from `block.midiIn` and never read back onto `block.midiOut`. Every device on the native engine therefore sees `false`, and any `setAllNotesOff()` it makes is discarded.

The engine signals discontinuity with explicit note-offs from `ClipMidiSource` instead, so the design question is whether the port grows a panic flag of its own (locate, track mute, launch discontinuity) or the SDK stops exposing one that only the Tracktion adapter can set.

Scope:
- Decide the contract: a port-level panic flag, or explicit note-offs only with the SDK accessor removed.
- If the flag stays, seed it from the port on the way in and write it back on the way out, for the same discontinuities the Tracktion adapter reports (`tracktion_PluginNode.cpp:212`: playhead jump, just-muted track).
- Keep the reading and the writeback free of allocation on the audio thread.

Acceptance:
- A device reading `isAllNotesOff()` sees the same value for the same discontinuity through both adapters, or the accessor is gone.
- A device that forwards the flag has it reach the next device in the chain on the native engine.
- Tests for locate, track mute and launch discontinuity on the engine path.

Arpeggiator panic handling added in #2413 is Tracktion-only until this is resolved, and it blocks the both-adapter acceptance in #2416.

lucaromagnoli:
Related: #2424 (#2347). The file pointers above are stale after that PR: `EngineMidiBufferView` is gone, and the flag is now split across `DeviceMidiInput::isAllNotesOff()`, which `EngineMidiInputView` answers with a constant false, and `DeviceMidiOutput::setAllNotesOff()`, which `EngineMidiOutputView` drops (`EngineMagdaDevice.cpp`). The Tracktion adapter reads and writes `MidiMessageArray::isAllNotesOff` on both sides. The decision in this issue is unchanged: either the engine's port grows a panic flag that is seeded on the way in and written back on the way out, or the two SDK accessors go.

lucaromagnoli:
Closed by #2429, merged into dev/0.20.0 as b703af2a.

Issue #2417, opened 2026-09-05
Preserve non-note MIDI through the arpeggiator


lucaromagnoli:
Follow-up to PR #2413: https://github.com/Conceptual-Machines/magda-core/pull/2413#discussion_r3940948459.

ArpeggiatorPlugin clears its input and generates only arp notes. Non-note messages such as mod wheel, expression, sustain, pitch bend, aftertouch and program change are discarded before they can reach the downstream instrument.

Scope:
- Define which non-note messages pass through, including channel routing (arp-generated notes currently use channel 1).
- Preserve timestamps, source IDs and relative ordering.
- Keep CC120/CC123 and the buffer panic flag consistent with the arp's own reset behavior.
- Forward messages using bounded storage or in-place compaction without appending while references into the input remain live. Account explicitly for long/SysEx messages and native port capacity.

Acceptance:
- Verify CC1/CC11/CC64, pitch bend, channel/poly aftertouch and program change alongside generated arp notes.
- Verify panic and coincident off/on ordering, metadata preservation, and capacity behavior through both hosts.
- No input references invalidated and no new audio-thread allocation from the forwarding strategy.

Deferred from #2413 because forwarding changes the device's output policy and buffer/capacity handling; adding an append branch before the current clear would lose the messages and reintroduce the reference hazard.

lucaromagnoli:
Related: #2424 (#2347). The forwarding strategy this issue worries about is now the plain shape: copy the chosen non-note messages from `context.midiIn` to `context.midiOut` alongside the generated notes. There is no clear to append before, and input references stay valid because the input is read-only. What remains is the policy (which messages, which channel) and the capacity accounting: on the native leg the output scratch is bounded by the port's byte budget (`EngineMidiOutputView`), so forwarded SysEx counts against it.

lucaromagnoli:
Closed by #2431, merged into dev/0.20.0 as 9886c296. Follow-up for the other three MIDI FX: #2432.

Issue #2415, opened 2026-09-05
Process arpeggiator input at event time rather than block start


lucaromagnoli:
Follow-up to PR #2413, review https://github.com/Conceptual-Machines/magda-core/pull/2413#discussion_r3940948314. Related scheduler work: #2362.

The arp consumes every input event before generating the block's output. A latch replacement or CC120/CC123 late in the block releases the old arp note at timestamp zero. It can also change the notes used for earlier steps in that block.

Simply copying the input timestamp onto the deferred note-off is insufficient: the current scheduler may already generate a replacement note-on earlier in the block, and a later old note-off can cut that new voice (especially for the same pitch).

Scope:
- Process held-note changes and panic messages in chronological order with arp steps and scheduled gate note-offs.
- Preserve timestamps in seconds, including host sub-block offsets, and define ordering for coincident events.
- Preserve the read-only input/reference safety and bounded audio-thread storage introduced in #2413.

Acceptance:
- At 48 kHz, put a latch replacement or panic 90% into a block and verify the prior note is not cut early.
- Cover same-pitch retrigger, multiple replacements, gate-off before the input event, and equal-time off/on ordering.
- Equivalent input timelines produce equivalent output across small, large and irregular callback partitions, with no stuck or duplicate notes.
- Cover both host adapters.

Deferred from #2413 because it changes input/scheduler ordering, beyond preserving notes through the existing clear.

lucaromagnoli:
Related: #2424 (#2347) split the SDK's MIDI into a read-only `midiIn` and a separate `midiOut`. The input-reference safety this issue asks to preserve is now the contract rather than a discipline: the arp reads `in` and writes `midi` (`ArpeggiatorPlugin.cpp`, `process()`), and there is no clear to order against. The scheduler ordering work itself is untouched.

lucaromagnoli:
Fixed by #2435.

Issue #2346, opened 2026-09-03
A MagdaDevice instrument's MIDI input is doubled by its wrapper rack on the TE leg


lucaromagnoli:
## Problem

Everything downstream of a MAGDA internal instrument on the TE leg sees every incoming MIDI event twice.

`InstrumentRackManager::wrapInstrument` wires two paths to the rack's MIDI output:

- `rackType->addConnection(synthId, 0, rackIOId, 0)` — the plugin's own MIDI output, unconditional (`InstrumentRackManager.cpp:84`).
- `rackType->addConnection(rackIOId, 0, rackIOId, 0)` — the raw input around the plugin, when `passRawMidiInput` (`:91-92`). For a plain instrument that is always on: `outputsMidi` is false so the node is `RawInputOnly`, which satisfies `passesRawMidiInput()` (`ChainRoutingModel.hpp:47-52, 88-95`).

The comment at `:82-83` states the assumption that makes this safe — *"a plain synth emits no MIDI, so the connection simply carries nothing"* — and it does not hold for MAGDA's own SDK instruments. `MagdaCompiledPolyInstrument::process` reads `context.midi` and never clears it (`MagdaCompiledPolyInstrument.cpp:478`), as does `MutableRingsPlugin::process` (`:329`). TE's `PluginNode` passes the array on in place after `applyToBuffer` (`tracktion_PluginNode.cpp:266`), so the wrapper's MIDI output is the input the device left behind. `SummingNode` then merges both paths (`tracktion_SummingNode.h:185-192`).

So pin 0 carries the input, the raw passthrough carries it again, and a MIDI-triggered device after the instrument fires twice per note.

## Why it is not caught

The null-diff corpus's hosted instruments substitute their own output rather than leaving the buffer alone (`tests/NullDiffHostedPlugin.cpp:345`), so the corpus never exercises the shape.

## Direction

An instrument's pin 0 should carry the instrument's own MIDI output only. Either the SDK instruments clear `context.midi` the way the MIDI devices do, or the wrapper stops connecting `synthId, 0` for a device that declares no MIDI output.

The same root as #2345 — the SDK's single MIDI buffer makes passthrough what happens when a device does nothing — so #2347 removes the class of bug rather than these instances.

## Priority

[...]

PR #2424, opened 2026-09-05
Split the device SDK's MIDI buffer into a read-only input and a separate output

### LINKED ISSUE #2347: Split the SDK's MIDI buffer into a read-only input and a separate output
## Problem

`DeviceProcessContext::midi` is one `DeviceMidiBuffer*` serving both directions (`MagdaDevice.hpp:104`), inherited from `te::MidiMessageArray`, which is what TE hands a plugin. Both adapters implement it that way: `EngineMagdaDevice` decodes the input into a scratch and writes the whole scratch back (`EngineMagdaDevice.cpp:360-375`, `:421-431`), and the TE wrapper hands the device `context.bufferForMidiMessages` directly (`TracktionMagdaDevicePlugin.cpp:200-202`).

So **passing MIDI through is what happens when a device does nothing**. It is not something a device opts into; it is the default, and every device has to remember to call `clear()` to avoid it. Three of the five MIDI devices remember (`ArpeggiatorPlugin.cpp:414`, `MidiStrumPlugin.cpp:353`, `MidiChordEnginePlugin.cpp:49`); the two sequencers made it conditional, which is #2345. The SDK instruments do not clear at all, which is #2346.

The contract also hides a distinction that matters:

- **what the device emits** — the arp's line, the sequencer's pattern. The device's business, and nothing else can know it.
- **whether the raw stream continues downstream unchanged** — a routing decision about the chain, which the user expresses with a toggle and the plan compiles into a `ChainMidiMerge`.

One buffer forces both through the same channel, so a device cannot state the first without implicitly deciding the second.

## Direction

Give `DeviceProcessContext` a `const` MIDI input and a separate MIDI output. A device reads what arrived and writes what it emits; a device that genuinely wants to transform what flows past it copies across explicitly, where it is visible in the source and to whoever reads the plan.

That makes the #2345 and #2346 classes unrepresentable rather than fixed case by case.

## Note

This does not move the thru decision to the device. Thru for a third-party plugin is still the host adding the raw stream, because JUCE replaces the host buffer with the plugin's declared MIDI output (`juce_VST3PluginFormatImpl.h:2512`) and no host can ask a plugin to pass its input through. The plan owns thru either way; the split stops a MAGDA device from silently doing it a second time.

## Scope

Touches every `MagdaDevice::process` implementation and both adapters. Worth sequencing after #2345, which is the user-visible bug.

## Background

Found while investigating #2345.

lucaromagnoli:
Closes #2347.

## Problem

`DeviceProcessContext::midi` was one mutable buffer serving both directions. Passing MIDI through was what happened when a device did nothing, so every MIDI device had to remember to `clear()`, and "did nothing to MIDI" was indistinguishable from "intentionally passed it on". #2345 (a thru device playing every note twice) was one instance of that class.

Related arpeggiator issues, made simpler by this contract but not done here: #2415 (input at event time), #2417 (forward non-note MIDI: now an explicit copy from `midiIn` to `midiOut`), #2418 (panic flag on the native port: `DeviceMidiInput::isAllNotesOff()` is a constant false there, `setAllNotesOff()` is dropped).

## Change

- `DeviceProcessContext` now carries a `const DeviceMidiInput* midiIn` and a `DeviceMidiOutput* midiOut`, empty on entry. Both null or both set. `DeviceProperties::producesMidi` says whether a device has a MIDI output at all.
- **Native adapter** (`EngineMagdaDevice`): one scratch per direction, each sized from its own port bound. The in-place `setEvent`/`removeEvent`/`sortByTimestamp` machinery and its swap-based hazards are gone.
- **Tracktion adapter** (`TracktionMagdaDevicePlugin`): read-only view over the node's `MidiMessageArray`; the device writes to a scratch that is swapped into the node's buffer only when the device declares `producesMidi`. Otherwise the buffer is left in place, which is how this engine carries the raw input past a device that emits no MIDI (`RawInputOnly` in `ChainRoutingModel`). A debug assert catches a device emitting MIDI it never declared.
- Arpeggiator, strum and both step sequencers declare `producesMidi` and drop their `clear()`. The five SDK instruments read `midiIn`.
- Tests follow the contract. A new engine test pins that a device's input never reaches its output port; the former "rewrite, reorder and drop" test becomes "forwards long messages by copying, never in place".
- `DEVICE_SDK.md` gains a paragraph on the MIDI contract.

Thru is unchanged: it is still the host's merge behind the device (`DeviceInfo::midiInThru`, `OpRole::ChainMidiMerge`), which is the only way a host can offer it over a plugin it does not write.

## Behaviour

No change for any shipped device on either engine. The MIDI effects already cleared their buffer; the instruments never wrote MIDI.

## Tests

- `make test`: 3612 passed, 13 skipped, 857,644 assertions.
- `make test-juce`: 440 suites, 0 failures; null-diff and real-project corpora green.

## Not in this PR

- On the Tracktion leg the thru toggle on a top-level MIDI effect still does nothing (MIDI effects are not rack-wrapped, so there is no host merge). Accepted in #2345 for the migration.
- The Chord Engine is a MIDI listener declared `DeviceType::MIDI` with the `midi-generator` tag, so it shows a thru toggle that does nothing and the model treats it as an emitter. Noted on #2314, the Chord Engine's port to MagdaDevice.
- #2346 unchanged.

https://claude.ai/code/session_01M3rnQFCXAVJzuRJstMu6At


magda/daw/audio/plugins/engine/EngineMagdaDevice.cpp
```diff
@@ -402,30 +319,23 @@ void EngineMagdaDevice::process(magda::engine::DeviceBlock& block) {
...
-    // What is in the scratch is what the device produced. A device that left
-    // its input in there would be sending a second copy of a stream the plan
-    // already merges behind it (#2345), and would be over this budget as soon
-    // as more than one source fed the track.
+    // The input cannot reach this port any more by construction (#2347); thru
+    // is the plan's merge behind the device.
     int bytesWritten = 0;
 
-    for (const auto& event : midiScratch_) {
+    for (const auto& event : midiOutScratch_) {
```
lucaromagnoli, line 340: [P2] Enforce producesMidi in the native adapter too

When a DeviceBlock has a MIDI output port, this loop forwards everything the SDK device wrote even if properties_.producesMidi is false. The Tracktion adapter now discards that same undeclared output and asserts, so the new SDK declaration has different semantics between engines. The EmittingDevice fixture in test_engine_device_layer.cpp already exercises this discrepancy: it leaves producesMidi at its default false while its tests require emitted events.

Please gate native output on properties_.producesMidi, update the emitting fixture to declare its output, and add coverage that an undeclared emitter cannot forward MIDI through the native adapter.

lucaromagnoli: Done in 68a9e42f: the native write-back returns early when `properties_.producesMidi` is false, `EmittingDevice` declares its output (with a `declared` flag for the negative case), and a new case pins that an undeclared emitter puts nothing on the port. Verified negatively: the new test fails with the gate removed. Full Catch2 and JUCE suites green locally. The branch also carries the dev merge with the serial Linux tests.

lucaromagnoli:
@codex

lucaromagnoli:
@codex

Issue #2416, opened 2026-09-05
Define arpeggiator seek, loop and held-key behavior across transport changes


lucaromagnoli:
Follow-up to PR #2413:
- https://github.com/Conceptual-Machines/magda-core/pull/2413#discussion_r3940948281
- https://github.com/Conceptual-Machines/magda-core/pull/2413#discussion_r3940948592
- https://github.com/Conceptual-Machines/magda-core/pull/2413#discussion_r3940948253
Coordinate with #2362, which also changes the step walk.

Current behavior:
- Without a host panic/reset signal, a backward timeline discontinuity can leave currentStep_ and a scheduled note-off ahead of transport; a large forward jump walks skipped steps on the audio thread.
- A playing-to-stopped edge clears all held notes and physicallyHeldCount_, so a still-held live chord does not continue in free-run mode.
- #2413 now handles the buffer-level panic flag and resets before fresh input. That covers flagged discontinuities, but does not distinguish seeks, loops, track mute, explicit panic or a genuine transport stop. The old comment claiming special loop treatment was removed; actual loop behavior in each host still needs verification.

Scope:
- Trace both hosts' discontinuity/loop/panic signals and define the device contract.
- Separate live-held notes from clip/latched state where necessary: blindly retaining every held note on stop can retain notes whose clip note-off will never arrive.
- Stop the old sounding note exactly once and re-anchor the scheduler on a seek without work proportional to skipped steps.
- Specify live free-run behavior on stop, loop continuity, restart, mute and explicit panic.

Acceptance:
- Tests for backward and large forward seeks with and without available host panic signaling.
- Bounded catch-up work at the fastest rate with ramp/swing/quantize.
- A live-held chord survives the intended stop-to-free-run transition and releases normally; clip notes do not hang on stop.
- Loop and restart tests through both adapters.

Deferred from #2413 because it changes transport and note-ownership semantics. Its current stop regression preserves existing behavior, not the final desired free-run policy.

lucaromagnoli:
One correction to "Current behavior": the held-key clearing is not only on the stop edge, and on the start edge it is new as of 01eff7d.

[...]

PR #2429, opened 2026-09-06
Carry the buffer panic flag through the native engine's MIDI port

### LINKED ISSUE #2418: Carry the buffer panic flag through the native engine's MIDI port

lucaromagnoli:
Closes #2418.

## Problem

The device SDK asks one question — `DeviceMidiInput::isAllNotesOff` — and only the Tracktion adapter answered it. The engine's ports are bare `juce::MidiBuffer`s with no flag beside the events, so `EngineMidiInputView` returned a constant `false` and `EngineMidiOutputView` dropped `setAllNotesOff`. Every device on the native engine read `false`, and the arpeggiator's panic handling from #2413 was Tracktion-only.

## Contract

The port grows the flag rather than the SDK losing the accessors. It is the device's only way to hear about a discontinuity the host signals without a CC event, and the fork raises one per plugin (`tracktion_PluginNode.cpp:212`) that a device has no other source for.

**A panic is only raised where the host re-asserts what should still be sounding on the same buffer.** That is the contract `ArpeggiatorPlugin` states at its `inputPanic` branch, and what `ClipMidiSource::playLane` does under `if (!block.continuous)` — chase the controllers, then the notes the instant is inside. A panic raised without that behind it tells a device to drop a chord nothing will give back.

Two sites qualify:

| site | chased by |
|---|---|
| `!BlockInfo::continuous` — locate, loop wrap, transport start | `playLane` → `chaseClip` |
| the arrangement being handed its track back (`hold.gained`) | same, via `lane.continuous = false` |

## Change

- **`DeviceBlock`** carries `midiInAllNotesOff` in and `midiOutAllNotesOff` out. One byte per MIDI slot in `PlanExecutor`, written by the op that owns the slot every block the way its buffer is — nothing allocates on the audio thread, and there is no stale flag to clear because nothing reads a slot its producer has not run.
- **Seeding at a device**: the port's flag OR'd with `!block.continuous`, which both hosts raise per device rather than passing along a chain.
- **Propagation**: `MergeMidi`, `MidiNoteGate`, the MIDI `Delay` and the `Fader`'s MIDI leg. The delay passes it undelayed — the line holds events, not the discontinuity that produced them, and a device is owed its panic on the block that caused it.
- **`ClipMidiSource`** raises it for the arrangement hand-back, through a new `EngineMidiSource::raisedAllNotesOff()` whose docs carry the chase rule.
- **Writeback**: a device that produces MIDI and says nothing drops the panic, matching what the fork does with its fresh output buffer. The adapter reads the flag back before the empty-block returns, since a panic with no events is exactly what a device raises when it has nothing left to say.
- `EngineMagdaDevice.hpp`'s divergence note loses the panic bullet; multi-out is the only one left.

## Two things deliberately not signalled

Both were in the first push and came out after review (76bb09a2) — each told a device to drop held notes where nothing re-asserted them.

**Track mute.** MAGDA's mute is a gain with the track still rendering, so its meters stay live and nothing is withheld from a device. Neither the mute nor the unmute leaves the block discontinuous, so a panic there dropped the arp's clip-held chord and left the track silent until the next onset. The fork raises one because `shouldTrackContentsBeProcessed()` goes false and the track's MIDI stops — a difference in what the two engines do with a muted track, not in how they carry the flag. Pinned by a test that asserts no panic across unmuted → muted → unmuted.

**Session slot launch.** A launch is self-describing: the slot it replaces owes note-offs for everything it started (`endSlot`), so the release is already in the stream. The launched slot starts at its own origin and chases nothing, and a slot still sounding beside it — reachable, since `forEachSlot` walks every slot with a handle and the engine does not stop A when B launches — was never re-asserted.

Re-asserting instead would mean re-emitting note-ons, restriking every synth on the track: for a mute, which is a volume change, and for a chord that never stopped sounding.

## Validation

Full Catch2 suite (3625 passed, 857,701 assertions) and the full JUCE suite green, null-diff corpus through both engines included.

Verified negatively:
- reverting the adapter and executor halves fails 8 of the targeted tests;
- the both-adapter JUCE case fails on the engine leg alone while the fork's leg still passes, which is exactly the divergence in the issue;
- reverting the `ClipMidiSource` hand-back raise fails its launch test;
- forcing `midiInAllNotesOff = true` fails the no-panic-on-mute test on 4 assertions, so it cannot pass vacuously.

New tests: locate and hand-back on the engine path; no panic on mute or on a slot launch; device-to-device forwarding; the adapter's read and writeback; and one JUCE case running the arpeggiator through both adapters over the same discontinuity.

Unblocks the both-adapter acceptance in #2416.

### DIFF
--- a/magda/engine/clip/ClipMidiSource.cpp
+++ b/magda/engine/clip/ClipMidiSource.cpp
@@ -501,12 +501,6 @@ bool ClipMidiSource::renderSession(juce::MidiBuffer& out, const BlockInfo& block
                 playLane(out, material, slot.midi, from, to);
             });
 
-            // A slot that begins sounding here started material the transport
-            // never moved for, which is the same discontinuity a device is
-            // owed as the arrangement's hold.gained below (#2418).
-            if (!status.soundingAtStart && status.playingAtEnd())
-                raisedAllNotesOff_ = true;
-
             // The session's own way for a note to end: a stop is not a hole,
             // not a snapshot swap and not the end of a span.
             //
--- a/magda/engine/clip/ClipMidiSource.cpp
+++ b/magda/engine/clip/ClipMidiSource.cpp
@@ -501,6 +501,12 @@ bool ClipMidiSource::renderSession(juce::MidiBuffer& out, const BlockInfo& block
                 playLane(out, material, slot.midi, from, to);
             });
 
+            // A slot that begins sounding here started material the transport
+            // never moved for, which is the same discontinuity a device is
+            // owed as the arrangement's hold.gained below (#2418).
+            if (!status.soundingAtStart && status.playingAtEnd())
+                raisedAllNotesOff_ = true;
+
             // The session's own way for a note to end: a stop is not a hole,
             // not a snapshot swap and not the end of a span.
             //
@@ -516,6 +522,7 @@ bool ClipMidiSource::renderSession(juce::MidiBuffer& out, const BlockInfo& block
 
 void ClipMidiSource::render(const BlockInfo& block, juce::MidiBuffer& out) {
     bytesUsed_ = 0;
+    raisedAllNotesOff_ = false;
 
     // What a previous block could not fit goes out first, before anything can
     // start a note of the same pitch behind it.
@@ -614,8 +621,10 @@ void ClipMidiSource::render(const BlockInfo& block, juce::MidiBuffer& out) {
         // ended when it lost the track and the timeline ran on underneath.
         // Resuming without a chase leaves a sustained pad silent, which is the
         // case playLane already answers for a locate.
-        if (hold.gained)
+        if (hold.gained) {
             lane.continuous = false;
+            raisedAllNotesOff_ = true;
+        }
     }
 
     // The session owns every sample. The arrangement owes note-offs on the
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -224,7 +224,6 @@ void PlanExecutor::reset() {
     audioSlots_.clear();
     midiSlots_.clear();
     midiSlotPanic_.clear();
-    wasInaudible_.clear();
     slotChannels_.clear();
     midiByteBounds_.clear();
     portOffsets_.clear();
@@ -276,7 +275,6 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
     insertForOp_.assign(numOps, nullptr);
     audioSourceForOp_.assign(numOps, nullptr);
     midiSourceForOp_.assign(numOps, nullptr);
-    wasInaudible_.assign(numOps, 0);
     meterForOp_.assign(numOps, nullptr);
     midiTapForOp_.assign(numOps, nullptr);
     paramWindowForOp_.assign(numOps, ParamTable::DeviceWindow{});
@@ -1433,14 +1431,6 @@ void PlanExecutor::renderOp(OpId id, const OpValue& published, const BlockInfo& 
 
         case OpKind::Device: {
             auto audio = audioOut(id, 0, numSamples);
-
-            // The block a track goes inaudible is the one its devices are owed
-            // a panic on, which is the edge the fork reads as
-            // TrackMuteState::wasJustMuted (#2418). Taken before anything below
-            // can leave early, so the edge is not missed on a silent block.
-            const bool justInaudible = value.trackInaudible && wasInaudible_[i] == 0;
-            wasInaudible_[i] = value.trackInaudible ? 1 : 0;
-
             const auto producesMidi =
                 op.outputs.size() > 1 && op.outputs[1].kind == SignalKind::Midi;
             juce::MidiBuffer* deviceMidiOut = nullptr;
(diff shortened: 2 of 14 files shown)


magda/engine/exec/PlanExecutor.cpp
```diff
@@ -1470,6 +1505,13 @@ void PlanExecutor::renderOp(OpId id, const OpValue& published, const BlockInfo&
             const auto window = paramWindowForOp_[static_cast<std::size_t>(i)];
             DeviceBlock deviceBlock{.audio = audio.getSubsetChannelBlock(0, blockWidth),
                                     .midiIn = &midiIn(op.inputs[1]),
+                                    // What reached the port, plus the two
+                                    // discontinuities the block itself carries:
+                                    // a playhead jump and a track just muted.
+                                    // Both hosts raise them per device rather
+                                    // than passing them along a chain (#2418).
+                                    .midiInAllNotesOff = !block.continuous || justInaudible ||
```
lucaromagnoli: [P2] Restore sustained clip notes after a mute panic

The new justInaudible panic makes ArpeggiatorPlugin discard its held clip notes via retainLiveHeldNotes(), but ClipMidiSource continues rendering a continuous block and does not chase those sustained notes again. Unmuting also leaves the block continuous. Consequently, muting and unmuting during a sustained clip chord leaves the arp silent until fresh note-ons arrive, even though the clip still considers the chord active. Continuing to process the track while muted does not reconstruct notes whose note-ons preceded the mute.

Please coordinate this panic with restoring the still-active source notes when appropriate, and add an integration test with ClipMidiSource feeding an arpeggiator across mute/unmute. The current probe test checks the flag edge but cannot catch lost held-note state.

magda/engine/clip/ClipMidiSource.cpp
```diff
@@ -501,6 +501,12 @@ bool ClipMidiSource::renderSession(juce::MidiBuffer& out, const BlockInfo& block
                 playLane(out, material, slot.midi, from, to);
             });
 
+            // A slot that begins sounding here started material the transport
+            // never moved for, which is the same discontinuity a device is
+            // owed as the arrangement's hold.gained below (#2418).
+            if (!status.soundingAtStart && status.playingAtEnd())
+                raisedAllNotesOff_ = true;
```
lucaromagnoli: [P2] Reassert continuing slots when raising a shared launch panic

This flag applies to the whole merged MIDI stream, but only the newly launched slot receives a discontinuous material block and chases its notes. If slot A is sustaining a chord and slot B starts on the same track, A remains continuous and emits no replacement note-ons. The downstream arpeggiator sees the panic, discards all clip-held notes, and reconstructs only the events in this block, so A disappears from its held chord despite still playing.

Please reassert the continuing material when raising this shared panic, or otherwise preserve its held-note state. Add a source-to-arpeggiator test that launches a second slot while the first sustains; the existing launch test only checks the boolean.

magda/engine/exec/PlanExecutor.cpp
lucaromagnoli: Right, and thank you — this was a real bug. Fixed in 76bb09a2, but in the other direction: the mute panic is gone rather than paired with a re-assertion.

The rule I should have applied from the start: **a panic is only safe where the host re-asserts what should still be sounding on the same buffer.** That is what `ArpeggiatorPlugin` states at its `inputPanic` branch ("the host's notes go and come back on the same buffer"), and what `ClipMidiSource::playLane` does — but only under `if (!block.continuous)`. Checking the four sites I added:

| site | chases? |
|---|---|
| `!block.continuous` (locate, wrap, start) | yes — `playLane` → `chaseClip` |
| arrangement `hold.gained` | yes — sets `lane.continuous = false` |
| session slot launch | no |
| track just inaudible | no |

So both of the ones you flagged, and only those.

I did not add the re-assertion you suggest, because for mute it would be worse than the bug. Re-asserting means re-emitting note-ons, which restrikes every synth on the track — for a mute, which in this engine is a gain with the track still rendering. And it would not match the fork either, which does not chase on unmute.

The deeper point is that the engine has no mute discontinuity to signal. The fork raises one because `shouldTrackContentsBeProcessed()` goes false and the track's MIDI stops; MAGDA deliberately keeps processing muted tracks so their meters stay live, so nothing is withheld from the device and there is nothing for it to recover from. Carrying the flag without carrying that behaviour is exactly the trade you identified: a silent divergence swapped for an audible bug. `OpValue::trackInaudible` and the executor's edge state are removed.

On the test: rather than the integration test you asked for, the decision is now pinned by its inverse — "muting a track raises no panic on its devices" renders a compiled plan across unmuted → muted → muted → unmuted through `resolvePlanValues` and asserts the device never sees a panic. Verified negatively: forcing `midiInAllNotesOff = true` fails it on 4 of its assertions. An arp-across-mute integration test would now assert that nothing happens, which the probe covers more directly.

If mute should become a genuine discontinuity, the fix is to stop the muted track's contents the way the fork does — a behaviour change well outside #2418, and one that would cost the live meters.

magda/engine/clip/ClipMidiSource.cpp
lucaromagnoli: Agreed, and confirmed reachable: `forEachSlot` (`SessionPlayback.hpp:216`) walks every slot of the track that has a handle, and nothing in the engine stops A when B launches — `SessionLauncher.hpp` says as much ("a slot's state has to survive another slot on the same track playing in between"), and the hand-over test drives it as two separate requests (`rig.handle.stop(at); rig.second.play(at);`). So A sustaining while B starts is a state the source has to be correct in, not one the launcher rules out.

Removed the session raise in 76bb09a2 rather than adding the re-assertion. Two reasons:

1. **A launch needs no flag.** It is self-describing: the slot it replaces owes note-offs for everything it started (`endSlot`, the invariant in `ClipMidiSource.hpp`), so a device downstream hears the release in the stream and releases normally. The flag adds nothing in the ordinary case.
2. **Re-asserting would be a retrigger.** Slot A's chord never stopped sounding; re-emitting its note-ons to satisfy the panic would restrike it on every synth downstream. That is audibly worse than the state loss.

The launched slot also chases nothing of its own — it starts at its own origin — so there was never a chase for this panic to ride on, which is the same test the mute case failed.

What remains on the source is the arrangement hand-back (`hold.gained`), which is the one launch-side discontinuity that does chase: it sets `lane.continuous = false`, so `playLane` re-asserts the controllers and the notes the instant is inside. A device that drops what it holds there gets it straight back on the same buffer.

Test updated to the inverse — "A launched slot raises no panic" still requires the launch actually happened (`noteOns().size() == 1`) so it cannot pass vacuously — and the hand-back test keeps the positive case. `EngineMidiSource::raisedAllNotesOff` now documents the rule, so the next source to raise one has the constraint in front of it.

PR #2430, opened 2026-09-06
Strum every note of a chord in Rings, not just the last

### LINKED ISSUE #2364: Strum a whole chord in Rings instead of its last note
## Problem

`MutableRingsPlugin::process` renders up to each note-on and then calls `noteOn` (magda/daw/audio/plugins/mutable/MutableRingsPlugin.cpp:329-336), which overwrites the single `currentNote_` and sets `strumPending_` (:180-183). The resonator consumes the strum once per internal 24-sample block (:212-224). Two note-ons at the same timestamp, which is every chord in a clip, leave one strum of the last note; note-ons closer together than one internal block collapse the same way.

## Scenario

A three-note chord in a clip into Rings with Polyphony 4 sounds one string.

## Direction

Queue note-ons and strum one per internal Process pass, or advance the resonator one internal block between same-timestamp note-ons so each gets its strum. Rings rotates voices per strum, so this is the whole of its chord support.

Pre-existing, carried through the port unchanged.

lucaromagnoli:
Closes #2364.

## The bug

Rings has no note-off and no per-voice gate: a note-on retunes the resonator and fires a strum, and the part rotates voices so earlier notes keep ringing. That rotation is the whole of its chord support.

`noteOn` wrote its note number straight into the single `currentNote_` the resonator reads, but the strum is only consumed once per internal 24-sample block. A chord in a clip arrives as note-ons at one timestamp, so the last of them overwrote the rest before any block ran. Notes closer together than one internal block collapsed the same way.

## The fix

Note-ons queue, and each internal block takes one off the front and strums it. A chord spreads over as many blocks as it has notes, which at the internal rate is under a millisecond, and the ordering is preserved. The queue is a fixed 16 entries, several times the four voices, and survives a host block boundary so a chord landing at the end of one still finishes.

`strumPending_` goes away with it: the strum is now decided at the block boundary from the queue rather than latched by whoever called `noteOn` last.

## Tests

`tests/devices/test_mutable_rings_chord.cpp`, three cases: a three-note chord at one timestamp, three notes spread four samples apart so they all land inside one internal block, and a six-note chord against four voices. Each asserts the render is not identical to its last note alone and carries more energy than it.

I checked they bite. Reverting the fix fails two of the three.

Worth recording, because it is worse than the issue says: an exactly coincident chord did not render one string, it rendered **silence** (energy 0.0). Notes a few samples apart did sound, which is presumably why it was reported as one string rather than none.

Full Catch2 suite green: 3615 passed, 13 skipped.

### DIFF
--- a/magda/daw/audio/plugins/mutable/MutableRingsPlugin.cpp
+++ b/magda/daw/audio/plugins/mutable/MutableRingsPlugin.cpp
@@ -157,7 +157,7 @@ struct MutableRingsPlugin::Impl {
         chunkPos_ = kBlock;
         std::memset(xL_, 0, sizeof(xL_));
         std::memset(xR_, 0, sizeof(xR_));
-        strumPending_ = false;
+        pendingCount_ = 0;
         currentNote_ = 60;
     }
 
@@ -178,9 +178,13 @@ struct MutableRingsPlugin::Impl {
 
     // Each note-on retunes and fires a one-block strum; Rings rotates voices so
     // earlier notes keep ringing (polyphony). No note-off: decay is the Damping.
+    //
+    // Queued rather than applied here, because the strum is consumed once per
+    // internal block: a chord arrives as note-ons at one timestamp, and writing
+    // the note straight in left only the last of them a string (#2364).
     void noteOn(int n) {
-        currentNote_ = n;
-        strumPending_ = true;
+        if (pendingCount_ < kMaxPendingNotes)
+            pendingNotes_[static_cast<size_t>(pendingCount_++)] = n;
     }
 
     void generate(float* outL, float* outR, int n) {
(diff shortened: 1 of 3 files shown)


magda/daw/audio/plugins/mutable/MutableRingsPlugin.cpp
```diff
@@ -212,17 +216,27 @@ struct MutableRingsPlugin::Impl {
...
         if (chunkPos_ >= kBlock) {
+            // One queued note per block, which is the rate Rings strums at.
+            bool strum = false;
+            if (pendingCount_ > 0) {
+                currentNote_ = pendingNotes_[0];
+                for (int i = 1; i < pendingCount_; ++i)
+                    pendingNotes_[static_cast<size_t>(i - 1)] =
+                        pendingNotes_[static_cast<size_t>(i)];
+                --pendingCount_;
+                strum = true;
```
lucaromagnoli, line 227: **[P2] Preserve earlier voices' pitches when draining the chord queue**

Calling `Part::Process()` with `strum=true` on consecutive internal blocks does not preserve the pitches just struck. In `rings/dsp/part.cc`, each strum first assigns `note_filter_.stable_note()` to the previously active voice before rotating to the new voice. That delayed stable-note buffer advances only in the non-strum branch of `NoteFilter::Process()`, so it cannot catch up while this queue drains.

I verified this with the actual upstream `NoteFilter`, initialized with the same constants as `Part::Init()`: consecutive strums for 60, 64, and 67 all return a stable note of 0 on a fresh instance. Consequently, the C and E voices are excited but then retuned to MIDI note 0 when the next queued note is processed; on an already-running instance the value can instead be an older pitch. The energy/non-identical-output tests pass despite losing the intended chord pitches.

Please preserve the previous MIDI pitch when rotating voices, or otherwise accommodate the upstream stable-note delay, and add pitch-sensitive regression coverage for the simultaneous chord.

— Codex

lucaromagnoli: Confirmed and fixed in 9860b2811, with the fork change in Conceptual-Machines/eurorack#1.

You were right on both counts, and the second one is the important part: my tests measured level, which cannot see this. One strum through a stale note filter is *louder* than three clean ones, it is just rumble instead of the chord, so every energy assertion passed while the pitches were being thrown away.

Measured at the fundamentals for a three-note chord at one timestamp, four voices:

| | note 60 | note 64 | 8 Hz (MIDI note 0) |
| --- | --- | --- | --- |
| before | 0.0014 | 0.0011 | 0.0480 |
| after | 0.0316 | 0.0428 | 0.0003 |

Two chords differing only in their top note rendered bit for bit identical before, which is the sharpest way to say it.

I went with filling the delay line on the sharp-edge branch rather than spacing the strums out. The line is eight blocks deep at `kSampleRate / kMaxBlockSize`, so waiting for it would put 4 ms between the notes of a chord, and it would still be reading an implementation detail rather than the pitch we actually sent.

Two things I got wrong on the way, in case they matter for the next review. My rig only read channel 0, and Rings splits partials between main and aux so either can sit at exact silence: that made single notes look silent at polyphony above one, which sent me chasing a bug that does not exist. It sums both channels now. And I had earlier concluded delta solo was irrelevant to the strum PR on the same kind of level-only evidence.

Tests now assert pitch. Reverting either fix fails the same two cases. Full Catch2 3616 passed, 13 skipped; JUCE suite green.

PR #2434, opened 2026-09-06
Emit an arpeggiator step at its swung position, not the raw one

### LINKED ISSUE #2362: Emit arpeggiator steps whose swung or quantized beat lands outside the block they were walked in

lucaromagnoli:
Closes #2362.

## The bug

The step walk in `ArpeggiatorPlugin::process` iterated the raw step grid, applied swing and quantize to get `swungBeat`, and emitted only when `swungBeat` fell inside the block. It advanced `currentStep_` either way, and the next block's catch-up loop tested the raw beat again, so a step whose warped position landed past the block end, or was pulled before its start, was consumed without ever playing.

At 1/16 and 120 bpm, swing at 50% shifts the odd steps by 0.0625 beats, three 512-sample blocks at 48 kHz. Nearly every odd step was dropped, so swing silenced half the pattern instead of delaying it. Quantize did the same in the other direction.

## The fix

`warpedStepBeat` folds swing and quantize into the step position, and the walk keys on that instead of the raw grid. Both warps are monotonic in the step index, so the walk stays ordered and a step whose position is past the block end simply stays pending until the block that holds it. The emit no longer needs an in-block guard: the catch-up leaves `currentStep_` on the first step at or after the block start.

The catch-up division starts a cycle early, because quantize can pull a step back across a cycle boundary. That keeps it bounded work rather than one pass per skipped step.

## Tests

`tests/devices/test_arpeggiator_swing.cpp`, two cases at 512-sample blocks: swing at 50% pushing the odd 1/16 steps three blocks on, and a three-note cycle where the quantize grid pulls step 1 back and step 2 forward across a block boundary.

Reverting the fix fails both, at 2 note-ons of 4 and 1 of 3.

### DIFF
--- a/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
+++ b/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
@@ -645,14 +645,17 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
     };
 
     // Where the step is actually played: swing on odd steps, then quantize
-    // pulling that toward a regular grid. Both are monotonic in the step
-    // index, so the walk below can key on this instead of the raw grid and a
+    // pulling that toward a regular grid. Neither can carry a step past its
+    // neighbour, so the walk below keys on this instead of the raw grid and a
     // step warped past the block end stays pending rather than being consumed
     // unplayed (#2362).
     auto warpedStepBeat = [&](int step) -> double {
         double beat = computeStepBeat(step);
+        // Half the gap to the next step, not half the nominal rate: Time Bend
+        // can compress two steps onto one beat, and a fixed offset would then
+        // swing the odd one past the even one that follows it.
         if (step % 2 == 1 && swingVal > 0.0f)
-            beat += static_cast<double>(swingVal) * stepBeats * 0.5;
+            beat += static_cast<double>(swingVal) * (computeStepBeat(step + 1) - beat) * 0.5;
 
         if (quantizeAmount > 0.0f && quantizeSubVal > 0) {
             double gridSpacing = cycleBeats / static_cast<double>(quantizeSubVal);
--- a/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
+++ b/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
@@ -644,6 +644,24 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
         return cycleStart + stepInCycle * stepBeats;
     };
 
+    // Where the step is actually played: swing on odd steps, then quantize
+    // pulling that toward a regular grid. Both are monotonic in the step
+    // index, so the walk below can key on this instead of the raw grid and a
+    // step warped past the block end stays pending rather than being consumed
+    // unplayed (#2362).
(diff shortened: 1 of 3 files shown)


magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -644,6 +644,24 @@ void ArpeggiatorPlugin::process(DeviceProcessContext& context) {
...
 
+    // Where the step is actually played: swing on odd steps, then quantize
+    // pulling that toward a regular grid. Both are monotonic in the step
+    // index, so the walk below can key on this instead of the raw grid and a
+    // step warped past the block end stays pending rather than being consumed
+    // unplayed (#2362).
+    auto warpedStepBeat = [&](int step) -> double {
+        double beat = computeStepBeat(step);
+        if (step % 2 == 1 && swingVal > 0.0f)
+            beat += static_cast<double>(swingVal) * stepBeats * 0.5;
```
lucaromagnoli: **[P2] Preserve step ordering when combining Time Bend and swing**

The new walk assumes warped step times remain ordered, but adding a fixed swing offset can exceed the gap between ramp-compressed steps. With four notes, 1/16 rate, Timing Depth -1, hard angle enabled, and 100% swing, steps 1 and 2 land at 0.125 and 0 beats respectively. Waiting for step 1 holds up step 2, which is then emitted before the current block's start because the loop no longer checks that lower bound. A regression probe using 512-sample blocks at 48 kHz reproduced note-on and note-off timestamps of -0.053333 seconds in block 5.

Please preserve chronological scheduling across combined ramp/swing settings and add a regression case covering this combination; the two new tests only cover the unbent grid.

— Codex

lucaromagnoli: Confirmed and fixed in 7d8ea7752. I reproduced the same -0.053333 s in block 5 before changing anything.

Swing now displaces a step by half the interval to the next step rather than half the nominal rate, so it can never reach its neighbour and the warped positions stay ordered by construction. Where Time Bend leaves no gap, swing is zero. On an unbent grid that interval is `stepBeats`, so the plain-grid behaviour is unchanged.

Third case added: 100 blocks at that setting, asserting every timestamp lands inside its own block and that note-ons are non-decreasing in absolute time.

PR #2435, opened 2026-09-06
Read the arpeggiator's input where the host put it

### LINKED ISSUE #2415: Process arpeggiator input at event time rather than block start

lucaromagnoli:
Fixes #2415.

## The bug

`ArpeggiatorPlugin::process()` consumed the whole input buffer before generating anything, so every held-note change applied from the top of the block. Two consequences, both in the issue:

- A latch replacement or a CC120/123 nine tenths of the way through a block released the note it replaced at timestamp 0 — up to a full block early.
- The steps that had already sounded earlier in that block were built from the chord that arrived *after* them.

## The fix

`process()` cuts the block where the input changes what is held, and runs the beat walk over each stretch between those events. A note is closed by the event that closed it, at that event's own time; a stretch plays the chord held while it lasts; and the arp anchors its grid where the chord arrived rather than wherever the block containing it began.

Ordering at one instant is: the non-note traffic the arp forwards, then the note-off the input caused, then the stretch's own steps — so a controller moved before a note-off still reaches the instrument first, and an old voice is always closed before the note that replaces it sounds.

## Two things that follow from the cut

Both are acceptance criteria on the issue:

- **A step is placed by the beat it is played on, not by its grid position.** One whose swing or quantize offset carries it past the end of a stretch is left for the stretch that holds it, rather than consumed and dropped. Before this, a swung step landing past a block boundary was lost outright, and *which* steps were lost depended on the block size — so the same input played different notes at different buffer sizes.
- **The host's sub-block offset is added wherever an input timestamp is read**, the non-note forwarder included, rather than only in the arp's own scan.

`MidiMagdaDevice::sendNoteOff` now takes a time within the block (seconds from the block start, like every other timestamp the SDK carries).

## Tests

`tests/devices/test_arpeggiator_event_time.cpp` — late latch replacement, late panic (CC120 and CC123), same-pitch retrigger, several replacements in one block, a step playing what was held when it sounded, and one phrase rendered over small (64), large (2048) and irregular partitions agreeing event for event with no stuck or duplicate notes.

`tests/devices/test_arpeggiator_event_time_juce.cpp` — the same panic driven through **both host adapters**, since the two stamp their MIDI differently (the fork carries seconds on its container, the engine's ports count samples).

Verified negatively: with the fix reverted all seven fail, both legs putting the note-off at sample 0 rather than at the panic's 57. The 27 existing arpeggiator cases pass unchanged, as do `make test` and `make test-juce` in full.

## Follow-up

The SDK's MIDI timestamps stay in seconds here. Switching them to integer sample offsets is worth doing on its own — it deletes six device-side `roundToInt(ts * sampleRate)` conversions and `midiTimeOffsetSeconds`, which both Tracktion nodes hard-code to 0 — and is a separate, mechanical change.

### DIFF
--- a/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
+++ b/magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
@@ -45,8 +45,17 @@ bool isLiveSource(const DeviceProcessContext& context, std::uint32_t sourceId) {
 /// device rather than the notes.
 class NonNoteForwarder {
   public:
-    NonNoteForwarder(const DeviceMidiInput& in, double offsetSeconds)
-        : in_(in), offsetSeconds_(offsetSeconds) {}
+    NonNoteForwarder(const DeviceMidiInput& in, double offsetSeconds, double blockDurationSecs)
+        : in_(in), offsetSeconds_(offsetSeconds), blockDurationSecs_(blockDurationSecs) {}
+
+    /// Where in the block the host put event @p index: its timestamp with the
+    /// host's sub-block offset added, inside the block it belongs to. The one
+    /// answer the arp reads an input time by, so a forwarded message and the
+    /// notes generated around it share a timeline (#2415).
+    double timeOf(int index) const {
+        return juce::jlimit(0.0, blockDurationSecs_,
+                            in_.message(index).getTimeStamp() + offsetSeconds_);
+    }
 
     /// Everything up to and including @p timeInBlock. A message coincident
     /// with a generated note goes out first: a controller moved on the beat
@@ -54,7 +63,8 @@ class NonNoteForwarder {
     void flushUpTo(DeviceMidiOutput& midi, double timeInBlock) {
         for (const int count = in_.size(); next_ < count; ++next_) {
             const auto& message = in_.message(next_);
-            if (message.getTimeStamp() + offsetSeconds_ > timeInBlock)
+            const double time = timeOf(next_);
+            if (time > timeInBlock)
                 return;
             // Channel messages only: getChannel() is 0 for SysEx and for
             // everything the transport carries.
@@ -63,6 +73,7 @@ class NonNoteForwarder {
 
             auto forwarded = message;  // short message: inline storage, no allocation
             forwarded.setChannel(1);
+            forwarded.setTimeStamp(time);
             midi.addEvent({std::move(forwarded), in_.sourceId(next_)});
         }
     }
@@ -73,9 +84,8 @@ class NonNoteForwarder {
 
   private:
     const DeviceMidiInput& in_;
-    /// The host's sub-block offset, added to every input timestamp the way the
-    /// arp adds it when placing the same events on its own timeline (#2415).
     double offsetSeconds_ = 0.0;
+    double blockDurationSecs_ = 0.0;
     int next_ = 0;
 };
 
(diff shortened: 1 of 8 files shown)


magda/daw/audio/plugins/ArpeggiatorPlugin.cpp
```diff
@@ -45,15 +45,16 @@ bool isLiveSource(const DeviceProcessContext& context, std::uint32_t sourceId) {
...
+        : in_(in), offsetSeconds_(offsetSeconds) {}
 
     /// Everything up to and including @p timeInBlock. A message coincident
     /// with a generated note goes out first: a controller moved on the beat
     /// belongs to the note that starts on it.
     void flushUpTo(DeviceMidiOutput& midi, double timeInBlock) {
         for (const int count = in_.size(); next_ < count; ++next_) {
             const auto& message = in_.message(next_);
-            if (message.getTimeStamp() > timeInBlock)
+            if (message.getTimeStamp() + offsetSeconds_ > timeInBlock)
```
lucaromagnoli: [P2] Apply the host offset to forwarded timestamps too

With a nonzero `midiTimeOffsetSeconds`, this comparison schedules controllers using the adjusted time, but the forwarded copy retains its original timestamp. Generated notes use the adjusted time, so a CC123 at 5 ms with a 10 ms offset is forwarded at 5 ms while its generated note-off lands at 15 ms. Current hosts use zero offsets, so this is conditional on a host supplying a nonzero offset. Could we set the forwarded timestamp consistently with the normalized event time and add a nonzero-offset regression covering both the controller and generated note-off?

lucaromagnoli: Fixed in 74a4e6a3.

`NonNoteForwarder` now owns the one reading of an input time — timestamp, plus the host offset, clamped into the block — and both the arp's own scan and the copies it forwards go through it, so the stamp on a forwarded message is the time it was scheduled by.

Regression added: `Arpeggiator reads a host's sub-block offset once, for everything it emits` (test_arpeggiator_event_time.cpp) drives a CC123 at 5 ms with a 10 ms offset and asserts the forwarded controller and the generated note-off both land at 15 ms. On the old forwarder it fails with the controller at 0.005 against the note-off's 0.015, which is the case you described.

While there: the arpeggiator tests had two `at()` helpers with opposite argument orders, so the shared one is now the only one.

Issue #2441, opened 2026-09-06
The null-diff corpus renders arrangement only, so no case can null a launched slot


lucaromagnoli:
Part of #1882, and what stands between #2306 and the first of the three kinds of case it asks for.

## The gap

#2306 says most of the launcher's surface is an ordinary null: "once two handles are playing the same clip from the same beat, the audio is an ordinary clip render and nulls like any other". No case can do that today, and the reason is not the model.

`ClipInfo` already carries `sceneIndex` and the session launch properties, so a `Case` can describe a clip that lives in a slot. What the corpus cannot do is launch it. Both legs drop session clips on purpose:

- The native leg skips any clip whose view is not `Arrangement` (`tests/NullDiffNativeLeg.cpp:469`).
- The fork does the same at the same point, in `ClipSynchronizer::syncArrangementClipToEngine`, which refuses a session clip and leaves it to the launcher.

That rule is right for what the corpus renders now. Its comment says so: every clip in the corpus is an arrangement clip, and without the skip the demo project reports sixty-four diagnostics for merely having a session view, every one of which would make it unmeasurable. So the corpus renders an offline arrangement, on both sides, and neither leg has a step that launches anything.

## What it would take

A case that declares a launch — a slot, a beat, and a quantization — and a runner step that drives it on both legs before the render, so the two are launched at the same sample rather than at whatever moment each engine happened to start. Once both handles are running the same clip from the same beat, the rest is the comparison the corpus already makes.

The two legs will not have been launched at the same sample unless something makes them, which is the whole of the work here.

## Why it is worth doing

- It is the only way #2306's post-launch cases get written, and #2306 calls them most of the launcher's surface.
- #2082's parity envelope is measured over the corpus, so until this exists the envelope covers no session playback at all.
- The launcher is, in #1894's words, "the most from-scratch subsystem and the least null-testable". The part of it that IS null-testable is currently untested for the reason above rather than on purpose.

## Not in scope

Launch timing and sequence behaviour, which #2306 names as its other two kinds and which are asserted rather than nulled. Those do not need the fork and are partly landed already.

lucaromagnoli:
Fixed by #2445.

PR #2442, opened 2026-09-06
Pin what the launcher does across tracks

### LINKED ISSUE #2306: Session launcher slice 7: behaviour cases for the launcher
Slice 7 of #1894, and the last of it. Depends on every slice above, and shares its corpus with #2081.

## Why this is not simply more null-diff cases

#1894 says the launcher is "the most from-scratch subsystem and the least null-testable", and that is a statement about what a null can prove here. A session render is not a function of the timeline: it is a function of when things were launched, and the fork and the native engine will not have been launched at the same sample unless something made them.

So the cases split into three kinds, and only the first is a null:

- **Post-launch audio.** Once two handles are playing the same clip from the same beat, the audio is an ordinary clip render and nulls like any other. This is most of the surface and it is worth having.
- **Launch timing**, which is asserted rather than nulled. A clip quantized to one bar starts on the bar, at the sample, under both engines and at every block size the #2078 gate uses. That is the assertion that catches a handle splitting on the wrong clock, and it does not need the fork to answer it.
- **Sequence behaviour.** Follow action chains, scene launches, a stop landing while a follow action is queued. These are assertions about a sequence of played ranges, and they need their own harness because there is no audio in which a wrong-but-plausible sequence looks different from a right one.

## What the cases have to cover

- A scene launch: N tracks launched as one event, all starting on the same beat, which is what `playSynced` is for.
- An empty slot in a scene stopping its track, which is `SessionLaunchService`'s rule and is where the two surfaces used to drift.
- A quantized stop arriving while a launch is queued on the same handle.
- Session and arranger on different tracks at the same time, playing together.
- A structural edit during playback, which is #2077's territory pointed at the launcher.
- Loop-record and slot recording are #1895's and are named here only so the boundary is on the record.

## The thing to write down

Where the two engines are deliberately different, the difference is declared with its mechanism rather than tolerated, the same way #2082 requires for the performance envelope and #2075 requires for render divergences. A launcher that "sounds right" and is not pinned is a launcher whose next regression is found by a user.

lucaromagnoli:
Part of #2306. Two of its three kinds of case; the third is #2441.

## What was missing

Every launcher suite builds its slots on a single track, which is the one thing a scene is not. `test_launch_handle.cpp` is one handle's state machine, `test_launch_requests.cpp` is the lane a request travels down, and both put every slot on `kTrack`. A scene that started seven of its eight tracks on one beat and the eighth on the next sounds like a scene, so nothing would have caught it.

Follow actions needed nothing: #2304 landed `test_follow_actions.cpp` with twenty cases over them.

## Across tracks

`tests/engine/test_launch_behaviour.cpp`:

- A scene starts every track it names on the beat it was asked for, in phase, in one gesture.
- A track already playing joins the leader's run rather than starting its own.
- The stop a scene carries for a track it holds no clip for lands on that same beat, not a block either side of it: the whole gesture drains before any handle advances.
- The scene case again at 64, 96, 512 and 4096, the sizes the invariance gate renders at. Beat one is sample 2048 whatever the host asked for.

## The two events that meet on one handle

Also uncovered, and not a multi-track question: every existing stop case stops a handle that is already playing. A handle holds one pending request, so a stop arriving over a queued launch is what was asked for and the slot never starts — whether the stop is quantized before that launch's own beat or after it.

## Two tracks rendering together

`test_session_playback.cpp` gains the other half of its own section claim. Its `SwitchRig` is one track handing itself between its two sections; `PairRig` is two tracks rendering together, one launched and one never asked for anything, which is what a project mid-performance normally is. A session that has taken one track has not taken the others, a handle advancing does not reach a track it does not belong to, and handing one track back leaves the other alone.

## What is left of #2306, and where it went

**Post-launch audio that nulls is #2441.** Not a scoping choice: both legs drop session clips on purpose (`NullDiffNativeLeg.cpp:469`, and `ClipSynchronizer` on the fork side), so the corpus renders an offline arrangement and neither leg has a step that launches anything. The model is not the blocker — `ClipInfo` already carries `sceneIndex` — the runner is.

**The empty-slot rule stays untested at the app layer.** `SessionLaunchService::launchScene` sits behind the `ClipManager` and `TrackManager` singletons; the launcher's half of the rule is asserted here, and the drift the issue worries about is already prevented by the rule living in one function both surfaces call. Nothing filed for it, and #2306 says so.

## Tests

3655 passed, 13 skipped.

### DIFF
--- a/tests/engine/test_launch_behaviour.cpp
+++ b/tests/engine/test_launch_behaviour.cpp
@@ -1,0 +1,292 @@
+#include <catch2/catch_approx.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <vector>
+
+#include "launch/SessionLauncher.hpp"
+
+/**
+ * @file test_launch_behaviour.cpp
+ * @brief A scene across tracks, and the events that meet on one handle (#2306).
+ *
+ * What the launcher does when more than one track is involved. test_launch_-
+ * handle.cpp is one handle's own state machine and test_launch_requests.cpp is
+ * the lane a request travels down; both build their slots on a single track,
+ * which is the one thing a scene is not.
+ *
+ * Assertions about which slot is playing on which beat rather than about audio.
+ * A scene that started seven of its eight tracks on one beat and the eighth on
+ * the next sounds like a scene, which is why it needs pinning here.
+ */
+
+using Catch::Approx;
+using namespace magda;
+using namespace magda::engine;
+
+namespace {
+
+constexpr int kBlockSize = 512;
+constexpr double kSampleRate = 48000.0;
+
+/// A quarter beat a block, so a launch and its first block are easy to tell
+/// apart from the one after it. Carried per sample too, since the sizes at the
+/// bottom of this file cut the same beats differently.
+constexpr double kBeatsPerBlock = 0.25;
+constexpr double kBeatsPerSample = kBeatsPerBlock / kBlockSize;
+
+BlockInfo blockOf(std::int64_t startSample, int numSamples) {
+    const auto endSample = startSample + numSamples;
+
+    BlockInfo block;
+    block.numSamples = numSamples;
+    block.sampleRate = kSampleRate;
+    block.monotonicSamples = {SamplePosition{startSample}, SamplePosition{endSample}};
+    block.playing = true;
+    block.continuous = startSample != 0;
+    block.beats = {static_cast<double>(startSample) * kBeatsPerSample,
+                   static_cast<double>(endSample) * kBeatsPerSample};
+    block.monotonicBeats = block.beats;
+    block.seconds = {static_cast<double>(startSample) / kSampleRate,
+                     static_cast<double>(endSample) / kSampleRate};
+    block.monotonicSeconds = block.seconds;
+    return block;
+}
+
+BlockInfo blockAt(int index) {
+    return blockOf(static_cast<std::int64_t>(index) * kBlockSize, kBlockSize);
+}
+
+/// The block covering @p beat, so a case names the beat it is rolling to rather
+/// than counting callbacks.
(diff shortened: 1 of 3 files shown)


tests/engine/test_launch_behaviour.cpp
```diff
@@ -0,0 +1,292 @@
...
+    REQUIRE(rig.at(0, 0).queuedState() == LaunchHandle::QueueState::playQueued);
+
+    {
+        LaunchRequestQueue::Gesture gesture(rig.requests);
+        gesture.stop(rig.key(0, 0), 2.0);
+    }
+    rig.rollTo(3.0, 1);
+
+    CHECK_FALSE(rig.playing(0, 0));
+    CHECK_FALSE(rig.at(0, 0).playedSampleRange().has_value());
```
lucaromagnoli, line 263: **[P2] Verify the cancelled launch never starts**

This advances past both scheduled events and then checks only that playback is stopped and `playedSampleRange()` is empty. Both checks also pass if the slot incorrectly plays from beat 1 to beat 2: `LaunchHandle::endRun()` sets the state to stopped and clears `run_`, so neither assertion retains evidence of that unwanted playback. The test therefore misses exactly the launch-then-stop regression described above it.

Please assert that `lastPlayedRange()` remains empty, or check playback between the two beats, so the case proves the queued launch was cancelled rather than eventually stopped.

— Codex

lucaromagnoli: Confirmed and fixed in d081e6c94. I checked it before changing anything: a slot that genuinely plays beat one to beat two and is stopped there reports `playState() == stopped` and an empty `playedSampleRange()`, so both of my assertions passed on exactly the shape the case was written to catch.

`lastPlayedRange()` is the discriminator, and it is now asserted empty on both stop cases: a stop clears the run it ended, so a slot whose launch was cancelled is the one with no last run at all.

PR #2443, opened 2026-09-06
Render with flush-to-zero, and clear what a plugin leaves that is not a number

### LINKED ISSUE #2240: The native engine has no denormal handling: no flush-to-zero, no output sanitising

lucaromagnoli:
Closes #2240.

## The gap

The native engine had no denormal handling anywhere: no flush-to-zero around a render, and nothing looking at what an op wrote. A device feeding its own output back through a filter decays into the denormal range and stays there, where on Intel every multiply is a microcode trap and a reverb tail is where a dropout comes from.

## Flush-to-zero

A per-thread CPU mode, so it is set on every thread that renders rather than once at startup:

- `PlanExecutor::process` and `ParallelPlanExecutor::process`, which between them cover the audio callback and the offline render.
- `RenderThreadPool::Worker::run`, once for the worker's whole life, since a worker does nothing but render.

Scoped rather than left on: the host's thread is not the engine's to hand back in a mode it did not ask for, and an offline render runs on a thread that goes on to do other things. There is a case for that too.

## The sanitising half, which is not the fork's rule

#2240 asked for this to be a decision rather than a transcription, so: a plugin's output is checked for values that are not numbers, and those are cleared. Before the dry signal is added back, so a plugin that produced nothing usable leaves the dry rather than poisoning it.

The fork instead clamps to a magnitude of 100, and only looks at all when the CPU raised its denormal flag during the call, on Intel. Both halves of that are wrong for what this guards against:

- **The gate is for denormals.** It says nothing about a NaN, so the fork's protection misses the case entirely on any plugin that does not also denormalise, and misses it on Apple silicon regardless.
- **The clamp treats it as a level.** A NaN is not one: it survives every gain, poisons every sum it reaches, and the track stays broken after the plugin that made it has been bypassed. An infinity does the same on the first multiply by zero. 100 is a number nothing derives, and +40 dBFS protects no speaker.

That is a declared divergence from the incumbent, and it is written down where the code is.

## Tests

`tests/engine/test_render_denormals.cpp` asserts the mode from inside a device, which is where the arithmetic it exists for happens: on the calling thread, on a pool thread, and off again after the block. The denormal is multiplied through `volatile` so the compiler cannot fold it into a constant that never met the CPU flag.

Two cases in `test_engine_external_device.cpp` cover a plugin that writes a NaN and an infinity: nothing non-finite leaves the device, and with the wrapper pair half dry, the dry signal is what those samples come back as.

Reverting either half fails its own cases. Full suite: 3651 passed, 13 skipped.

### DIFF
--- a/tests/engine/test_render_denormals.cpp
+++ b/tests/engine/test_render_denormals.cpp
@@ -1,0 +1,180 @@
+#include <catch2/catch_test_macros.hpp>
+#include <limits>
+#include <memory>
+
+#include "NullDiffGain.hpp"
+#include "core/TrackInfo.hpp"
+#include "exec/ParallelPlanExecutor.hpp"
+#include "exec/PlanExecutor.hpp"
+#include "exec/RenderThreadPool.hpp"
+#include "plan/PlanCompiler.hpp"
+
+/**
+ * @file test_render_denormals.cpp
+ * @brief Flush-to-zero is on wherever a block is rendered (#2240).
+ *
+ * A device that feeds its own output back through a filter decays into the
+ * denormal range and stays there, where on Intel every multiply is a microcode
+ * trap and a reverb tail is where a dropout comes from. The mode is a per-thread
+ * CPU flag, so it has to be set on every thread that renders rather than once at
+ * startup: the audio callback's, the offline render's, and each of the pool's.
+ *
+ * Asserted from inside a device, because that is where the arithmetic the mode
+ * exists for actually happens.
+ */
+
+using namespace magda::engine;
+
+namespace {
+
+constexpr int kBlockSize = 512;
+
+/**
+ * @brief A denormal multiplied here and now, and what came back.
+ *
+ * Through volatile, so the compiler cannot fold it into a constant that never
+ * met the CPU's flush-to-zero bit: a folded answer is the same on every thread
+ * and would make this file pass without the mode being set anywhere.
+ */
+float multiplyDenormal() {
+    volatile float tiny = std::numeric_limits<float>::denorm_min();
+    volatile float one = 1.0f;
+    return tiny * one;
+}
+
+/// Records, from inside the render, what a denormal multiply came to.
+class DenormalProbe final : public EngineDevice {
+  public:
+    void process(DeviceBlock&) override {
+        product = multiplyDenormal();
+        rendered = true;
+    }
+
+    void setMidiInputBoundBytes(int) override {}
+    void setMidiOutputBoundBytes(int) override {}
+
+    bool forwardsMidiInput() const override {
+        return false;
+    }
+
+    float product = std::numeric_limits<float>::quiet_NaN();
+    bool rendered = false;
+};
+
+/// One track carrying one device, feeding the master.
+RenderPlan planWithDevice(magda::DeviceId device) {
+    magda::TrackInfo track;
+    track.id = 1;
+    track.type = magda::TrackType::Media;
+    track.name = "Track";
+    track.chain.fxChainElements.emplace_back(magda::nulldiff::gainDevice(device));
+
+    magda::TrackInfo master;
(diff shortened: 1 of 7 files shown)


tests/engine/test_render_denormals.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    // The workers set their own, since the flag belongs to the thread and the
+    // executor's caller is not the one running the op.
+    RenderThreadPool pool(2, false);
+    ParallelPlanExecutor executor(&pool);
+    prepared(executor, plan, bindings);
+
+    juce::AudioBuffer<float> output(2, kBlockSize);
+    output.clear();
+    executor.process(values, oneBlock(), output);
```
lucaromagnoli, line 229: **[P2] Ensure the worker test actually executes on a worker**

`RenderThreadPool::render()` calls `job.takeWork()` on the calling thread as well as waking the workers. With this single-device plan, the caller can therefore execute the probe itself, under `ParallelPlanExecutor::process()`'s own denormal guard. A temporary thread-identity assertion showed that this happened in 27 of 30 runs: the original assertions still passed, but no worker's floating-point mode was tested. Those runs cannot detect missing denormal handling in `Worker::run()`.

Please force worker participation, for example using synchronized independent devices as in the existing parallel-executor rendezvous test, and verify that the denormal probe actually ran on a worker.

— Codex

lucaromagnoli: Confirmed and fixed in 413c9f17f. `RenderThreadPool::render` calls `job.takeWork()` on the calling thread as well as notifying the workers, so with one device the caller usually took it and the case ran under `ParallelPlanExecutor::process`'s own guard. It named the worker and mostly did not use one.

It now uses two devices that stay inside `process()` until both are, which one thread cannot do, following the rendezvous shape in test_parallel_executor.cpp. Each records the thread it ran on and what a denormal came to there, so the case asserts that a worker rendered a device rather than assuming it.

Checked both directions: removing the scope from `Worker::run` fails it on every run of three, and it passed ten runs in a row with the scope in place.

Issue #2447, opened 2026-09-06
An offline render has no way to become a file


lucaromagnoli:
Part of #1882, and what the rest of #2248 waits on.

## What is missing

`renderOffline` writes into an `OfflineRenderSink`, and there is no sink in the product that writes a file. The two that exist are both in tests: `BufferSink` in `NullDiffNativeLeg.cpp` and `CollectingSink` in `test_offline_render.cpp`. Nothing the native engine renders can become an export, a bounce or a freeze.

The export dialog still goes through Tracktion (`TracktionOfflineRenderTask` in `OfflineRenderHelper.cpp`), and there is no CLI at all. So this is not a gap anybody hits today; it is a gap the cutover (#1897) hits on the day the fork is removed, because at that point nothing can write a rendered file.

## What it needs

- A sink that writes a file, going through `PcmQuantiser` (#2248) for the float-to-fixed-point step so nothing rounds unquantised float on its own. WAV and FLAC, 16, 24 and 32-bit float, which is what the export dialog already offers.
- A tail and trim contract. The Tracktion path renders pre-roll and trims it afterwards by re-reading the file (`MainWindowExport.cpp`); the native render already has `tailSeconds` and a latency-trimmed sink, so the second pass should not be needed.
- The dither choice reaching it: off, TPDF or shaped, defaulting to TPDF whenever the target is 16- or 24-bit. That is the half of #2248 that could not land with the unit, since there was nothing to wire it to.
- The export dialog and any CLI pointed at it rather than at the fork.

## Verification

A render of a known signal comes back off disk as the samples that went in, at each depth. At 16-bit with dither the file's noise floor is flat and its quantisation harmonics are gone, which is the measurement `test_pcm_quantiser.cpp` already makes on the buffer and this would make on the file.

lucaromagnoli:
The sink landed in #2451: `magda/engine/io/AudioFileSink` writes WAV or FLAC at 16 or 24 bits through `PcmQuantiser`, or 32-bit float straight through, with TPDF as the default wherever the target quantises. The tail and trim contract holds -- the file is what the sink was handed, no pre-roll written and no second pass reading it back -- and there is a case pinning that against a plan reporting latency.

What is left here is the wiring: the export dialog and a CLI pointed at it rather than at `TracktionOfflineRenderTask`. Both wait on the app-side bridge (#1897), since `renderOffline` still has no caller outside tests.

PR #2448, opened 2026-09-06
Add the unit that turns float into fixed point, dither included

### LINKED ISSUE #2248: Dither on 16-bit and 24-bit export (native render path)
MAGDA does not dither. Every 16-bit and 24-bit export and CLI render is rounded straight from float. The `OfflineRenderHelper` never enables the renderer's dither flag, the output device flag is off, and the post-FX stage that `TrackChain.hpp` says exists partly for dithering has nothing in it. At 16-bit this is the one place the omission is audible: quiet fades and reverb tails pick up quantisation harmonics instead of a noise floor.

This is for the native engine's render path only. Nothing goes into the Tracktion path.

### Where it applies

| Path | Depth | Dither |
| --- | --- | --- |
| Export and CLI render to 16-bit | 16 | yes, default on |
| Export and CLI render to 24-bit | 24 | yes, correct and free; inaudible |
| Export to 32-bit float | 32f | no quantisation, nothing to do |
| Internal bounces and freezes | 32f by default | nothing to do; stays float between stages |
| Live output | float to driver | driver's business, not the DAW's |

### Design

One small unit, `magda_pcm` or similar, that owns the float-to-integer conversion and the dither together, so export, CLI render and any future bounce-to-16 share the same path. The writer must never round unquantised float on its own.

- **TPDF dither**: two uniform randoms summed, scaled to one LSB of the target depth, added before rounding. The default for "dither on".
- **Noise shaping** (optional mode, "shaped"): error-feedback filter with the published Lipshitz/Vanderkooy/Wannamaker F-weighted coefficients (9 taps), pushing the noise energy above 10 kHz.
- Applied once, at the end of the offline render, immediately before the integer write, keyed on the target depth.
- Deterministic seed so renders are reproducible in the null-diff corpus.
- Export dialog and CLI: a dither switch (off / TPDF / shaped), defaulting to TPDF whenever the target is 16-bit or 24-bit.

Write it rather than vendor it: JUCE has no dither hook, the Faust libraries and Signalsmith do not cover it, and SoX is LGPL and far too large for one stage. If a named algorithm is wanted later, the Airwindows dithers (MIT) are single self-contained files and NJAD is the one worth lifting.

### Test

Render a -90 dBFS sine to 16-bit with and without dither. Without: the spectrum shows harmonics of the quantisation. With TPDF: a flat noise floor and no harmonics. With shaping: the floor is lower below 10 kHz and higher above. Fixed seed, asserted on band energies, not on exact samples.

### Also

Until this lands, nothing in the manual, the export dialog or the site should say the export is dithered.

lucaromagnoli:
Closes #2248. The wiring it also asked for is #2447.

## The gap

MAGDA does not dither. Every 16-bit render rounds straight from float, and at 16 bits that is the one place the omission is audible: a sine one LSB tall becomes a three-level staircase, which is a square wave, which is a tone nobody played.

## The unit

`PcmQuantiser` (`magda/engine/io/`) owns the float-to-fixed-point step and the dither together, so a writer never rounds unquantised float on its own. Three modes:

- `none` — round and nothing else, for a target with no quantisation to hide.
- `tpdf` — two uniform randoms summed, scaled to one LSB. The default wherever the target is fixed point.
- `shaped` — TPDF fed back through the F-weighted nine-tap curve from Lipshitz, Vanderkooy and Wannamaker, so the same noise is spent where there is less hearing to spend it on.

The generator is an xorshift with a fixed seed and a `reset()`, so a render of one project produces one file and the null-diff corpus can carry it. Full scale is held rather than wrapped: a dither step on top of it is over the top of the format, and the alternative is a sign flip.

## Measured, not asserted

On band energies at a fixed seed, which is what the issue asked for. What makes dither right is the shape of what it leaves behind, and a test pinning samples would be pinning the generator.

| | measured |
|---|---|
| undithered, three odd harmonics against the rest of the band | **64%** of it |
| TPDF against undithered at those harmonics | **26x lower**, a floor in their place |
| the tone against its own third harmonic, dithered | **1000x** — the sine survives |
| shaped against flat below 10 kHz | **4.7x lower** |
| shaped against flat above 10 kHz | **58x higher** |

Nine cases in `tests/engine/test_pcm_quantiser.cpp`, including that the same seed renders the same samples twice, that every sample lands on the target's grid, and that 24 bits is 256 steps finer than 16.

## What is not here

The switch in the export dialog and the CLI. Neither can have one yet: the native render writes into a sink and there is no sink in the product that writes a file, the export dialog still renders through Tracktion (which #2248 rules out), and there is no CLI. That wiring is #2447, and it carries the dither choice with it.

Full suite: 3669 passed, 13 skipped.


magda/engine/io/PcmQuantiser.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * listening tests were run.
+ */
+constexpr float kFWeighted[PcmQuantiser::kShaperTaps] = {2.412f,  -3.370f, 3.937f,  -4.174f, 3.353f,
+                                                         -2.205f, 1.281f,  -0.569f, 0.0847f};
+
+/// The largest magnitude a signed sample of @p bits can hold, as a float. The
+/// negative side reaches one step further, and the positive side is what a
+/// conversion has to stay inside.
+float fullScaleFor(int bits) {
+    return static_cast<float>((1ULL << static_cast<unsigned>(bits - 1)) - 1ULL);
```
lucaromagnoli: **P2 — Preserve the quantised codes through the PCM writer**

Using `2^(bits - 1) - 1` as the scale leaves the returned floats off the binary PCM grid. For example, a 16-bit code of -1 is returned as `-1/32767`; JUCE's `AudioFormatWriter::convertFloatsToInts` followed by the WAV writer's 16-bit conversion stores that as -2. This introduces another quantisation error after the dither and outside the shaper's feedback. The current grid test checks against the same scale, so it misses the storage mismatch. Please return integer codes suitable for writing directly, or define and verify a float-to-writer conversion that preserves every quantised code, with 16/24-bit file round-trip coverage.

— Codex

magda/engine/io/PcmQuantiser.cpp
```diff
@@ -0,0 +1,107 @@
...
+                // round took away, so feeding it forward is what puts the noise
+                // where the curve wants it.
+                for (auto tap = 0; tap < kShaperTaps; ++tap)
+                    value += kFWeighted[tap] * history[tap];
+            }
+
+            const auto wanted = value;
+
+            if (mode_ != DitherMode::none)
+                value += nextUniform() + nextUniform();
```
lucaromagnoli: **P2 — Use double precision for the 24-bit quantisation arithmetic**

At larger 24-bit sample values, `float` spacing is half a target LSB, so adding TPDF in float rounds away part of the noise before `nearbyint` applies the intended quantisation. Reproducing this arithmetic with the default generator over 200,000 draws, integer-valued input 6000000 remains unchanged about 94% of the time, whereas adjacent input 6000001 remains unchanged about 44%; ideal TPDF gives 75% for either. The resulting noise therefore depends on the signal's code parity. Please perform scaling, noise addition, rounding and error feedback in double precision, and exercise actual 24-bit processing at higher levels; the existing 24-bit test only compares `lsb()`.

— Codex

magda/engine/io/PcmQuantiser.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // [-0.5, 0.5).
+    constexpr float kInverse = 1.0f / 16777216.0f;
+    return (static_cast<float>(state_ >> 8U) * kInverse) - 0.5f;
+}
+
+void PcmQuantiser::process(juce::AudioBuffer<float>& buffer, int numSamples) {
+    const auto channels = juce::jmin(buffer.getNumChannels(), errors_.getNumChannels());
+    const auto count = juce::jmin(numSamples, buffer.getNumSamples());
+
+    for (auto channel = 0; channel < channels; ++channel) {
```
lucaromagnoli, line 105: **P2 — Keep each channel's random sequence independent of block partitioning**

The channel-major loop consumes one shared RNG state, so changing the render block size assigns different random draws to each channel. With identical seed and 1024 samples of stereo silence, reproducing this loop as one 1024-sample call versus two 512-sample calls changes 204 output codes in each channel. This conflicts with `OfflineRenderRequest`'s explicit guarantee that block size is only a batching choice and the rendered audio is sample-identical. Please use independently seeded per-channel RNG state, or consume draws in a sample-major order, and add a stereo test comparing different block partitions for both TPDF and shaped modes.

— Codex

lucaromagnoli (commented):
Reviewed the quantiser implementation and its tests. The existing [2248] tests pass (9 cases, 4,367 assertions). Three correctness issues remain in PCM storage compatibility, 24-bit arithmetic, and block-size reproducibility.

— Codex

magda/engine/io/PcmQuantiser.cpp
lucaromagnoli: Confirmed, and worse than the finding says. writeFromAudioSampleBuffer converts float to int32 through roundToInt(INT_MAX * sample) (juce_AudioFormatWriter.cpp:73) and the format then keeps the top bits, so the effective factor is 2^(bits-1) - 2^(bits-31) rather than 2^(bits-1). No float can carry a chosen code through that path at any scale: I measured code 17134 written and 17133 read back.

Taken your first option. processToCodes fills the int32 array AudioFormatWriter::write takes, the target's bits at the top, and the header says that path is the only one a writer may use after this unit. A 16- and 24-bit round trip through a WAV in memory now asserts every code comes back exactly. Fixed in 1711d02ef.

magda/engine/io/PcmQuantiser.cpp
lucaromagnoli: Confirmed, and your figures reproduce exactly. Before fixing, I made the arithmetic float again and measured a code left where it was 93.9% of the time on 6000000 and 44.5% on 6000001, against the 75% a triangular distribution gives either.

Scaling, the noise, the round and the error feedback are all in double now, and the case asserts 75% within four points on both codes rather than only comparing lsb(). Fixed in 1711d02ef.

magda/engine/io/PcmQuantiser.cpp
lucaromagnoli: Confirmed. One generator per channel now, seeded through splitmix's finaliser so two channels one apart are unrelated, and reset() restarts each of them.

The new case renders 1024 samples of stereo in one call and in two of 512 and requires them identical, in both dithered modes. Checked it bites: sharing one state across the channels fails it. Fixed in 1711d02ef.

magda/engine/io/PcmQuantiser.cpp
```diff
@@ -0,0 +1,152 @@
...
+                // dither, so what is fed back is everything the round and the
+                // dither together did to it. Shaping only the round would
+                // leave the larger half of the added noise flat.
+                state.errors[0] = wanted - rounded;
+            }
+
+            // Clipped where the writer clips. A code past the end is stored as
+            // the end whatever this does, and letting it through would put the
+            // shaper's feedback a step away from what the file holds.
+            const auto code = juce::jlimit(-maxCode_, maxCode_, rounded);
```
lucaromagnoli: **P2 — Preserve the most-negative PCM code on the integer path**

Now that output uses `AudioFormatWriter::write`, the symmetric clipping limit is no longer the writer's limit: JUCE's `Int16::setAsInt32LE` / `Int24::setAsInt32LE` retain the high bits without the symmetric clamp used by `setAsFloatLE`. With dither off, input `-1.0f` is therefore unnecessarily changed from valid code -32768 to -32767 at 16 bits, and from -8388608 to -8388607 at 24 bits. Even already-representable PCM cannot round-trip unchanged at negative full scale. Please clamp to `[-scale_, maxCode_]` and extend the integer WAV round-trip test to include both extrema; the existing 0.8-amplitude tone never exercises this boundary.

— Codex

lucaromagnoli (commented):
Second pass on 1711d02ef: the three previous findings are addressed. All 12 [2248] test cases pass (9,495 assertions). One remaining PCM boundary issue is noted inline.

— Codex

magda/engine/io/PcmQuantiser.cpp
lucaromagnoli: Confirmed and fixed. `AudioData::Int16::setAsInt32LE` is `*data = (uint16)(newValue >> 16)`, a shift with no clamp, so the symmetric limit was mine rather than the writer's and it was throwing away a code the format holds.

Clamped to `[-scale_, maxCode_]`. The round trip now writes both extrema at 16 and 24 bits with dither off and requires them back unchanged: before the fix, -1.0 came out as -0.99997 at both depths, which is exactly the code you named.

lucaromagnoli (commented):
Rechecked 146294cb1. The negative full-scale boundary finding is fixed: the clamp now preserves -scale_, and the new 16/24-bit WAV round-trip test covers both extrema. No further actionable findings in this pass.

Validation: all 13 [2248] test cases pass (9,513 assertions), and git diff --check is clean. This remains a quantiser-unit review; product export integration is outside this PR's stated scope.

— Codex

PR #2450, opened 2026-09-06
Declare the Chord Engine for what it is: a listener

### LINKED ISSUE #2427: Chord Engine is declared a MIDI emitter but only listens
## Problem

The Chord Engine reads the chain's MIDI to detect chords and emits nothing of its own. `MidiChordEnginePlugin::applyToBuffer` is a transparent passthrough (`MidiChordEnginePlugin.cpp:41`); it clears the buffer only while audition is muted.

It is declared as a MIDI generator anyway:

- `TrackManager.cpp:389` creates it with `deviceType = DeviceType::MIDI`.
- `BaseDevicePack.cpp:193` tags it `midi-generator`.

So `DeviceInfo::emitsMidi()` (`DeviceInfo.hpp:465`) is true and the model treats it as a device with MIDI in and MIDI out:

- `PluginCapabilities.cpp:95` shows it a MIDI thru toggle. On the Tracktion leg the toggle does nothing: the plugin passes its input through regardless.
- `ChainRoutingModel.hpp:97` gives it `MergeRawInputAndPluginOutput` while thru is on (the default), so on the native leg the plan would put a `ChainMidiMerge` behind it and every note would arrive twice: once through the plugin, once through the merge. It has no native adapter today (no `createDevice` in its spec), so this is latent.

## Direction

Declare it for what it is: a device that consumes MIDI and emits none, so the model gives it `RawInputOnly` and no thru toggle. That means either a `DeviceType` other than `MIDI` for it, or letting `emitsMidi()` distinguish a MIDI listener from a MIDI generator. Check what `DeviceType::MIDI` drives in the slot UI (`DeviceSlotHeaderControls`, `DeviceSlotContentLayout`) before changing the type, since it also hides the mod and macro buttons.

When it gets a native adapter, the SDK contract from #2347 makes the passthrough explicit: a listener reads `midiIn` and writes nothing.

## Background

Found while implementing #2347 (PR #2424).

lucaromagnoli:
Closes #2427. With #2449, this unblocks #2314.

## The bug

`MidiChordEnginePlugin::applyToBuffer` is a transparent passthrough: it reads the chain's MIDI to detect chords and writes none, clearing the buffer only while audition is muted.

It was declared a MIDI generator anyway — `TrackManager` created it with `DeviceType::MIDI`, `BaseDevicePack` tagged it `midi-generator` — and `DeviceInfo::emitsMidi()` reads the type, so the model gave it a MIDI output it never writes to.

Two things followed:

- its **thru toggle did nothing**, because the plugin passes its input on regardless;
- `ChainRoutingModel` gave it `MergeRawInputAndPluginOutput` while thru is on, which is the default, so a plan compiled for a chord track would put a `ChainMidiMerge` behind a device that produces nothing and **every note would arrive twice**.

The second was latent only because the plan skipped chord tracks. #2449 gives them a way in, which is why this comes with it.

## The fix

`DeviceType::Analysis` plus `canReceiveMidi` — a transparent tap that reads MIDI. `emitsMidi()` false, no thru toggle, and the plan's transparent path keeps its full stereo width rather than narrowing the chain.

**Migrated, not just fixed at the creation site.** `deviceType` is saved per device (`TrackSerializer.cpp:439`), so every project with a chord track carries the old declaration. The migration joins the existing chain in `ProjectSerializer`, runs on the staged model, and is idempotent. It walks every chain rather than the chord track's own, because the device shows in the browser and can be dropped anywhere, racks included.

**The slot UI does not move.** Two places keyed "hide the mod button, hide the macro button, drop the meter strip" on `deviceType == MIDI`. Both now also read `traits.isChordEngine`, which already existed and is keyed on the `chord-engine` tag, so the device's slot looks exactly as it did.

## Tests

- A saved Chord Engine is declared for what it is, and the migration is idempotent.
- The migration leaves an arpeggiator's declaration alone — it does produce MIDI.
- A MIDI listener reads the chain and gets no thru toggle, asserted on both halves: the type alone would have taken its MIDI input away with its output.
- The registry no longer tags it a MIDI generator, and still tags it a chord engine.

`make test` (858k assertions) and `make test-juce` both pass.

## One consequence worth naming

`addDeviceToTrack` keeps MIDI generators off the master track by type and by tag, and the Chord Engine now matches neither, so it can be added there. An analyser on master is what that rule is meant to allow; it was blocked as a side effect of the wrong declaration.

### DIFF
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -386,7 +386,15 @@ TrackId TrackManager::createTrack(const juce::String& name, TrackType type) {
         engine.uniqueId = "midichordengine";
         engine.fileOrIdentifier = "midichordengine";
         engine.isInstrument = false;
-        engine.deviceType = DeviceType::MIDI;
+
+        // A transparent tap that reads the chain's MIDI and writes none
+        // (#2427). Not DeviceType::MIDI, which is the role of a device that
+        // produces MIDI: DeviceInfo::emitsMidi() reads the type, so declaring
+        // it that way gave the engine a MIDI output it never writes to, a thru
+        // toggle that does nothing, and a merge behind it that would double
+        // every note the moment a chord track compiles.
+        engine.deviceType = DeviceType::Analysis;
+        engine.canReceiveMidi = true;
         engine.format = PluginFormat::Internal;
         addDeviceToTrack(trackId, engine);
 
--- a/magda/daw/core/TrackManager.cpp
+++ b/magda/daw/core/TrackManager.cpp
@@ -15,6 +15,7 @@
 #include "Config.hpp"
 #include "DeviceState.hpp"
 #include "DrumGridPads.hpp"
+#include "LegacyDeviceAliases.hpp"
 #include "ModulatorEngine.hpp"
 #include "PluginCapabilities.hpp"
 #include "PluginPreferences.hpp"
@@ -2216,6 +2217,12 @@ DeviceId TrackManager::addDeviceToPostFx(TrackId trackId, const DeviceInfo& devi
     applyCachedCapabilitiesToDevice(newDevice);
     if (daw::audio::isInternalAnalysisPlugin(newDevice.pluginId))
         newDevice.deviceType = DeviceType::Analysis;
+
+    // This section stamps its own ids from its own counter and so does not go
+    // through prepareNewDevice, which is where the main chain's insertions are
+    // corrected. The declaration a browser drop carries is wrong here for the
+    // same reason it is wrong there (#2427).
+    legacy_devices::normalizeChordEngineRole(newDevice);
 
     auto& elements = track->chain.postFxChainElements;
     insertIndex = std::clamp(insertIndex, 0, static_cast<int>(elements.size()));
--- a/magda/daw/project/serialization/ProjectSerializer.cpp
+++ b/magda/daw/project/serialization/ProjectSerializer.cpp
@@ -379,7 +379,8 @@ bool ProjectSerializer::loadAndStage(const juce::File& file, StagedProjectData& 
         // DeviceType::MIDI device, which is the role of one that produces MIDI,
         // and it produces none. Before the parameter migrations for no reason
         // of its own; it touches nothing they read.
-        legacy_devices::migrateChordEngineRole(outData.tracks, outData.masterTrack.get());
+        legacy_devices::normalizeChordEngineRoleInProject(outData.tracks,
+                                                          outData.masterTrack.get());
 
         // Then any device that renumbered its parameters, so a saved link still
         // addresses the parameter it was made against (#2079). After the
--- a/magda/daw/project/serialization/ProjectSerializer.cpp
+++ b/magda/daw/project/serialization/ProjectSerializer.cpp
@@ -375,6 +375,12 @@ bool ProjectSerializer::loadAndStage(const juce::File& file, StagedProjectData& 
                                                        outData.automationLanes,
                                                        outData.automationClips);
 
+        // Then the Chord Engine's declared role (#2427). It was created as a
+        // DeviceType::MIDI device, which is the role of one that produces MIDI,
+        // and it produces none. Before the parameter migrations for no reason
+        // of its own; it touches nothing they read.
+        legacy_devices::migrateChordEngineRole(outData.tracks, outData.masterTrack.get());
+
         // Then any device that renumbered its parameters, so a saved link still
         // addresses the parameter it was made against (#2079). After the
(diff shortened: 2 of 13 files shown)


magda/daw/core/TrackManager.cpp
```diff
@@ -386,7 +386,15 @@ TrackId TrackManager::createTrack(const juce::String& name, TrackType type) {
...
         engine.isInstrument = false;
-        engine.deviceType = DeviceType::MIDI;
+
+        // A transparent tap that reads the chain's MIDI and writes none
+        // (#2427). Not DeviceType::MIDI, which is the role of a device that
+        // produces MIDI: DeviceInfo::emitsMidi() reads the type, so declaring
+        // it that way gave the engine a MIDI output it never writes to, a thru
+        // toggle that does nothing, and a merge behind it that would double
+        // every note the moment a chord track compiles.
+        engine.deviceType = DeviceType::Analysis;
```
lucaromagnoli, line 397: [P2] Normalize the Chord Engine role in the shared insertion path

This fixes chord-track creation, but adding a Chord Engine from the browser still creates `DeviceType::MIDI`: its registry `browserCategory` remains `"MIDI"`, and the browser/drop handlers derive the device type from that category. `prepareNewDevice` only normalizes plugins tagged `analysis`, which the Chord Engine is not. Consequently, a newly dropped Chord Engine retains the ineffective thru toggle and the duplicate-note merge in the native plan until the project is saved and reopened. Please normalize its type and MIDI-input capability in the shared insertion path as well.

magda/daw/project/serialization/ProjectSerializer.cpp
```diff
@@ -375,6 +375,12 @@ bool ProjectSerializer::loadAndStage(const juce::File& file, StagedProjectData&
                                                        outData.automationLanes,
                                                        outData.automationClips);
 
+        // Then the Chord Engine's declared role (#2427). It was created as a
+        // DeviceType::MIDI device, which is the role of one that produces MIDI,
+        // and it produces none. Before the parameter migrations for no reason
+        // of its own; it touches nothing they read.
+        legacy_devices::migrateChordEngineRole(outData.tracks, outData.masterTrack.get());
```
lucaromagnoli: [P2] Apply the role migration when loading chain and rack presets

Older chain/rack presets also persist the Chord Engine's `DeviceType::MIDI`, but `PresetManager::loadChainPreset` and `loadRackPreset` deserialize their contents directly and never call `loadAndStage`. Those loaders run the existing device/parameter migrations but not this new one, so inserting an old preset reintroduces the ineffective thru toggle and duplicate-note routing even in an otherwise migrated project. Please expose and apply this migration to chain and rack preset fragments too.

lucaromagnoli (commented):
Two paths still retain the old Chord Engine declaration. The rebuilt focused tests passed (30 cases, 220 assertions).

magda/daw/core/TrackManager.cpp
lucaromagnoli: Fixed in 8c5e0a1e. You were right — six call sites derive a device type from the browser category, and `prepareNewDevice` is the one path they all go through, so the rule now sits beside the analysis-tagged normalization already there rather than being repeated at each.

Covered by a test through the real add path (`DeviceApiLive`) that inserts a Chord Engine declared `DeviceType::MIDI`, as the browser hands it over, and asserts it comes back Analysis, consuming MIDI and emitting none. Reverting the call fails it.

magda/daw/project/serialization/ProjectSerializer.cpp
lucaromagnoli: Fixed in 8c5e0a1e. Both loaders now call it, beside the retired-device migration they already ran.

Exposed the way that one is: `normalizeChordEngineRoleInChain` and `normalizeChordEngineRoleInRack` alongside the project walker, so the rule is stated once and applied from four places — the project loader, the two preset loaders, and every insertion. Renamed off "migrate" while doing it, since correcting a browser-derived declaration is not a migration.

Fragment and rack-fragment tests added.

lucaromagnoli:
Fixed in d6c796c7.

Post-FX stamps its ids from its own counter and does not go through `prepareNewDevice`, so it repeats that function's tail — `applyCachedCapabilitiesToDevice` plus the analysis-tag normalization — and missed the Chord Engine line added with it.

Checked the other direction too, so this is the last of them: three places mint a device id — `prepareNewDevice`, `addDeviceToPostFx`, and `duplicateTrack` — and the last copies devices that are already correct.

Test added through `addDeviceToPostFx`; reverting the call fails it. `make test` (858k assertions) and `make test-juce` both pass.

Issue #1897, opened 2026-07-27
Cutover: bridge rewrite, dual-engine release, fork removal


lucaromagnoli:
Part of #1882. The final phase.

Cutover and removal of the TE fork.

- Replace the sync-layer internals with compiler passes: ClipSynchronizer, AudioBridge, PluginManagerSync, RackSyncManager, MidiInputRouter internals. The model stops being mirrored into an engine object graph; it becomes the compiler input.
- Port the 23 engine-level JUCE test suites to the native engine (the Catch2 suite is engine-free and unaffected).
- Flip the default behind the flag, ship at least one release cycle dual-engine, then remove third_party/tracktion_engine (which also drops AirWindows and the rest of the unused payload).
- Cutover-candidate and default-engine are different dates: the audible-bug tail is discovered in use, not scheduled.

## Slices

- [ ] #2551
- [ ] #2552
- [ ] #2553
- [ ] #2554
- [ ] #2555
- [ ] #2556
- [ ] #2557

lucaromagnoli:
I recommend treating these as separate gates rather than one completion event:

1. **Cutover candidate:** parity corpus and interactive QA pass; all required project migrations are available.
2. **Opt-in dual-engine release:** native engine is selectable, with engine choice recorded in diagnostics/crash reports.
3. **Default flip:** native engine becomes default, but TE remains available as a fallback for at least one release.
4. **TE removal:** fallback usage and native-only blocker counts meet an explicit threshold; TE-only project fixtures still load through migration tests.

“Candidate,” “default,” and “fork removal” should have different exit criteria and dates. Suggested criteria include:
- zero known data-loss/project-load regressions;
- defined acceptable audio-difference classes from #1896;
- launcher/recording/manual QA matrix completed across supported platforms;
- no remaining production code outside the TE adapter includes TE;
- rollback path tested before the default flip.

Fork removal should be its own final PR so it is reviewable as a dependency/boundary deletion rather than mixed with bridge rewrites.

lucaromagnoli:
Acceptance addition: cutover requires performance parity, not just sample parity. The benchmark suite (see #1896) must show the native parallel executor matching or beating TE CPU-per-block on the project corpus at matched buffer sizes. The single-threaded executor is a debugging reference and is not eligible for cutover.

PR #2451, opened 2026-09-06
Give an offline render a sink that writes a file

lucaromagnoli:
Part of #2447.

## What was missing

`renderOffline` wrote into an `OfflineRenderSink`, and no sink in the product wrote a file: the two that existed were `BufferSink` and `CollectingSink`, both in tests. Nothing the native engine rendered could become an export, a bounce or a freeze.

## What this adds

`magda/engine/io/AudioFileSink` writes WAV or FLAC at 16 or 24 bits through `PcmQuantiser` (#2248), or at 32-bit float straight through.

- Fixed point reaches the writer as codes on the integer path. `writeFromAudioSampleBuffer` converts through `INT_MAX` and lands codes one low, which would be a second rounding after the dither and outside the shaper's feedback.
- Dither defaults to TPDF wherever the target quantises and to none at float, ignoring an explicit ask at 32-bit float. That is the half of #2248 that could not land with the unit, since there was nothing to wire it to.
- Rate and channel count come from the `RenderContext` rather than from the caller: they are what the plan was prepared with, and a header that disagrees is a file that plays at the wrong speed or the wrong width.
- A spec no format holds (FLAC above 24 bits, a depth nothing writes) and a file that cannot be opened or replaced are refused at `create`, so a render nothing can store is never run.

## Tail and trim

The file is what the sink was handed. The range, the tail and the plan's latency are settled in `renderOffline`, so nothing renders a pre-roll and nothing reads the file back to cut one off, which is what `MainWindowExport` does on the Tracktion path today.

## Verification

`tests/engine/test_audio_file_sink.cpp`, read back off the disk rather than asserted against numbers written in the test: the codes the quantiser chose at each depth in both containers, a float file holding the render's own samples, the dither choice reaching the file, a shaped render staying shaped, block splits leaving no trace, the width and rate coming from the context, the refusals, replace rather than append, and a block narrower than the file refused rather than stored.

One case in `test_offline_render.cpp` covers the trim contract end to end: a plan reporting 333 samples of latency renders to a file whose first sample is the range's first sample.

Full engine suite green locally (933 cases).

## Not in this PR

Pointing the export dialog and a CLI at this. Both wait on the app-side bridge (#1897): there is no native `AudioEngine` implementation yet, so `renderOffline` still has no caller outside tests. That is what remains open on #2447.


lucaromagnoli (commented):
Reviewed the file sink and its JUCE writer/stream interactions. All 10 [2447] cases pass (102,627 assertions). A separate probe linked against the compiled sink reproduced the two failure-path issues below; no repository source changes were made.

— Codex

magda/engine/io/AudioFileSink.cpp
```diff
@@ -0,0 +1,133 @@
...
+    const auto floating = spec.bitDepth >= 32;
+
+    auto format = formatFor(spec.format);
+    if (!format->getPossibleBitDepths().contains(spec.bitDepth))
+        return nullptr;
+
+    // Replaced rather than appended to: a FileOutputStream over an existing
+    // file starts at its end, so a second render into the same name would be
+    // written after the first one's header.
+    if (destination.existsAsFile() && !destination.deleteFile())
```
lucaromagnoli: **P1 — Preserve the previous destination when creating the writer fails**

The destination is deleted before the format has accepted the complete context or the replacement stream/writer has been created. I reproduced this with an existing file and `{format=flac, bitDepth=24}` plus a nine-channel context: `create()` returns null because the FLAC encoder rejects the channel count, but the previous export has already been replaced by an empty file. Thus a refused export destroys existing audio before any render runs. Please create the replacement in a sibling temporary file and preserve the original until the replacement is ready to commit; checking bit depth alone cannot protect against channel/rate or subsequent I/O failures. Add a refusal test with an existing destination and assert that its contents survive.

— Codex

magda/engine/io/AudioFileSink.cpp
```diff
@@ -0,0 +1,133 @@
...
+        codeChannels_[static_cast<std::size_t>(channel)] =
+            codes_.data() + (static_cast<std::ptrdiff_t>(channel) * numSamples);
+
+    scratch_.setSize(numChannels_, numSamples, false, false, true);
+}
+
+bool AudioFileSink::close() {
+    // Destroying the writer is what finishes the file: it writes the header it
+    // could not write until the length was known, and flushes the encoder.
+    writer_.reset();
```
lucaromagnoli, line 219: **P1 — Include finalisation failures in the close result**

`failed_` only records synchronous block-write failures. JUCE's FileOutputStream buffers writes, WAV rewrites its header during destruction, and FLAC flushes pending encoded samples during destruction, so disk errors can first occur inside `writer_.reset()` and are discarded here. A probe writing 16 stereo samples to WAV, then setting the process file-size limit to zero immediately before `close()`, returned `true` with `samplesWritten()==16` while the destination was zero bytes. Disk-full/quota failures can therefore be reported as successful exports. Please retain observable stream/finalisation error state through writer destruction and include it in the sticky close result; cover a failure that occurs only on final flush, for both containers.

— Codex

magda/engine/io/AudioFileSink.cpp
lucaromagnoli: Fixed. The render goes to a `juce::TemporaryFile` beside the destination now, and `close()` is the one moment the destination changes: a refusal in `create()` and a failure during the render both leave what was already there. Your FLAC-plus-nine-channels case is a test, which reads the existing file back after the refusal and compares it, and checks the directory for leftovers.

That case also turned up a leak on the way. `createWriterFor(unique_ptr&, options)` is documented to leave the stream alone on failure, and `~FlacWriter` empties the pointer and then drops it without deleting, so the open file leaked. `create()` reclaims it, with a comment saying a JUCE bump has to re-read that: a version that deleted it instead would make the reclaim a double free, and the FLAC refusal case is what would catch it.

magda/engine/io/AudioFileSink.cpp
lucaromagnoli: Fixed. The writer writes through a stream that reports into something outliving it, and `close()` reads that report after `writer_.reset()`.

The flag alone would not have been enough. `FileOutputStream` keeps only its last operation's result, and the WAV writer's header seek flushes the buffer internally and drops that failure, so the header write that follows resets the status to ok: a truncated file with nothing having returned false. So `close()` also checks length, since the bytes the writer handed over are what the file has to be.

Covered for both containers with a 512-byte `RLIMIT_FSIZE` around a render that fits entirely in the stream buffer, so the only thing that runs out of room is the finish. I mutation-checked it: with the length check removed the case fails, `close()` returning true after the truncated file has already replaced the previous export.

lucaromagnoli (commented):
Rechecked 7d7052592. Both previous findings are addressed: the sibling temporary file preserves the destination on refusal/failure, and close() observes finalisation errors before committing it.

Recompiled and reran the original probe against the updated sink: nine-channel FLAC is refused with the previous contents intact; denying the final WAV flush now makes close() return false. All 12 [2447] tests pass (108,809 assertions), including the new WAV/FLAC failure cases. git diff --check is clean. No further actionable findings in this pass.

— Codex

Issue #2446, opened 2026-09-06
The native plan skips the chord track, so nothing on it renders


lucaromagnoli:
Found while starting #2314, which names this work but describes it as a missing ClipSnapshot lane. It is not: it is one line in the plan compiler.

## What happens

`Compiler::computeTrackOrder` excludes chord tracks from the topological order outright (`magda/engine/plan/PlanCompiler.cpp:413`):

```cpp
// The chord track is a monitor-only progression lane and never renders.
std::vector<bool> skipped(numTracks, false);
for (std::size_t i = 0; i < numTracks; ++i)
    skipped[i] = tracks_[i].type == TrackType::Chord;
```

A skipped track contributes no edges, is not counted into `remaining`, and is never emitted, so the plan carries nothing for it: no clip ops, no device ops, no route to master.

## Why that is wrong

The comment states a decision the rest of the app does not make. `TrackManager::createTrack` builds a chord track with **two devices** — a Chord Engine and a 4OSC — routes it to master like any other track, and seeds its `muted` flag from the chord-preview preference (`TrackManager.cpp:336-395`). That mute is the audition toggle: `TrackManager::isChordTrackMuted()` returns the chord track's own `muted`, and the Chord Engine reads it to decide whether to drop playback MIDI before it reaches the instrument (`MidiChordEnginePlugin.cpp:119`).

So the chord track is a monitor lane that sounds. It sounds on the fork today, which makes it a parity requirement by #1896's rule: anything the current engine delivers that users depend on is a parity requirement by default, and an exclusion needs an argument. There is none on this line — the comment asserts the exclusion rather than arguing it.

## What it takes

- Drop the skip, so a chord track compiles like any other: its clips play, its devices are emitted, and its mute is an ordinary track mute, which is exactly what the audition toggle wants to be.
- Fix the Chord Engine's declared role first, or the chain routes it wrong the moment it compiles. `DeviceInfo::emitsMidi()` is true for anything whose `deviceType` is `DeviceType::MIDI`, and `TrackManager.cpp:389` sets exactly that, with `BaseDevicePack.cpp:193` tagging it `midi-generator`. The device reads MIDI and emits none, so under #2347's contract it wants `takesMidiInput` without `producesMidi`, which is `RawInputOnly` and no thru toggle. This is #2427's finding, which was closed into #2314.
- Plan goldens, and a null-diff case for a chord track playing through to master.

## What this is not

[...]

PR #2452, opened 2026-09-06
Port the Chord Engine to MagdaDevice: a listener, not a plugin

### LINKED ISSUE #2314: Port the MIDI Chord Engine to MagdaDevice: values, not lookups

lucaromagnoli:
Closes #2314. The last of #2299's MIDI device ports, and the one #2312's recipe could not take directly.

## What the service was actually for

The issue's premise is that the port needs the chord track's content in `ClipSnapshot`. It does not. The device calls exactly **one** method on `DeviceTrackContext` — `isChordTrackMuted()`, which returns the chord track's own `muted` flag — and nothing anywhere reads a chord under the playhead.

What actually blocked it was the plan skipping chord tracks (#2446) and the device's declared role (#2427). Both are in, so this is now a value:

`AudioBridge` pushes the mute where it already applies it, which is the same event — the audition toggle *is* the chord track's mute. `DeviceTrackContext::isChordTrackMuted` is deleted; the interface keeps its other method for external plugins.

## The audio thread

Under #2347 a device declaring no MIDI output gets a **const** `midiIn`, so the old `applyToBuffer` trick — clearing the chain's buffer while audition is off — isn't available. It also isn't needed. That clear did two things:

- **stopped the downstream instrument** — already done by the mute itself, since the audition toggle is a real track mute (`AudioBridge.cpp:369`);
- **kept playback out of the detection** — the device's own state, which it can simply decline to record.

The transport half of the condition now comes off the block (`DeviceProcessContext::isPlaying`) rather than `edit.getTransport()`.

## Where detection lives

Unchanged. It was already engine-neutral — a 30 Hz timer over a lock-free FIFO, touching nothing of Tracktion's — so it stays on the device, which two `MagdaDevice`s already do (`FaustPlugin`, `FaustInstrumentPlugin`). Only `process()` belongs to the audio thread.

## Blast radius

Small, because the class keeps its name. Four `dynamic_cast` sites become `deviceFromPlugin` — the same call the Strum and Arpeggiator use — and every signature, nested type (`AIProgression`, `Listener`) and UI binding stays put. Registration moves from `.createPlugin` to `.createDevice`, matching the Oscilloscope, which is the other parameterless Analysis device.

## Tests

New device test for what changed: it records what goes past and writes nothing, and audition-off keeps playback out of the detection while still detecting when the transport is stopped. Reverting the gate fails it.

`test_device_services_juce` asserted the poll this removes; it now asserts the contract that replaced it.

`make test` (868k assertions), `make test-juce` and `make debug` all pass.

### DIFF
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -18,6 +18,8 @@
 #include "plugin_manager/ExternalPluginState.hpp"
 #include "plugins/DeviceServices.hpp"
 #include "plugins/InternalPluginRegistry.hpp"
+#include "plugins/MidiChordEnginePlugin.hpp"
+#include "plugins/tracktion/TracktionMagdaDevicePlugin.hpp"
 #include "session/SessionMonitorPlugin.hpp"
 
 namespace magda {
@@ -359,6 +361,26 @@ void AudioBridge::syncRecordArmedToTE(TrackId trackId) {
     }
 }
 
+namespace {
+
+/// Push a chord track's audition state to the Chord Engines on it (#2314).
+///
+/// The device used to poll a `DeviceTrackContext` for this; an engine-built one
+/// has no session to poll, so the host pushes it where it already applies the
+/// mute. The audition toggle is the track's own mute, which is why this sits
+/// beside `setMute` rather than anywhere of its own.
+void pushChordAudition(te::AudioTrack& teTrack, const magda::TrackInfo& trackInfo) {
+    if (trackInfo.type != magda::TrackType::Chord)
+        return;
+
+    for (auto* plugin : teTrack.pluginList)
+        if (auto* engine = magda::daw::audio::tracktion_adapter::deviceFromPlugin<
+                magda::daw::audio::MidiChordEnginePlugin>(plugin))
+            engine->setChordTrackMuted(trackInfo.muted);
+}
+
+}  // namespace
+
 void AudioBridge::trackPropertyChanged(int trackId) {
     // Track property changed (volume, pan, mute, solo, recordArmed) - sync to Tracktion Engine
     auto* track = getAudioTrack(trackId);
@@ -368,6 +390,7 @@ void AudioBridge::trackPropertyChanged(int trackId) {
             // Sync mute/solo to track
             track->setMute(trackInfo->muted);
             track->setSolo(trackInfo->soloed);
+            pushChordAudition(*track, *trackInfo);
 
             // Sync freeze state
             if (trackInfo->frozen != track->isFrozen(te::AudioTrack::individualFreeze)) {
@@ -966,6 +989,7 @@ void AudioBridge::syncAll() {
             // Sync mute/solo state to TE (essential on project load)
             teTrack->setMute(track.muted);
             teTrack->setSolo(track.soloed);
+            pushChordAudition(*teTrack, track);
 
             // Sync audio output routing (group/aux targets now exist from first pass)
             trackController_.setTrackAudioOutput(track.id, track.audioOutputDevice);
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -363,20 +363,39 @@ void AudioBridge::syncRecordArmedToTE(TrackId trackId) {
 
 namespace {
 
-/// Push a chord track's audition state to the Chord Engines on it (#2314).
-///
-/// The device used to poll a `DeviceTrackContext` for this; an engine-built one
-/// has no session to poll, so the host pushes it where it already applies the
-/// mute. The audition toggle is the track's own mute, which is why this sits
-/// beside `setMute` rather than anywhere of its own.
+/**
+ * @brief Push a chord track's audition state to the Chord Engines on it (#2314).
+ *
+ * The device used to poll a `DeviceTrackContext` for this; an engine-built one
+ * has no session to poll, so the host pushes it where it already applies the
+ * mute. The audition toggle is the track's own mute, which is why this sits
+ * beside `setMute` rather than anywhere of its own.
+ *
+ * @param teTrack   the engine track whose plugins carry the devices
+ * @param trackInfo the model track it was built from; anything but a chord
+ *                  track returns immediately
+ */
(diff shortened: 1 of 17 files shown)


magda/daw/audio/AudioBridge.cpp
```diff
@@ -368,6 +390,7 @@ void AudioBridge::trackPropertyChanged(int trackId) {
             // Sync mute/solo to track
             track->setMute(trackInfo->muted);
             track->setSolo(trackInfo->soloed);
+            pushChordAudition(*track, *trackInfo);
```
lucaromagnoli, line 421: [P2] Initialize audition state after device insertion

The replacement push runs on track-property changes and full sync, but `trackDevicesChanged` only calls `syncTrackPlugins`, which never pushes this value. Adding a Chord Engine to an already-muted chord track therefore leaves its `chordTrackMuted_` at the default `false`, so playback enters detection until another property change or full sync. This also affects normal chord-track creation: `createTrack` sends `notifyTracksChanged` before adding the default devices, so the initial full sync has no Chord Engine to update. Please push the current audition state after per-track plugin synchronization as well.

magda/daw/audio/AudioBridge.cpp
```diff
@@ -359,6 +361,26 @@ void AudioBridge::syncRecordArmedToTE(TrackId trackId) {
...
+///
+/// The device used to poll a `DeviceTrackContext` for this; an engine-built one
+/// has no session to poll, so the host pushes it where it already applies the
+/// mute. The audition toggle is the track's own mute, which is why this sits
+/// beside `setMute` rather than anywhere of its own.
+void pushChordAudition(te::AudioTrack& teTrack, const magda::TrackInfo& trackInfo) {
+    if (trackInfo.type != magda::TrackType::Chord)
+        return;
+
+    for (auto* plugin : teTrack.pluginList)
```
lucaromagnoli: [P2] Deliver audition state to rack-contained Chord Engines

This loop visits only the track's top-level plugins. A Chord Engine placed inside a rack lives in the rack type's plugin list, and `deviceFromPlugin` only unwraps `TracktionMagdaDevicePlugin`; it does not descend into `RackInstance`. Those devices never receive `setChordTrackMuted`, even during full sync or when the audition toggle changes, and consequently keep detecting playback while audition is off. Please traverse the track's device mappings or recursively visit rack contents so every Chord Engine on the chord track receives the value.

lucaromagnoli (commented):
Two gaps in the replacement mute-state delivery path. Both new device tests passed (8 assertions), but they set the mute directly and do not exercise host delivery.

magda/daw/audio/AudioBridge.cpp
lucaromagnoli: Fixed in d3d9aeaa. It descends into racks now, the way `BottomPanel::findChordEngine` already did — I had the precedent in front of me and did not apply it.

magda/daw/audio/AudioBridge.cpp
lucaromagnoli: Fixed in d3d9aeaa. Pushed after `syncTrackPlugins` in `trackDevicesChanged` as well.

Confirmed the creation ordering you describe: `createTrack` notifies at `TrackManager.cpp:367` and adds the chord devices at 400 and 411, so the full sync that follows has no engine to reach — the device-sync push is what catches it.

Tested through the bridge on exactly that order: the track is muted first, then the device sync has to carry it. `true` can only have come from the push, since the device's own default is `false`. Reverting the call fails the test.

magda/daw/audio/AudioBridge.cpp
```diff
@@ -359,6 +361,45 @@ void AudioBridge::syncRecordArmedToTE(TrackId trackId) {
...
+    // Racks too. `deviceFromPlugin` unwraps the host's device adapter and
+    // nothing else, so a Chord Engine dropped inside a rack lives in the rack
+    // type's own plugin list and a top-level walk never reaches it.
+    for (auto* plugin : teTrack.pluginList) {
+        push(plugin);
+
+        if (auto* rackInstance = dynamic_cast<te::RackInstance*>(plugin))
+            if (rackInstance->type != nullptr)
+                for (auto* innerPlugin : rackInstance->type->getPlugins())
+                    push(innerPlugin);
```
lucaromagnoli: [P2] Traverse nested racks when pushing audition state

The revised walk reaches devices in a top-level rack, but `push(innerPlugin)` only unwraps a device adapter; it does not descend into another `RackInstance`. `RackSyncManager::loadRackContents` represents nested racks as separate rack types and inserts their `RackInstance` into the outer rack, so a Chord Engine inside a nested rack still never receives `setChordTrackMuted` and continues detecting playback with audition off. Please make this traversal recursive (or visit the track's mapped devices). The new `Chord Audition Push` integration test passes, but only exercises a top-level Chord Engine; a nested-rack case would catch this remaining gap.

lucaromagnoli: Fixed in b4bb3e73. Recursive now, unguarded the way `chain_walk`'s own descent is — racks nest as a tree, and a rack containing itself would already be a cycle in the model.

You were right about the test too. It now builds the case you describe — an engine inside a rack inside a rack — and asserts over *every* engine the track carries rather than naming one, so what it holds is that none is missed. Its finder does its own descent rather than calling the walk under test, which would have passed for exactly the reason the walk was wrong. Reverting to one level of descent fails it.

Issue #2279, opened 2026-09-01
Insert capture for bounce: a session, an immutable capture, and a playback that fails to construct


lucaromagnoli:
Part of #1893, and the half of #2245 that came back out of it.

A bounce cannot run a hardware insert: an offline render chews through an hour of timeline in a minute and no outboard answers at that speed. So it either renders at real time or plays back a recording of what the hardware said during a normal pass, and the incumbent does the latter (`InsertCapturePlugin`, a hidden plugin sitting behind the insert).

`EngineInsert` is already the seam for this and needs nothing added: an implementation that returns captured samples and one that returns what an interface just handed it satisfy the same two calls, and which one a render binds is what decides.

## Why this is its own issue

A first attempt landed in #2278 and was taken back out. It was one dual-mode object that owned the live hardware, the real-time capture, the replay, two sample-rate domains, coverage, failure status, MIDI storage and latency conversion, and each review round fixed a symptom of that rather than the shape.

The finding that settled it: a capture can be unfit to replay -- taken at another sample rate, or with a gap where a pass seeked over the window -- and `EngineInsert` has no way to express that. The executor binds a pointer and calls it. A `usable()` flag is a rule some caller has to remember to check before binding, and the render that forgot would write wrong audio into a file with nothing downstream able to see it.

## The shape

Three types instead of one, so the failure cannot be ignored:

- **`InsertCaptureSession`** owns the live insert and writes capture data during a normal playback pass. It is the only thing that touches hardware and the only thing that writes.
- **`InsertCapture`**, immutable once taken: the audio, the MIDI, the rate it was captured at, the window it covers, the round trip, and whether it is complete. Nothing can edit it after the fact, so nothing has to defend against half-written state.
- **`InsertCapturePlayback`**, which consumes a validated capture for a render context and either constructs or does not. No live-hardware pointer, no mode, no validity flag: a render that cannot be served has nothing to bind, which is what makes the failure impossible to miss.

## What has to be decided in it

- **Rates.** A capture is written at whatever rate the device is open at and read at whatever rate the export asked for. Positions and the round trip survive that if they are held in seconds; the samples do not, and need either a resampler at construction or a refusal. Refusing is honest -- re-capturing is a live pass -- but the product question is whether an export at another rate should silently need one.
- **Where the samples live.** In memory sized to the window, or on disk like the incumbent's. A minute of stereo at 44.1 kHz is 21 MB, and several inserts over a long export is the case that decides it.
[...]

PR #2454, opened 2026-09-06
Capture what a hardware insert says, so a bounce can play it back

lucaromagnoli:
Part of #2279.

A render pulls an hour of timeline in a minute and no outboard answers at that speed, so a bounce plays back a recording of a live pass. `EngineInsert` was already the seam for it; what was missing was anything to bind to it.

## Three types, not one

The shape #2279 settled on after #2278 came back out. A capture can be unfit to replay -- taken over a window a pass seeked across, or narrower than the render -- and the executor binds a pointer and calls it, so fitness cannot be a flag a caller remembers to read.

- **`InsertCaptureSession`** forwards every `send` and `receive` to the live insert, so the pass sounds exactly as it would without it, and writes down what came back. Room for the whole window is taken at construction, so the write path never allocates. Coverage is a bit per sample rather than a count: a pass that seeked leaves a hole, and a hole is not a short total. The progress counter read from another thread counts distinct samples, so a pass that looped covers its window once.
- **`InsertCapture`** is that recording, immutable once taken: audio, MIDI, the rate it was taken at, the window and the round trip. The last two in seconds, so reading the capture at another rate keeps the delay the pass actually had.
- **`InsertCapturePlayback`** either constructs or does not. Null for an incomplete capture, a window it does not cover, or fewer channels than the render. A render that cannot be served has nothing to bind.

MIDI is bounded at construction, since the write path may not allocate: 1100 events a second, which is past what a DIN return can carry at 31250 baud. A dropped or oversized message makes the capture incomplete rather than silently thinner.

## The two calls the issue left open

**Rates: resampled, not refused.** `InsertRenderCaptureService` already records at the device rate and resamples on completion, so refusing would be a parity regression: an export at 96 kHz of a session tracked at 44.1 would need a second live pass. The playback resamples at construction, off the audio thread, through `cubicLagrange` -- the curve a file at another rate is read through, so nothing is resampled to a quality a listener could name.

**Storage: memory sized to the window, not disk.** The incumbent writes temp files. What a disk-backed capture needs is a FIFO and a writer thread off the audio thread, which is what recording (#1895) is about to build; the header says it goes through here when it exists. A stereo minute at 44.1 kHz is 21 MB.

## Verification

`tests/engine/test_insert_capture_session.cpp`, all offline against a stub insert that answers with a tone derived from where the block is:

- a pass writes what the hardware returned, and runs through it block for block
- a pass that seeked over the window leaves a capture no playback can be built on
- a pass over the same window twice covers it once
- replay lands sample for sample where the pass received it, and reports the same round trip
- how a bounce cuts its blocks is not in the output
- a 96 kHz render of a 48 kHz capture keeps the tone
- a window it does not cover, and a capture narrower than the render, are both refused
- a program change comes back two bytes long rather than three
- a block the transport is not moving through returns silence, not the last thing in the buffer

Full engine suite green locally: 945 cases, 679,259 assertions.

## Not in this PR

The session owner -- the equivalent of `InsertRenderCaptureService` orchestrating a capture pass around a bounce -- which is app-side and waits on the bridge (#1897). That is what keeps #2279 open.


magda/engine/insert/InsertCaptureSession.cpp
```diff
@@ -0,0 +1,156 @@
...
+        const auto at = blockStart + metadata.samplePosition;
+        if (at < 0 || at >= windowSamples)
+            continue;
+
+        if (midiCount_ >= static_cast<int>(midi_.size())) {
+            midiOverflowed_.store(true, std::memory_order_relaxed);
+            break;
+        }
+
+        auto& event = midi_[static_cast<std::size_t>(midiCount_)];
```
lucaromagnoli, line 116: **P2 — Replace MIDI when overwriting a captured timeline range**

Audio is overwritten when a pass loops or seeks back, but MIDI is always appended. Replaying one 128-sample window twice, with one returned event at sample zero on each pass, produces a complete capture with two events at sample zero. A bounce therefore doubles note-ons/controllers (and retains messages that disappeared on the later pass), although its audio is from the later pass. Repeated loops can also exhaust the fixed event capacity despite the final window needing very few events. Please give MIDI the same overwrite semantics as audio for revisited ranges, or reject such captures consistently, while keeping the receive path allocation-free. The existing loop test has no MIDI.

— Codex

magda/engine/insert/InsertCapturePlayback.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+            const auto sampleAt = [&](int offset) {
+                return from[std::clamp(index + offset, 0, std::max(0, last))];
+            };
+
+            to[at] = cubicLagrange(sampleAt(-1), sampleAt(0), sampleAt(1), sampleAt(2), fraction);
+        }
+    }
+
+    return resampled;
```
lucaromagnoli, line 152: **P2 — Filter before downsampling the captured return**

Cubic interpolation alone does not remove content above the destination Nyquist frequency. At a 2:1 reduction, every interpolation position is integral, so this simply keeps every other source sample. The compiled playback turns a unit-amplitude 30 kHz tone captured at 96 kHz into an 18 kHz tone at 48 kHz, with RMS 0.707107: ultrasonic return content becomes full-level audible aliasing. Please use a rate converter with an anti-alias filter for rate reductions and add a high-frequency downsampling test; the current 100 Hz upsampling test cannot detect this.

— Codex

magda/engine/insert/InsertCaptureSession.cpp
```diff
@@ -0,0 +1,156 @@
...
+        std::min(static_cast<int>(audio.getNumChannels()), audio_.getNumChannels());
+    const auto count = static_cast<int>(to - from);
+    const auto sourceOffset = static_cast<int>(from - blockStart);
+
+    for (auto channel = 0; channel < channels; ++channel)
+        audio_.copyFrom(channel, static_cast<int>(from),
+                        audio.getChannelPointer(static_cast<std::size_t>(channel)) + sourceOffset,
+                        count);
+
+    markCovered(from, to);
```
lucaromagnoli, line 92: **P2 — Require coverage of every recorded audio channel**

`markCovered()` marks the entire time span complete even when the preceding `min()` copied fewer channels than the capture owns. A stereo session receiving a mono block over its whole window yields `complete()==true`, and `InsertCapturePlayback::create` accepts it for a stereo render because it checks the allocated channel count. The unrecorded second channel is silently exported as zero (or stale audio after a revisit). I reproduced this with one 128-sample mono block into a two-channel session. Please mark a short audio return incomplete or track coverage per channel, accounting explicitly for MIDI-only receive calls, so allocated width cannot masquerade as captured width.

— Codex

lucaromagnoli (commented):
Reviewed the capture/session/playback implementation and the executor seam. All 10 [2279] tests pass (6,492 assertions). A separate probe linked against the compiled implementation reproduced the three issues below. No repository source changes were made.

— Codex

magda/engine/insert/InsertCapturePlayback.cpp
```diff
@@ -0,0 +1,198 @@
...
+ * an outboard return carries plenty up there. Linear phase, so the delay it
+ * adds is exactly half its length and comes straight back off.
+ *
+ * 102 taps at the design below, run once over the capture rather than per
+ * block, and only when the export asks for a lower rate than the pass had.
+ */
+juce::AudioBuffer<float> bandLimited(const juce::AudioBuffer<float>& source, double sourceRate,
+                                     double cutoff) {
+    const auto coefficients = juce::dsp::FilterDesign<float>::designFIRLowpassKaiserMethod(
+        static_cast<float>(cutoff), sourceRate, 0.05f, -80.0f);
```
lucaromagnoli: **P2 — Use an odd-length symmetric FIR and compensate its actual delay**

These Kaiser parameters return order 101 (102 taps). In this vendored JUCE, `designFIRLowpassWindowMethod` special-cases `i == order / 2` as the sinc centre even for odd orders, making the resulting coefficients asymmetric. Also, `length / 2` advances 51 samples, whereas a symmetric 102-tap FIR would delay 50.5. A probe of the compiled 96-to-48 kHz playback, excluding the boundaries, gives a 10 kHz unit sine RMS 0.731606 instead of 0.707107 and a phase advance of 0.3386 radians; the 30 kHz input still aliases at RMS 0.0258365 (about -29 dB relative attenuation despite the -80 dB design). The new `rms < 0.05` test passes this defective response. Please choose an even filter order/odd tap count (or generate correct coefficients explicitly), remove `(length - 1) / 2` samples of delay, and test passband gain/timing as well as meaningful stopband attenuation.

— Codex

lucaromagnoli (commented):
Rechecked d98dd4b9b. The MIDI overwrite/capacity and short-channel-return findings are addressed. All 14 [2279] cases pass (6,503 assertions), and git diff --check is clean. The new anti-alias filter needs one further correction, reproduced against the compiled playback below.

— Codex

magda/engine/insert/InsertCaptureSession.cpp
lucaromagnoli: Fixed. Coverage is a write number per sample now instead of a bit: each `receive` that writes stamps its range, events carry the stamp they arrived under, and `take()` keeps only the ones whose sample still belongs to that write. Audio is replaced by the copy and MIDI by the same fact, and a message that disappeared on the later pass goes with it.

Capacity too: the room is compacted in place before it is called full, so ten passes over a one-event stretch end with one event rather than an overflow. Two cases, since as you say the old loop test had no MIDI in it: the later pass's note surviving alone, and ten loops through a capture with room for four events.

magda/engine/insert/InsertCaptureSession.cpp
lucaromagnoli: Fixed. A return with fewer channels than the capture owns now writes nothing and flags itself, so the span stays uncovered, the capture is incomplete, and `create` builds no playback for it. Your repro is a case: mono blocks over the whole window leave `missingSamples() == kWindowSamples`.

On the MIDI-only point: the capture is sized to the channel count it was given rather than forced to at least one, so an insert whose return is MIDI alone owns zero audio channels and its receive calls still cover the window. A narrow return is a return short of the width that was recorded, which that is not.

magda/engine/insert/InsertCapturePlayback.cpp
lucaromagnoli: Fixed, and you were right that the existing test could not see it. A linear-phase low-pass now runs before the interpolation whenever the rate goes down, with the cutoff placed so the stopband starts at the target's Nyquist, and its delay comes straight back off rather than shifting the capture against the render. It runs once over the capture rather than per block, and only when the export asks for a lower rate than the pass had.

See the thread below for the second half of this: the first version of that filter used `designFIRLowpassKaiserMethod`, whose coefficients are asymmetric at the order it picked, so it was still aliasing at -29 dB. The taps are generated in the unit now, and the case measures the pass band as well as the stop band.

magda/engine/insert/InsertCapturePlayback.cpp
lucaromagnoli: Confirmed on all three counts, including that my own case was too loose to see it. `designFIRLowpassKaiserMethod` returns order 101, and the window method takes the centre as `order / 2` whatever the parity, so index 50 gets the DC value while the rest is a sinc about 50.5: the kernel is asymmetric and has no one delay to remove.

The taps are generated in the unit now, odd by construction, so the centre is a tap and the delay is exactly `(length - 1) / 2`. Same Kaiser window and rejection, normalised to unity at DC, and the cutoff is derived so the stopband starts at the target's Nyquist rather than sitting at a factor of 0.45.

Measured on the compiled playback, 96 to 48 kHz: 10 kHz comes back at RMS 0.708293 with no sample further than 1.2e-05 from where it should be, against 0.7316 and 0.34 radians before; 30 kHz is at RMS 3.1e-06, against 0.0258. Two cases replace the one: the pass band asserted sample by sample, which is what catches a shift, and the stop band tightened to `< 0.005`, which the asymmetric kernel fails at 0.026.

lucaromagnoli (commented):
Rechecked dce9d61d4. The symmetric odd-length FIR and corrected delay address the remaining finding. Relinked the signal probe against the updated playback: at 96-to-48 kHz, the 10 kHz tone now has RMS 0.707098 with effectively zero phase shift, and the 30 kHz alias is down to RMS 0.00000310. The additional 192-to-48 kHz probes also show strong suppression above the destination Nyquist.

All 15 [2279] tests pass (6,762 assertions), and git diff --check is clean. No further actionable findings in this pass.

— Codex

Issue #35, opened 2026-01-17
Agent-to-agent communication


lucaromagnoli:
Agents communicate via orchestrator for collaborative tasks.

PR #2456, opened 2026-09-06
Hold the launch de-click fix with impulse material

### LINKED ISSUE #2444: The fork de-clicks a launched slot's own first sample
Found by #2441's first session case, which put an impulse on the slot's first sample and measured a full-scale residual.

## What happens

`SlotControlNode::processSubSection` applies `applyAudioStartDeClick(pc.buffers.audio, launchFadeSamples)` on every block where the handle went from stopped to playing (`tracktion_SlotControlNode.cpp:325`). That helper reads the block's first sample as a discontinuity and subtracts a decaying copy of it over the fade:

```cpp
const auto discontinuity = dest[0];
...
const auto correction = 0.5f * (1.0f + std::cos (pi * phase));
dest[i] -= discontinuity * correction;
```

It asks nothing about where in its material the clip is. A slot launched from its own beginning is not resuming mid-material, so its first sample is the material's own attack, and the de-click removes it and leaves 256 samples of tail in its place.

The engine has the same de-click, ported sample for sample (`FadeCurves.cpp`, `StartDeClick`), and one extra question in front of it (`ClipVoice.cpp`):

```cpp
const auto beginsAtItsOwnStart = windowStart <= event.span.seconds.start + (0.5 / sampleRate_);
if (!sounded_ || !block.continuous) {
    if (beginsAtItsOwnStart)
        deClick_.reset();
    else
        deClick_.begin(region, clip.launchFadeSamples);
}
```

That guard was written for the arrangement, where #2040 found the same thing on an impulse sitting on a clip edge and the incumbent's arranger path does not de-click. Its launcher path does.

## What was measured

`session.launch` with impulse material, a slot launched on the render's first sample. The engine renders the impulse; the fork renders `1.0 - 1.0 = 0` at sample 0 and a raised-cosine decay from -0.99996 downwards after it. Peak residual 0 dBFS.

The case ships playing a tone instead, whose file is windowed at both ends, so the slot's first sample is silence and neither de-click has a step to take. It nulls at -130.5 dB.

## Why it is worth a line

Nothing MAGDA ships is broken by it today in a way a user would call a bug -- a drum one-shot launched from a slot loses its transient and gains a click-length tail, which reads as the clip being soft rather than as a defect. What it is, is a divergence between the two engines on the one path #2306 calls an ordinary null, and it is the engine that is right.

## Not in scope

Fixing the fork. The guard belongs in `SlotControlNode`, which is upstream's, and the engine already has it. This is here so the difference is written down and the corpus case that would have found it says why it plays a tone.

lucaromagnoli:
Closes #2444. **Draft: do not merge before Conceptual-Machines/tracktion_engine#35**, and the submodule pointer wants moving to `magda_minimal`'s head once that lands rather than the branch commit it names now.

## The fix is in the fork

`SlotControlNode` subtracted a decaying copy of the block's first sample whenever a slot went from stopped to playing, so a slot whose first sample is a transient lost it and gained a click-length tail — a drum one-shot launched from a slot loses its crack. The engine already asks first whether there is a step to take, and leaves a voice that begins at its own start alone (`ClipVoice.cpp`). tracktion_engine#35 gives the fork the same question, and keeps the de-click where there really is a boundary: a scene launch joining a run already playing, or a clip with an offset.

## What changes here

`session.launch` goes back to **impulses**. That is the material that found the bug and the only material that holds it fixed: an impulse puts a full-scale transient on exactly the sample the two engines disagreed about. The case shipped playing a tone because a windowed file's first sample is silence — which passes either way and proves nothing about this.

Measured on the corpus, with the fork patched and unpatched:

| | peak | rms |
|---|---|---|
| impulses, unpatched | **0.0 dBFS** | -32.6 |
| impulses, patched | **-133.0 dB** | -169.9 |

So the corpus is the negative verification: revert the fork commit and this case fails at full scale.

`make test` (976k assertions) and `make test-juce` both pass.

## A correction to the issue

#2444 says fixing the fork is out of scope, on the grounds that `SlotControlNode` is upstream's. That was written from a wrong assumption: `third_party/tracktion_engine` is a Conceptual-Machines fork that already carries its own PRs, so it is ours to patch.

### DIFF
--- a/tests/NullDiffCorpus.cpp
+++ b/tests/NullDiffCorpus.cpp
@@ -2400,21 +2400,15 @@ std::vector<Case> buildCorpus(const juce::File& scratchDirectory) {
                              launching(plainTrack()));
         value.endBeat = 8.0;
 
-        // A tone, whose file is windowed at both ends, so the slot's first
-        // sample is silence. Both engines take a step out of a launched clip's
-        // first samples and only one of them asks first whether there is a step
-        // to take: the fork's SlotControlNode de-clicks whatever the block
-        // begins with, while the engine leaves a voice that begins at its own
-        // start alone, on the grounds that what looks like a step there is the
-        // material's own attack (ClipVoice.cpp). Impulses put a full-scale
-        // transient on exactly that sample, so an impulse case here would be
-        // measuring the launch instant, which #2306 asserts and this does not.
-        // The difference itself is #2444.
-        //
-        // Nothing interpolates between the two at one rate and ratio one, so the
-        // ordinary floor still applies: the tone is here for what its first
-        // sample is, not for what an interpolator would do to it.
-        const auto source = writeSource(scratchDirectory, "session", tone());
+        // Impulses, which put a full-scale transient on the slot's very first
+        // sample. That is the sample both engines take a step out of when a
+        // launch begins, and the one they used to disagree about: the fork
+        // de-clicked whatever a launched block began with and removed the
+        // transient, where the engine leaves a voice that begins at its own
+        // start alone (ClipVoice.cpp). Fixed in the fork for #2444, and this
+        // material is what holds it fixed -- a tone would pass either way,
+        // because its file is windowed and its first sample is silence.
+        const auto source = writeSource(scratchDirectory, "session", impulses());
         value.sources.push_back(source);
 
         // Sixteen beats against an eight-beat render, so the run never reaches


lucaromagnoli:
Re-pointed in 982a7c41. The submodule now names `magda_minimal`'s head (`c54563de`, the squash of #35) rather than the fix branch commit, so nothing here tracks a branch any more.

Content-identical to what was measured — `git diff` between the two commits is empty — and re-verified against the merged pointer: `session.launch` still nulls at a peak of -133.0 dB and the JUCE suite passes.

The blocker in the description is cleared; this is mergeable as it stands.

third_party/tracktion_engine
```diff
@@ -1 +1 @@
-Subproject commit ddb43f367db103ddf5988b3f1dd20be7846d1b7d
+Subproject commit c54563ded0c02dee7dbd1dc380e71fcae8aaf06a
```
lucaromagnoli: [P2] Preserve de-clicking for clips with a nonzero source offset

The pinned fork change's `beganAtMaterialStart` check in `SlotControlNode.cpp:332-336` compares the playing range's start with `playStartTime`, but neither value includes the clip's source offset. For a normal launch, `LaunchHandle::advance` sets `playStartTime` to the playing range's start, so this comparison succeeds even when `WaveNodeRealTime` starts reading mid-file via `clip.getOffsetInBeats()`. That disables the existing launch de-click for a trimmed clip beginning at a nonzero waveform value, potentially introducing a click, despite the comment and PR description saying offset clips retain it. Please account for the source offset before skipping the correction and add an offset-launch regression case. The new zero-offset impulse case passes at -133.0 dB, but does not exercise this boundary.

lucaromagnoli: You are right on the facts, and I have corrected the comment and the description — but I have not made the behaviour change, because measuring it first said it would create a divergence rather than remove one.

**The naming was wrong.** On a plain launch `LaunchHandle::advance` sets `playStartTime` to the played range's start, so the test succeeds whatever the source offset is. It does not mean "began at its own material start"; it means "began here rather than being joined in progress by a scene launch". Renamed to `joinedRunInProgress` in [tracktion_engine#36](https://github.com/Conceptual-Machines/tracktion_engine/pull/36), and the comment no longer claims offset clips keep their de-click.

**On accounting for the offset.** I built the offset-launch case you asked for and rendered it through both engines before deciding:

```
native.wav    first samples: 0.5 0.5 0.5 0.5 ...
incumbent.wav first samples: 0.5 0.5 0.5 0.5 ...
```

Neither de-clicks a trimmed start. MAGDA's engine draws the line in the same place — `ClipVoice::beginsAtItsOwnStart` asks whether the voice begins where the *clip* begins, not where the file does, which a trimmed clip satisfies. So accounting for the offset in the fork alone would make it de-click where the engine does not.

Whether either should de-click a trimmed start is a fair question and I think a real one, but it is one question for both engines: filed as #2457 with the measurement.

**The case is not in this PR**, and that is worth flagging rather than burying: a trimmed launched slot also renders **23 samples apart** between the engines at -113.5 dB, which is a trim-alignment difference and would have failed the case for a reason that has nothing to do with the de-click. Filed as #2458 with the working case and where to start looking.

Issue #2462, opened 2026-09-06
Recording slice 4: MIDI recording, takes, and the notes a loop leaves open


lucaromagnoli:
Slice 4 of #1895. Depends on slices 1 and 2. Parallel to slice 3: the same pass, the other kind of material.

## What the fork does today

`TracktionEngineWrapper::finalizeMidiRecording` (`daw/engine/TracktionEngineWrapperRecording.cpp:799`) reads TE's recorded `MidiClip` back out into the model: notes into `MidiNote` in beats, controller events split so pitch wheel becomes `MidiPitchBendData` and anything under 128 becomes `MidiCCData`, one `MidiTake` per loop pass out of `getTakeSequence`, and an active take that is the last full pass, stepping back one when the final pass has fewer notes than the one before it. Everything else TE recorded is dropped.

So MIDI recording is already defined in model terms. What it is missing is a producer that is not TE.

## Scope

- Capture: what the live MIDI source (slice 1) saw, sample-stamped, through the write path (slice 2). Nothing model-shaped is built on the audio thread.
- Conversion at materialisation: sample stamps to beats through the tempo map, which is where a pass recorded across a tempo change either lands right or does not. Beats are the domain, and a recorded note is stored as `startBeat` and `lengthBeats`, never as seconds.
- Notes still held when the pass ends, and notes held across a loop wrap: what length a note gets when its note-off never arrives, and which pass owns a note that started in the previous one.
- One `MidiTake` per loop pass, with the shipped active-take rule (last full pass, back one when it is thinner).
- CC and pitch bend carried at the same resolution they arrived at; what the model has no field for is dropped here rather than silently in a converter.
- Program changes and other channel messages: today they do not survive, and this slice either keeps that or states what changes.

## Not in this slice

The audible path. An armed track's live MIDI reaches its instrument through the input op in slice 1; nothing about recording changes what is heard, and if it does that is a bug in this slice.

Previews (slice 5), slots (slice 6), swap carry (slice 7).

## Verification

Offline, feeding the source a known event stream against a driven transport:

- notes land on the beats they were played on, at a fixed tempo and across a tempo change inside the pass
- a note held past the end of the pass gets a defined length rather than a zero-length or a runaway one
- a note held across a loop wrap ends up in one pass, once
- one take per pass, and the active take rule at a full final pass and a cut-short one
- CC and pitch bend survive with their positions; a message the model cannot hold is dropped at a named place
[...]

Issue #2461, opened 2026-09-06
Recording slice 3: an arrangement audio take, from armed to clip


lucaromagnoli:
Slice 3 of #1895. Depends on slices 1 and 2: an input to record and a path to disk.

## What the fork does today

TE records into clips of its own, and `TracktionEngineWrapper::recordingFinished` (`daw/engine/TracktionEngineWrapperRecording.cpp:83`) unpicks them afterwards: it reads the takes out of the clip's `IDs::TAKES` tree, resolves each one through `SourceFileReference`, builds a `std::vector<AudioTake>`, picks the active take, calls `removeFromParent()` on TE's clip, and creates a MAGDA clip which then syncs a new TE clip back into the edit.

The rules that survive that dance are the ones this slice owns, and they are MAGDA's rather than TE's:

- Loop recording is loop-aligned: one take per pass, all sharing the clip start, and a lead-in recorded before the loop is not kept. `AudioTake` has no per-take offset and comping assumes take 0 at t=0, and that is a model decision, not an engine one.
- The active take is the last full pass, stepping back one if the final pass is short (under 95% of the longest), because only the last pass can be cut off mid-way.
- Start alignment is a head correction: the fork adds silence or drops samples at the top of the file (`adjustSamples`) from `WaveInputDevice::getAdjustmentSeconds()`, which is the user's record adjustment plus the device manager's, clamped to +/- 3 s.

The dance itself does not survive. The native path writes files and creates one model clip; nothing creates an engine-side clip that then has to be deleted.

## Scope

- A take as an engine object: a writer opened at a timeline position, fed by slice 2, closed on stop, reporting what it wrote and whether anything was lost.
- Where the take starts. The transport position at record start, less the input latency slice 1 declared. Count-in needs nothing new -- `TransportClock` already holds `countInBeats` and `ClickGenerator` already sounds it -- but the take must start at the count-in's end, not at its start.
- Loop passes: a file per pass, aligned to the loop, with the shipped lead-in rule kept as it is.
- Where the files go: the project's recordings directory, the same place `MagdaEngineBehaviour::getFileForNewAudioRecording` names today.
- Materialisation on the message thread: one `AudioClipModel` with its `takes` and `currentTakeIndex`, built when the pass finishes. No model object is created from the audio thread, and no clip exists until recording ends.
- A take whose write path reported an overrun or a failed write is surfaced as such rather than quietly becoming a short clip.

## Not in this slice

[...]

Issue #2460, opened 2026-09-06
Recording slice 2: the capture queue and the writer thread


lucaromagnoli:
Slice 2 of #1895. Depends on the engine core. Independent of slice 1: one gets the input in, this one gets samples out to disk, and slice 3 is where they meet.

## What the fork does today

`WaveInputRecordingThread` (`playback/devices/tracktion_WaveInputDevice.cpp`) owns a `BlockQueue`: 32 pre-made `QueuedBlock`s, a free list and a pending list, each behind its own `juce::CriticalSection`. The audio thread calls `addBlockToRecord`, which takes `freeQueueLock`, and if the free list is empty does this:

```cpp
QueuedBlock* findFreeBlock()
{
    const juce::ScopedLock sl (freeQueueLock);
    if (firstFree == nullptr)
        return new QueuedBlock();
    ...
}
```

then copies the block through `buffer.setSize (...)` and takes `pendingQueueLock` to append. Two locks the writer thread also holds while draining, and an allocation on the audio thread whenever the disk falls behind. When the backlog passes 500 blocks it logs "Audio recording can't keep up!" once and carries on; a write that fails logs and stops the transport from a timer.

That is the mechanism, not a design to copy: the failure mode it is built around is exactly the one the native write path may not have.

## Scope

The write path, on its own, with no notion of a take or a clip:

- A per-writer SPSC FIFO, capacity fixed at prepare, that the audio thread copies into without a lock and without an allocation.
- A writer thread that drains it to disk and is the only thread that touches a file.
- Sample-stamped MIDI through the same path, bounded the way `kMaxMidiBytesPerPort` bounds a port, so a burst cannot allocate either.
- An overrun that is reported as lost samples rather than absorbed. A take with a hole in it is data loss, and the consumer has to be able to see it happened rather than find a shorter file than it expected.
- The writer's own failure (disk full, file gone) surfacing as a fact about the take, for slice 3 to act on.

`tap/SampleRing.hpp` is the wrong primitive to reuse here and it is worth saying why: its contract is that the writer never blocks and the reader sees the present, which means overwriting what the reader has not taken. That is right for a meter and for the retro-capture rings in #2311; for a take it silently drops the middle of a performance.

## Not in this slice

Where the samples came from, what file they go into, and what the file becomes. Also disk-backed insert capture, which #2279 deferred to this path (`InsertCaptureSession` is memory-only and says so); it queues behind this rather than shaping it.

## Verification

[...]

PR #2467, opened 2026-09-06
Let a monitoring track hear the input the device just captured

### LINKED ISSUE #2459: Recording slice 1: a live input reaching an input op
Slice 1 of #1895. Depends on the engine core and on nothing else in this subsystem, which is why it is first.

## What is already here

Arming is structural, the way the design notes said it would be. The model carries `recordArmed` and `InputMonitorMode` (`daw/core/TrackInfo.hpp`), `monitorsInput()` is the gate, and `PlanCompiler` already emits an `OpKind::AudioInput` or `OpKind::MidiInput` for a track whose input route is `External`, tags it `LivenessDomain::Live`, and leaves it out entirely for a track that is neither armed nor monitoring. `PlanBindings` already has `audioInputs` and `midiInputs` keyed by `TrackId`.

So the plan shape is done. What is missing is everything behind the binding.

## What is missing

`EngineSession::process(int numSamples, juce::AudioBuffer<float>& output)` takes an output buffer and nothing else. The device callback's input channels have no way into a render, no object stands for a live input, nothing resolves `TrackInfo::audioInputDevice` or `midiInputDevice` to actual hardware channels, and input latency appears nowhere in the plan.

## Scope

- The input reaching the session: the callback's input channels arriving with the block they belong to, rather than a source reading a buffer somebody else filled at another time.
- A live audio source behind `OpRole::LiveAudioInput`: an `EngineAudioSource` reading the channels its track names. The strings are the ones #2272 has to resolve on the output side (`"stereo:<device>"` for a pair, a bare device name for mono), and the mono-into-stereo and stereo-into-mono rules get stated once, here.
- A live MIDI source behind `OpRole::LiveMidiInput`: an `EngineMidiSource` fed from the MIDI callback, with events sample-stamped against the block. The fork's preview path reads the transport position off an atomic at message-arrival time (`MidiInputRouter::TrackMidiPreviewConsumer`), which is where its timing slop comes from; a sample stamp taken against the block the event lands in is what slices 4 and 6 need to exist.
- Monitor semantics as they actually ship: `monitorsInput()` is armed or `In`, so `Auto` is audible only while armed. That belongs in the compiler, which already asks the question, not in a runtime flag.
- Input latency declared by the host on the source, and deliberately not handed to the compensation pass: what arrives has already happened, so delaying the rest of the graph to match it would put playback and the click behind the live signal rather than align anything. It is the number a recorded take is placed by (#2461), which is the only thing that can act on it.

## Not in this slice

Anything that writes. No queue, no file, no take: a monitored input is audible and nothing else. Previews, session slots and plan-swap carry are their own slices.

## Verification

Offline, against a stub input frame filled with a known signal:

- a track monitoring input renders it; a track neither armed nor monitoring compiles without the op at all
- `Auto` is silent unarmed and audible armed, and `In` is audible either way
- a mono input into a stereo track and a stereo pair into a mono track, at the gains the fork uses
- MIDI events land at their sample offsets within the block, and the port's byte budget (`kMaxMidiBytesPerPort`) is respected when a burst exceeds it
- the latency the host declares is what the source reports, and a plan carrying an input op resolves to no delay at all

lucaromagnoli:
Slice 1 of #1895, closes #2459.

## What was missing

Less than it looked like. `PlanCompiler` already emits an `AudioInput` or `MidiInput` op for a track whose input route is External, tags it `LivenessDomain::Live`, and leaves it out entirely for a track that is neither armed nor monitoring; `PlanBindings` already has `audioInputs` and `midiInputs` keyed by track, and `PlanExecutor` already renders them through whatever is bound. Arming was structural before this and still is.

What was missing was anything to bind. `EngineSession::process(int numSamples, juce::AudioBuffer<float>& output)` took an output buffer and nothing else, so the callback's input had no way into a render at all.

## What this adds

`magda/engine/io/LiveInput`:

- **`LiveInputFeed`** holds the callback's input and narrows it to each block. A callback a loop wraps inside renders as two blocks, and the input is not timeline material and knows nothing about that, so the window is handed out per block rather than read per callback. It is closed when the callback ends, so a source reached between callbacks reads silence rather than the last one's samples.
- **`LiveAudioInput`** reads the channels its track was pointed at. A destination wider than the map repeats the map's last channel, which is what `WaveInputDeviceNode::process` does (it copies its last read channel across the remaining destination channels), so a mono input is heard in both ears rather than only the left. A channel the device does not have is silence and is counted, since a wrong channel map is otherwise indistinguishable from a quiet room.
- **`LiveMidiInput`** takes one device's events or every device's, at the offsets the host stamped. The fork collects messages between callbacks and places them relative to the first one's timestamp, which its own comment calls "near enough for live stuff" (`MidiInputDeviceNode::processSection`); a stamp against the block the event arrived in is what #2462 needs to record a note where it was played. A burst past the port's byte budget is dropped and counted rather than overrunning `kMaxMidiBytesPerPort`.

`EngineSession::process` takes a `LiveInputBlock` alongside the output it fills, defaulted for the hosts that have no inputs and for every offline render. `liveInputs()` hands the host the feed to build sources against, which it binds through the `createAudioInput` / `createMidiInput` hooks the factory already had.

Which hardware channels a track's `audioInputDevice` names, and which device its `midiInputDevice` names, stays the host's to resolve: the engine knows channel indices and an opaque source id, never a device name.

## One decision, and it reverses what the issue said

Input latency is declared on the source and never handed to the compensation pass. The issue asked for it to be carried by the plan so the monitored path was "compensated like any other latency", and that is wrong: what arrives has already happened, so declaring it would delay every other branch to match, putting playback and the click behind the live signal rather than aligning anything. The number exists to place a recorded take, which is #2461's job. #2459 has been corrected to say so, and a case pins it: a plan carrying an input op resolves to no delay at all.

## Verification

`tests/engine/test_live_input.cpp`, eight cases, offline throughout:

- the compiler gate over all six monitor-mode and armed combinations, including that `Auto` alone is silent until the track is armed
- the callback's own samples channel for channel, and a mono input in both ears
- a channel the device does not have reading silence, and being counted
- a block of a split callback reading its own part of the input, both as a unit and through a real session with a loop set, where the two pieces of the wrapped callback read the input in order
- MIDI at the offsets the host stamped, filtered to one device or merged across all, re-offset inside a split callback
- a 1000-event burst dropping exactly what the byte budget cannot hold
- silence before the first callback and after the last one ends

Full engine suite green locally: 958 cases, 680,001 assertions.

## Not in this PR

Anything that writes. No queue, no file, no take: a monitored input is audible and nothing else. The capture queue and writer thread are #2460, which is independent of this and can go in either order.

### DIFF
--- a/magda/engine/exec/EngineSession.cpp
+++ b/magda/engine/exec/EngineSession.cpp
@@ -6,6 +6,25 @@
 
 namespace magda::engine {
 
+namespace {
+
+/// Closes the feed's callback however process() leaves, so a source reached
+/// between callbacks reads nothing rather than the last one's samples.
+struct ScopedLiveInput {
+    explicit ScopedLiveInput(LiveInputFeed& feed) : feed_(feed) {}
+
+    ~ScopedLiveInput() {
+        feed_.endCallback();
+    }
+
+    ScopedLiveInput(const ScopedLiveInput&) = delete;
+    ScopedLiveInput& operator=(const ScopedLiveInput&) = delete;
+
+    LiveInputFeed& feed_;
+};
+
+}  // namespace
+
 EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> plan,
                                              const RenderContext& context,
                                              const RuntimeStateIds& modelIds, PlanValues values) {
@@ -156,6 +175,20 @@ void EngineSession::publishClips(std::shared_ptr<const ClipSnapshot> clips) {
 
 void EngineSession::process(int numSamples, juce::AudioBuffer<float>& output,
                             const LiveInputBlock& input) {
+    // The executor asserts on a block longer than the plan was prepared for.
+    // Here the same number also says how much buffer there is to write into, so
+    // it is held to what the caller actually provided rather than trusted: the
+    // pieces of a split callback are views onto this, and a view past the end
+    // of it is not a wrong answer, it is someone else's memory.
+    jassert(numSamples <= output.getNumSamples());
+    numSamples = std::min(numSamples, output.getNumSamples());
+
+    // Before the clear, and before anything else touches the output: a host may
+    // hand the same memory for input and output, and the feed's copy is what
+    // makes that the caller's business rather than a silent loss of the take.
+    liveInputs_.beginCallback(input, numSamples);
+    const ScopedLiveInput scopedInput(liveInputs_);
+
     output.clear();
 
     PublishedRender::ScopedAccess<farbot::ThreadType::realtime> render(published_);
@@ -172,19 +205,6 @@ void EngineSession::process(int numSamples, juce::AudioBuffer<float>& output,
     // still asked at the next real one.
     if (numSamples <= 0)
         return;
-
-    // The executor asserts on a block longer than the plan was prepared for.
-    // Here the same number also says how much buffer there is to write into, so
-    // it is held to what the caller actually provided rather than trusted: the
-    // pieces of a split callback are views onto this, and a view past the end
-    // of it is not a wrong answer, it is someone else's memory.
-    jassert(numSamples <= output.getNumSamples());
-    numSamples = std::min(numSamples, output.getNumSamples());
-
-    // What the device captured, held once for the callback and narrowed to
-    // each block below. Ended before returning, so a source reached outside a
-    // callback reads nothing rather than the last one's samples.
-    liveInputs_.beginCallback(input, numSamples);
 
     // Values and plans travel separately and are swapped one after the other,
     // so for the moment between the two the ones in flight can belong to the
@@ -236,8 +256,6 @@ void EngineSession::process(int numSamples, juce::AudioBuffer<float>& output,
             (*render)->click->render(transport->tempo, transport->click, segment.block,
                                      segment.countingIn, output, segment.startSample);
     }
-
-    liveInputs_.endCallback();
 }
 
 }  // namespace magda::engine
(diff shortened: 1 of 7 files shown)


magda/engine/exec/EngineSession.cpp
```diff
@@ -180,6 +181,11 @@ void EngineSession::process(int numSamples, juce::AudioBuffer<float>& output) {
     jassert(numSamples <= output.getNumSamples());
     numSamples = std::min(numSamples, output.getNumSamples());
 
+    // What the device captured, held once for the callback and narrowed to
```
lucaromagnoli: **P2 — Preserve input that aliases the callback's output buffer**

`LiveInputBlock::audio` is a borrowed view, but `process()` clears `output` before `beginCallback()` reads that view, and the feed never copies the samples. A host using an in-place input/output buffer therefore loses the entire live input. I reproduced this through a real monitoring session with `session.process(64, shared, {blockOf(shared), {}})`: the source receives 0 at sample 3 instead of the input's 3, and the master output is entirely silent. Both integration tests use separate input and output allocations, so they miss this case. Please preserve overlapping input in preallocated storage before clearing or rendering output (moving beginCallback alone is insufficient because it only stores pointers), and add an integration test for shared/overlapping channel storage.

— Codex

lucaromagnoli (commented):
Reviewed the new live-input feed, sources and EngineSession integration. All 8 [live-input] tests pass (474 assertions), and git diff --check is clean. A separate real-session test reproduces the buffer-aliasing issue below. No repository source changes were made.

— Codex

lucaromagnoli (commented):
Rechecked f09ff3ccd. The shared-buffer finding is fixed: the feed copies input into preallocated storage before output is cleared, and scoped cleanup closes the feed on all exits. The new real-session shared-buffer regression passes.

All 10 [live-input] tests pass (615 assertions). No further actionable findings in this pass. Review scope is the live-input fix; the separate in-progress recording changes were not reviewed.

— Codex

Issue #2311, opened 2026-09-02
Capture rings on the tap seam: retro audio/MIDI capture and listen-pass thumbnails


lucaromagnoli:
## Summary

Capture rings on the engine's tap seam: the SampleRing idiom (SPSC, writer never blocks, reader sees the present) with longer windows, sized in bars. Master ring always on (configurable length, ~60 s stereo = ~23 MB), per-track rings on demand, a MIDI ring on live input. These cover exactly the session journal's two blind spots: live external audio input, and external-plugin output as actually heard.

## Verbs

- capture the last N bars of the master (or a track) as audio, straight to the timeline: #1996's recorder is UI over this ring, with its drag-to-timeline flow
- retrospective MIDI record: keep what was played while not recording (pairs with #1895's live MIDI path)
- during a loop, the journal indexes ring offsets per iteration, so "keep pass 7's audio" is a direct slice

## Listen-pass thumbnails

While the transport plays, the master ring persists small compressed thumbnails (Opus, background thread, roughly 15 MB per hour of listening). They give any future consumer instant recorded sound for a past listen pass through the preview-player idiom (the transport is never touched).

## Consumers

- #1996 looping buffer recorder
- the mixing agent's single-pass tap render: rings on every track's post-fader tap replace OfflineMixAnalysis's N+1 offline solo passes (its own documented follow-up), fix the resident-stem memory cost, and add a live "analyze what I just heard" mode
- the session journal recorder (#2310): rings cover its two non-reconstructible boundaries, live external input and external-plugin output as heard

## Budgets

- master ring RAM is the budget item; length is a setting
- per-track rings opt-in only: twenty always-on rings would be half a gig, and that version is refused by design
- CPU per ring is a memcpy per block

## Depends on

- native engine tap ops (post-cutover); the session journal issue for indexing

PR #2473, opened 2026-09-07
Add the capture queue and the writer thread

### LINKED ISSUE #2460: Recording slice 2: the capture queue and the writer thread

lucaromagnoli:
Slice 2 of #1895, closes #2460. Independent of slice 1: one gets the input in,
this one gets samples out to disk, and slice 3 is where they meet.

`RecordStream` is one take's queue, single producer and single consumer, with the
room taken at construction so a write on the audio thread is a copy and nothing
else. `RecordThread` is its counterpart to `PrefetchThread`: one thread rather
than one per take, polling rather than woken from the callback, and draining
every stream that has anything before it sleeps at all, because a round that is
behind here is a hole in a recording rather than a glitch.

Not `tap/SampleRing.hpp`, and the header says why: a ring lets the writer
overwrite what the reader has not taken, which is right for a meter and for the
retro capture buffers in #2311. A take is the past, so this refuses to overwrite
and reports what it had to drop instead. Sample stamped MIDI goes through the
same path, bounded by event count and by three bytes a message, so a burst
cannot allocate either. A sink that refuses a write breaks the take once and the
queue keeps draining, leaving #2461 to decide what a broken take becomes.

## Verification

Offline throughout: a synthetic producer, a consumer that may or may not keep up,
and no take, no file and no input anywhere in the cases.

| What #2460 asks for | Case |
| --- | --- |
| nothing lost when the writer keeps up, and in order | A drained stream holds every sample, in order; A take crossing the end of the queue comes back unbroken |
| an overrun reported with a count, not blocked or overwritten | A stalled writer costs samples, and says how many |
| capacity is what was set and does not grow under load | A stalled writer costs samples, and says how many |
| MIDI keeps its sample stamps, burst included | MIDI keeps the sample it was played on; MIDI a stream cannot hold is dropped and counted |
| a failed write reported once and reaching the asker | A write the sink refuses breaks the take rather than the queue |

Full suite green: 3740 cases, 997128 assertions. `[record]` green over 30
consecutive runs, which is the case with a real background thread behind it.

## One divergence from the issue

#2460 says capacity is fixed at prepare. It is fixed at construction here, which
is stricter and is what makes a write a copy and nothing else. Slice 3 reads that
contract, so it is worth a word now if you would rather it match the wording.

### DIFF
--- a/magda/engine/io/RecordStream.cpp
+++ b/magda/engine/io/RecordStream.cpp
@@ -1,0 +1,151 @@
+#include "io/RecordStream.hpp"
+
+#include <algorithm>
+
+namespace magda::engine {
+
+namespace {
+
+int roundedCapacity(int wanted, int floorSamples) {
+    return juce::nextPowerOfTwo(std::max(floorSamples, wanted));
+}
+
+}  // namespace
+
+RecordStream::RecordStream(RecordSink& sink, const RecordStreamSettings& settings)
+    : sink_(sink),
+      capacity_(roundedCapacity(settings.capacitySamples, 1024)),
+      mask_(static_cast<std::uint64_t>(capacity_) - 1),
+      chunkSamples_(std::max(1, settings.chunkSamples)) {
+    if (settings.numChannels > 0)
+        audio_.setSize(settings.numChannels, capacity_);
+
+    midi_.resize(static_cast<std::size_t>(roundedCapacity(settings.midiCapacityEvents, 64)));
+    midiMask_ = midi_.size() - 1;
+}
+
+bool RecordStream::writeAudio(juce::dsp::AudioBlock<const float> audio, int numSamples) {
+    const auto channels = audio_.getNumChannels();
+    if (numSamples <= 0 || channels == 0)
+        return true;
+
+    const auto write = audioWrite_.load(std::memory_order_relaxed);
+    const auto read = audioRead_.load(std::memory_order_acquire);
+    const auto free = static_cast<std::uint64_t>(capacity_) - (write - read);
+
+    if (free < static_cast<std::uint64_t>(numSamples)) {
+        samplesLost_.fetch_add(numSamples, std::memory_order_relaxed);
+        return false;
+    }
+
+    const auto offset = static_cast<int>(write & mask_);
+    const auto first = std::min(numSamples, capacity_ - offset);
+    const auto rest = numSamples - first;
+    const auto sourceChannels = static_cast<int>(audio.getNumChannels());
+
+    for (auto channel = 0; channel < channels; ++channel) {
+        // A stream given fewer channels than it holds writes silence into the
+        // rest rather than leaving the last take's samples there.
+        if (channel >= sourceChannels) {
+            audio_.clear(channel, offset, first);
+            if (rest > 0)
+                audio_.clear(channel, 0, rest);
+            continue;
+        }
+
+        const auto* source = audio.getChannelPointer(static_cast<std::size_t>(channel));
+        audio_.copyFrom(channel, offset, source, first);
+        if (rest > 0)
+            audio_.copyFrom(channel, 0, source + first, rest);
+    }
+
+    audioWrite_.store(write + static_cast<std::uint64_t>(numSamples), std::memory_order_release);
+    return true;
+}
+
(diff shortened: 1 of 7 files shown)


magda/engine/io/RecordStream.cpp
```diff
@@ -0,0 +1,151 @@
...
+    auto worked = false;
+
+    if (audio_.getNumChannels() > 0) {
+        const auto read = audioRead_.load(std::memory_order_relaxed);
+        const auto write = audioWrite_.load(std::memory_order_acquire);
+
+        if (const auto available = write - read; available > 0) {
+            const auto take = contiguousAudioToDrain(read, available);
+            const auto offset = static_cast<std::size_t>(read & mask_);
+            const auto chunk = juce::dsp::AudioBlock<const float>(audio_).getSubBlock(
```
lucaromagnoli: **[P1] Avoid mutating the audio buffer from the consumer**

Although the block uses `const float`, passing mutable `audio_` selects JUCE’s mutable-buffer constructor, which calls `getArrayOfWritePointers()` and writes its non-atomic `isClear` flag. The producer’s `copyFrom()` concurrently writes that same flag, creating a data race during normal recording even when the sample regions do not overlap. The queue indices do not synchronize these concurrent metadata writes. Construct the block through `std::as_const(audio_)` to select the read-only overload.

— Codex

lucaromagnoli: Confirmed and fixed in 70c2b3a46. `getArrayOfWritePointers` sets `isClear`, and `copyFrom` on the audio thread writes the same non-atomic bool, so drain was racing it on every chunk. Now built through `std::as_const`, which takes the `getArrayOfReadPointers` overload.

Issue #2463, opened 2026-09-06
Recording slice 5: the recording pass reported without a poll


lucaromagnoli:
Slice 5 of #1895. Depends on slices 3 and 4, which produce the pass this reports.

## What the fork does today

`RecordingPreview` (`daw/audio/midi/RecordingNoteQueue.hpp`) is already the right model: one growing, beat-domain description of the pass in flight, living outside `ClipManager` because no clip exists until recording finishes. It carries the track, whether the target is the arrangement or a slot, `startBeat`, `currentLengthBeats`, the notes so far and a peak per tick for audio.

How it is filled is what does not survive. Notes arrive through `RecordingNoteQueue`, pushed by a `te::InputDeviceInstance::Consumer` that reads the transport position off an atomic at message-arrival time and drained by `drainRecordingNoteQueue()`; the length grows because `updatePreviews()` is called once a frame; audio peaks come from a metering buffer on the side. Three different clocks for one pass, none of them the one the samples were written against.

## Scope

- The pass as a tap, following `LaunchTap` and `ValueTap`: written by the audio thread, read by whoever draws, no poll and no frame timer deciding what the length is.
- Length from the transport the samples were stamped against, so what is drawn and what is written cannot disagree.
- Notes as they are captured, positioned the same way slice 4 positions them, so the preview and the finished clip do not resolve a beat differently.
- Audio peaks off the same signal the take is written from, rather than a separate metering path.
- Input meters on armed and monitoring tracks, which today come from `WaveInputDevice::levelMeasurer` and have no equivalent behind an input op.
- Both targets, arrangement and slot, on one reading; slice 6 fills in the slot half.

## Not in this slice

Drawing any of it. The UI keeps the shape it consumes today.

## Verification

Offline, driving a pass and reading the tap:

- the reported length matches the samples written for the pass, at several block sizes
- a note appears in the preview at the beat it later occupies in the finished clip
- the reading a block leaves is that block's own, never a mix of two
- a pass that never starts (armed, transport stopped) reports nothing rather than a zero-length pass
- an input meter reads the input a monitoring track is hearing

lucaromagnoli:
Done in #2527, merged into `dev/0.20.0` as 8629a05b.

[...]

PR #2475, opened 2026-09-07
Record an arrangement audio take, from armed to clip

### LINKED ISSUE #2461: Recording slice 3: an arrangement audio take, from armed to clip

lucaromagnoli:
Slice 3 of #1895, closes #2461. Where slices 1 and 2 meet: the input a monitoring track hears, through the queue a writer thread drains, into the files a clip is made of.

## What is here

**`TakeRecorder`** is one take, fed a block at a time on the audio thread. It reads the channels its track named and no others, so a mono input is a mono file and what a stereo track makes of it stays playback's business.

The rules it keeps are the model's rather than any engine's:

- **The head is corrected by the input's declared latency.** What arrives has already happened, and a take holds what happened, so a positive latency drops that many arrivals and a negative adjustment pads the head with that much silence. The clip lands on the record position rather than being moved back to meet the file.
- **A loop wrap opens the next pass a latency later**, for the same reason: the samples the wrap belongs to have not arrived yet. A negative adjustment cannot make a boundary retroactive, and the header says so.
- **A first pass that did not begin on the loop start is a lead-in, not a take.** `AudioTake` has no offset of its own, so a pass that cannot share the clip start would play in the wrong place. It is dropped and its file deleted once a wrap has proved the take loop-aligned.
- **The pass that plays is the last full one**, stepping back when the final one is under 95% of the longest, since only the last pass can be cut off mid-way.
- **A count-in is not part of the take**: it is time before the play position, and the take starts where it ends.

Nothing makes a model object on the audio thread. `finish()` is where passes become an `AudioClipModel`, on the thread that stopped the take, and a take that lost samples or could not write says so on the result rather than passing as a shorter clip.

**`TakeFileSink`** is the record thread's end of it: a file opened at a pass's first sample rather than when the take was armed, so a pass that captured nothing leaves nothing behind, and pass boundaries travel on a lane of their own because the queue carries only audio. It writes through `AudioFileSink`, which grew an `AudioBlock` overload and a `failed()` getter so the engine has one file-writing path rather than two.

The takes reach the callback through a feed beside the clips and outside every plan: arming a track compiles a plan, but starting a recording is not a structural edit and must not cost one.

## Verification

`tests/engine/test_take_recorder.cpp`, offline throughout: a synthetic input whose every sample says which arrival it is, a driven transport, no session and no thread.

- what the file holds is what the input carried, sample for sample
- the first recorded sample is the one under the playhead at record start, at a positive, a negative and no declared latency
- with a count-in set, the file starts at the count-in's end
- three loop passes are three files, all loop-length, and the clip is loop-aligned
- the active take is the last full pass, and the last-but-one when the final pass is cut short
- a lead-in leaves no file and no take
- a stop mid-pass keeps what was recorded up to it
- a mono input is recorded mono
- an overrun reaches the finished take rather than passing as a shorter file

Full Catch2 suite passes (3736 passed, 13 skipped); app and JUCE test targets build clean.

## Not in this slice

MIDI (#2462), what the app draws while the pass is running (#2463), session slots, and surviving a recompile mid-take.

### DIFF
--- a/magda/engine/io/TakeRecorder.cpp
+++ b/magda/engine/io/TakeRecorder.cpp
@@ -1,0 +1,238 @@
+#include "io/TakeRecorder.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <span>
+#include <utility>
+
+namespace magda::engine {
+
+namespace {
+
+/// A wrap anchors the cursor exactly on the loop start, so anything further off
+/// than this is a locate rather than a pass boundary.
+constexpr double kLoopStartTolerance = 1.0e-6;
+
+/// How short the final pass has to be to count as a stop rather than a take.
+constexpr double kFullPassFraction = 0.95;
+
+int takeChannels(const TakeRecorderSettings& settings) {
+    return std::max<int>(1, static_cast<int>(settings.channels.size()));
+}
+
+RenderContext fileContext(const RenderContext& context, const TakeRecorderSettings& settings) {
+    return {.sampleRate = context.sampleRate,
+            .maxBlockSize = context.maxBlockSize,
+            .numChannels = takeChannels(settings)};
+}
+
+RecordStreamSettings queueFor(const TakeRecorderSettings& settings) {
+    auto queue = settings.stream;
+    queue.numChannels = takeChannels(settings);
+    return queue;
+}
+
+bool atLoopStart(const BlockInfo& block, const LoopRange& loop) {
+    return loop.valid() && std::abs(block.beats.start - loop.startBeat) <= kLoopStartTolerance;
+}
+
+/// The last full pass, stepping back one if the final pass was cut short. Only
+/// the last one can be, which is what makes the rule this simple.
+std::size_t activeTake(std::span<const RecordedPass> passes) {
+    auto longest = 0.0;
+    for (const auto& pass : passes)
+        longest = std::max(longest, pass.durationSeconds);
+
+    const auto last = passes.size() - 1;
+    if (passes.size() > 1 && passes[last].durationSeconds < longest * kFullPassFraction)
+        return last - 1;
+
+    return last;
+}
+
+}  // namespace
+
+TakeRecorder::TakeRecorder(const LiveInputFeed& feed, const RenderContext& context,
+                           TakeRecorderSettings settings)
+    : settings_(std::move(settings)),
+      input_(feed, settings_.channels, settings_.latencySamples),
+      sink_(settings_.directory, settings_.name, settings_.file, fileContext(context, settings_)),
+      stream_(sink_, queueFor(settings_)) {
+    scratch_.setSize(takeChannels(settings_), std::max(1, context.maxBlockSize), false, true,
+                     false);
+
+    if (settings_.latencySamples < 0)
+        pad_.setSize(takeChannels(settings_), -settings_.latencySamples, false, true, false);
+
+    latencySeconds_ = context.sampleRate > 0.0
+                          ? static_cast<double>(settings_.latencySamples) / context.sampleRate
+                          : 0.0;
+}
+
+void TakeRecorder::capture(const BlockInfo& block, bool countingIn, const LoopRange& loop) {
+    if (state_ == State::stopped)
+        return;
+
+    // A count-in is time before the play position and a stop is where a take
+    // ends, so neither is part of one.
+    if (!block.playing || countingIn) {
+        if (state_ == State::rolling)
+            stop();
+
+        return;
+    }
+
+    if (state_ == State::waiting) {
+        start(block, loop);
+    } else if (!block.continuous) {
+        if (!atLoopStart(block, loop)) {
+            stop();
+            return;
+        }
+
+        openPass(loop);
+    }
+
+    write(block);
+
+    // What a take holds is what happened, and what happened at the end of this
+    // block arrives a latency later: the tail runs on by exactly as much as the
+    // head gave up.
+    lastBeat_ = block.beatAtTime(block.seconds.end - latencySeconds_);
+}
+
+void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop) {
+    state_ = State::rolling;
+    rolling_.store(true, std::memory_order_relaxed);
+
+    startBeat_ = block.beats.start;
+    startedAtLoopStart_ = atLoopStart(block, loop);
+    headDrop_ = std::max(0, settings_.latencySamples);
+
+    // A negative latency asks for samples from before the take was armed.
+    // Silence stands in for them, which is what leaves the first real sample
+    // where it happened.
+    if (const auto padded = pad_.getNumSamples(); padded > 0) {
+        stream_.writeAudio(juce::dsp::AudioBlock<const float>(std::as_const(pad_)), padded);
+        written_ += padded;
+        captured_.store(written_, std::memory_order_relaxed);
+    }
+}
+
+void TakeRecorder::openPass(const LoopRange& loop) {
+    wrapped_ = true;
+    loopStartBeat_ = loop.startBeat;
+    loopEndBeat_ = loop.endBeat;
+
+    // The samples the wrap belongs to are still a latency away. A negative one
+    // puts that boundary in the past, where the only thing left is to end the
+    // pass here and be late by the adjustment.
+    pendingSplit_ = std::max(0, settings_.latencySamples);
+}
+
+void TakeRecorder::stop() {
+    state_ = State::stopped;
+    rolling_.store(false, std::memory_order_relaxed);
+}
+
+void TakeRecorder::write(const BlockInfo& block) {
+    const auto numSamples = std::min(block.numSamples, scratch_.getNumSamples());
+    if (numSamples <= 0)
+        return;
+
+    input_.render(block, juce::dsp::AudioBlock<float>(scratch_).getSubBlock(
+                             0, static_cast<std::size_t>(numSamples)));
+
+    const auto captured = juce::dsp::AudioBlock<const float>(std::as_const(scratch_));
+
+    auto offset = 0;
+    if (headDrop_ > 0) {
+        const auto dropped = std::min(headDrop_, numSamples);
+        headDrop_ -= dropped;
+        offset += dropped;
+    }
+
+    for (;;) {
+        if (pendingSplit_ == 0) {
+            sink_.markPassEnd(written_);
+            pendingSplit_ = -1;
+        }
+
+        if (offset >= numSamples)
+            break;
+
+        auto chunk = numSamples - offset;
+        if (pendingSplit_ > 0)
+            chunk = std::min(chunk, pendingSplit_);
+
+        stream_.writeAudio(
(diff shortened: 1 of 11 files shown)


magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,238 @@
...
+    placeClip(result_);
+    return result_;
+}
+
+void TakeRecorder::placeClip(RecordedTake& take) const {
+    // Loop-aligned the moment a wrap happened: every pass that survived covers
+    // the loop, so that is where the clip goes and how long it is.
+    if (wrapped_) {
+        take.startBeat = loopStartBeat_;
+        take.lengthBeats = std::max(0.0, loopEndBeat_ - loopStartBeat_);
```
lucaromagnoli, line 246: **P2 — Place the clip according to the pass actually captured**

`wrapped_` becomes true when the transport wraps, before the latency-delayed file boundary is reached. If recording starts at beat 1 inside a [0, 2] loop and stops 64 samples after the wrap with 128 samples of input latency (8 kHz, 120 bpm), only the lead-in has been captured. `finish()` retains that single pass, but this branch places it at beat 0 with length 2 beats, so its audio plays at the wrong timeline position. This reproduces in an isolated test. Please base loop alignment on the captured pass boundary and preserve the original position for a retained lead-in, or discard it consistently.

— Codex

magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+void TakeRecorder::openPass(const LoopRange& loop) {
+    wrapped_ = true;
+    loopStartBeat_ = loop.startBeat;
+    loopEndBeat_ = loop.endBeat;
+
+    // The samples the wrap belongs to are still a latency away. A negative one
+    // puts that boundary in the past, where the only thing left is to end the
+    // pass here and be late by the adjustment.
+    pendingSplit_ = std::max(0, settings_.latencySamples);
```
lucaromagnoli: **P2 — Preserve pending boundaries across subsequent loop wraps**

When the loop is shorter than the declared input latency, each wrap replaces `pendingSplit_` before it can reach zero. No pass boundary is ever emitted. With a 64-sample loop and 128-sample latency, an isolated test recording ten iterations produces one 512-sample file with no alternative takes instead of eight 64-sample passes. Please retain outstanding delayed boundaries, for example as absolute arrival positions, so a later wrap cannot erase an earlier one.

— Codex

magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,238 @@
...
+    auto offset = 0;
+    if (headDrop_ > 0) {
+        const auto dropped = std::min(headDrop_, numSamples);
+        headDrop_ -= dropped;
+        offset += dropped;
+    }
+
+    for (;;) {
+        if (pendingSplit_ == 0) {
+            sink_.markPassEnd(written_);
```
lucaromagnoli: **P2 — Report a rejected pass boundary on the finished take**

`markPassEnd()` returns false when its 64-entry queue is full, but the result is ignored and the pending split is cleared. The audio queue need not be full: recording 70 passes of 64 samples before draining uses only 4,480 samples of its default 65,536-sample capacity. An isolated test produces 65 passes instead of 70 while reporting `samplesLost == 0` and `failed == false`; the remaining passes are silently concatenated. Please propagate boundary overflow to the finished result or prevent it from losing pass boundaries.

— Codex

magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    // Where the pass ends, in the take's own samples: the timeline it has
+    // covered so far. The latency does not enter it -- the head correction
+    // already put sample i of the take at i samples past its start -- and the
+    // boundary is handed over now rather than counted down to, so a loop
+    // shorter than one cannot overtake the last one still pending.
+    if (firstBoundary_ < 0)
+        firstBoundary_ = arrivals_;
+
+    if (!sink_.markPassEnd(arrivals_))
```
lucaromagnoli: **P2 — Keep negative-latency boundaries independent of disk timing**

With negative latency, the head padding puts the queued audio ahead of `arrivals_`, so by the time this boundary is published the sink may already have written past it. The sink can only close at its current position; it cannot split audio already written. An isolated reproduction at 8 kHz with a two-beat loop at 120 BPM and -128 samples of latency gives a first pass of 8,000 samples when draining only at finish, but 8,128 samples when draining after each 64-sample callback. Identical input therefore produces different pass contents depending on disk scheduling. Please ensure the boundary cannot fall behind already-published audio, preserving a consistent policy for negative latency.

— Codex

magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+    const auto kept = numSamples - offset;
+    if (kept <= 0)
+        return;
+
+    stream_.writeAudio(
+        captured.getSubBlock(static_cast<std::size_t>(offset), static_cast<std::size_t>(kept)),
+        kept);
+    written_ += kept;
+    capturedBeats_ += beatsFor(block, kept);
```
lucaromagnoli: **P2 — Measure clip length at the captured audio’s timeline position**

`beatsFor(block, kept)` counts beats at input arrival time, but latency correction places those samples earlier on the timeline. These durations differ across a tempo change. In an isolated reproduction at 8 kHz, starting at beat 0 with 128-sample latency, stepping from 120 to 240 BPM at beat 1, and capturing 8,000 arrivals produces a 7,872-sample file. The file covers 2.936 beats from the clip start, but `lengthBeats` is 2.968 because this accumulation subtracts the dropped head at the initial tempo instead of measuring the compensated tail through the tempo map. Please derive the length from the captured audio’s timeline interval rather than summing arrival-block beat durations.

— Codex

magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,267 @@
...
+    // Where the loop ends, and when the samples belonging to it are delivered:
+    // a latency earlier, which is what makes this block the last chance to name
+    // the boundary before its audio is queued.
+    const auto untilEnd = std::llround(block.offsetForBeat(loop.endBeat));
+    const auto delivered = untilEnd + settings_.latencySamples;
+
+    if (delivered < 0 || delivered >= block.numSamples)
+        return;
+
+    markBoundary(arrivals_ + untilEnd, loop);
```
lucaromagnoli: **P2 — Handle loop changes after predicting a pass boundary**

This commits a pass boundary before the transport actually reaches the loop end. If looping is disabled after the prediction but before the wrap, that boundary remains queued and the uninterrupted recording is split into alternative takes and placed at the old loop range. I reproduced this at 8 kHz and 120 BPM with -128-sample latency: enable a [0, 2] loop, record 7,936 arrivals, disable looping, then record another 8,064 arrivals. No wrap occurs, but the result has alternative takes and a length of 2 beats instead of one continuous 16,128-sample file spanning 4.032 beats. Please define and enforce a policy for loop changes after prediction, so a boundary for a wrap that never happens cannot silently turn continuous audio into loop takes.

— Codex

magda/engine/io/TakeRecorder.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+/// The last full pass, stepping back one if the final pass was cut short. Only
+/// the last one can be, which is what makes the rule this simple.
+std::size_t activeTake(std::span<const RecordedPass> passes) {
+    auto longest = 0.0;
+    for (const auto& pass : passes)
+        longest = std::max(longest, pass.durationSeconds);
+
+    const auto last = passes.size() - 1;
+    if (passes.size() > 1 && passes[last].durationSeconds < longest * kFullPassFraction)
```
lucaromagnoli: **P2 — Exclude first-pass padding from the full-pass comparison**

The negative-latency policy deliberately makes the first pass longer than the loop, but `activeTake()` still uses the longest file as the reference for deciding whether the final pass was cut short. Once the padding exceeds roughly 5.3% of the loop length, a complete final pass is rejected. An isolated reproduction at 8 kHz and 120 BPM with a two-beat loop, -512-sample latency, and three complete iterations produces pass lengths of 8,512, 8,000, and 8,000 samples; `currentTakeIndex` is 1 and `file` selects take_2.wav instead of the complete final take_3.wav. Please assess completeness against the loop duration or exclude the first pass’s extra padding from the reference duration.

— Codex

PR #2478, opened 2026-09-07
Declare a device's sidechain as a port, and model its source

### LINKED ISSUE #2329: Declare a device's sidechain as a port, and model its source
Sidechain today is two facts and a convention. `DeviceInfo::canSidechain` is inferred (Faust: `inputCount > 2` in `FaustPlugin.cpp`; compiled: `wantsSidechain()`; external: the bus layout). `SidechainConfig` is `{None|Audio|MIDI, sourceTrackId}` on `DeviceInfo` and `RackInfo`. And the key reaches a MagdaDevice as a channel offset into its own buffer (`DeviceProcessContext::sidechainInputChannel`), because that is how the compiled dynamics DSPs are written.

The native plan is ahead of the model: a `Device` op already has three input slots `[audio, MIDI, sidechain]`, and `DeviceBlock::sidechain` is its own block, zero channels when unconnected. Modifiers are ahead of devices too: `ModTapPoint` lets each modifier choose pre-FX or post-fader on its source, while a device sidechain always taps post-fader, pre-mute (`trackSidechainTap_`).

## The decision: a sidechain is a declared port with a modelled source

**Declared, not inferred.** `DeviceProperties` says what the device wants: `sidechain = {None, Audio(channels), Midi}`. Faust devices declare it in the dsp rather than by having more inputs than outputs; compiled devices already have `wantsSidechain()`; externals keep reading the bus layout; ported MagdaDevices set it in their properties. `DeviceInfo::canSidechain` becomes a projection of that declaration, populated by the processor with the rest of the device's facts.

**Delivered as a port.** A MagdaDevice receives the key as `DeviceBlock::sidechain`, the plan's slot, not as extra channels of its own buffer. Compiled DSPs that read the key from channels 3 and 4 get it copied there by `MagdaCompiledEffect`, one place instead of a convention every device has to know. The TE host does the same copy in `MagdaDevicePlugin`.

**A source, not a track id.** `SidechainConfig` grows:

- `tapPoint`, reusing `ModTapPoint` (PreFx / PostFader), default PostFader so nothing changes for existing projects. These are the two points both engines have; the note in `ModInfo.hpp` about not inventing a third stands.
- `gainDb` and `listen`. Gain is a `Gain` op on the sidechain edge. Listen replaces the slot's output with the key it was handed, a value on the process op like delta and mute, so it never rebuilds anything. Filtering stays in the DSP.
- Later, without changing the shape: a source variant for a hardware input or a rack chain. Not in this issue. The point is that a source is a struct, so adding a kind of source is not a rewrite.

`RackInfo::sidechain` is the trigger and follower source for a rack's modifiers, a different thing that happens to share the struct. Give it its own type or leave it as is, but the device sidechain's new fields do not leak into it.

## Consequences

- The compiler resolves `tapPoint` against `trackSidechainTap_` and `trackTriggerTap_`, which it already keeps separately.
- Serialisation: new fields with defaults; old projects read as PostFader, 0 dB, listen off.
- Undo and MCP: sidechain edits go through the existing sidechain commands; the MCP device surface exposes the new fields and the declared port.

Related: #1929 (Faust extra-channel sidechain), #1591 (Sidechain device), #52 (original sidechain routing).

lucaromagnoli:
Sidechain was two facts and a convention: `DeviceInfo::canSidechain` inferred from a channel count, `SidechainConfig` a `{type, sourceTrackId}` pair, and the key reaching a `MagdaDevice` as extra channels of its own buffer. The native plan was already ahead of it — a `Device` op has three input slots and `DeviceBlock::sidechain` is its own block — and modifiers were ahead of devices, choosing their tap point per listener while a device sidechain always tapped post-fader.

Closes #2329. (Merges to `dev/0.20.0`, so the issue is closed by hand.)

## Declared, not inferred

`magda::SidechainPort` (`{None | Audio(channels) | MIDI}`) is what a device states in `DeviceProperties::sidechain`, and it replaces the `canSidechain` bool everywhere.

- **Faust** declares it in the dsp: `declare magda_sidechain "audio";`. The key's width is the inputs past the outputs, and the declaration is frozen into `FaustState` at compile time so `properties()` never reads the editable source off another thread. Extra inputs without the declaration are the device's own — there's a test for that, and the Faust agent's prompt says so now.
- **Compiled devices** override `MagdaCompiledEffect::sidechainPort()` instead of `wantsSidechain()`.
- **Externals** still read the bus layout, now for the key's width as well as its existence.
- `DeviceInfo::canSidechain` becomes `DeviceInfo::sidechainPort`, populated by the processor with the rest of the device's facts.

Ring Mod gains the declaration it never had. That also fixes its fork wiring: with only two declared inputs, `guessSidechainRouting` took the two-input branch and summed the track's own stereo into one channel.

## Delivered as a port

`DeviceProcessContext::sidechainInputChannel` is gone; `sidechain` (read-only channel pointers) and `numSidechainChannels` replace it, indexed from `startSample` like the audio.

- `MagdaCompiledEffect` copies the key into the dsp inputs after the device's own, so no dsp knows where a host keeps one.
- `FaustPlugin` does the same for the inputs its source declared.
- `TracktionMagdaDevicePlugin` splits the fork's buffer once, off a live channel split cached in two atomics and refreshed wherever the fork asks the plugin about its shape — the runtime Faust device recompiles into a different one.
- `EngineMagdaDevice` hands `DeviceBlock::sidechain` straight through: a port on both sides, with nothing in between deciding where in a buffer a key belongs.

## A source, not a track id

`SidechainConfig` grows three fields:

- **`tapPoint`**, reusing `ModTapPoint`. The compiler resolves it against `trackTriggerTap_` and `trackSidechainTap_`, which it already keeps apart — pre-FX is what the source played, post-fader is what it sends, and riding the source fader is the whole difference.
- **`gainDb`**, a `Gain` op (`OpRole::DeviceSidechainGain`) on the sidechain edge, emitted wherever a key is connected so moving the trim is a value and never a recompile. Pinned by a fingerprint test.
- **`listen`**, an `OpValue` flag like the delta solo: the device still runs, so its own metering and smoothing carry on, and what leaves the slot is the key it was handed.

Serialisation writes them only when they are not the defaults, and reads a legacy `canSidechain` as a mono port; an old project opens as post-fader, 0 dB, listen off, which is what it always sounded like. `RackInfo::sidechain` keeps `type` and `sourceTrackId` only — its serializer cannot carry the new fields.

## Reach

The slot's sidechain menu offers the tap point, a trim submenu and a listen toggle, through new `TrackManager` setters that share one walk with `setSidechainSource`. `DeviceDto` reports the declared port and the whole source over the remote API.

## Not in here

- No MCP **write** operation for sidechain. None exists today, and the issue says edits go through the existing sidechain commands.
- The TE leg still taps post-fader at unity. The new fields are native-engine behaviour, which is why their defaults are exactly what keeps null-diff parity.

## Testing

`make test` — 3750 passed, 13 skipped, 0 failed. `make test-juce` — all passed.

New coverage: the pre-FX tap and the trim's op identity in the compiler; the trim and listen in the executor (listen verified negatively — reverting the resolution makes it fail); a `sidechain-prefx` plan golden beside the existing `sidechain` one; the round trip of the new fields and of a pre-#2329 project; `SidechainPort::Kind` in the persisted enum pins; and, in the Faust suite, that four inputs without a declaration take no key.

### DIFF
--- a/magda/daw/ui/components/chain/slot/DeviceSlotSidechainControls.cpp
+++ b/magda/daw/ui/components/chain/slot/DeviceSlotSidechainControls.cpp
@@ -18,6 +18,30 @@ struct TrackEntry {
     juce::String name;
 };
 
+/// One thing the menu can do, looked up by item id rather than decoded out of
+/// an id range. Two unbounded lists of tracks and a handful of fixed actions
+/// cannot share one id space: with enough tracks the source ids run into the
+/// action ids, and picking a track changes the tap point instead.
+struct MenuAction {
+    enum class Kind { Clear, SetSource, SetTapPoint, SetTrim, ToggleListen };
+
+    Kind kind = Kind::Clear;
+    magda::TrackId trackId = magda::INVALID_TRACK_ID;
+    magda::SidechainConfig::Type type = magda::SidechainConfig::Type::None;
+    magda::ModTapPoint tapPoint = magda::ModTapPoint::PostFader;
+    float gainDb = 0.0f;
+    bool listen = false;
+};
+
+using MenuActions = std::vector<MenuAction>;
+
+/// Adds @p action and returns the item id that runs it. Ids are 1-based: a
+/// PopupMenu reports 0 for "nothing was chosen".
+int addAction(MenuActions& actions, MenuAction action) {
+    actions.push_back(action);
+    return static_cast<int>(actions.size());
+}
+
 /// The trim steps the menu offers. A menu cannot drag a value, and a key's trim
 /// is a matching decision rather than a ride, so a handful of steps is what it
 /// is for; anything finer belongs to whoever automates it (#2329).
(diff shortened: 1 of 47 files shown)


magda/engine/exec/PlanValues.cpp
```diff
@@ -389,6 +389,33 @@ void Resolver::resolveOp(OpId id, OpValue& value) {
...
+        }
+
+        case OpRole::DeviceProcess: {
+            const auto* device = findDevice(*track, key);
+            if (device == nullptr) {
+                report(id, "no device " + std::to_string(key.deviceId) +
+                               " in the model, monitoring its own output");
+                break;
+            }
+            value.listensToSidechain = device->sidechain.listen && device->sidechain.isActive();
```
lucaromagnoli: [P2] Only enable sidechain listening for an audio source

On a device that accepts both audio and MIDI sources, enable Listen on an audio sidechain and then select a MIDI source. setSidechainSource preserves the listen flag, and isActive() remains true for MIDI, so this expression still enables listening. The compiler supplies no audio sidechain block for a MIDI source, so PlanExecutor clears the device output after processing. The MIDI source menu also hides the Listen toggle, making the silence difficult to undo. Require sidechain.type == SidechainConfig::Type::Audio when resolving this flag (or clear it when switching source types).

magda/daw/ui/components/chain/slot/DeviceSlotSidechainControls.cpp
```diff
@@ -69,28 +79,62 @@ void showDeviceSlotSidechainMenu(const magda::DeviceInfo& device,
...
+         onSidechainChanged = std::move(onSidechainChanged)](int result) {
             if (result == 0)
                 return;
 
+            auto& trackManager = magda::TrackManager::getInstance();
+
             if (result == 1) {
-                magda::TrackManager::getInstance().clearSidechain(deviceId);
+                trackManager.clearSidechain(deviceId);
+            } else if (result == 300 || result == 301) {
```
lucaromagnoli: [P2] Keep action IDs separate from MIDI source IDs

MIDI source entries receive IDs starting at 200 with no upper bound. With at least 101 available source tracks, the 101st entry has ID 300, so selecting it now changes the tap point instead of selecting that MIDI source. IDs 301, 310, and 400–406 collide similarly with later entries. These branches intercept the source IDs even when the audio-only Key controls are hidden. Allocate action IDs outside the generated source ranges, or use separate callbacks for these actions.

magda/engine/exec/PlanValues.cpp
lucaromagnoli: Confirmed and fixed in 8e5bd5c6. Reproduced it as a test first: with `listen` set and a MIDI source, the device output went to 0.0.

The guard is now `SidechainConfig::listensToKey()` — `listen && type == Audio && sourceTrackId != INVALID` — read at resolution rather than cleared when the source type changes. That way no path into the model can produce the silence (MCP, deserialization, undo), and switching audio → MIDI → audio remembers the setting instead of dropping it. `tests/engine/test_render_plan_executor.cpp` covers it; reverting the guard fails it with `0.0f == Approx(0.5)`.

magda/daw/ui/components/chain/slot/DeviceSlotSidechainControls.cpp
lucaromagnoli: Confirmed and fixed in 8e5bd5c6. Worth noting the audio list had the same defect before this branch: it starts at 100 and the handler reads `result >= 100 && result < 200`, so at 100 sources the 101st audio entry is id 200 and gets applied as MIDI source 0.

Rather than move the action ids, the id space is gone: each item is appended to a `std::vector<MenuAction>` as the menu is built and its id is that index + 1, so the callback looks the action up instead of decoding a range. No unbounded list can run into anything, and it fixes the pre-existing audio/MIDI collision with the same change.

PR #2483, opened 2026-09-07
Qualify auto on pointers and const-bind read-only references

lucaromagnoli:
readability-qualified-auto across `magda/`. 73 files, no behaviour change.

`auto` deducing a pointer now says `auto*`, so a reader can tell a pointer from
a value without going to the initialiser, and references that are never written
bind as `const auto&`.

Built clean locally on top of #2482.

clang-tidy formatted its own edits (`-format -style=file`), so the pointers came
out matching the project's `PointerAlignment: Left` rather than being reflowed
by the commit hook afterwards.

### DIFF
--- a/magda/daw/audio/plugins/MidiStrumPlugin.cpp
+++ b/magda/daw/audio/plugins/MidiStrumPlugin.cpp
@@ -302,8 +302,8 @@ void MidiStrumPlugin::scheduleStrum() {
     // between them, and one hung voice each (#2363).
     scheduleReleaseAll();
 
-    auto* begin = ordered_.begin();
-    auto* end = begin + heldCount_;
+    auto begin = ordered_.begin();
+    auto end = begin + heldCount_;
     std::copy(held_.begin(), held_.begin() + heldCount_, begin);
 
     const int ord = displayIndex(kOrder);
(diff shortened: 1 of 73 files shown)


magda/daw/audio/plugins/MidiStrumPlugin.cpp
```diff
@@ -302,8 +302,8 @@ void MidiStrumPlugin::scheduleStrum() {
     // between them, and one hung voice each (#2363).
     scheduleReleaseAll();
 
-    auto begin = ordered_.begin();
-    auto end = begin + heldCount_;
+    auto* begin = ordered_.begin();
```
lucaromagnoli: **P1 — Preserve iterator deduction for std::array**

`ordered_` is a `std::array`, and MSVC represents its iterator as `std::_Array_iterator`, not a raw pointer. Changing this declaration to `auto*` therefore breaks the Windows build: the CI run reports C3535/C2440 at this line because it cannot deduce or initialize a pointer from that iterator. This remains present in the current head after the duplicate-const cleanup.

Restore `auto` for both `begin` and `end`, or initialize the pointer with `ordered_.data()`.

— Codex

Issue #2485, opened 2026-09-08
Implement track playback mode in the native engine


lucaromagnoli:
The native engine has no concept of `TrackPlaybackMode`. `grep -rn 'playbackMode' magda/engine` returns nothing.

TE gates arrangement playback on `AudioTrack::playSlotClips`, which `ClipSynchronizer::syncPlaybackModeToEngine` writes from `TrackInfo::playbackMode` (`magda/daw/audio/session/ClipSynchronizer.cpp:319-327`). A track in Session mode plays its launcher slots, and TE's `ArrangerLauncherSwitchingNode` drops its arrangement clips. The native engine instead decides arranger vs launcher purely from live launch handles, so a Session-mode track with nothing launched plays its arrangement.

Both legs are self-consistent. They are not the same behaviour, and the persisted per-track mode is a real feature with UI behind it (the orange transport button), so the native engine is the side that is missing something.

## How it surfaces

`project.retrospect` in the real-project corpus (`tests/engine/test_real_project_corpus_juce.cpp`). Its two unmuted tracks were saved with `"playbackMode": 1`:

```
project.retrospect: a render did not sound: native 4.0 dBFS, incumbent -inf dBFS, floor -40.0 dBFS
```

The TE render is digitally silent (`peak=0` over the full 230087 samples); native renders normally.

Two things it is not:

- Not a regression. Reproduced with `magda/` reverted to `dev/0.20.0`.
- Not the hosted plugins. Those tracks host two Retrospect VST3 instances, but bypassing them changes nothing, and the plugins report clean (`missing=0 disabled=0 err='' in=2 out=2 enabled=1`). TE drops the clips upstream of every plugin.

It also only shows up locally: CI has no Retrospect installed, so the corpus skips the case and the gap stays invisible there.

## What closes it

The native engine reads `TrackInfo::playbackMode` and gates its arranger the way TE does, so a Session-mode track with no launched clip renders silence rather than its arrangement.

Until then `project.retrospect` cannot hold, and belongs on the corpus calibration list with this issue named against it rather than left failing.

Part of #1882.

PR #2488, opened 2026-09-08
Gate the native arranger on the track's playback mode (#2485)

lucaromagnoli:
Closes #2485.

## What

The native engine had no concept of `TrackPlaybackMode`. `sectionHold` decided how much of a block the arrangement sounds from the launch handles alone, so a Session-mode track with nothing launched played its arrangement where the fork's `playSlotClips` drops it.

The rule now: a track's arrangement sounds a block only when the track is in Arrangement mode **and** no launch handle holds the section. Session mode silences it launched or not. Launched slots play through their handles exactly as before.

- The mode rides in the clip snapshot (`ClipLane` → `TrackClipPlayback`, printed by the dump, `kVersion` 2), so a flip applies at a block boundary.
- A flip has no handle to say what the block before was, so each arrangement source keeps the mode it last rendered under. Without that a Session-mode track fires its lost edge every callback, restarting the de-click and resending note-offs.
- Both sources apply the hold whether or not a handle feed exists; a null table is an empty handle range and the mode still decides.

## Corpus

`project.retrospect` was saved with two tracks in Session mode **and** active session clips. The app relaunches those on play; the corpus never did, which is why the incumbent rendered silence and native rendered the arrangement. `loadFixture` now derives a launch at the render's start beat from each track's `activeSessionClipId`, and refuses a saved active clip that names no session clip. `project.retrospect` comes off the calibration list #2486 put it on against this issue.

A code-built case, `session.mode.holds.slot`, pins the retrospect shape beside an audible neighbour so the runner has a comparison rather than two silences.

## Tests

- `test_session_playback.cpp`: Session mode with nothing launched sounds neither section (handles published and not); flipping to Session mid-play carries the arrangement down and owes one note-off, with the block after fully silent; flipping back ramps in and re-chases; a launched slot on a Session-mode track plays and releasing it leaves the track silent. Audio and MIDI.
- `test_clip_snapshot.cpp`: the lane's mode reaches the track and the dump.
- Negative check: with the `sectionHold` change reverted, the new engine cases and the new corpus case fail.
- `magda_tests` and `make test-juce` green on this tree.

**Not verified here:** Retrospect is not scanned on this Mac, so the corpus skipped `project.retrospect`. It needs one `make test-juce` on the Mac that has it; the expected outcome is the case rendering on both legs and the calibration list holding in both directions.

https://claude.ai/code/session_012q17kwk7LKPovRoPtrBWjX

### DIFF
--- a/magda/daw/core/TrackInfo.hpp
+++ b/magda/daw/core/TrackInfo.hpp
@@ -34,8 +34,6 @@ struct SendInfo {
 
 enum class InputMonitorMode { Off, In, Auto };
 
-enum class TrackPlaybackMode { Arrangement, Session };
-
 /**
  * @brief Track data structure containing all track properties
  */
--- a/magda/daw/core/TrackTypes.hpp
+++ b/magda/daw/core/TrackTypes.hpp
@@ -268,4 +268,8 @@ inline bool canHaveChildren(TrackType type) {
     return traitsOf(type).canHaveChildren;
 }
 
+/// Which of a track's two bodies of material plays: its arrangement, or its
+/// session slots. Persisted by ordinal; the engine gates on it (#2485).
+enum class TrackPlaybackMode { Arrangement, Session };
+
 }  // namespace magda
--- a/magda/engine/clip/ClipAudioSource.cpp
+++ b/magda/engine/clip/ClipAudioSource.cpp
@@ -151,17 +151,30 @@ void ClipAudioSource::render(const BlockInfo& block, juce::dsp::AudioBlock<float
 
     renderMaterial(block, out);
 
-    // Only a source that shares its track with a session has a section to lose
-    // it to. One built without the feed is the whole track.
-    if (section_ == Section::Arrangement && handles_ != nullptr)
+    // Every Arrangement source has a mode to check now (#2485), handles or not.
+    if (section_ == Section::Arrangement)
         applySectionHold(out);
 }
 
 void ClipAudioSource::applySectionHold(juce::dsp::AudioBlock<float> out) {
     const auto numSamples = static_cast<int>(out.getNumSamples());
 
-    const LaunchHandleFeed::Reader handles(*handles_);
-    const auto hold = sectionHold(handles.get(), trackId_, numSamples);
+    // A track missing from the snapshot is Arrangement: nothing to silence.
+    bool session = false;
+    const ClipSnapshotFeed::Reader snapshot(clips_);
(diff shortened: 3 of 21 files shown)


lucaromagnoli:
Found one correctness issue:

**[P2] Keep one snapshot for the entire audio render** — `magda/engine/clip/ClipAudioSource.cpp:164`. `renderMaterial()` releases its snapshot before `applySectionHold()` acquires another. A concurrent publication can therefore render material from one snapshot and gate it using another snapshot’s mode. For example, switching from Session to Arrangement while replacing arrangement content can expose the old content for one block. Acquire the snapshot once in `render()` and pass the track or mode through both stages. MIDI already uses one snapshot throughout its render.

Architecturally, the shared `sectionHold()` function is a good place for the decision, and keeping mode changes independent of render-plan compilation makes sense. Two improvements deserve attention:

- **Block consistency:** each audio and MIDI source independently reads the snapshot and remembers the previous mode. This does not guarantee they observe a transition in the same block. Longer term, resolve track playback state once per block and share it across both sources.
- **Dependency scope:** importing all of `TrackInfo.hpp` into `ClipSnapshot.hpp` just for `TrackPlaybackMode` pulls track chains and view settings into the clip playback interface. Move the enum into a lightweight shared types header.

Validation: the existing test binary passed **244 clip test cases, 69,305 assertions**. Those tests publish between renders, so they do not cover the snapshot race. Retrospect plugin parity remains unverified.

lucaromagnoli:
Addressed in d237e623.

- **One snapshot per audio render (P2):** `ClipAudioSource::render()` now acquires the snapshot once and passes the track to both `renderMaterial()` and `applySectionHold()`, so the material and the mode that gates it always come from the same publish. The MIDI source already worked this way. Structural rather than tested: a race between two acquisitions inside one render is not reproducible deterministically, and there is no longer a second acquisition to race.
- **Dependency scope:** `TrackPlaybackMode` moved into `core/TrackTypes.hpp`, which has no includes of its own. `ClipSnapshot.hpp` and `ClipSnapshotCompiler.hpp` include that instead of `TrackInfo.hpp`.
- **Block consistency across a track's audio and MIDI:** a property of per-source snapshot reads that predates this PR and that the mode inherits. Resolving playback state once per block and sharing it is a new mechanism, so it is #2490 rather than part of this change.

Retrospect parity stays unverified on this Mac, as the description says; it needs one `make test-juce` on the Mac that has it scanned.

PR #2489, opened 2026-09-08
Return the condition, clamp with std::min and std::max

lucaromagnoli:
The fifth clang-tidy batch: `readability-simplify-boolean-expr`, `readability-use-std-min-max` and `readability-inconsistent-declaration-parameter-name`. An `if` that returns true or false returns the condition instead, a guarded assignment becomes a clamp, and a declaration's parameter names match the definition's.

Follows #2482, #2483, #2484 and #2487. 60 files.

## The clamps are not the identity they look like

An earlier version of this description claimed they were. They are not, and review caught it.

`std::min(a, b)` is `b < a ? b : a`, so argument order decides what happens to a NaN. The fixer wrote the incoming value first, which propagates NaN into the accumulator where the guarded assignment had dropped it. Over `[-0.8, 0.9, NaN, 0.1]` a peak pair came out `[0.1, 0.1]` instead of `[-0.8, 0.9]` — a persisted waveform cache losing real peaks, in `WaveformPeakCache`, `AudioThumbnailManager` and `SamplerUI`.

Fixed in 01448c292 by putting the accumulator first at all 21 sites, not only the float ones. The comparison is then the one the `if` performed whatever the type, rather than a rule that holds until something feeds it a NaN.

The boolean simplifications also include De Morgan on negated conjunctions, which changes how a line reads without changing what it does:

```
-if (!(loopEnabled && loopEndSample > loopStartSample))
+if (!loopEnabled || loopEndSample <= loopStartSample)
```

## Scope

`magda/engine/` is excluded. The native engine is being worked for #2485 and a mechanical rewrite of 100+ files underneath that costs more than it saves. It gets its own sweep once #2485 lands.

## Verification

- Catch2: 3772 cases, 998426 assertions, 0 failures.
- `magda_juce_tests`: all 453 groups pass, corpus included. This is the suite no earlier batch in this rollout was run against, which is how #2487's use-after-free reached review.
- Artifact scan (nested `std::move`, stacked `const`, `&&&`, repeated `static`) clean on all four counts.
- No build warning lands on any line this batch changed.
- Every new `std::min`/`std::max` checked for mixed argument types, since `size_t` is `unsigned long` here and `unsigned long long` on MSVC. The only two pairing distinct expressions are `size_t`/`size_t` and `int`/`int`.

### DIFF
--- a/magda/daw/audio/WaveformPeakCache.cpp
+++ b/magda/daw/audio/WaveformPeakCache.cpp
@@ -150,8 +150,8 @@ std::unique_ptr<WaveformPeakCache> WaveformPeakCache::computeAndWrite(
                 float maxVal = -1.0f;
                 for (int k = s; k < bucketEnd; ++k) {
                     const float v = samples[k];
-                    minVal = std::min(v, minVal);
-                    maxVal = std::max(v, maxVal);
+                    minVal = std::min(minVal, v);
+                    maxVal = std::max(maxVal, v);
                 }
                 // Empty bucket (only at EOF) collapses to silence.
                 if (minVal > maxVal) {
@@ -239,8 +239,8 @@ WaveformPeakCache::MinMax WaveformPeakCache::getMinMaxForRange(int channel, juce
     for (juce::int64 b = firstBucket; b <= lastBucket; ++b) {
         const auto bMin = chPeaks[static_cast<size_t>(b * 2)];
         const auto bMax = chPeaks[static_cast<size_t>(b * 2 + 1)];
-        minI = std::min(bMin, minI);
-        maxI = std::max(bMax, maxI);
+        minI = std::min(minI, bMin);
+        maxI = std::max(maxI, bMax);
     }
 
     if (minI > maxI) {
--- a/magda/daw/audio/WaveformPeakCache.cpp
+++ b/magda/daw/audio/WaveformPeakCache.cpp
@@ -150,10 +150,8 @@ std::unique_ptr<WaveformPeakCache> WaveformPeakCache::computeAndWrite(
                 float maxVal = -1.0f;
                 for (int k = s; k < bucketEnd; ++k) {
                     const float v = samples[k];
-                    if (v < minVal)
-                        minVal = v;
-                    if (v > maxVal)
-                        maxVal = v;
+                    minVal = std::min(v, minVal);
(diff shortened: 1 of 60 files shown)


magda/daw/audio/WaveformPeakCache.cpp
```diff
@@ -150,10 +150,8 @@ std::unique_ptr<WaveformPeakCache> WaveformPeakCache::computeAndWrite(
                 float maxVal = -1.0f;
                 for (int k = s; k < bucketEnd; ++k) {
                     const float v = samples[k];
-                    if (v < minVal)
-                        minVal = v;
-                    if (v > maxVal)
-                        maxVal = v;
+                    minVal = std::min(v, minVal);
+                    maxVal = std::max(v, maxVal);
```
lucaromagnoli: **P2 — Preserve NaN handling in waveform peak accumulation**

Putting the incoming sample first changes behavior for floating-point audio containing NaNs: the old guarded assignments ignored those samples, whereas these calls replace both accumulated extrema with NaN and the next finite sample resets both peaks. A standalone C++ reproduction with `[-0.8, 0.9, NaN, 0.1]` yields `[0.1, 0.1]` here instead of the previous `[-0.8, 0.9]`, so the persisted waveform cache loses real peaks. JUCE's floating-point WAV reader passes these values through without sanitizing them. Please use `std::min(minVal, v)` and `std::max(maxVal, v)` to preserve the previous behavior, and apply the same accumulator-first ordering to the sample reductions in `AudioThumbnailManager.cpp` and `SamplerUI.cpp`.

— Codex

lucaromagnoli: Confirmed and fixed in 01448c292.

Reproduced your case standalone: over `[-0.8, 0.9, NaN, 0.1]` the value-first form gives `[0.1, 0.1]` and the accumulator-first form gives `[-0.8, 0.9]`, matching the guarded assignment.

I put the accumulator first at all 21 sites rather than only the three you named, integer ones included. The float sites are where NaN can reach today, but the ordering is only correct by accident on the others, and a later batch would have to re-derive which is which. `WaveformPeakCache`, `AudioThumbnailManager` and `SamplerUI` are all covered.

Worth noting the PR body claimed these clamps were "the identity they look like". That was wrong and I have corrected it.

PR #2526, opened 2026-09-08
Make the clang-tidy pre-push hook actually fail

lucaromagnoli:
Seven clang-tidy sweeps have landed and nothing was enforcing any of them. The
hook that existed could not fail: it ended its command with `|| true`, and
clang-tidy exits 0 for a plain warning regardless. It also hardcoded the Homebrew
path and printed "clang-tidy not found, skipping" everywhere else.

## What changes

The hook now lives in `.pre-commit-hooks/clang-tidy.sh` and takes pass/fail from
the exit code, which `--warnings-as-errors` is what makes meaningful.

`CMAKE_OSX_SYSROOT` is named so the compile database carries `-isysroot`. Apple's
`/usr/bin/c++` resolves the SDK through xcrun and records nothing, which left
Homebrew's clang-tidy unable to find `TargetConditionals.h`. The Makefile worked
around it by computing the path per invocation; a hook cannot. Only a blank value
is filled, so an explicitly chosen SDK still wins, and every database consumer
benefits, editors included.

## Why not pocc/pre-commit-hooks

It was the obvious choice and it does not work here. Its clang-tidy hook decides
pass/fail from whether stderr is empty, and clang-tidy always writes an
"N warnings generated." tally. The regex meant to strip that matches the singular
"warning generated" only:

```python
re.sub(rb"[\d,]+ warning \S+\s+", b"", self.stderr)   # vs "214 warnings generated."
```

So every file in this codebase fails it. pre-commit cannot set the `CLANG_TIDY`
environment variable, so a wrapper cannot be interposed either.

## What is enforced

The bugprone/cert half of `.clang-tidy`, minus five checks that still report,
measured over `magda/` without engine:

| check | findings |
|---|---|
| implicit-widening-of-multiplication-result | 31 |
| incorrect-roundings | 4 (the deliberate ClipWaveformPainter pixel roundings) |
| unused-return-value | 2 |
| return-const-ref-from-parameter | 2 |
| reserved-identifier | 1 |

The rest of that tier is at zero. Sweeping these five is the next batch and is
what unblocks `WarningsAsErrors`.

Verified both directions: a clean file passes, and a file with a real finding
exits 1 with the diagnostic.

## One correction to .clang-tidy

It claimed `bugprone-exception-escape` "stays enabled and currently reports zero"
while disabling it fourteen lines above. It reports 44, so the disable was right
and the paragraph was not. Rewritten to say what re-enabling would cost; the
noise table's count corrected from 33 to 44.

`magda/engine` is excluded while the native engine is still being written.

## Note

The hook runs at `pre-push`, where `build-check` and `test-check` already sit,
because clang-tidy compiles a full TU per file and is too slow for every commit.
Easy to move if you would rather have it on commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -329,14 +329,44 @@ jobs:
       run: |
         set -euo pipefail
         base="${{ github.event.pull_request.base.sha || github.event.before }}"
-        # Unset, or all-zeros on a branch's first push: nothing to compare against.
-        if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
-          echo "No base commit to diff against; nothing to analyse."
-          exit 0
-        fi
 
-        # The checkout is depth-1, so the base commit is not in this clone yet.
-        git fetch --no-tags --depth=1 origin "$base"
+        # A branch's first push reports the all-zero SHA. Treating that as
+        # "nothing changed" would let a branch created, pushed once and merged
+        # carry findings the whole way: pull_request events only fire for PRs
+        # into main, so a PR into dev/0.X.0 never gets a second chance to lint.
+        # Find where this branch actually left the shared history instead.
+        if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+          git fetch --no-tags --depth=200 origin \
+            '+refs/heads/main:refs/remotes/origin/main' \
+            '+refs/heads/dev/*:refs/remotes/origin/dev/*'
+
+          # Only the shared branches above. Ranging over every remote ref would
+          # let this branch's own ref win the merge-base, and diffing a branch
+          # against itself reports nothing to do.
+          best="" best_count=""
+          head_sha=$(git rev-parse HEAD)
+          for ref in $(git for-each-ref --format='%(refname)' \
+                         refs/remotes/origin/main refs/remotes/origin/dev); do
+            mb=$(git merge-base HEAD "$ref" 2>/dev/null) || continue
+            # A ref that already contains HEAD gives no useful range.
+            [ "$mb" = "$head_sha" ] && continue
+            n=$(git rev-list --count "$mb..HEAD" 2>/dev/null) || continue
+            if [ -z "$best_count" ] || [ "$n" -lt "$best_count" ]; then
+              best_count="$n" best="$mb"
+            fi
+          done
+
+          if [ -z "$best" ]; then
+            echo "Could not find a branch point to diff against." >&2
+            echo "Refusing to report clean without having analysed anything." >&2
+            exit 1
+          fi
+          echo "First push; using branch point $best ($best_count commit(s) ahead)."
+          base="$best"
+        else
+          # The checkout is depth-1, so the base commit is not in this clone yet.
+          git fetch --no-tags --depth=1 origin "$base"
+        fi
 
         # Two dots rather than three: a merge base needs history neither side has
         # here. It can pull in commits that landed on the base branch meanwhile,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -340,17 +340,17 @@ jobs:
 
         if [ -n "$base" ]; then
           # The checkout is depth-1, so the base commit is not in this clone yet.
-          git fetch --no-tags --depth=1 origin "$base"
+          git fetch --no-tags --no-recurse-submodules --depth=1 origin "$base"
         elif [ "$shared" = true ]; then
           # A shared branch. Its commits arrived through PRs that were analysed
           # on their own branches, and diffing back to where dev left main would
           # re-analyse an entire release. Range over what this push added.
           base="${{ github.event.before }}"
           if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
-            git fetch --no-tags --depth=2 origin "${{ github.ref_name }}"
+            git fetch --no-tags --no-recurse-submodules --depth=2 origin "${{ github.ref_name }}"
             base="HEAD~1"
           else
-            git fetch --no-tags --depth=1 origin "$base"
+            git fetch --no-tags --no-recurse-submodules --depth=1 origin "$base"
           fi
         else
           # A feature branch, where github.event.before cannot be trusted.
@@ -358,15 +358,21 @@ jobs:
           # in-flight runs: a push touching no C++ can cancel the only run that
           # would have analysed the C++ still sitting at HEAD, then find nothing
           # in its own narrow range. So diff the whole branch, every push.
+          # --no-recurse-submodules throughout: the checkout configures submodule
+          # recursion, so a plain fetch here went off to fetch faust's and
+          # tracktion's, some of which fail on refs their remotes no longer
+          # serve, and took the step down with them before clang-tidy ever ran.
+          #
           # HEAD first, and in its own fetch. checkout leaves it a shallow root,
           # and merge-base cannot walk a history that stops at the first commit,
           # so deepening only main/dev leaves every merge-base below failing and
           # the branch point unfindable. Adding this branch as another refspec to
           # the fetch below does not deepen HEAD either; it has to be its own
           # invocation.
-          git fetch --no-tags --deepen=200 origin
+          git fetch --no-tags --no-recurse-submodules --deepen=200 origin \
+            "${{ github.ref_name }}"
 
-          git fetch --no-tags --depth=200 origin \
+          git fetch --no-tags --no-recurse-submodules --depth=200 origin \
             '+refs/heads/main:refs/remotes/origin/main' \
             '+refs/heads/dev/*:refs/remotes/origin/dev/*'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -331,14 +331,33 @@ jobs:
         CLANG_TIDY_MAX_TUS: '40'
       run: |
         set -euo pipefail
-        base="${{ github.event.pull_request.base.sha || github.event.before }}"
+        base="${{ github.event.pull_request.base.sha }}"
 
-        # A branch's first push reports the all-zero SHA. Treating that as
-        # "nothing changed" would let a branch created, pushed once and merged
-        # carry findings the whole way: pull_request events only fire for PRs
-        # into main, so a PR into dev/0.X.0 never gets a second chance to lint.
-        # Find where this branch actually left the shared history instead.
-        if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+        case "${{ github.ref_name }}" in
+          main | dev/*) shared=true ;;
+          *)            shared=false ;;
+        esac
+
+        if [ -n "$base" ]; then
+          # The checkout is depth-1, so the base commit is not in this clone yet.
+          git fetch --no-tags --depth=1 origin "$base"
+        elif [ "$shared" = true ]; then
+          # A shared branch. Its commits arrived through PRs that were analysed
+          # on their own branches, and diffing back to where dev left main would
+          # re-analyse an entire release. Range over what this push added.
+          base="${{ github.event.before }}"
+          if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+            git fetch --no-tags --depth=2 origin "${{ github.ref_name }}"
+            base="HEAD~1"
+          else
+            git fetch --no-tags --depth=1 origin "$base"
+          fi
+        else
+          # A feature branch, where github.event.before cannot be trusted.
+          # The first push reports the all-zero SHA, and concurrency cancels
+          # in-flight runs: a push touching no C++ can cancel the only run that
+          # would have analysed the C++ still sitting at HEAD, then find nothing
+          # in its own narrow range. So diff the whole branch, every push.
           git fetch --no-tags --depth=200 origin \
             '+refs/heads/main:refs/remotes/origin/main' \
             '+refs/heads/dev/*:refs/remotes/origin/dev/*'
@@ -364,11 +383,8 @@ jobs:
             echo "Refusing to report clean without having analysed anything." >&2
             exit 1
           fi
-          echo "First push; using branch point $best ($best_count commit(s) ahead)."
+          echo "Branch point $best ($best_count commit(s) ahead)."
           base="$best"
-        else
-          # The checkout is depth-1, so the base commit is not in this clone yet.
-          git fetch --no-tags --depth=1 origin "$base"
         fi
 
         # Two dots rather than three: a merge base needs history neither side has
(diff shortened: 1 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -69,27 +69,21 @@ repos:
...
-                  /opt/homebrew/opt/llvm/bin/clang-tidy "$file" -p=cmake-build-debug --config-file=.clang-tidy --format-style=file || true
-                fi
-              done
-            else
-              echo "clang-tidy not found, skipping..."
-            fi
+        entry: .pre-commit-hooks/clang-tidy.sh
         language: system
-        files: \.(cpp|hpp|h)$
+        files: \.cpp$
```
lucaromagnoli: **P2 — Analyze header-only pushes too**

This filter means pre-commit never invokes the hook when a push changes only `.h`/`.hpp` files. The script comment saying headers are analysed through their TUs cannot take effect because no TU is passed in that case, so a new finding in an inline or template-heavy header reaches the remote unchecked. Please include header changes and map them to at least one compile-database TU (or conservatively analyse the affected TU set).

— Codex

.pre-commit-hooks/clang-tidy.sh
```diff
@@ -0,0 +1,88 @@
...
+    *.cpp) ;;
+    *) continue ;;  # headers are analysed through the TUs that include them
+    esac
+
+    # A file added since the last configure has no database entry, and clang-tidy
+    # would guess at the flags rather than say so. pre-commit passes repo-relative
+    # paths while the database records absolute ones, so match on the suffix.
+    if ! grep -q "/$file\"" "$DB" 2>/dev/null; then
+        echo "skipping $file: no compile_commands.json entry, run 'make debug'"
+        continue
```
lucaromagnoli: **P2 — Fail when a changed source is absent from the compile database**

This path prints a warning and continues without changing `status`, so the hook exits 0 even though it analysed nothing. I reproduced that by passing a missing `.cpp` path: it printed the skip message and returned success. Since `build-check` has already regenerated the database, a missing entry at this point either remains unconfigured or is not part of the build; silently accepting it leaves exactly the new-source gap this gate is meant to close. Please mark the run failed when a requested source has no database entry.

— Codex

.github/workflows/ci.yml
```diff
@@ -308,6 +315,49 @@ jobs:
...
+      env:
+        CLANG_TIDY: clang-tidy-20
+        BUILD_DIR: build
+      run: |
+        set -euo pipefail
+        base="${{ github.event.pull_request.base.sha || github.event.before }}"
+        # Unset, or all-zeros on a branch's first push: nothing to compare against.
+        if [ -z "$base" ] || [ "$base" = "0000000000000000000000000000000000000000" ]; then
+          echo "No base commit to diff against; nothing to analyse."
+          exit 0
```
lucaromagnoli: **P2 — Do not skip lint on a branch’s first push**

For a newly created branch, `github.event.before` is the all-zero SHA, so this exits successfully without analysing any file. This workflow only receives `pull_request` events for PRs targeting `main`; ordinary PRs into `dev/0.20.0` therefore have no later full-PR lint run, and a branch pushed once can carry clang-tidy findings indefinitely. Please use a real comparison base for first pushes (or conservatively analyse all eligible TUs) instead of treating that case as clean.

— Codex

.pre-commit-hooks/clang-tidy.sh
```diff
@@ -0,0 +1,143 @@
...
+        else
+            echo "$file has no compile_commands.json entry." >&2
+            echo "Run 'make debug'. If it belongs to no target, say so in" >&2
+            echo "UNBUILT_OK in $0 rather than leaving it unanalysed." >&2
+            status=1
+        fi
+        ;;
+    *.h | *.hpp)
+        mapfile -t includers < <(
+            grep -rl --include='*.cpp' "include.*$(basename "$file")" magda 2>/dev/null | head -n "$MAX_TUS_PER_HEADER"
```
lucaromagnoli: **P2 — Follow transitive includes when mapping headers**

The direct-include search still silently misses header-only changes. For example, `magda/daw/music/ChordTypes.hpp` is included by `ChordEngine.hpp`, `ChordCache.hpp`, and other headers, but by no `.cpp` directly. Passing that changed header to this script prints “no translation unit includes ChordTypes.hpp” and exits 0, even though many compiled TUs consume it transitively. Please resolve the dependency graph recursively (or use Ninja/depfile data), and fail rather than report clean when a non-allowlisted header cannot be mapped to a TU.

— Codex

.pre-commit-hooks/clang-tidy.sh
```diff
@@ -0,0 +1,158 @@
...
+# One header can pull in a TU another already did, and so can a .cpp alongside
+# its own header.
+if [ ${#targets[@]} -gt 0 ]; then
+    mapfile -t targets < <(printf '%s\n' "${targets[@]}" | sort -u)
+fi
+
+if [ "$MAX_TUS" -gt 0 ] && [ ${#targets[@]} -gt "$MAX_TUS" ]; then
+    echo "note: ${#targets[@]} translation units affected, analysing the first $MAX_TUS."
+    echo "      Set CLANG_TIDY_MAX_TUS=0 to analyse all of them."
+    targets=("${targets[@]:0:$MAX_TUS}")
```
lucaromagnoli: **P1 — Do not cap explicitly changed source files**

The cap is applied after the directly changed `.cpp` targets and header-derived targets are merged. As a result, a push touching nine source files locally analyzes only the first eight; CI similarly drops directly changed files after 40. A clang-tidy failure in one of those skipped files will therefore pass the gate. Please retain every explicit `.cpp` target and apply the cap only to the additional TUs introduced by header dependency expansion (or leave the CI run uncapped).

— Codex

.pre-commit-hooks/tus-for-headers.py
```diff
@@ -0,0 +1,108 @@
...
+        return 2
+
+    tus = set()
+    for outputs in found.values():
+        for output in outputs:
+            source = mapping.get(output)
+            # Tests and third_party compile these headers too, but the checks
+            # here target shipping code, matching .clang-tidy's own scope.
+            if source and source.startswith("magda/") and not source.startswith("magda/engine/"):
+                tus.add(source)
```
lucaromagnoli: **P1 — Fail when filtering removes every dependent TU**

`unmapped` is computed before this production-TU filter, so finding a header only in an excluded test or engine output counts as success even when this loop emits no TU. This is reachable today: `magda/daw/audio/midi/MonoNoteProcessor.hpp` appears in Ninja dependencies only through `tests/clips/test_mono_note_processor.cpp`; invoking this mapper for that header prints nothing and exits 0, and the shell consequently runs clang-tidy on nothing while reporting success. Please validate each header after output-to-source mapping and scope filtering, and fail (or deliberately analyze an available test TU) when no eligible TU remains.

— Codex

.pre-commit-hooks/clang-tidy.sh
```diff
@@ -0,0 +1,182 @@
...
+        )
+    else
+        mapfile -t expanded < <(printf '%s\n' "${expanded[@]}" | sort -u)
+    fi
+fi
+
+if [ "$MAX_TUS" -gt 0 ] && [ ${#expanded[@]} -gt "$MAX_TUS" ]; then
+    echo "note: changed headers reach ${#expanded[@]} further translation units,"
+    echo "      analysing the first $MAX_TUS. Set CLANG_TIDY_MAX_TUS=0 for all."
+    expanded=("${expanded[@]:0:$MAX_TUS}")
```
lucaromagnoli: **P1 — Preserve at least one TU for every changed header when capping**

The mapper now chooses candidates per header, but this cap is applied after those sets are flattened and sorted, so one broad header can consume the whole budget and leave another changed header unanalysed. For example, mapping `TypeIds.hpp` together with `MonoNoteProcessor.hpp` currently yields 465 TUs; the latter header’s only candidate, `tests/clips/test_mono_note_processor.cpp`, sorts last. The local first-8 and CI first-40 therefore contain no TU that includes `MonoNoteProcessor.hpp`, recreating the silent pass for that header. Please preserve one representative TU per changed header before spending the remaining cap on fan-out (which likely requires retaining the header-to-TU grouping from the mapper).

— Codex

.github/workflows/ci.yml
```diff
@@ -308,6 +315,84 @@ jobs:
...
+    - name: Run clang-tidy on changed files
+      env:
+        CLANG_TIDY: clang-tidy-20
+        BUILD_DIR: build
+        # Higher than the hook's 8: a push waits on the hook, nobody waits on
+        # this, and a change to a widely included header deserves the coverage.
+        CLANG_TIDY_MAX_TUS: '40'
+      run: |
+        set -euo pipefail
+        base="${{ github.event.pull_request.base.sha || github.event.before }}"
```
lucaromagnoli: **P1 — Diff the whole branch because earlier push runs are cancelled**

On a push event this limits clang-tidy to the latest push, but the workflow has `cancel-in-progress: true`. If push A changes a C++ file and push B follows while A is still running (for example, B only edits the PR description/config/docs), B cancels the only run that would analyze A; its replacement diffs only `A..B`, finds no C++ files, and succeeds without ever checking the C++ change still present at HEAD. Please diff push runs from the branch point/base branch through HEAD, as the first-push path already does, or otherwise retain completed coverage for earlier commits before cancelling their runs.

— Codex

.github/workflows/ci.yml
```diff
@@ -308,6 +315,100 @@ jobs:
...
+          else
+            git fetch --no-tags --depth=1 origin "$base"
+          fi
+        else
+          # A feature branch, where github.event.before cannot be trusted.
+          # The first push reports the all-zero SHA, and concurrency cancels
+          # in-flight runs: a push touching no C++ can cancel the only run that
+          # would have analysed the C++ still sitting at HEAD, then find nothing
+          # in its own narrow range. So diff the whole branch, every push.
+          git fetch --no-tags --depth=200 origin \
```
lucaromagnoli: **P1 — Deepen the feature branch before looking for its merge base**

`actions/checkout` defaults to depth 1, and this fetch deepens only `main`/`dev/*`. The checked-out feature `HEAD` remains listed as a shallow root, so Git cannot traverse even its first parent and every `git merge-base HEAD "$ref"` below fails. I reproduced the same shape with this PR head: after fetching the head at depth 1 and `dev/*` at depth 200, `git merge-base HEAD origin/dev/0.20.0` exits 1. Consequently every feature-branch push reaches “Could not find a branch point” and fails CI. Please deepen/unshallow the feature ref too (or set an adequate `fetch-depth` on checkout) before computing these merge bases.

— Codex

.pre-commit-hooks/tus-for-headers.py
```diff
@@ -0,0 +1,172 @@
...
+            continue
+        if line[0] != " ":
+            # "target: #deps N, deps mtime M (VALID|STALE)"
+            current = line.split(":", 1)[0]
+            stale = line.endswith("(STALE)")
+            continue
+        if current is None or stale:
+            continue
+        dep = line.strip()
+        if dep[dep.rfind("/") + 1:] not in names:
```
lucaromagnoli: **P1 — Parse Ninja dependency paths portably on Windows**

This basename extraction recognizes only `/`. Ninja dependency paths on Windows commonly use `\\`, so an absolute dependency such as `C:\\repo\\magda\\Foo.hpp` is compared whole against the wanted basename `Foo.hpp` and rejected; the later absolute-path test also treats `C:\\...` as relative. In addition, `line.split(":", 1)` above truncates an absolute output target at the drive colon. The result is that changed headers map to no TUs, get the non-blocking “nothing compiles” note, and silently pass the local hook on Windows. Please normalize both separators/absolute paths and split target records on the `: #deps ` marker rather than the first colon.

— Codex

.pre-commit-hooks/tus-for-headers.py
```diff
@@ -0,0 +1,185 @@
...
+            while tus:
+                tu = tus.pop(0)
+                if tu not in seen:
+                    seen.add(tu)
+                    chosen.append(tu)
+                    break
+            if not tus:
+                remaining.remove(tus)
+            if max_tus > 0 and len(chosen) >= max_tus:
+                break
```
lucaromagnoli: outsource to a function

.pre-commit-hooks/tus-for-headers.py
```diff
@@ -0,0 +1,185 @@
...
+    return chosen, total
+
+
+def in_scope(source):
+    """Shipping code, matching .clang-tidy's own scope."""
+    return source.startswith("magda/") and not source.startswith("magda/engine/")
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
```
lucaromagnoli, line 148: break down into smaller helpers

.pre-commit-config.yaml
```diff
@@ -69,27 +69,24 @@ repos:
...
-            fi
+        entry: .pre-commit-hooks/clang-tidy.sh
         language: system
+        # Headers too: a push that changes only an inline or template-heavy
+        # header would otherwise go unanalysed. The script maps each one to the
+        # translation units that include it.
         files: \.(cpp|hpp|h)$
+        exclude: ^magda/engine/
+        pass_filenames: true
         stages: [pre-push]
```
lucaromagnoli, line 90: **P1 — Install the pre-push hook that owns this gate**

The repository’s supported setup never installs this stage. Both `setup.sh:120` and `docs/development/code-style.md:163` run plain `pre-commit install`, which installs the `pre-commit` Git hook by default; this config has no `default_install_hook_types`, while clang-tidy is registered only for `pre-push`. A developer following either instruction therefore gets no `.git/hooks/pre-push`, and this local gate never runs. Please add `pre-push` to `default_install_hook_types` and update the installer/docs to install that hook type explicitly (including a migration instruction for existing clones).

— Codex

.pre-commit-hooks/clang-tidy.sh
```diff
@@ -0,0 +1,185 @@
...
+    *.h | *.hpp) headers+=("$file") ;;
+    esac
+done
+
+# Database membership is resolved by the same code that reads it, rather than a
+# grep here: the file records absolute native paths, so matching a relative one
+# by string has to assume a separator, and on Windows that marks every file
+# missing and passes the lot.
+if [ ${#sources[@]} -gt 0 ]; then
+    if ! absent=$(python3 "$resolver" --missing --build-dir "$BUILD_DIR" "${sources[@]}"); then
```
lucaromagnoli: **P1 — Use the Python executable that setup actually supports**

The installer accepts `python`/`pip` when `python3`/`pip3` are absent (`setup.sh:111-116`), but this hook hardcodes `python3` here and again for header mapping. That is a common Windows installation shape: setup can successfully install the new pre-push hook, then every C++ push fails at this line with `python3: command not found`, before clang-tidy runs. Please resolve an interpreter with the same `python3`-then-`python` fallback used by setup (or otherwise arrange a known Python executable for the hook); the resolver shebang should follow the same contract for direct use.

— Codex

Issue #2464, opened 2026-09-06
Recording slice 6: session slot recording and the arrangement capture


lucaromagnoli:
Slice 6 of #1895. Depends on slices 3 and 4, and on the launcher (#1894), which is done.

## What the fork does today

Two separate things, both keyed off polling.

Slot recording is `armSessionSlotRecording`, `beginArmedSessionSlotRecordings` and `finishSessionSlotRecordings` in `daw/engine/TracktionEngineWrapperRecording.cpp`: an empty slot on an armed track is marked as a target, TE's recording targets are set through the bridge, and `finalizeSessionSlotAudioRecording` / `finalizeSessionSlotMidiRecording` turn what TE recorded into a slot clip.

Session-to-arrangement capture is `SessionRecorder` (`daw/audio/session/SessionRecorder.hpp`), which watches session clips play and writes arrangement clips from what it saw. It sees them by asking: a `getPlayState_` callback per clip, a `getLaunchTime_` callback per track, and `updatePreviews()` once a frame. What was heard and what was captured come from two different clocks, and the difference between them is the drift.

## Scope

- Recording into a slot begins when the slot's launch actually fires, which is a quantized beat that may land inside a block, not when Record was pressed. `LaunchHandle::advance` already splits a block at that beat and `LaunchTap` already reports it; the take starts on the same sample the slot started sounding on.
- The run's end: stopped, re-launched, or a follow action moving on. The take ends where the run ended.
- A slot that already holds a clip is not a recording target, as today.
- Session-to-arrangement capture built from the launcher's own sample-stamped launch and stop events rather than from a poll of play state. One clock authority, which is what removes heard-versus-captured drift as a class rather than tightening it.
- A scene launch capturing as one event across tracks, not N observations that may fall on either side of a frame.

## Not in this slice

Follow actions themselves, which the launcher owns. Comping. Overdubbing into an occupied slot, which is not a feature today.

## Verification

Offline, driving the launcher and the transport:

- a slot launched on a beat inside a block records from that beat, not from the block boundary
- a scene launch across several tracks produces takes that all start on the same sample
- a run stopped mid-clip keeps what played, and a re-launch starts a second take rather than extending the first
- an arrangement capture's clip starts at the beat the launch tap reported for that run, over a corpus of launches at several quantizations
- a capture over a transport loop places each run once
- an occupied slot is left alone

lucaromagnoli:
Merged in #2533.

PR #2527, opened 2026-09-08
Report the recording pass without a poll

lucaromagnoli:
Slice 5 of #1895, closes #2463. The pass a take is recording, published by the block that recorded it rather than assembled a frame later out of three different clocks.

## What is here

**`tap/RecordTap.hpp`** is the fourth tap, following `LaunchTap`: where the pass started, how much it covers, the notes it has captured and the peaks its audio reached. Both takes own one and hand it out through `TakeCapture`, so a reader asks the take that was published rather than a table kept beside it.

- **The length is the take's own.** For audio, the beat its written samples reach; for MIDI, the beat its last block reached. Nothing counts frames, so what is drawn and what is written cannot come apart.
- **A note is converted the way `finish()` converts it** -- the block's own map, off the pass origin the take will use -- so the beat in the preview is the beat in the finished clip. A held note reaches the pass end and stops where it came up.
- **A wrap opens the next pass on the boundary the take actually took.** A pass end the take could not hold is two passes run together in the preview as well.
- **Peaks come off the samples the take wrote,** at the position it wrote them, rather than from a metering path alongside.
- **Both targets are on one reading.** Arrangement or slot, with the scene; slice 6 (#2464) fills in the slot half.

### The read protocol

This is the part worth naming, because the pass is bigger than a word and `ValueTap`'s one-load answer does not stretch to it.

The scalars a wrap resets together sit behind a sequence the reader retries over, and that window is those stores and nothing else, which is what makes the retry converge.

The notes are not in it: an array is as long as the pass is, and a reader copying one inside a window would be racing the writer over a growing amount of work. Each note carries the pass that recorded it instead, and a reader takes the ones belonging to the pass it is reporting. A wrap during a read therefore costs notes off the end of the list, never a note in the wrong place.

The peaks carry neither and are the one thing here that can be ragged -- a wrap inside a read redraws a few ticks of the last pass's audio, for the frame it takes to ask again. Peaks are presentation and place nothing.

### Input meters

`OpRole::LiveInputMeter` is a `Meter` op between the live input and the track's mix, so the level tap machinery that already exists covers it and the input reaches the track through the thing that measures it. The incumbent reads this off `WaveInputDevice::levelMeasurer`, which had no equivalent behind an input op.

## Verification

`tests/engine/test_record_tap.cpp`, offline throughout: the tap on its own, then both takes driven a callback at a time.

- a pass reports the length it captured, at three block sizes, audio and MIDI
- a note is drawn at the beat the finished clip holds it at, at a fixed tempo and across a tempo change inside the pass
- a held note is as long as the pass until it comes up
- an armed track with a stopped transport, and one counting in, report nothing rather than a pass of no length
- a wrap draws the pass it opened, and starts the peaks again
- a reader running beside the writer never sees one pass's start beside another's notes
- peaks are the samples the take wrote, and what a pass has no room for is counted
- a take nobody draws reports its length and nothing lost

The input meter case is in `test_live_input.cpp`, where the session rig already is: the meter is the input's only consumer, and it reads the level the monitoring track is hearing.

### DIFF
--- a/magda/engine/io/TakeRecorder.cpp
+++ b/magda/engine/io/TakeRecorder.cpp
@@ -96,7 +96,7 @@ void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop) {
 
     passOrigin_ = 0;
     passOriginBeat_ = startBeat_;
-    pendingBoundary_ = -1;
+    pendingCount_ = 0;
     tap_.open(startBeat_);
 
     // A negative latency asks for samples from before the take was armed.
@@ -131,16 +131,21 @@ void TakeRecorder::openPass(const BlockInfo& block, const LoopRange& loop) {
 
     // The pass the tap draws is the pass the file will be, so it turns over
     // where the sink does: at the boundary, not at the wrap that named it.
-    pendingBoundary_ = boundary;
-    pendingStartBeat_ = block.beats.start;
+    if (pendingCount_ < kPendingCapacity)
+        pending_[(pendingHead_ + pendingCount_++) % kPendingCapacity] = {boundary,
+                                                                         block.beats.start};
 }
 
 void TakeRecorder::openTapPass(const BlockInfo& block) {
-    passOrigin_ = pendingBoundary_;
+    const auto& pass = pending_[pendingHead_];
+
+    passOrigin_ = pass.boundary;
     passOriginBeat_ =
-        block.beatAtTime(startSeconds_ + (static_cast<double>(pendingBoundary_) / block.rate()));
-    pendingBoundary_ = -1;
-    tap_.open(pendingStartBeat_);
+        block.beatAtTime(startSeconds_ + (static_cast<double>(pass.boundary) / block.rate()));
+    tap_.open(pass.startBeat);
+
+    pendingHead_ = (pendingHead_ + 1) % kPendingCapacity;
+    --pendingCount_;
 }
 
 void TakeRecorder::stop() {
@@ -176,16 +181,20 @@ void TakeRecorder::write(const BlockInfo& block) {
     stream_.writeAudio(keptBlock, kept);
 
     // The pass's own audio, at the position it was written at, so a peak and
-    // the samples behind it are the same stretch. A block the pass boundary
-    // falls inside is split there, the way the sink splits it.
+    // the samples behind it are the same stretch. Every boundary this block
+    // reaches is taken, the way the sink takes them: a loop shorter than the
+    // block puts more than one inside it.
     auto from = 0;
-    if (pendingBoundary_ >= 0 && block.rate() > 0.0 && written_ + kept > pendingBoundary_) {
-        from = static_cast<int>(pendingBoundary_ - written_);
-        if (from > 0)
-            tap_.addPeaks(keptBlock.getSubBlock(0, static_cast<std::size_t>(from)), from,
-                          written_ - passOrigin_);
+    while (pendingCount_ > 0 && block.rate() > 0.0 &&
+           written_ + kept > pending_[pendingHead_].boundary) {
+        const auto at = static_cast<int>(pending_[pendingHead_].boundary - written_);
+        if (at > from)
+            tap_.addPeaks(keptBlock.getSubBlock(static_cast<std::size_t>(from),
+                                                static_cast<std::size_t>(at - from)),
+                          at - from, (written_ + from) - passOrigin_);
 
         openTapPass(block);
+        from = at;
     }
 
     tap_.addPeaks(keptBlock.getSubBlock(static_cast<std::size_t>(from),
--- a/magda/engine/io/TakeRecorder.cpp
+++ b/magda/engine/io/TakeRecorder.cpp
@@ -38,13 +38,13 @@ std::vector<std::int64_t> passLengths(std::span<const RecordedPass> passes) {
 
 }  // namespace
 
-TakeRecorder::TakeRecorder(const LiveInputFeed& feed, const RenderContext& context,
+TakeRecorder::TakeRecorder(const LiveInputFeed& feed, const RenderContext& context, RecordTap& tap,
                            TakeRecorderSettings settings)
     : settings_(std::move(settings)),
       input_(feed, settings_.channels, settings_.latencySamples),
       sink_(settings_.directory, settings_.name, settings_.file, fileContext(context, settings_)),
       stream_(sink_, queueFor(settings_)),
-      tap_(RecordMaterial::audio, settings_.tap) {
+      tap_(tap) {
     scratch_.setSize(takeChannels(settings_), std::max(1, context.maxBlockSize), false, true,
                      false);
 
--- a/magda/engine/io/TakeRecorder.cpp
+++ b/magda/engine/io/TakeRecorder.cpp
@@ -124,13 +124,16 @@ void TakeRecorder::openPass(const BlockInfo& block, const LoopRange& loop) {
     if (firstBoundary_ < 0)
         firstBoundary_ = boundary;
 
+    // One acceptance, and the file and the preview both follow it: a boundary
+    // refused here is refused for both, which is what keeps them the same
+    // passes rather than two opinions about where a pass ended.
     if (!sink_.markPassEnd(boundary)) {
         ++boundariesLost_;
         return;
     }
 
-    // The pass the tap draws is the pass the file will be, so it turns over
-    // where the sink does: at the boundary, not at the wrap that named it.
+    // The preview turns over where the sink does: at the boundary, not at the
+    // wrap that named it.
     if (pendingCount_ < kPendingCapacity)
         pending_[(pendingHead_ + pendingCount_++) % kPendingCapacity] = {boundary,
                                                                          block.beats.start};
--- a/magda/engine/io/TakeRecorder.cpp
+++ b/magda/engine/io/TakeRecorder.cpp
@@ -43,7 +43,8 @@ TakeRecorder::TakeRecorder(const LiveInputFeed& feed, const RenderContext& conte
     : settings_(std::move(settings)),
       input_(feed, settings_.channels, settings_.latencySamples),
       sink_(settings_.directory, settings_.name, settings_.file, fileContext(context, settings_)),
-      stream_(sink_, queueFor(settings_)) {
+      stream_(sink_, queueFor(settings_)),
+      tap_(RecordMaterial::audio, settings_.tap) {
     scratch_.setSize(takeChannels(settings_), std::max(1, context.maxBlockSize), false, true,
                      false);
 
@@ -72,7 +73,7 @@ void TakeRecorder::capture(const BlockInfo& block, bool countingIn, const LoopRa
             return;
         }
 
-        openPass(loop);
+        openPass(block, loop);
     }
 
     write(block);
@@ -93,6 +94,10 @@ void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop) {
     startedAtLoopStart_ = atLoopStart(block, loop);
     headDrop_ = std::max(0, settings_.latencySamples);
 
+    passOrigin_ = 0;
+    passOriginBeat_ = startBeat_;
+    tap_.open(startBeat_);
+
     // A negative latency asks for samples from before the take was armed.
     // Silence stands in for them, which is what leaves the first real sample
     // where it happened.
@@ -103,7 +108,7 @@ void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop) {
     }
 }
 
-void TakeRecorder::openPass(const LoopRange& loop) {
+void TakeRecorder::openPass(const BlockInfo& block, const LoopRange& loop) {
     loopStartBeat_ = loop.startBeat;
     loopEndBeat_ = loop.endBeat;
 
@@ -118,13 +123,24 @@ void TakeRecorder::openPass(const LoopRange& loop) {
     if (firstBoundary_ < 0)
         firstBoundary_ = boundary;
 
-    if (!sink_.markPassEnd(boundary))
+    if (!sink_.markPassEnd(boundary)) {
         ++boundariesLost_;
+        return;
+    }
+
+    // The pass the tap draws is the pass the file will be, so it opens on the
+    // boundary that was taken. It is counted from what has been written rather
+    // than from the boundary itself, so a peak and the length beside it are the
+    // same stretch of audio.
+    passOrigin_ = written_;
+    passOriginBeat_ = endBeat_;
+    tap_.open(block.beats.start);
 }
 
 void TakeRecorder::stop() {
     state_ = State::stopped;
     rolling_.store(false, std::memory_order_relaxed);
+    tap_.close();
 }
 
 void TakeRecorder::write(const BlockInfo& block) {
@@ -148,9 +164,15 @@ void TakeRecorder::write(const BlockInfo& block) {
     if (kept <= 0)
         return;
 
-    stream_.writeAudio(
-        captured.getSubBlock(static_cast<std::size_t>(offset), static_cast<std::size_t>(kept)),
-        kept);
+    const auto keptBlock =
+        captured.getSubBlock(static_cast<std::size_t>(offset), static_cast<std::size_t>(kept));
+
+    stream_.writeAudio(keptBlock, kept);
+
+    // The pass's own audio, at the position it was written at, so a peak and
+    // the samples behind it are the same stretch.
+    tap_.addPeaks(keptBlock, kept, written_ - passOrigin_);
+
     written_ += kept;
     captured_.store(written_, std::memory_order_relaxed);
 
@@ -159,6 +181,8 @@ void TakeRecorder::write(const BlockInfo& block) {
     // arrived, and across a tempo change the two are not the same length.
     if (const auto rate = block.rate(); rate > 0.0)
         endBeat_ = block.beatAtTime(startSeconds_ + (static_cast<double>(written_) / rate));
+
+    tap_.extend(std::max(0.0, endBeat_ - passOriginBeat_));
 }
 
 RecordedTake TakeRecorder::finish() {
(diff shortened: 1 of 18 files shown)


magda/engine/tap/RecordTap.cpp
```diff
@@ -0,0 +1,179 @@
...
+    into.notes.clear();
+    for (std::size_t at = 0; at < numNotes; ++at) {
+        const auto identity = notes_[at].identity.load(std::memory_order_relaxed);
+        if (static_cast<std::uint32_t>(identity >> 32U) != into.pass)
+            break;
+
+        into.notes.push_back({static_cast<int>(identity & 0x7fU),
+                              static_cast<int>((identity >> 8U) & 0x7fU),
+                              notes_[at].startBeat.load(std::memory_order_relaxed),
+                              notes_[at].lengthBeats.load(std::memory_order_relaxed)});
```
lucaromagnoli: [P2] Read each reused note slot coherently

The pass check at lines 161–163 does not protect the subsequent position and length loads. A reader can accept the old identity, then the writer can wrap and overwrite this slot before those loads, so the reading contains the old pass's pitch/velocity with the next pass's coordinates. There is also a window where noteOn has overwritten the coordinates but has not yet replaced the identity. Please use a coherent per-slot publication/read protocol that detects an overwrite spanning the payload read; tagging the identity alone does not provide the documented guarantee. A concurrency test should vary and validate positions and lengths across passes as well as pitch.

magda/engine/io/TakeRecorder.cpp
```diff
@@ -118,13 +123,24 @@ void TakeRecorder::openPass(const LoopRange& loop) {
...
         ++boundariesLost_;
+        return;
+    }
+
+    // The pass the tap draws is the pass the file will be, so it opens on the
+    // boundary that was taken. It is counted from what has been written rather
+    // than from the boundary itself, so a peak and the length beside it are the
+    // same stretch of audio.
+    passOrigin_ = written_;
+    passOriginBeat_ = endBeat_;
```
lucaromagnoli: [P2] Advance the preview at the accepted audio boundary

With positive input latency, written_ is behind arrivals_, but the file boundary accepted above is at arrivals_. Resetting passOrigin_ to written_ and opening the tap immediately therefore assigns the remaining tail of the previous file to the new preview. For example, with 128 samples of latency, the next 128 samples after the transport wraps still go into the previous pass's file but become the new preview's first peaks and length. Please split/advance the tap when the written audio reaches the accepted boundary so the preview represents the same pass as the sink. The existing wrap test uses zero latency and cannot expose this.

magda/engine/tap/RecordTap.cpp
```diff
@@ -0,0 +1,179 @@
...
+    auto& slot = notes_[at];
+    slot.startBeat.store(beat, std::memory_order_relaxed);
+    slot.lengthBeats.store(0.0, std::memory_order_relaxed);
+    slot.identity.store(packIdentity(pass_, note, velocity), std::memory_order_relaxed);
+
+    // The count is what publishes the slot, so it is stored last.
+    numNotes_.store(at + 1, std::memory_order_release);
+
+    held_[heldIndex(channel, note)] = static_cast<std::uint32_t>(at);
+    heldSlots_.push_back(static_cast<std::uint32_t>(at));
```
lucaromagnoli: [P2] Handle a repeated note-on before replacing the held index

Two note-ons for the same channel and pitch append two held slots, but this assignment replaces the only index through which noteOff can find the first one. After the corresponding note-off, the first slot remains in heldSlots_ and extend() keeps lengthening it to the end of the pass. This also disagrees with PassWalk::add(), which replaces the earlier pending note for that channel/pitch. Please handle the existing held entry consistently with the finished-take conversion and cover note-on, note-on, note-off followed by further pass extension.

lucaromagnoli (commented):
Found three issues in the recording preview. The current build is up to date and all 17 existing [2463] tests pass; the cases below are not covered.

magda/engine/tap/RecordTap.cpp
lucaromagnoli: Fixed in 4975e8cd.

A slot is now cleared before it is filled and published by the identity written last, and a reader that sees the identity change either side of reading the position drops the note. That closes both windows you named -- the slot taken over by the next pass, and `noteOn` having written the coordinates but not yet the identity within one pass.

The cross-pass concurrency case now carries positions and lengths as well as pitch (every note of pass p sits in `[p*1000, p*1000+8)` with a length no greater than the one it was closed at), but it does not reproduce the tear on its own: the window is two loads wide and pass churn is not enough to land in it. So there is a second case, "A note is read as one note, not two halves of a slot", which strikes one pitch four million times over -- the slot is rewritten in place, and the identity carries the velocity while the position is the same number -- so the two disagreeing is the tear. It fails on the first read within a second without the protocol, and passes with it.

magda/engine/io/TakeRecorder.cpp
lucaromagnoli: Fixed in 4975e8cd.

`openPass` now holds the accepted boundary rather than turning the tap over at the wrap, and `write` opens the pass when the written audio reaches it, splitting the block the boundary falls inside the way the sink splits it. So the preview pass and the file are the same samples at any latency.

Covered by "An audio pass turns over where the file does, not where the wrap is", at 128 samples of latency: the pass still reads as 1 a block past the wrap, and once the audio reaches the boundary its first peak is the material at arrival 4128 onward rather than 4000. Both halves fail against the old code.

magda/engine/tap/RecordTap.cpp
lucaromagnoli: Fixed in 4975e8cd.

`noteOn` now rewrites the slot the pitch is already in rather than appending a second one, which is what `PassWalk::add` does when it finishes the take -- the earlier pending note is replaced. Nothing is orphaned in `heldSlots_`, so `extend` has nothing to keep lengthening.

Covered by "A pitch struck twice replaces the note that was already down": note on, note on, note off, then two more beats of pass. The preview holds one note, at the second strike's velocity, and its start and length are compared against the finished take's rather than against constants. It reads 2 == 1 against the old code.

magda/engine/io/TakeRecorder.cpp
```diff
@@ -118,13 +124,29 @@ void TakeRecorder::openPass(const LoopRange& loop) {
...
 
-    if (!sink_.markPassEnd(boundary))
+    if (!sink_.markPassEnd(boundary)) {
         ++boundariesLost_;
+        return;
+    }
+
+    // The pass the tap draws is the pass the file will be, so it turns over
+    // where the sink does: at the boundary, not at the wrap that named it.
+    pendingBoundary_ = boundary;
```
lucaromagnoli: [P2] Preserve every pending audio boundary for short loops

Each wrap replaces pendingBoundary_. When the loop is shorter than the positive input latency, another wrap arrives before written audio reaches the previous boundary, so that boundary is overwritten and the preview never advances. The sink correctly retains all of these boundaries. I reproduced this with AudioRig using a 64-sample loop, 128-sample latency and 640 samples delivered: finish() produces 8 valid file passes, while the preview still reports pass 1 and lengthBeats = 0.128 instead of 0.016. Please queue the accepted pending boundaries and consume them as written audio reaches them rather than retaining only the latest one. All 20 existing [2463] tests pass; this isolated short-loop reproduction fails.

lucaromagnoli: Fixed in d3cd2ee2. You were right and I had waved this one through when I wrote the first fix -- I noticed the overwrite and reasoned it was pathological, without checking that the sink keeps all of them.

The boundaries are queued and taken as the written audio reaches them, in a lane as deep as the sink's own. It cannot outrun that: this retires a boundary when `written_` passes it and the sink retires one only once the record thread has consumed past it, so a boundary the sink accepted always has room here. A block can now cross more than one, and its peaks are split at each.

Your reproduction is the test: 64-sample loop, 128 samples of latency, 640 delivered. It reads as pass 8 at one loop of length; against the single-slot version it reads pass 1 at 0.128, which is your numbers exactly.

lucaromagnoli:
The repeated findings point to an architectural issue within the recording-preview design. I should have raised this after the earlier findings rather than treating each as an isolated fix.

Two problems connect them:

- **Recording semantics are implemented twice.** The preview tracks held notes and pass boundaries separately from the finished take. The repeated-note and latency bugs came from those implementations disagreeing.
- **The publication protocol lacks a consistent guarantee.** Pass scalars, note identities, note coordinates and peaks use different synchronization rules. Making each field atomic prevents data races; it does not make a reading coherent. The latest isolated concurrency test also reproduced mixed coordinates when the same pitch and velocity are retriggered: the identity returns to the same value, so comparing it before and after the payload read does not detect the replacement.

I recommend pausing incremental fixes and establishing two design rules: recording and preview consume the same accepted pass/note transitions, and readers receive a validated snapshot—or retain their previous valid reading when validation fails. Audio-thread work must remain bounded and allocation-free. The contract should explicitly distinguish coherent recording metadata from any intentionally approximate waveform presentation.

The current approach could implement those rules with proper revisions and shared transition logic; it does not necessarily need replacing wholesale. But a per-slot counter alone fixes only the latest race. The broader issue is who owns recording truth, and what consistency the preview promises.

magda/engine/tap/RecordTap.cpp
```diff
@@ -0,0 +1,210 @@
...
+            break;
+
+        const auto startBeat = slot.startBeat.load(std::memory_order_relaxed);
+        const auto lengthBeats = slot.lengthBeats.load(std::memory_order_relaxed);
+
+        // The slot could have been taken over between the identity and the two
+        // loads above. A length that grew under the read is the same note being
+        // held; a slot that changed hands is not this note at all.
+        std::atomic_thread_fence(std::memory_order_acquire);
+        if (slot.identity.load(std::memory_order_relaxed) != identity)
```
lucaromagnoli: [P1] Detect every note-slot replacement, including identical retriggers

Retriggering the same pitch and velocity within one pass restores the same packed identity. A reader can load the previous strike's startBeat, then the replacement's lengthBeats, and accept both because the identity compares equal. The temporary zero does not prevent this: the reader can miss it entirely. This is an ABA problem in the publication protocol.

An isolated concurrency test using steadily advancing note positions and constant pitch/velocity reproduced 6 mixed readings on this revision. All 23 existing [2463] tests pass; the retrigger test varies velocity and therefore does not exercise this identical-identity case.

Use a per-slot revision that changes on every replacement, with a validated payload read. This addresses the specific race; the broader ownership and snapshot-consistency concerns are described in the architectural comment.

lucaromagnoli:
Agreed, and thanks for stopping the incremental loop — you were right that the four findings were one problem twice over. Done in 461c48f6.

## Recording and preview consume the same transitions

`io/TakeNotes.hpp` holds what a channel message is and which pitches are down. Both halves drive it: `PassWalk` when it finishes the take, `PreviewNotes` as the pass runs.

The duplication was worse than I had realised once I went looking. The wire rule for a note on at zero velocity had two implementations — a hand-rolled `kindOf` on the recorded bytes and JUCE's `isNoteOn`/`isNoteOff` on the live message — and the replace-on-restrike rule had two more. That is exactly where the repeated-note bug came from.

`PreviewNotes` is now the one road from a transition to a published note, so a case about the tap drives what the recorder drives rather than a parallel composition. The tap publishes and nothing else: no held table, no replacement rule, no note semantics of its own.

For the audio boundary: the preview's lane is `TakeFileSink::kBoundaryCapacity` rather than a second number, and the invariant that keeps the two the same passes is written down — the preview retires a boundary when the audio is offered past it and the sink no earlier, so what the sink accepted always has room.

## A validated snapshot, and what is deliberately not one

Which pass it is, where it starts, how many notes it holds and what each of them is are published behind one revision and handed over only if it did not move. `read` returns false and leaves the caller's reading untouched when it cannot settle.

The revision turns once per **block** rather than once per note (`RecordTap::Change`, nesting, so the recorder wraps a block and the individual calls still stand alone). That matters twice: the unit a reader can land either side of is the same unit the take consumes, and a reader retries at most once a block instead of once a note. My first cut turned it per note and starved readers under a hot writer, which the reworked concurrency cases caught by counting settled reads.

Two things stay outside the snapshot and the header says so rather than leaving it to be found:

- **A note's length** is measured from a start that is in the snapshot, so it can be a block behind without being attached to the wrong note. What is not promised is that two notes' lengths are from the same block.
- **The peaks** are presentation, not metadata. A wrap landing inside a read can redraw a few ticks of the previous pass for one frame. That is the only approximation admitted, and it places nothing.

Audio-thread cost is unchanged in kind: a structural change is a handful of stores, a block's extension is one store per note still down, and nothing allocates.

## Tests

New: a read that cannot settle leaves the reading it was given; a block's events reach a reader together or not at all. The concurrency cases now assert that reads settled and that notes were examined, so they cannot pass by reading nothing — the earlier versions could have.

One thing I will not overclaim: the guard on `open()` closes a window a few instructions wide that the note guards mask in any sequence where notes follow an open, and no probabilistic test I wrote isolates it. It is there because it is correct, not because a case proves it.

magda/engine/tap/RecordTap.cpp
lucaromagnoli: Fixed in eed25c74, then subsumed by the redesign in 461c48f6.

The immediate fix was a per-slot revision turned on every replacement, which is what you asked for. The architectural pass then replaced it with one tap-wide revision covering the pass and every note together, so there is one publication protocol rather than three — the ABA is closed the same way, by a count rather than by what the note is.

Your reproduction shape is now a standing case: one pitch at one velocity, struck repeatedly without coming up, the position advancing and encoded in the length. It is deliberately unbatched so what it exercises is the guard on the call itself rather than the block wrapper. It mixes readings within a second without the count and not at all with it.

magda/engine/tap/RecordTap.cpp
```diff
@@ -0,0 +1,173 @@
...
+        into.scene = settings_.scene;
+        into.pass = pass;
+        into.startBeat = startBeat;
+        into.notesLost = notesLost;
+        into.notes.swap(into.staging);
+
+        // Outside the snapshot on purpose. A length is the pass or a note as of
+        // some block, and neither can be attached to the wrong pass or the
+        // wrong note: the start it is measured from is in the snapshot.
+        into.lengthBeats = lengthBeats_.load(std::memory_order_relaxed);
```
lucaromagnoli: [P2] Read pass length before validating the snapshot

A wrap can occur after the revision check at lines 139–141 but before this load. The reader then returns the previous pass's start and notes with the new pass's length, potentially shrinking the old overlay to zero. Keeping the start in the snapshot does not associate a subsequent unvalidated length load with that start.

Please capture lengthBeats_ before the final revision validation and commit the captured value with the other metadata. Length may still update independently within a pass, but a pass replacement must invalidate the reading.

An isolated concurrency test reproduced this on the current revision: the writer opens and extends each pass inside one RecordTap::Change, while the reader checks that each successfully read length belongs to the reported pass. All 26 existing [2463] tests pass.

lucaromagnoli: Fixed in the commit above. You are right, and the hole was in the argument I wrote rather than only in the code: "a length is safe outside the snapshot because the start it is measured from is inside it" holds within a pass and not across one, and I did not write down the half that does not hold.

`lengthBeats_` is captured with the other metadata and committed from the capture, so only the peaks are outside the check now. The length still grows without turning the revision — doing that would send a reader round on every block it landed in — and what makes that safe is stated as the narrower thing it is: a start cannot move while its pass stands, and a wrap sends the reader round like any other change.

Your reproduction is the case: the writer opens and extends each pass inside one `Change`, each pass named twice over by where it starts and by a length one beat past that, and the reader checks the two agree. It fails on the first read without the fix and runs clean with it.

magda/engine/tap/RecordTap.hpp
```diff
@@ -0,0 +1,326 @@
...
+     * than once a note. Nesting is the ordinary case -- the calls below take
+     * one each -- and only the outermost turns the revision.
+     *
+     * Audio thread, for the length of a block and no longer.
+     */
+    class Change {
+      public:
+        explicit Change(RecordTap& tap) : tap_(tap) {
+            if (tap_.changing_++ == 0)
+                tap_.revision_.fetch_add(kChanging, std::memory_order_release);
```
lucaromagnoli: [P2] Order the odd revision before the following payload stores

The opening fetch_add(..., memory_order_release) does not supply the writer-side fence needed by this snapshot protocol. The payload stores that follow are relaxed, so a reader observing one of those stores has no synchronization edge forcing its final revision load to observe the opening increment. The C++ memory model therefore permits changed payload fields to be accepted with the previous even revision.

Please add a release fence after marking the outermost revision odd, before modifying the payload. When a reader observes a payload modification, that fence pairs with the existing acquire fence before the final revision check, ordering the opening increment before that check. The closing release increment can remain as it is. See the fence-to-fence synchronization rule: https://eel.is/c++draft/atomics.fences#2

This finding is based on memory-order analysis, not a reproduced runtime failure. All 60 recording tests passed on this revision, and the preceding pass-length fix looks correct.

lucaromagnoli: Fixed in 88f15183. You are right, and I checked the reasoning rather than taking it: `release` on the increment is a one-way barrier — it orders what precedes it and permits the relaxed payload stores after it to be hoisted ahead — so a reader observing one of those stores had no edge forcing its second look to see the increment. `std::atomic_thread_fence(std::memory_order_release)` after the increment supplies it, pairing with the acquire fence the reader already takes, and the closing increment is left alone for the reason you give.

No test can demonstrate this and I have not pretended to add one. What I could check is that the fence is really there rather than folded away, and on this ARM64 build it is: the `Change` constructor now emits two calls where the destructor emits one, and `MidiTakeRecorder.cpp.o` carries the `atomic_thread_fence` helper where it previously carried none at all — so before the fix that translation unit had no writer-side barrier of any kind. Worth noting that makes it a hardware-reachable reordering here, not only an abstract-machine one.

The header now says which end each fence is at and why release on the increment is not enough, so the next person does not have to re-derive it. Full suite green (1,821,278 assertions).

PR #2532, opened 2026-09-09
Put back five checks that only a newer clang-tidy has

lucaromagnoli:
Fixes dev, which is red on the clang-tidy gate.

## What broke

#2530 turned `WarningsAsErrors: 'bugprone-*,cert-*'` on and, in the same
change, deleted five exclusions from the pre-push hook. The first push to dev
after that failed on findings the sweep never saw:

| check | finding on dev |
|---|---|
| `bugprone-signed-bitwise` | `channel &= ~1` in `Config.hpp`, `1 << fftOrder_` and `1 << 11` in `SpectrumAnalyzerUI` |
| `bugprone-throwing-static-initialization` | `const juce::Identifier DRUM_CHAIN_TYPE("CHAIN")` -- the constructor can throw |
| `bugprone-float-loop-counter` | `for (float db = kMaxDb; db >= kMinDb; db -= 20.0f)` |

## Why the sweep missed them

The justification in #2530 was that no check exists by those names. Locally
that is true and checkable:

```
$ clang-tidy --checks='-*,bugprone-signed-bitwise' --list-checks
Error: no checks enabled.
$ clang-tidy --checks='-*,hicpp-signed-bitwise' x.cpp
warning: use of a signed integer operand with a binary bitwise operator [hicpp-signed-bitwise]
```

That is LLVM 20.1.4, where the check lives under `hicpp-`. The CI runner has a
clang-tidy new enough to carry the `bugprone-` spellings, which is where #2529
measured its 63/15/4/2/1 -- those counts were right and I read them as
measuring something else.

The trap underneath: **clang-tidy silently ignores an unknown check name.** A
sweep that reports zero and a check that is not there produce identical output,
so "I ran it and got nothing" is not evidence the name is wrong. Verifying a
check's existence against one machine's `--list-checks` is not enough when the
gate runs somewhere else.

## The fix

The five exclusions go back, in `.pre-commit-hooks/clang_tidy.py` and in
`.clang-tidy` both, so `make lint` and the editor agree with the gate rather
than diverging by LLVM version. The findings are real where they report and
deserve triaging on that version -- `1 << 11` on an int is idiomatic here and
the JUCE `Identifier` pattern is everywhere -- but that is a batch of its own,
not something to discover through a red dev.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -22,6 +22,11 @@ Checks: >-
   -bugprone-misplaced-widening-cast,
   -bugprone-nondeterministic-pointer-iteration-order,
   -bugprone-macro-parentheses,
+  -bugprone-signed-bitwise,
+  -bugprone-throwing-static-initialization,
+  -bugprone-derived-method-shadowing-base-method,
+  -bugprone-float-loop-counter,
+  -bugprone-unchecked-string-to-number-conversion,
   cert-oop54-cpp,
   performance-*,
   -performance-enum-size,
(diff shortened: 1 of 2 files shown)


lucaromagnoli:
On what the green checks here do and do not prove.

`build-and-test-macos` passed, but its gate step logged:

```
Branch point d49cba52e4a3fe46936fa61252c621df8c082c4d (1 commit(s) ahead).
No changed C++ files under magda/ to analyse.
```

This PR touches only `.clang-tidy` and `clang_tidy.py`, so the gate analysed
nothing and passed trivially. That is the same shape of non-evidence that let
the original mistake through, so it should not be read as confirmation.

What is confirmed is the mechanism, on a check LLVM 20.1.4 does have:

```
--checks='-*,hicpp-*'                       --warnings-as-errors='hicpp-signed-bitwise'  -> exit 1
--checks='-*,hicpp-*,-hicpp-signed-bitwise' --warnings-as-errors='hicpp-signed-bitwise'  -> exit 0
```

So a `-check` entry in the list does suppress it even when `--warnings-as-errors`
still names it, which is the only step in doubt. The version-specific names
themselves can only be confirmed on the runner, and the first dev push that
touches C++ under `magda/` is what will actually exercise them.

PR #2533, opened 2026-09-09
Record into a slot on the beat its launch fires

lucaromagnoli:
Slice 6 of #1895, closes #2464. Session slot recording and the
session-to-arrangement capture, both off the launcher's own sample-stamped run
edges rather than off a poll of play state.

## The edge

`LaunchHandle::advance` already cuts a block at the sample a launch fires on.
`SplitStatus` now says so — `runBeganAt` and `runEndedAt`, an event sample and a
bound — and `advanceLaunchHandles` publishes both down `SlotRunQueue` for the
publishing thread. A re-launch reports an end and a begin on the same sample,
because the run it began is not the run it ended.

`slotRun()` is the audio thread's own reader of the same answer, off the block
status the launcher wrote. `EngineSession::process` therefore advances the
launcher before the takes capture: a take following a slot starts on the sample
its launch fired on, and that is decided there.

## The take

A take carrying a `SlotRunTarget` follows that slot's run instead of the
transport: it starts on the sample the launch fired on, ends where the run
ended, and ignores the count-in and the transport's loop, since a run is on
monotonic time — a wrap or a locate is none of its business. A re-launch closes
the take rather than extending it, which is the rule an arrangement take already
keeps for a second play. Audio and MIDI alike; the MIDI take narrows the block's
events to the run's window, so a note played before the launch instant in the
same block is not in it.

## The slot

An empty slot had no handle at all: the compiler only emitted a
`SessionSlotPlayback` for a slot holding a clip, so a launch on the one kind of
slot you can record into was dropped. `ClipLane::recordSlots` is how a host arms
one, and the compiler is where an occupied slot is refused — with a diagnostic,
its own clip untouched. That is where "a slot that already holds a clip is not a
recording target" now lives.

## The capture

`SessionCapture` folds the run edges into arrangement spans. A scene is one
event because its runs share an origin, not because they were observed in one
frame; a run the transport looped inside is placed once, at the beat it was
launched on, and measured on the monotonic axis where no wrap takes anything
back. Arming after a launch keeps the beat the run began on, as the fork's
`SessionRecorder` does.

## Verification

Offline, driving the launcher and the transport:

- a launch on a beat 32 samples inside a block records from that beat (audio);
  a note played before that instant in the same block is not in the take (MIDI)
- a scene of four tracks gives four takes whose first arrival is one sample, and
  four spans sharing one origin
- a run stopped mid-block keeps what played; a re-launch is a second take
- capture start beats over a corpus of quantizations (1, 1.5, 2, 3.25, 4)
- a run across three transport loop passes is placed once
- a run a follow action ended is captured to there
- an occupied armed slot keeps its clip and is not a target

Full non-JUCE suite: 3830 passed, 13 skipped, 0 failed. JUCE suite including the
real-project null-diff corpus: all passed.

## Not in this slice

Follow actions themselves, comping, overdubbing into an occupied slot.

### DIFF
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -254,6 +254,23 @@ class EngineSession {
         return clock_.positionBeats();
     }
 
+    /// The same instant on the count no wrap or locate takes back, which is
+    /// what a capture measures a run's length on (launch/SessionCapture.hpp).
+    double monotonicBeats() const {
+        return clock_.monotonicBeat();
+    }
+
+    /**
+     * @brief The run edges the launcher publishes, for the capture (#2464).
+     *
+     * Drained on the publishing thread by SessionCapture. Outside every epoch
+     * beside the request lane it answers: a launch made while a plan was
+     * compiling still has its edge reported when the new one is live.
+     */
+    SlotRunQueue& slotRuns() {
+        return runs_;
+    }
+
     /// Callbacks in which a loop was too short to render as separate blocks.
     int loopWrapOverflows() const {
         return clock_.loopWrapOverflows();
@@ -375,6 +392,9 @@ class EngineSession {
     /// epoch, like the feed it is read beside: a request made while a plan was
     /// being compiled is still a request when the new one is live.
     LaunchRequestQueue requests_;
+
+    /// What came of them: where each run began and ended, for the capture.
+    SlotRunQueue runs_;
 
     /// The cursor. Not published and not swapped: it's where the timeline
     /// is, a property of the session rather than of any plan, and a plan
--- a/magda/engine/exec/EngineSession.hpp
+++ b/magda/engine/exec/EngineSession.hpp
@@ -3,6 +3,7 @@
 #include <farbot/RealtimeObject.hpp>
 #include <memory>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "clip/ClipSnapshotFeed.hpp"
@@ -254,12 +255,6 @@ class EngineSession {
         return clock_.positionBeats();
     }
 
-    /// The same instant on the count no wrap or locate takes back, which is
-    /// what a capture measures a run's length on (launch/SessionCapture.hpp).
-    double monotonicBeats() const {
-        return clock_.monotonicBeat();
-    }
-
     /**
      * @brief The run edges the launcher publishes, for the capture (#2464).
      *
@@ -269,6 +264,21 @@ class EngineSession {
      */
     SlotRunQueue& slotRuns() {
         return runs_;
+    }
+
+    /**
+     * @brief Ends no block could stamp, and forget them (#2464).
+     *
+     * A slot retired while it was sounding: publishClips() drops the handle,
+     * and the run it was playing ends there rather than at whatever the next
+     * thing to notice happens to be. Hand each to SessionCapture::apply. On the
+     * publishing thread, like the publish that produced them.
+     *
+     * Kept until asked for rather than dropped like a full lane: one per slot
+     * deleted while it played, which is a user's edit and not a rate.
+     */
+    std::vector<SlotRunEvent> takeRetiredRuns() {
+        return std::exchange(retired_, {});
     }
 
     /// Callbacks in which a loop was too short to render as separate blocks.
@@ -395,6 +405,10 @@ class EngineSession {
 
     /// What came of them: where each run began and ended, for the capture.
     SlotRunQueue runs_;
+
+    /// Ends this thread stamped rather than a block: a slot retired while it
+    /// was sounding (@ref takeRetiredRuns).
+    std::vector<SlotRunEvent> retired_;
 
     /// The cursor. Not published and not swapped: it's where the timeline
     /// is, a property of the session rather than of any plan, and a plan
--- a/magda/engine/io/TakeRecorder.cpp
+++ b/magda/engine/io/TakeRecorder.cpp
@@ -53,6 +53,12 @@ TakeRecorder::TakeRecorder(const LiveInputFeed& feed, const RenderContext& conte
 }
 
 void TakeRecorder::capture(const BlockInfo& block, bool countingIn, const LoopRange& loop) {
+    // A slot take is on its run's clock, not the transport's (#2464).
+    if (settings_.slot) {
+        captureRun(block);
+        return;
+    }
+
     if (state_ == State::stopped)
         return;
 
@@ -76,7 +82,7 @@ void TakeRecorder::capture(const BlockInfo& block, bool countingIn, const LoopRa
         openPass(block, loop);
     }
 
-    write(block);
+    write(block, 0, block.numSamples);
 
     // Timeline the take has covered, which is what a pass boundary is counted
     // in. Not the samples written: those are the same stretch read a latency
@@ -84,12 +90,65 @@ void TakeRecorder::capture(const BlockInfo& block, bool countingIn, const LoopRa
     arrivals_ += block.numSamples;
 }
 
-void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop) {
+void TakeRecorder::captureRun(const BlockInfo& block) {
+    if (state_ == State::stopped)
+        return;
+
+    // A stopped transport passes no timeline, so the run makes no progress and
+    // the take holds nothing. Not a stop: the run has not ended.
+    if (!block.playing)
+        return;
+
+    const auto run = slotRun(*settings_.slot);
+
+    // The slot was retired or refilled: the take ends where it stood.
+    if (run.gone) {
+        if (state_ == State::rolling)
+            stop();
+
+        return;
+    }
+
+    auto from = 0;
+    auto closing = false;
+
+    if (state_ == State::waiting) {
+        // An end in this block belongs to the run before this one.
+        if (!run.beganAt)
+            return;
+
+        from = run.beganAt->value;
+        start(block, LoopRange{}, from);
+    } else if (run.endedAt) {
+        // A launch in this block begins the next take, not this one.
+        closing = true;
+    }
+
+    const auto to = closing ? run.endedAt->value : block.numSamples;
+
+    if (to > from) {
+        write(block, from, to);
+        arrivals_ += to - from;
+    }
+
+    if (closing)
+        stop();
+}
+
+void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop, int from) {
     state_ = State::rolling;
     rolling_.store(true, std::memory_order_relaxed);
 
     startBeat_ = block.beats.start;
     startSeconds_ = block.seconds.start;
+
+    // A take that begins inside the block begins where the launch fired, not
+    // at the block's own start (#2464).
+    if (from > 0 && block.rate() > 0.0) {
+        startSeconds_ = block.seconds.start + (static_cast<double>(from) / block.rate());
+        startBeat_ = block.beatAtTime(startSeconds_);
+    }
+
     endBeat_ = startBeat_;
     startedAtLoopStart_ = atLoopStart(block, loop);
     headDrop_ = std::max(0, settings_.latencySamples);
@@ -157,7 +216,7 @@ void TakeRecorder::stop() {
     tap_.close();
 }
 
-void TakeRecorder::write(const BlockInfo& block) {
+void TakeRecorder::write(const BlockInfo& block, int from, int to) {
     const auto numSamples = std::min(block.numSamples, scratch_.getNumSamples());
     if (numSamples <= 0)
         return;
@@ -167,14 +226,18 @@ void TakeRecorder::write(const BlockInfo& block) {
 
     const auto captured = juce::dsp::AudioBlock<const float>(std::as_const(scratch_));
 
-    auto offset = 0;
+    // The input is rendered whole; only what the take keeps of it is the
+    // window the caller named.
+    auto offset = std::clamp(from, 0, numSamples);
+    const auto end = std::clamp(to, offset, numSamples);
+
     if (headDrop_ > 0) {
-        const auto dropped = std::min(headDrop_, numSamples);
+        const auto dropped = std::min(headDrop_, end - offset);
         headDrop_ -= dropped;
         offset += dropped;
     }
 
-    const auto kept = numSamples - offset;
+    const auto kept = end - offset;
     if (kept <= 0)
         return;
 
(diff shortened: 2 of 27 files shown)


magda/engine/launch/SessionCapture.cpp
```diff
@@ -0,0 +1,70 @@
...
+void SessionCapture::update(SlotRunQueue& runs) {
+    runs.drain([this](const SlotRunEvent& event) {
+        // Any edge closes what was in flight. An end closes its own run; a
+        // launch closes one whose end a full queue dropped
+        // (SlotRunQueue::overflows), which ended here at the latest.
+        if (const auto found = runs_.find(event.key); found != runs_.end()) {
+            // Unless the slot was emptied and refilled while it sounded: that
+            // run ended when its handle was retired, which nothing stamped, so
+            // it is dropped rather than placed at a length nobody measured.
+            if (found->second.incarnation == event.incarnation)
```
lucaromagnoli: [P1] Include handle retirement in the timestamped run lifecycle

Deleting a playing slot retires its handle without publishing an end event. This leaves its run in runs_: if capture is disarmed later, finish() extends the span through time when the slot no longer sounded; if a replacement incarnation launches first, this branch silently discards the previous captured run. The event stream therefore cannot reconstruct what actually played across an ordinary slot deletion/refill. Retirement needs an authoritative end boundary in the same lifecycle as launch and stop, rather than dropping or guessing the old span in the consumer.

magda/engine/io/TakeRecorder.cpp
```diff
@@ -76,20 +82,73 @@ void TakeRecorder::capture(const BlockInfo& block, bool countingIn, const LoopRa
...
 }
 
-void TakeRecorder::start(const BlockInfo& block, const LoopRange& loop) {
+void TakeRecorder::captureRun(const BlockInfo& block) {
+    if (state_ == State::stopped)
+        return;
+
+    // A stopped transport passes no timeline, so the run makes no progress and
+    // the take holds nothing. Not a stop: the run has not ended.
+    if (!block.playing)
```
lucaromagnoli: [P1] Observe run transitions even while transport is stopped

advanceLaunchHandles() still applies requests and advances handles in stopped transport blocks, but this early return skips their transient run edges. Reproduction: start a slot take, pause transport, issue an immediate slot stop while paused, then resume transport. The stop edge has already been overwritten by a later blockStatus(), so the recorder remains rolling and resumes writing although the slot has ended. The same early return exists in MidiTakeRecorder::captureRun(). Run lifecycle observation needs to be independent of whether the current block contributes samples; otherwise both recorders can permanently miss a boundary.

magda/engine/exec/EngineSession.hpp
```diff
@@ -254,6 +254,23 @@ class EngineSession {
         return clock_.positionBeats();
     }
 
+    /// The same instant on the count no wrap or locate takes back, which is
+    /// what a capture measures a run's length on (launch/SessionCapture.hpp).
+    double monotonicBeats() const {
+        return clock_.monotonicBeat();
```
lucaromagnoli: [P1] Synchronize capture boundaries with the audio-thread event stream

This accessor returns TransportClock::monotonicBeat_, a plain double documented as audio-thread-only and updated by advance(). SessionCapture runs on the publishing thread and disarm() needs this timestamp, so using this accessor for its intended purpose while processing audio introduces a C++ data race. The capture boundary also needs ordering relative to the queued run edges. Please define a synchronized boundary/timestamp contract for capture instead of exposing the mutable audio-thread clock directly.

magda/engine/launch/SessionCapture.cpp
```diff
@@ -0,0 +1,70 @@
...
+        return;
+
+    const auto length = monotonicBeat - run.startMonotonicBeat;
+
+    // A run that ended on the sample it began on played nothing.
+    if (length <= 0.0)
+        return;
+
+    captured_.push_back(CapturedRun{
+        .key = key, .startBeat = run.startBeat, .lengthBeats = length, .origin = run.origin});
```
lucaromagnoli: [P2] Preserve stable material identity in collected runs

Run tracks an incarnation, but CapturedRun drops it and returns only a reusable SlotKey plus timing. If clip A finishes playing and that slot is refilled with clip B before collect() is consumed, the result cannot tell the host which material produced the span; resolving the current slot can capture B instead of A. Two incarnations captured before collection are likewise indistinguishable by material. The capture contract needs a stable material reference/identity retained through collection, with an explicit way for the host to resolve it, rather than requiring consumers to infer historical material from the current slot.

lucaromagnoli (commented):
Architecture review: these findings point to an incomplete run-lifecycle contract. Please resolve how retirement, transport-independent run transitions, capture boundaries, and material identity are represented before adding recorder-specific conditional handling. I would block merging on the P1 findings.

Validation: the existing focused [capture],[2464] tests passed (18 cases, 119 assertions). The findings below are from source review; those tests do not cover these scenarios.

magda/engine/launch/SessionCapture.cpp
lucaromagnoli: Fixed in 98c0854d. Retirement now has an end in the same lifecycle as launch and stop, stamped by the one thing that knows it: `RuntimeStateStore::publishHandles` reads where the run had been advanced to (`playedMonotonicRange()->end`, `playedSampleRange()->end`) and emits an `ended` event for every handle it is about to drop that was sounding. Read on the far side of `feed.publish`, so the audio thread can no longer reach the handle — safe on that thread by the same rule that makes the erase safe there.

They travel on their own channel (`EngineSession::takeRetiredRuns`) rather than down the lane, because the lane has one writer and it is the audio thread; `SessionCapture::apply` takes one.

The discard you spotted is gone too, and not by ordering the two channels: a run is now held under the handle that played it (`RunKey{key, incarnation}`) rather than under the slot, so the refill's launch and the retired run's end can arrive in either order and both spans survive. Test: "a run is held under the handle that played it" — it fails with `1 == 2` if the key is the slot alone.

magda/engine/io/TakeRecorder.cpp
lucaromagnoli: Fixed in 98c0854d, in both recorders. The `run.gone` check moved above the transport question, and a stopped block now contributes `samples = 0` rather than being skipped, so the edges are read wherever they land:

```cpp
const auto samples = block.playing ? block.numSamples : 0;
```

Your reproduction is the test ("A slot stop asked for while the transport is stopped ends the take", audio and MIDI): without the fix the recorder is still `rolling()` and the audio file is 16000 samples instead of 8000. The launch half had the same hole — a launch applied while paused left the take permanently `waiting` and `finish()` came back empty — and the one change covers both.

Note for the record: on a stopped block the launcher's event always lands on sample 0 (`eventAtMonotonicBeat` returns `start()` on an empty seconds range), so `start()` stamps the paused cursor. If the user then resumes elsewhere, a slot take's `startBeat` is the old cursor while its audio comes from the new position. Left as is: a session clip is placed by its scene, not by a beat, so that field is informational for a slot take, and it matches what the arrangement path does.

magda/engine/exec/EngineSession.hpp
lucaromagnoli: Fixed in 98c0854d — the accessor is deleted. You are right on both counts: `positionBeats_` is `std::atomic<double>` and documented readable anywhere, `monotonicBeat_` is a plain double documented audio-thread, and I exposed the second as if it were the first.

The boundary now comes from the lane rather than from a second clock. `SlotRunQueue::reachedBeat()` is stored by `advanceLaunchHandles` **after** that block's edges (release), and `reached()` is read **after** a drain (acquire), so a boundary can never be a beat whose edges the capture has not folded in. `disarm()` takes no argument and drains first.

That is the ordering half as well as the race half: test "disarming takes the edges before it reads the boundary" — a run that stopped two beats before the button was pressed. Without the drain-first it reads 5 beats instead of 4, which is exactly the drift a poll has.

magda/engine/launch/SessionCapture.cpp
lucaromagnoli: Fixed in 98c0854d. `CapturedRun` carries the `incarnation` that played it, and the capture keys its in-flight runs by it, so two incarnations captured before a `collect()` are distinguishable and neither displaces the other.

On resolving it: the incarnation is the store's own handle identity, published to the request lane by `publishHandles` via `setIncarnations`, so the host already receives it at the moment it publishes the clips that decided it — that is where a (slot, incarnation) → material map is built, and the header now says so. I stopped short of putting a `SourceId` or `ClipId` in `CapturedRun` itself: the engine deliberately never names a clip (`RecordedTake` does not either, because an event names a `SourceId` and the pool that hands those out is the host's), so carrying one here would be the first place it did.

magda/engine/launch/SessionCapture.cpp
```diff
@@ -0,0 +1,77 @@
...
+        run.capturing = true;
+}
+
+void SessionCapture::disarm() {
+    // Before the boundary is read, so it cannot be a beat whose edges have not
+    // been folded in: a run that stopped on its own ends where it stopped
+    // rather than where the button was pressed.
+    update();
+
+    const auto until = lane_.reached();
```
lucaromagnoli: [P1] Bind the disarm boundary to the events actually consumed

The atomic accessor removes the data race, but update() followed by reached() does not establish the claimed ordering. drain() snapshots written_ once. The audio thread can then publish a stop and update reached_ after that snapshot, before this load. Disarm sees the newer boundary while the stop remains unconsumed, finishes the run beyond its actual end, and marks it non-capturing; consuming the stop later cannot correct the collected span.

A standalone check deterministically interleaved a producer advance after drain captured its cursor: the drain consumed only the begin, reached() returned beat 5, and the stop at beat 4 was still queued. Acquire/release makes the preceding writes visible; it does not make the already-completed drain consume them. Please define a coherent event/boundary publication contract, such as a boundary associated with a committed queue position or an ordered capture-boundary event, rather than treating a separate timestamp as the drain's boundary.

magda/engine/launch/SessionCapture.cpp
```diff
@@ -0,0 +1,77 @@
...
+void SessionCapture::update() {
+    lane_.drain([this](const SlotRunEvent& event) { apply(event); });
+}
+
+void SessionCapture::apply(const SlotRunEvent& event) {
+    const RunKey of{event.key, event.incarnation};
+
+    // Held under the handle that played it, so a run whose end arrives after
+    // the next incarnation's launch is still the run that ended.
+    if (const auto found = inFlight_.find(of); found != inFlight_.end()) {
```
lucaromagnoli, line 31: [P1] Preserve retirement ends that arrive before their queued begin

The new retirement channel is still order-dependent for the same incarnation. A slot can launch on the audio thread, be deleted before capture drains the begin, and have its retirement delivered through apply() first. Since inFlight_ has no matching run yet, apply(end) drops the end. update() subsequently consumes the queued begin and leaves the retired run sounding indefinitely. This is explicitly allowed by the header's claim that order against update() does not matter.

Standalone reproduction using the updated code: arm; queue begin at beat 1; apply retirement end at beat 4; update. Actual result: sounding=1, captured=0 (expected 0 and 1). Disarming at beat 8 then produces a seven-beat span instead of three. Keying by incarnation handles a replacement's begin overtaking an old end, but not an end overtaking its own begin. Retirement must participate in an ordered lifecycle contract, or the consumer must explicitly retain and reconcile unmatched boundaries. The current tests deliver the old begin before retirement and miss this case.

lucaromagnoli (commented):
Re-review of 98c0854d: the paused-transport recording issue is addressed, and CapturedRun now retains incarnation identity. Two P1 ordering gaps remain. The lifecycle contract still needs to order run edges, retirement, and capture boundaries coherently; separate delivery channels and an independent atomic timestamp do not establish that ordering.

Validation: 23 focused tests passed (142 assertions). Additional standalone checks against the updated SessionCapture/SlotRunQueue reproduced retirement-before-begin loss and demonstrated a drain returning a watermark beyond an unconsumed edge. No repository code changed.

magda/engine/launch/SessionCapture.cpp
lucaromagnoli: You are right, and my ordering argument was wrong: acquire made the writes visible, it never made the already-finished drain consume them. Fixed in e6aaf7fa.

The boundary is now handed out by the drain that earned it, read before the cursor:

```cpp
template <typename Fn> double drain(const Fn& fn) {
    const auto until = reached_.load(std::memory_order_acquire);
    const auto end = written_.load(std::memory_order_acquire);
    ...
    return until;
}
```

Every edge up to `until` is taken by this same drain (its cursor is loaded after `until`, so it is at least what it was when `until` was stored), and anything the drain missed belongs to a later block, whose beats are ≥ `until` — so a boundary can never pass an end still queued. `SlotRunQueue::reached()` is deleted: pairing the beat with a drain is the contract, so there is now no accessor to pair it with anything else.

Pinned with a threaded case, following test_record_tap.cpp's own read-protocol tests: one writer publishing runs of a known length (edges, then the beat the block reached) while the reader disarms and re-arms, and no captured span may outlast the run it came from. With the beat read after the cursor instead it reports a 63-beat span for a run that sounded 3 beats, on every run. It asserts on the longest span after joining rather than per span, so a regression reports a failure instead of aborting the binary.

magda/engine/launch/SessionCapture.cpp
lucaromagnoli: Correct, and the header claim was wrong in the direction I had not tested. Fixed in e6aaf7fa: an end with no run to close waits in `unmatchedEnds_` and is applied once the drain has caught up with its launch.

Two details that matter, both learned by getting them wrong first:

- It is applied **after** the drain, not at the first launch that matches. Two launches of one handle can be queued together (a re-launch is `ended` + `began` on one sample, same incarnation), and the retirement's end belongs to the run the lane leaves in flight, not to the first one.
- A launch that ends a run of its own must **not** clear the waiting end. I had an `unmatchedEnds_.erase` there as hygiene; it made exactly the two-launch case fail (one span instead of two, second run sounding for ever). A waiting end can only ever match a launch queued before the retirement, since nothing can launch a handle that has gone, so there is nothing for it to wrongly close.

Your reproduction is now a case of its own — arm, queue the launch, apply the retirement end, then update — and without the fix it reports your numbers exactly: `sounding == 1`, `captured == 0`. A second case covers the two queued launches: 2 spans of 2 beats, and `sounding == 0`.

magda/engine/launch/SessionCapture.cpp
```diff
@@ -0,0 +1,99 @@
...
+    reconcile();
+}
+
+void SessionCapture::apply(const SlotRunEvent& event) {
+    const RunKey of{event.key, event.incarnation};
+
+    // Held under the handle that played it, so a run whose end arrives after
+    // the next incarnation's launch is still the run that ended.
+    if (const auto found = inFlight_.find(of); found != inFlight_.end()) {
+        finish(of, found->second, event.monotonicBeat);
```
lucaromagnoli, line 32: [P1] Do not apply a retirement end to an earlier run whose relaunch is still queued

The unmatched-end fix only handles retirement arriving when no run is tracked. If capture already knows an earlier run on this handle, this branch immediately closes that earlier run with the retirement timestamp, even though the retirement belongs to a later run whose relaunch has not been drained.

Reproduction against e6aaf7fa:
1. Arm; queue begin at beat 1; publish reached beat 2; update() so that run is tracked.
2. Queue ended at beat 3 and began at beat 3 for a relaunch of the same incarnation; publish reached beat 5, without draining.
3. Delete the slot and apply its retirement end at beat 5, then update().

Expected: spans [1,3) and [3,5), each two beats. Actual: one span starting at 1 with length 4. Retirement first closes the old run at 5; the queued end at 3 then becomes unmatched, and reconcile() uses it to close the new run at its own start, discarding that span.

The new two-launch test leaves both launches undrained, so it misses this partially consumed history. This is still a lifecycle contract issue: (slot, incarnation) names the handle, not the particular run an end closes. Retirement must be ordered after that handle's queued transitions or identify its exact run; immediate application plus unmatched-end reconciliation cannot establish either guarantee.

lucaromagnoli (commented):
Re-review of e6aaf7fa: the previous clock-ordering race is addressed, but one P1 retirement-ordering gap remains. A handle incarnation identifies a handle rather than an individual run, and the separate retirement channel can still close the wrong run. Please resolve the run identity/ordering contract rather than extending unmatched-end bookkeeping around individual cases.

Validation: existing focused tests passed (26 cases, 156 assertions). A standalone reproduction compiled against the latest SessionCapture source produced one four-beat span instead of two two-beat spans. No repository code changed.

magda/engine/launch/SessionCapture.cpp
lucaromagnoli: Right again, and it makes the unmatched-end bookkeeping the wrong shape rather than incomplete. Fixed in 49f5ba81, and the fix deletes code:

```cpp
void SessionCapture::apply(const SlotRunEvent& event) {
    update();      // the lane first, always
    fold(event);
}
```

The ordering guarantee you asked for is available for free, and it comes from the publish rather than from anything here: retiring a handle waits for the block the callback is in, so by the time an end reaches the capture, nothing more can ever be published for that handle. One drain therefore folds *every* transition it had, and what the drain leaves in flight is the run this end closes — not "an earlier run this handle happened to be on".

That subsumes the previous round's case as well: an end can no longer arrive before its own launch, so `unmatchedEnds_` and `reconcile()` are gone. One state fewer, and the ordering holds by construction instead of case by case.

Your reproduction is now a case of its own — first run tracked, relaunch queued and undrained, then the retirement — and without the drain it reports your numbers exactly: one span of 4.0 beats where there should be two of 2.0. The earlier two-launch case stays, and it is the one that catches the mirror-image mistake (I had an `erase` that cleared a waiting end, which is why that case exists).

Full non-JUCE suite 3840 passed / 13 skipped; JUCE suite including the null-diff corpus all passed.

PR #2535, opened 2026-09-09
Pin the clang-tidy gate, and hold it advisory until the engine pass

lucaromagnoli:
## The problem

There is more than one self-hosted macOS runner and they do not carry the same
LLVM:

| runner | LLVM |
|---|---|
| `macOS-ARM64` | 20.1.4 |
| `LRomagnoli-UKMAC25` | newer -- it has the `bugprone-` spellings of checks that are `hicpp-`/`cert-` on 20 |

The gate took whichever clang-tidy the runner had, so the findings it reported
depended on which machine picked up the job, and nothing in the log said which
one that was. The step's own comment asserted the opposite -- "it is the machine
the pre-push hook runs on ... rather than two releases that disagree about what
a finding is" -- which held when there was one runner.

**The divergence is silent.** clang-tidy ignores an unknown check name rather
than complaining, so a release missing a check reports clean and passes. That
is exactly how #2530 landed: one runner answered "no checks enabled" for five
names, I read that as proof they were fictitious and deleted the exclusions,
and dev went red when the same code met the runner that has them. #2529 had
measured those five correctly on the other machine.

## The fix

The hook prints what it resolved, on every run:

```
clang-tidy 20 (/opt/homebrew/opt/llvm/bin/clang-tidy)
```

and honours `CLANG_TIDY_REQUIRE_MAJOR`. CI sets it to `20`, the version every
finding count recorded in `.clang-tidy` was measured on, and prefers the
`llvm@20` keg explicitly since it is keg-only and so on nobody's PATH. A runner
without it fails with what to install:

```
This gate is pinned to LLVM 21 and found 20.
Findings differ between releases, so an unpinned gate reports a different set
depending on which runner takes the job. Install LLVM 21 and point CLANG_TIDY
at it, or change CLANG_TIDY_REQUIRE_MAJOR if the pin is meant to move.
```

Moving the pin later means re-triaging the tier, not just editing the number --
LLVM 21 adds five checks currently excluded, worth roughly 85 findings by
#2529's sample.

## Note for whoever merges

If `LRomagnoli-UKMAC25` has no `llvm@20`, its next run fails with the message
above rather than silently analysing with something else. `brew install llvm@20`
there settles it. That failure is the point of the pin, but it is better known
in advance than discovered on a red dev.

## Verification

- `clang_tidy.py` parses, `ci.yml` is valid YAML
- no pin set: prints the version, analyses, exits 0
- pin matching (20): analyses, exits 0
- pin mismatched (21): exits 1 with the message above, before analysing anything

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -151,9 +151,16 @@ Checks: >-
 # out of scope by construction. Triaging them was considered and dropped
 # (2026-09-09); reviving it means moving the pin in ci.yml first.
 #
-# The bug-catching tier is at zero over all 705 magda/ translation units,
-# magda/engine included, so it is binding from here on.
-WarningsAsErrors: 'bugprone-*,cert-*'
+# Off until magda/engine is written and swept (decided 2026-09-09). The tier is
+# at zero over all 705 magda/ translation units, engine included, so this is
+# ready to carry 'bugprone-*,cert-*' - but gating shipping code while the engine
+# is still being written sets a bar that moves the moment it lands, and the
+# sweeps have excluded that directory throughout.
+#
+# The gate still runs and still reports; it is advisory, via --advisory in
+# .pre-commit-config.yaml and CLANG_TIDY_ADVISORY in ci.yml. Restore all three
+# together, after the engine pass.
+WarningsAsErrors: ''
 
 # Header filter. Anchored on a leading slash: the old unanchored `tests` matched
 # third_party/JUCE/.../unit_tests/*.h, and it never covered magda/engine at all.
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -146,8 +146,10 @@ Checks: >-
 # --list-checks saying the names were unknown, and the first push to dev failed
 # on them: an unknown check name is silently ignored, so a clean sweep and an
 # absent check look exactly alike. The findings are real where they report -
-# `1 << 11`, a file-scope juce::Identifier, a float loop counter - and want
-# triaging on that version before the exclusions come off.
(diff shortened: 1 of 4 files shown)


lucaromagnoli:
Added: the gate is now **advisory** until `magda/engine/` is written and swept.

Gating shipping code while a whole directory has never been analysed sets a bar
that moves the moment the engine lands. Every sweep in this rollout excluded
`magda/engine/`, so "the tier is at zero" is a statement about the other 705
translation units, not about what the gate meets next.

The analysis is the reason the step exists, so it keeps running and keeps
printing findings. Only the blocking waits, across three switches meant to be
restored together:

| where | now | restore to |
|---|---|---|
| `.clang-tidy` | `WarningsAsErrors: ''` | `'bugprone-*,cert-*'` |
| `.pre-commit-config.yaml` | `clang_tidy.py --advisory` | drop the flag |
| `ci.yml` | `CLANG_TIDY_ADVISORY: '1'` | drop the variable |

An advisory run says so on every finding:

```
clang-tidy findings above. ADVISORY ONLY - not failing.
Gating is off until magda/engine is written and swept; see WarningsAsErrors in .clang-tidy.
```

That banner is deliberate. The hook this rollout replaced ended its command with
`|| true` and read as a passing gate through six sweeps -- an advisory run that
is quiet about being advisory is the same failure with better manners.

Verified by reintroducing the `BindingRegistry` defect that took dev red:
gating mode exits 1, `--advisory` exits 0, `CLANG_TIDY_ADVISORY=1` exits 0, and
findings print in all three.

Issue #2465, opened 2026-09-06
Recording slice 7: a take that survives a plan swap


lucaromagnoli:
Slice 7 of #1895. Depends on slice 3. The one slice here whose failure loses a take rather than making a noise.

## Why it is a slice

The design conclusion for the epic draws the line by a litmus: state that can be recomputed from the model, the transport and the parameter lanes is recomputed at a swap, and everything else is accumulated history that has to be carried. An in-flight recording is the clearest case on the carry side. It is a file handle, a write position, the take's start, and a queue with samples in it that no one has written yet.

`RuntimeStateStore` already owns the objects a plan renders through, keyed by section-aware model identity, so that recompiling a plan does not rebuild an instrument. A take has to be checked out of the same store for the same reason, and for higher stakes: an instrument rebuilt mid-note clicks, a take rebuilt mid-pass is a split file or a hole.

Any edit recompiles. Adding a device to another track, renaming something, moving a clip: none of them are about the armed track, and all of them swap the plan under it.

## Scope

- Recording state declared as carried state, checked out of the store by the same identity the differ matches on, so the swap moves nothing.
- A recompile during a pass leaves the writer, the queue and the position untouched, and the file is continuous across the swap sample.
- The edits that genuinely invalidate a take have one defined outcome rather than several: the track loses its input, the track is deleted, arm is switched off mid-pass. In each case the take closes cleanly and what exists is materialised. A file is never left open and a partial take is never silently dropped.
- Nothing is allocated or freed on the audio thread at the swap, which is the store's existing contract and is worth restating because a writer looks like something that wants closing at exactly the wrong moment.

## Not in this slice

The swap-to-self null test the design notes propose as mechanical enforcement of carried state generally. That is validation infrastructure and belongs to #1896; this slice needs its own cases regardless.

## Verification

Offline, recording across driven recompiles:

- a pass recorded over N swaps produces one file whose content is continuous, with no gap, no repeat and no discontinuity at any swap sample
- a swap that touches the armed track's own chain (a device added in front of the input) still carries the take
- disarming mid-pass closes the take and materialises what was recorded
- deleting the recording track mid-pass closes the file and loses nothing that was already written
- a swap during a pass neither allocates nor frees on the audio thread

lucaromagnoli:
Merged in #2542.

PR #2542, opened 2026-09-09
Carry a take across a plan swap

lucaromagnoli:
Slice 7 of #1895, closes #2465. A recording in flight is carried across a plan
swap, and the edits that end one end it cleanly.

## Why it is in the store

An in-flight take is not recomputable from the model, the transport or the
parameter lanes. It is a file handle, a write position, the take's start and a
queue holding samples nobody has written yet. So it goes where a device instance
goes: `RuntimeStateStore`, keyed by `TakeKey` — the identity the differ matches
an input op on, minus the op, since a track has one live audio input and one
live MIDI input and so at most two takes.

Any edit recompiles. Adding a device to another track, a rename, a clip moved:
none of them are about the armed track, and all of them swap the plan under it.
None of them touch the take, so the writer, the queue and the position are the
same objects on both sides and the file is continuous across the swap sample.

## What ends a take

`RuntimeStateIds` gains the takes the model still says may run: a track armed,
with an input of that material. It is the one entry there naming something
*allowed to continue* rather than something that exists, because the plan cannot
answer it — a track with monitoring on still compiles the same input op after
you disarm it.

The three edits that end a take each have one outcome: disarm, the input taken
away, the track deleted. Each closes the take and hands it back through
`takeClosedTakes()` with the tap it was writing to, so what was recorded up to
the edit becomes a clip. A file is never left open and a partial take is never
dropped. A second take started on a key that already has one displaces the first
the same way rather than destroying it under the callback.

## Ordering

- **An edit is one boundary for the plan and for the takes.** `PreparedRender`
  carries the model's eligible `TakeKey`s, so a block feeds a take only when the
  epoch it is rendering names it. The recording set says what exists; the epoch
  says what may record. Without that, a callback could hold the new plan and the
  old set at once and go on feeding a take the edit had ended — for more than
  one callback if the publishing thread was descheduled.
- A take leaves the callback's set first, in a publish that waits for the block,
  and only then is it released from the store.
- Takes are closed after the plan swap, since the edit only counts once its plan
  is playing.
- The tap travels with the take, because `finish()` writes to it. `startTake`
  therefore takes a factory rather than a built recorder: closing whatever the
  key was already recording takes that key's tap away, so the tap to build
  against does not exist until the close has happened.
- A take no edit reached is not touched at all — no publish, no wait.

Retention is the store's rather than an order its caller has to keep:
`releaseDeleted` will not evict a take or a tap the store still holds, whatever
the model IDs say, on the same reading the live plan already gets there. A
store-level case drives that eviction directly and fails without it.

`EngineSession::recordingFeed()` is gone. The session is the only thing that
publishes the callback's set now, which is what makes the ordering above an
invariant rather than a convention.

## Verification

`tests/engine/test_take_across_swap.cpp`, driving a real `EngineSession`:

- a pass over ten swaps is one file whose every sample follows the one before it,
  with no gap, no repeat and no discontinuity at any swap sample
- a device added in front of the armed track's own input still carries the take
- disarming mid-pass closes it and materialises 256 samples
- a track that loses its input closes the take it was feeding
- deleting the recording track closes the file with nothing lost
- an edit that ends no take publishes nothing to the callback
- a second take on the same key closes the one it displaces
- a swap during a pass neither allocates nor frees on the audio thread
- the store keeps the tap of a take it is still holding
- a take the live plan's model does not name is not fed
- a take carried and then stopped against callbacks running on their own thread
  takes no further sample once `stopTake` has returned
- a callback standing inside the window between the plan going live and the
  takes the edit ended leaving the set: the ended take gains nothing, a take on
  an untouched track keeps going

The last one is new machinery: `tests/AllocationWatch` replaces the global
`operator new`/`delete` with counters armed per thread, so a case can assert that
a block allocates and frees nothing. It forwards to malloc and free, so nothing
else in the binary behaves differently for it being linked in, and it self-tests
(`allocationWatchWorks()`) rather than passing vacuously where a sanitizer
runtime supplies its own.

All twelve were checked against a broken engine: naming no take at all fails the
four carry cases, naming every take fails disarm and input-gone, one deliberate
allocation in `process()` fails the allocation case, dropping the eviction guard
fails the store case, dropping the eligibility gate fails both the gate case and
the in-window one (20 of 20), and removing the wait from `stopTake` fails the
concurrent stop 12 of 12.

Full non-JUCE suite: 3852 passed, 13 skipped, 0 failed. JUCE suite including the
real-project corpus: all passed.

The last two run callbacks on a thread of their own and publish against them.

One stops the take rather than disarming it, deliberately: with the gate in place
a disarmed take goes quiet at the swap, so only an explicit stop leaves blocks
being fed right up to the publish inside `stopTake`, where the wait is still the
only thing between `finish()` closing the file and a callback writing to it.

The other stands inside the window rather than racing it. `EngineSession` carries
one hook, `betweenPlanAndTakesForTest`, called after the plan is live and before
`closeUnnamedTakes` — empty otherwise, one null check per publish. It is the only
one, and deliberately: every other ordering in `publish()` is a wait, and a wait
cannot be observed from outside, while this is a gap. The case parks a callback
from within the hook, so the recording set is still the pre-edit one by
construction, and reads two consecutive blocks. Clean over 30 consecutive runs of
the tag.

## Not in this slice

The swap-to-self null test #2465 hands to #1896. Nothing here starts a recording
from the model — the host side of arm-and-roll is still ahead.

### DIFF
--- a/magda/engine/exec/EngineSession.cpp
+++ b/magda/engine/exec/EngineSession.cpp
@@ -111,10 +111,9 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
     // it would be counting from one again by its next block (#2122).
     live_->executor.clearUnboundValueTaps();
 
-    // After the swap, because until it the plan playing was the one the edit
-    // has not happened to yet, and before the release below, because that is
-    // what frees a tap a take still writes to. A take the edit did not reach is
-    // not touched at all -- no publish, no wait, nothing moved (#2465).
+    // The takes this edit ended (#2465). After the swap, since the edit only
+    // counts once its plan is playing. Before releaseDeleted, since that frees
+    // the tap a take writes to as it closes.
     closeUnnamedTakes(modelIds);
 
     // Safe only now: before the swap, everything about to be destroyed was
@@ -134,9 +133,8 @@ void EngineSession::publishTakes() {
 
 void EngineSession::startTake(const TakeKey& key, const RecordTapSettings& settings,
                               const std::function<std::unique_ptr<TakeCapture>(RecordTap&)>& make) {
-    // First, and before the tap: what this key was recording leaves with its
-    // own tap, so the take about to be built gets a fresh one rather than the
-    // one somebody is about to be handed.
+    // Whatever this key was already recording, closed first. Its tap leaves
+    // with it, so the take built below gets a fresh one.
     if (auto displaced = stopTake(key); displaced.take != nullptr)
         closed_.push_back(std::move(displaced));
 
@@ -153,8 +151,8 @@ ClosedTake EngineSession::stopTake(const TakeKey& key) {
     if (released.take == nullptr)
         return {};
 
-    // After the release and before the caller has it: the set published here is
-    // the one without it, so when this returns the callback is out of it.
+    // The set without it, published before the caller has it. Once this
+    // returns the callback is out of the take, so finishing it is safe.
     publishTakes();
     return ClosedTake{key, std::move(released.take), std::move(released.tap)};
 }
@@ -164,8 +162,8 @@ void EngineSession::closeUnnamedTakes(const RuntimeStateIds& modelIds) {
     if (unnamed.empty())
         return;
 
-    // Every one of them out of the store first, then one publish for the lot:
-    // a scene of armed tracks deleted together is one wait, not one each.
+    // All of them out of the store, then one publish: a scene of armed tracks
+    // deleted together costs one wait rather than one each.
     std::vector<ClosedTake> closed;
     closed.reserve(unnamed.size());
     for (const auto& key : unnamed) {
--- a/magda/engine/exec/EngineSession.cpp
+++ b/magda/engine/exec/EngineSession.cpp
@@ -115,6 +115,9 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
     // it would be counting from one again by its next block (#2122).
     live_->executor.clearUnboundValueTaps();
 
+    if (betweenPlanAndTakesForTest)
+        betweenPlanAndTakesForTest();
+
     // The takes this edit ended (#2465). After the swap, since the edit only
     // counts once its plan is playing.
     closeUnnamedTakes(modelIds);
--- a/magda/engine/exec/EngineSession.cpp
+++ b/magda/engine/exec/EngineSession.cpp
@@ -112,8 +112,7 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
     live_->executor.clearUnboundValueTaps();
 
     // The takes this edit ended (#2465). After the swap, since the edit only
-    // counts once its plan is playing. Before releaseDeleted, since that frees
-    // the tap a take writes to as it closes.
+    // counts once its plan is playing.
     closeUnnamedTakes(modelIds);
 
     // Safe only now: before the swap, everything about to be destroyed was
--- a/magda/engine/exec/EngineSession.cpp
+++ b/magda/engine/exec/EngineSession.cpp
@@ -37,6 +37,10 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
     prepared->values = std::move(values);
     prepared->context = context;
 
+    // Sorted already: a std::set of them is ordered by the same operator< the
+    // block's lookup uses.
+    prepared->takes.assign(modelIds.takes.begin(), modelIds.takes.end());
+
     // The metronome is prepared against the device, not the plan, so it is
     // shared with the epoch it replaces unless the device changed. Sharing is
     // what keeps a click that is sounding from being cut in half by an edit
@@ -304,9 +308,16 @@ void EngineSession::process(int numSamples, juce::AudioBuffer<float>& output,
 
         // Before the plan and outside it: a take holds the input the device
         // captured, not what the track's chain went on to make of it.
+        //
+        // The set says which takes exist and the epoch says which may record,
+        // and the epoch is the one this block is rendering. So an edit that
+        // ended a take stops feeding it on the block that first renders its
+        // plan, rather than on whichever later block the publishing thread got
+        // the reduced set out by (#2465).
         if (takes)
-            for (auto* recorder : *takes.get())
-                recorder->capture(segment.block, segment.countingIn, transport->loop);
+            for (const auto& entry : *takes.get())
+                if (std::binary_search((*render)->takes.begin(), (*render)->takes.end(), entry.key))
+                    entry.take->capture(segment.block, segment.countingIn, transport->loop);
 
         // Where the transport is, for the thread that reads ahead of it. A
         // relaxed store of a double, before the block rather than after: the
--- a/magda/engine/exec/EngineSession.cpp
+++ b/magda/engine/exec/EngineSession.cpp
@@ -111,6 +111,12 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
     // it would be counting from one again by its next block (#2122).
     live_->executor.clearUnboundValueTaps();
 
+    // After the swap, because until it the plan playing was the one the edit
+    // has not happened to yet, and before the release below, because that is
+    // what frees a tap a take still writes to. A take the edit did not reach is
+    // not touched at all -- no publish, no wait, nothing moved (#2465).
+    closeUnnamedTakes(modelIds);
+
     // Safe only now: before the swap, everything about to be destroyed was
     // still reachable from the plan the audio thread was rendering. The plan
     // that is live goes in as well, so what it names survives however stale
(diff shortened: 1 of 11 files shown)


lucaromagnoli:
[P2] End takes at the same callback boundary as the plan edit

Checked 5e9e21db. The tap-retention clarification and test are useful, but the publication gap remains in `EngineSession::publish()` around lines 97–116:

1. `published_.nonRealtimeReplace(prepared)` makes the disarmed/deleted/input-less track's new epoch visible.
2. A callback can acquire that new epoch and the old `RecordingFeed` before `closeUnnamedTakes()` publishes the reduced recording set.
3. That callback still calls `capture()` on the ended take, adding samples after the edit became effective. If the publishing thread is descheduled between these operations, this can last more than one callback.

The separate waits protect object lifetime, but do not make these two state changes one recording boundary. Moving take removal before the plan swap would instead allow recording to end while the old plan is still live.

Suggested direction: keep the recorder, writer, queue, and tap owned outside the plan in `RuntimeStateStore`. Add an immutable snapshot of the model's eligible `TakeKey`s to `PreparedRender`, populated before publication. Give each recording-feed entry its key alongside its recorder pointer, and gate `capture()` using eligibility from the render epoch already held by that callback. A sorted vector with an allocation-free lookup would suffice. Eligibility must come from `modelIds.takes`, since a monitoring input op can survive disarm.

With that arrangement, the first callback on the new epoch cannot capture an ended take even if the old recording feed is still published. The existing later removal/wait can then handle retirement and handoff safely. Unrelated edits retain the same recorder objects, and starting/stopping a take can still use the separate recording feed without recompiling the graph.

This needs a small change to what the session publishes atomically, not a broader engine redesign or moving take ownership into `RenderPlan`.

Please add a deterministic concurrent regression test that pauses the publishing thread after the epoch swap but before recording-feed removal, runs a callback, and verifies that a disarmed/deleted/input-less track's take gained no samples. Also verify that an eligible take continues through that interval. The current nine `[2465]` tests pass (93 assertions), but do not exercise this interleaving.

lucaromagnoli:
Confirmed and fixed in eb4da104, taking the suggested direction.

`PreparedRender` now carries the model's eligible `TakeKey`s, sorted, assigned from `modelIds.takes` before the plan is published. `RecordingTakes` entries carry their key, and the capture loop feeds a take only when the epoch that block is rendering names it — a `std::binary_search` over a vector, so no allocation, which the existing allocation case already holds the block to. The later removal and its wait still do retirement and handoff, and an unrelated edit still keeps the same recorder objects.

So the first block on the new epoch is already the first block that stops feeding an ended take, whether or not the publishing thread has got the reduced set out.

Agreed that removing takes before the swap is the same bug mirrored — recording would stop while the plan that still allows it is playing. Noted that in the commit so it does not get re-proposed.

**On the deterministic regression.** A genuinely concurrent interleaving is not deterministically assertable from outside: with a background thread driving `process()`, a test cannot observe which block was the first on the new epoch, so "no sample after the swap" has no fixed expectation to compare against, and the failing case only appears when the publishing thread is descheduled.

What is deterministic is the invariant the race violates, so that is what the new case pins — "A take the live plan's model does not name is not fed": disarm and publish (which closes the rig's take and leaves an epoch naming no take on that key), then start a take against that epoch. It is in the callback's set and is fed nothing; the finished take is empty. Removing the `binary_search` gate fails it, which is the same defect the race produced. If you want a stress case on top of that I will add one, but it would be a smoke test rather than a regression.

**Second thing this turned up.** `ClipLane::recordSlots` says which slot a take records into, not whether one may run, and it does not consult `TrackInfo::recordArmed`. So a slot take is retained by the track's arm like any other take of that track's input, and a host that arms a slot without arming its track gets a recording the next edit ends. Documented on `RuntimeStateIds::takes` rather than changed: `collectRuntimeStateIds` is built from `TrackInfo` and the slot arming arrives through a different call, and the host side of arm-and-roll is not written yet.

Full non-JUCE suite: 3850 passed, 13 skipped, 0 failed. JUCE suite including the real-project corpus: all passed.

lucaromagnoli:
Fair, and I withdraw the "not deterministically assertable" claim — it was true of the assertion I had reached for, not of the property. Added in 83b37ec4.

The rig can now run callbacks on a thread of their own. Twenty edits are published against it, and then the take is **stopped rather than disarmed**. That distinction turned out to matter: with the eligibility gate in place a disarmed take goes quiet at the swap, so the wait inside the close is no longer what protects it, and a case built on disarm passes with the wait removed. Only an explicit stop leaves blocks being fed right up to the publish inside `stopTake`, which is where the wait is still the only thing between `finish()` closing the file and a callback writing to it.

What it asserts: the take has taken no further sample twenty milliseconds and some hundreds of callbacks after `stopTake` returned. Deterministic, and load-bearing — removing the publish from `stopTake` fails it 12 times out of 12, and the case is clean over 30 consecutive runs. It also still requires the file to be one unbroken stretch of the input, which holds whatever the interleaving was.

Two things I checked on the way and think are worth recording:

- The first shape of this case exercised the concurrency but caught nothing when the wait was removed, because the take object stays alive in the `ClosedTake` and `finish()` racing `capture()` is not deterministically visible in the file. Asserting on `capturedSamples()` rather than on the audio is what made it bite.
- The gate has narrowed what the waits are for. For a take an *edit* ended, the epoch stops feeding it before the set is republished, so the wait there now protects the handoff rather than the writing. For an explicit stop it is still doing the original job. Both are now covered.

Full non-JUCE suite: 3851 passed, 13 skipped, 0 failed.

lucaromagnoli:
Both were right. Fixed in 01bfdb5d — the window is now forced rather than waited for.

A device that parks the callback inside its own `process()` closes both gaps at once. A parked callback holds the recording feed for as long as it stands there, so it is a callback the publishing thread genuinely has to wait for, stopped where a real one would be, and blocks are let through one at a time.

The case parks a callback, disarms, and publishes from another thread — which cannot get past its own swap while that callback is parked. It then lets blocks through until one of them renders the edit's plan. That block is *found*, not assumed: a witness device the edit adds ahead of the parking one in the chain says which plan the parked block belongs to, because how far the publishing thread got by then is the scheduler's business. That block held the recording set from before the edit, and fed the take nothing.

- gate removed: fails 20 of 20
- gate in place: 25 consecutive runs of the tag, no flakes

Two things worth reporting rather than quietly fixing:

**My first version of this was wrong in exactly the way you warned about.** It assumed the block after the swap was the edit's, which is only true when the publishing thread reaches its swap before the parked block is released. Standalone it passed; under suite load it failed at `128 == 64` — the extra block. The witness is what removes the assumption.

**Two fixture bugs, both of which turned a failure into a hang or an abort instead of a report.** `stopCallbacks` now stops parking before it joins, or a case that fails an assertion waits forever on a callback that cannot return — which is what happened the first time I ran the negative check, and it cost me a run I had to kill. And the devices clear the factory's pointer to them when they are destroyed, since deleting the armed track destroys them; without that, the delete-track case aborted on a freed mutex. A test rig whose failure mode is a deadlock is worse than no rig, so both are fixed rather than worked around.

Also worth flagging: an earlier attempt to cover this by disarming under a running callback passes with the wait removed, because with the gate in place a disarmed take is already quiet by then. That is why the concurrent stop case stops rather than disarms, and why this one asserts on the epoch boundary rather than on the wait.

Full non-JUCE suite: 3852 passed, 13 skipped, 0 failed. JUCE suite including the real-project corpus: all passed.

lucaromagnoli:
**Update: both findings below are addressed in cdcca319.** I reviewed the new commit and ran `[2465]`: all 12 tests pass (162 assertions). The witness pointer is removed, and the test hook holds the publisher after epoch publication and before recording-feed cleanup while callbacks run. The test checks both that the ended take stops receiving samples and that an eligible take continues. The original comment below refers to 01bfdb5d; I had not rechecked the PR head immediately before posting it.

---

Follow-up test review of 01bfdb5d: all 12 `[2465]` tests pass (148 assertions), but two issues remain in the new concurrency test.

**[P2] Synchronize publication of the witness pointer** — `tests/engine/test_take_across_swap.cpp:448`

`witnessHasRendered()` reads `factory_.witness` on the test thread while `InputFactory::createDevice()` assigns it on the async publishing thread (line 275). The first poll can happen before the new epoch has rendered, and the 20 ms sleep establishes no happens-before relationship. The atomic flag inside `WitnessDevice` does not make publication of the pointer itself safe. Use an atomic pointer with release/acquire publication, or preferably a shared atomic witness flag created before starting the publishing thread and passed into the device.

**[P2] Explicitly hold the publisher between epoch publication and recording-feed cleanup** — `tests/engine/test_take_across_swap.cpp:875–879`

The witness proves that the parked callback rendered the new plan, but does not prove that it acquired the old recording feed. After the previous callback is released, the publisher can swap the epoch and remove the ended take before the next callback acquires the feed. Running callbacks back to back makes that less likely but does not exclude it. In that interleaving, the sample-count assertion passes even without the eligibility gate, so this is still not a deterministic regression test of the original window.

Suggested direction: add a narrow test synchronization point immediately after epoch publication and before recording-feed cleanup. Hold the publisher there, run a callback on the new epoch while the old recording feed is necessarily still present, and assert that the ended take gains no samples while an eligible take continues. Release the publisher with failure-safe cleanup. Verify that removing the eligibility gate reliably fails this test.

These are test-harness findings, not a request for a broader engine redesign. The architectural issue was publishing plan state and recording eligibility independently despite their shared boundary. Putting eligibility in the render epoch addresses that: `RuntimeStateStore` can continue owning the long-lived recorder/writer/queue/tap, the epoch decides which takes may capture, and the later recording-feed removal/wait makes retirement safe.

lucaromagnoli:
Both correct. Fixed in cdcca319, and the second one is why the first existed.

**[P2] witness pointer** — right, `factory_.witness` was written by `createDevice` on the publishing thread and read on the test thread with nothing ordering them, and the atomic inside the device did nothing for the pointer to it. The witness is gone entirely (see below), so that race is gone with it. The factory's pointer to the parking device is now `std::atomic<ParkingDevice*>` for the same reason — `createDevice` runs on whichever thread published the plan that needed one.

**[P2] the parked callback may hold the new feed** — also right, and it is the more important of the two. The witness proved the block rendered the new plan and said nothing about which recording set it was holding. In the interleaving where the publisher got past the cleanup first, the case passed with the gate removed too. My 20-of-20 was the audio thread reliably winning a race, not a guarantee.

Taking the suggested direction. `EngineSession` gains `betweenPlanAndTakesForTest`, called on the publishing thread after the plan is live and before `closeUnnamedTakes` — empty otherwise, one null check per publish. It is deliberately the only such hook: every other ordering in `publish()` is a wait, and a wait cannot be observed from outside, while this one is a gap.

The case now stands inside it rather than trying to catch it from outside. It parks a callback from within the hook, so the recording set is still the pre-edit one by construction, and reads two consecutive blocks:

- the take the edit ended gains nothing across them
- a second take, on a track the edit did not touch, keeps going — the other half you asked for

No witness, no async publish, no sleep, and no failure-safe cleanup needed because nothing throws between park and release.

- gate removed: fails 20 of 20
- gate in place: 30 consecutive runs of the tag, no flakes

Everything the previous version needed to work around the race — the witness device, the `std::async`, the 20 ms sleep, the unwind guard — came out, so the case is about 40 lines shorter than the one it replaces.

Agreed on the framing in your last paragraph, and that is where it has landed: `RuntimeStateStore` owns the recorder, writer, queue and tap; the epoch decides which takes may capture; the later removal and its wait make retirement and handoff safe.

Full non-JUCE suite: 3852 passed, 13 skipped, 0 failed. JUCE suite including the real-project corpus: all passed. clang-tidy gate clean.

Issue #2148, opened 2026-08-22
Collapse search loops to any_of / find_if / contains in api, project, music, agents and engine


lucaromagnoli:
Part of #2143.

Collapse search loops and the small map/filter shapes in `daw/api`, `daw/engine` (TE wrapper), `daw/project`, `daw/music`, `daw/media_db`, `magda/agents` and `magda/engine`. `engine/plan/PlanCompiler.cpp` already shows the house style (`any_of` predicates at 57-92, `find_if` then `found == end() ? nullptr : &*found` at 311, `count_if` at 508); the rest should read the same way.

## daw/engine (TE wrapper)
- `TracktionEngineWrapperInit.cpp:249, 262` and `TracktionEngineWrapperDevices.cpp:76, 202`: four byte-identical "any channel index under limit -> enable" loops -> one `enableDevicesForChannels()` on `ranges::any_of`
- `TracktionEngineWrapperPlugins.cpp:691` (`ranges::find` with `createIdentifierString` projection), `753` (`all_of`)
- `TracktionEngineWrapperRecording.cpp:471` (`any_of(targets | views::values, &Target::active)`), `TracktionEngineWrapperTracks.cpp:124` (`trackMap_ | views::keys | ranges::to`)
- `TempoSequenceRippleMath.hpp:84` -> `ranges::equal(a, b, nearBeat, &T::beat, &T::beat)`

## daw/music
- `ChordSuggestionEngine.cpp:363` (`all_of`), `245` (`max_element` on map by `.second`, guard empty), `592` (3-way `views::zip`), `1009` (`filter | to`), `1194, 1199` (`ranges::replace`), plus ~15 sorts with single-key comparators
- `ChordEngine.cpp:176` (`ranges::equal`), `121` (`ranges::min` with projection, guard empty), `274, 283` (`ranges::contains`)
- `ChordTypes.hpp:94, 96`, `ScaleDetector.hpp:15, 17, 95, 158, 166, 187` (`transform | ranges::to<std::set>`)

## daw/api, project, media_db, cli
- `api/osc_feedback_live.cpp:202, 295, 304, 596` (find surface by peer, one projection)
- `api/device_api_live.cpp:104, 111` (`filter(nonNull) | transform | to`), `api/remote_subscriptions.cpp:66` (`ranges::equal(*lhs, *rhs, deepEquals)`)
[...]

Issue #2147, opened 2026-08-22
Collapse search loops to any_of / find_if / contains in daw/ui


lucaromagnoli:
Part of #2143.

Collapse flag-and-break / early-return search loops to `ranges::find_if`, `any_of`, `all_of`, `contains` across `daw/ui` (44 verified sites in panels/views/dialogs/state, 48 in components), plus the sort-by-member and min/max shapes that share the same projections.

## Shared predicate first
`isAudioFile(const juce::File&)` is hand-rolled as a loop in 8 places: `views/SessionView.cpp:3785, 4058`, `components/chain/custom_ui/SamplerUI.cpp:554`, `DrumGridUI.cpp:553`, `ImpulseResponseUI.cpp:180`, `components/pianoroll/PianoRollGridComponent.cpp:2707`, `components/tracks/TrackContentPanel.cpp:3284`, `panels/content/ChordClipContent.cpp:743` (MIDI variant), `TrackChainContent.cpp:634` (preset variant). One predicate + `ranges::any_of(files, ...)`.

## panels, views, dialogs, state
- `panels/content/PluginBrowserContent.cpp:967, 1010, 1110, 1122`: find plugin by key then mutate; one `keyOf(plugin)` projection serves all four. Also 557/564 (filter+transform specs).
- `views/SessionView.cpp:1720, 3268, 3319`: index-of trackId -> one `trackIndexOf()` on `ranges::find` + `distance`; `2552, 2931, 3501` (any_of); `2758, 2806` (8 duplicated lines -> `filter(isSessionClip) | ranges::to`)
- `panels/content/ChordClipContent.cpp:713`, `DrumGridClipContent.cpp:1385`, `MidiEditorContent.cpp:794` (bind `getClipsOnTrack()` to a local first, it returns by value), `inspector/TrackInspector.cpp:1580`, `interaction/ArrangementHitTester.cpp:33`, `panels/state/PanelController.cpp:311` (`all_of` + `contains`)
- `dialogs/AISettingsDialog.cpp:407, 1505, 1655`; `windows/MainWindowCommands.cpp:67` (`none_of && any_of`)
- sorts with projections: `state/TimelineController.cpp:1156, 1181, 1393` (`&TimelineMarker::positionBeats`), `panels/content/AIChatConsoleContent.cpp:2380, 2446`, `inspector/clip/sections/ChordProgressionSection.cpp:80`, `themes/UserTheme.cpp:131, 173`
[...]

PR #2545, opened 2026-09-09
Collapse daw/ui search loops to ranges algorithms

lucaromagnoli:
Part of #2147. Not the whole issue yet; the remaining sites are listed at the bottom and are coming as further commits on this branch.

Every call site takes a named predicate or projection, so the first layer reads as a sentence rather than as a lambda body:

```cpp
return anyEnabledMod(ctx.deviceMods) || anyEnabledMod(ctx.rackMods) ||
       anyEnabledMod(ctx.trackMods) || anyMacro(ctx.deviceMacros) ||
       anyMacro(ctx.rackMacros) || anyMacro(ctx.trackMacros);
```

## One predicate for droppable files

The eight hand-written extension lists disagreed with each other:

| site | extensions |
|---|---|
| `SessionView` | wav aiff aif mp3 ogg flac **m4a wma opus** |
| `TrackContentPanel`, `MediaExplorerContent`, `SamplerUI`, `DrumGridUI` | wav aiff aif mp3 ogg flac |
| `ImpulseResponseUI` | wav aif aiff flac ogg, **no mp3** |

No fixed list is right on every platform either, because `registerBasicFormats()` adds `CoreAudioFormat` on macOS and `WindowsMediaAudioFormat` on Windows. So `ui/utils/AudioFileTypes.hpp` asks the format manager. `isAudioFile` and `isMidiFile` are callable objects rather than overloaded functions, which is what lets them pass by name to `any_of` instead of through a wrapper lambda.

This fixes a real bug: `SessionView` was accepting `.opus` drops that no build can decode.

## The duplication the sweep exposed

Each of these is now behind one name:

- `SessionView::trackIndexOf`, six copies of the same index-of walk
- `TrackHeadersPanel::droppableHeaderIndexAt`, three copies of the same hit test
- `CurveEditorBase::setPointComponentSelected`, four copies
- `ParamLinkResolver::hasActiveLinks`, six loops plus two identical `ControlTarget` constructions
- `PluginBrowserContent`, nine re-inlined copies of the uniqueId-or-name key rule that `preferenceIdentifierForPlugin` already implemented

`ChainPanel::calculateInsertIndex` lost its trailing "after last element" branch: `find_if` plus `distance` already returns `size()` when nothing matches.

## Two things found on the way

`DrumGridRoles` is in `daw/audio`, not `daw/ui`. The core and audio sweep closed without it and it is the same two-loop shape, so it is fixed here.

Four files used `std::ranges::distance` without including `<iterator>`, relying on `<algorithm>` to pull it in. That is the same transitive-include trap that broke `std::inserter` on the C++23 bump, and it would compile here and fail on a different standard library.

## Verification

Debug build clean. Catch2: 3859 cases, 1946211 assertions, 0 failures. JUCE suite green, including all nine real-project corpus cases through both engines.

`readability-use-anyofallof` now reports nothing in `daw/ui`. Its one remaining hit is `daw/project/ProjectInfo.hpp`, which belongs to #2148.

## Still to come on this branch

`AISettingsDialog` (3), the `filter | ranges::to` pair in `SessionView`, `PianoRollContent`'s min/max trio, `ClipComponent`'s zip, the `AIChatConsoleContent` sorts, `DeviceSlotHeaderControls`' reverse loop, `PluginBrowserContent` spec filter/transform, and a few singles in `ChordClipContent`, `MidiEditorContent` and `TrackInspector`.

Deliberately not converted: the `ModulatorEditorPanel` sites the issue calls "the ParamLinkResolver twins" are nested accumulates over `m.links` with a per-outer-element `m.value` in today's code, not searches. That is a fold, and joining would lose the outer element.

### DIFF
--- a/magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
+++ b/magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cmath>
 #include <functional>
+#include <ranges>
 #include <utility>
 
 #include "../../../audio/AudioBridge.hpp"
@@ -1113,10 +1114,8 @@ void TrackHeadersPanel::trackPropertyChanged(int trackId) {
         // Update send labels from track data
         if (track->sends.size() == header.sendLabels.size()) {
             // Same count — update levels in-place (avoids destroying labels mid-drag)
-            for (size_t i = 0; i < header.sendLabels.size(); ++i) {
-                float levelDb = gainToDb(track->sends[i].level);
-                header.sendLabels[i]->setValue(levelDb, juce::dontSendNotification);
-            }
+            for (const auto& [label, send] : std::views::zip(header.sendLabels, track->sends))
+                label->setValue(gainToDb(send.level), juce::dontSendNotification);
         } else {
             // Send count changed — full rebuild
             rebuildSendLabels(header, trackId);
@@ -2005,9 +2004,9 @@ void TrackHeadersPanel::rebuildSendLabels(TrackHeader& header, TrackId trackId) 
             const auto* track = TrackManager::getInstance().getTrack(trackId);
             if (!track)
                 return;
-            for (size_t i = 0; i < header.sendLabels.size() && i < track->sends.size(); ++i) {
-                if (track->sends[i].busIndex == busIndex) {
-                    float newLevel = dbToGain(static_cast<float>(header.sendLabels[i]->getValue()));
+            for (const auto& [label, send] : std::views::zip(header.sendLabels, track->sends)) {
+                if (send.busIndex == busIndex) {
+                    const float newLevel = dbToGain(static_cast<float>(label->getValue()));
                     UndoManager::getInstance().executeCommand(
                         std::make_unique<SetSendLevelCommand>(trackId, busIndex, newLevel));
                     break;
@@ -2157,13 +2156,15 @@ juce::Rectangle<int> TrackHeadersPanel::getResizeHandleArea(int trackIndex) cons
 }
 
 bool TrackHeadersPanel::isResizeHandleArea(const juce::Point<int>& point, int& trackIndex) const {
-    for (int i = 0; i < static_cast<int>(trackHeaders.size()); ++i) {
-        if (getResizeHandleArea(i).contains(point)) {
-            trackIndex = i;
-            return true;
-        }
-    }
-    return false;
+    const auto handleContainsPoint = [this, point](int i) {
+        return getResizeHandleArea(i).contains(point);
+    };
+    const auto indices = std::views::iota(0, static_cast<int>(trackHeaders.size()));
+    const auto match = std::ranges::find_if(indices, handleContainsPoint);
+    if (match == indices.end())
+        return false;
+    trackIndex = *match;
+    return true;
 }
 
 void TrackHeadersPanel::layoutMeterColumn(TrackHeader& header, juce::Rectangle<int>& workArea,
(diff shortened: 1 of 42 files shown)


magda/daw/ui/panels/content/PluginBrowserContent.cpp
```diff
@@ -35,7 +37,7 @@ namespace magda::daw::ui {
 namespace {
 
 juce::String preferenceIdentifierForPlugin(const PluginBrowserInfo& plugin) {
-    return plugin.uniqueId.isNotEmpty() ? plugin.uniqueId : plugin.name;
+    return preferenceIdentifierForPlugin(plugin);
```
lucaromagnoli: **P1 — Restore the preference key calculation**

This call resolves to the function currently being defined, so every use recurses without a base case. refreshPluginList() calls it for each plugin when the browser gets an engine, which makes opening or activating the plugin browser stack-overflow, or hang after tail-call optimization. Please restore the uniqueId/name calculation here.

magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
```diff
@@ -1113,10 +1114,8 @@ void TrackHeadersPanel::trackPropertyChanged(int trackId) {
         // Update send labels from track data
         if (track->sends.size() == header.sendLabels.size()) {
             // Same count — update levels in-place (avoids destroying labels mid-drag)
-            for (size_t i = 0; i < header.sendLabels.size(); ++i) {
-                float levelDb = gainToDb(track->sends[i].level);
-                header.sendLabels[i]->setValue(levelDb, juce::dontSendNotification);
-            }
+            for (const auto& [label, send] : std::views::zip(header.sendLabels, track->sends))
```
lucaromagnoli, line 1117: **P2 — Keep the ranges API within the advertised toolchain floor**

The public prerequisites still promise GCC 11.1+, but libstdc++ did not implement std::views::zip until GCC 13.1: https://gcc.gnu.org/onlinedocs/libstdc%2B%2B/manual/manual/status.html. A documented GCC 11 or 12 build will therefore fail here even though CMake can select C++23 mode. Please either raise the supported standard-library/compiler floor and audit the other new C++23 ranges calls, or keep these loops on APIs available at the current floor.

magda/daw/ui/panels/content/PluginBrowserContent.cpp
lucaromagnoli: Fixed in 42730e3a2. You are right, and it was worse than a bad refactor: replacing the nine re-inlined copies of the uniqueId-or-name rule also rewrote the body of the helper that implements it, so `preferenceIdentifierForPlugin` returned a call to itself. Restored the calculation.

Worth recording why it survived: the full build passed and both suites passed, because nothing in either opens the plugin browser. I also swept the whole branch for the same replace-the-definition mistake. The only other self-calls are `TimelineController::handleEvent` dispatching between event overloads, which predates this branch.

magda/daw/ui/components/tracks/TrackHeadersPanel.cpp
lucaromagnoli: Fixed in 42730e3a2. The floor I wrote was the language gate, not the library one, which is exactly the distinction the C++23 audit had already flagged for libc++ and I failed to apply to libstdc++.

Audited what actually needs C++23 from the library across `magda/` and `tests/`: only `std::views::zip` (5 sites) and `std::ranges::contains` (4). `ranges::to` appears once but only inside a comment explaining why `toJuce` exists.

Raised the floor rather than reverting the calls, since GCC 11 and 12 were never built here and CI only exercises 13.3. Now documented as GCC 13+, Clang 16+ with libc++, AppleClang 16+ or MSVC 19.36+, with the reason stated so the next person does not re-derive it from the CMake gate. Same correction in README_CN.md and CONTRIBUTING.md.

One caveat: GCC 13.1 is from the libstdc++ status page you linked, and AppleClang 16 is what I verified locally, but the MSVC 19.36 number is from the STL changelog and only 19.50 is actually exercised in CI.

Issue #2554, opened 2026-09-09
Cutover slice 4: the sync layer becomes compiler passes


lucaromagnoli:
Slice 4 of #1897. Depends on slices 1 to 3, and is the one the epic's first bullet is about.

Slice 1 leaves `NativeAudioEngine` delegating roughly 25 of the interface's 72 methods to a TE instance it owns, because they are not engine concerns and narrowing the interface first would have produced a refactor with nothing audible at the end of it. This is where that is paid off.

## Scope

- The model stops being mirrored into an engine object graph and becomes the compiler's input. `ClipSynchronizer`, `AudioBridge`, `PluginManagerSync`, `RackSyncManager` and `MidiInputRouter` lose their internals to compiler passes.
- `getAudioBridge()` and `getMidiBridge()` leave the interface, or stop returning TE-coupled types. They are the two that make a non-TE implementation impossible to write honestly.
- The rest of the non-engine surface moves off `AudioEngine` to the services that own it: plugin discovery and exclusion, groove templates, `MagdaApi&`, `PluginWindowManager*`, sampler media, the tempo-ripple command. An engine interface that answers for a plugin scan is an interface two engines have to agree about for no reason.
- The TE instance `NativeAudioEngine` holds goes away.

## Why it is not first

Because it is worth doing once, against a native engine that is actually running. Narrowing the seam before anything drove it would be guessing at what the second implementation needs, and the guesses would be found wrong by slices 1 to 3 anyway.

## Verification

- The native engine holds no TE object and the app builds with the fork's headers unreachable from `NativeAudioEngine`.
- Every case in the corpus still nulls where it did, and the smoke tests still pass by hand.

Issue #2551, opened 2026-09-09
Cutover slice 1: the app plays through the native engine


lucaromagnoli:
Slice 1 of #1897, and the one that makes every slice after it testable by ear.

## Why it is a slice, and why it is first

Nothing outside the test binaries has ever run the native engine. `magda::engine::EngineSession` has no caller in `magda/daw/`; the only non-test match for the name is a collision (`HeadlessEngineSession` in the CLI wraps the TE-backed `magda::AudioEngine`). Everything the engine can do — seven subsystems, a corpus of 74 cases — it does offline, driven by a harness.

That is the wrong place to be for what is left. The remaining questions in #1895 and #2466 are ones an offline corpus cannot answer: whether a monitor round trip is playable, whether an input latency figure is the right one, whether a plugin behaves when a real host drives it at a real block size. Those need an interface and headphones, and there is currently no way to put the engine in front of either.

The seam already exists and #1885 built it. `magda::AudioEngine` (`magda/daw/engine/AudioEngine.hpp`) is an abstract interface with 72 pure virtuals, constructed through `createDefaultAudioEngine()`, with exactly one implementation and two construction sites (`MainWindow.cpp:586`, `magda_cli_main.cpp:45`). A second implementation is all that stands between the engine and a real device.

## What makes it affordable

The app-side half is already built and was built for this:

- `EngineDeviceFactory` (`magda/daw/audio/plugins/engine/`) resolves a Device op against the app's own catalogs, internal SDK devices **and external plugins**, sync or async (#2174, #2243). Plugins work on day one; that is the point of going this way rather than porting more subsystems first.
- `RuntimeStateStore` and the plan differ carry DSP state, instances and delay lines across a recompile, so an edit during playback is a swap rather than a rebuild.
- `EngineSession::process(numSamples, output, LiveInputBlock)` is already the shape an `AudioIODeviceCallback` wants.

What does not exist app-side is the compile itself: `compileRenderPlan` has no caller outside tests.

## Scope

- `NativeAudioEngine : AudioEngine`, chosen in `createDefaultAudioEngine` by a setting, so both engines ship in one binary and the choice is made at construction rather than at build time.
- A plan compiled from `TrackManager` and `ClipManager`, republished when the model changes. The differ decides what moves; nothing is rebuilt that did not change.
- Driven from the device callback into `EngineSession::process`.
- Device ops bound through `EngineDeviceFactory`.
- Transport, tempo, time signature, loop.
[...]

Issue #2466, opened 2026-09-06
Recording slice 8: parity cases for recording


lucaromagnoli:
Slice 8 of #1895, and its last: the parity cases for what the other seven ported, written here rather than in #1896 for the reason pinned on this issue -- a case is worth what its author's knowledge of the feature is worth, and that is highest on the day the port lands.

## The problem this slice has and the others do not

Recording has no null-diff shape by default. The corpus renders a project through both engines and compares files; a recording pass consumes something no project holds, and hardware does not replay.

So the first job here is an input that both legs can be fed deterministically. The fork already has the ingredients: `VirtualMidiInputDevice` is how MAGDA injects programmatic MIDI today (`daw/audio/midi/QwertyMidiKeyboard.cpp` registers the QWERTY keyboard as one), and `HostedAudioDevice` is the audio-side candidate for feeding known samples into the fork's input path. The native leg needs the equivalent behind slice 1's source. Once both legs record the same synthetic signal from the same transport, everything below is a file comparison like any other corpus case.

## Cases

- A take of a known signal comes back off disk as the samples that went in, on both legs.
- The first recorded sample is the same sample on both legs, at a declared input latency, positive and negative, and with a count-in set.
- A loop-recorded pass count produces the same take count with the same boundaries, and the same active take.
- MIDI recorded across a tempo change lands on the same beats, and a note held across the loop wrap resolves the same way.
- A slot recording starts at the same beat the launch did, at each quantization.
- A session-to-arrangement capture places its clips at the same beats.
- A monitored input is audible at the same level and the same latency through both engines, which is the one case that is a render rather than a file.

## What stays hands-on

The issue already budgets it, and it is worth naming so the corpus is not asked to cover it: what a driver reports as its input latency and whether that number is right, the audible round trip on real hardware, monitor switching with a real interface, and the feel of arming and punching while playing. Those are the afternoon with an interface plugged in, alongside the one #2245 wants for inserts.

lucaromagnoli:
Re-scoping this, with the argument, because the case list does not hold up against what slices 1-7 actually landed.

**Six of the seven cases already have native assertions.**

| Case here | Already asserted |
|---|---|
| A known signal comes back off disk as the samples that went in | `test_take_recorder.cpp:295` |
| First recorded sample at a declared latency, ±, and with a count-in | `test_take_recorder.cpp:323`, `:362`, `:765` |
[...]

PR #2558, opened 2026-09-09
Give the app a second engine it can select

lucaromagnoli:
First increment of #2551. `NativeAudioEngine` implements `AudioEngine` and is chosen in `createDefaultAudioEngine`, so both engines ship in one binary and switching between them is a setting rather than a build.

**The seam only.** All 80 methods forward to a `TracktionEngineWrapper` this owns, so selecting the native engine changes nothing audible yet. That is deliberate: it makes the first step verifiable on its own, and the step is not a small one — nothing outside the test binaries had ever constructed a second `AudioEngine`.

## Why this had to come first

`magda::engine::EngineSession` has no caller in `magda/daw/`. The only non-test match for the name is a collision (`HeadlessEngineSession` in the CLI wraps the TE-backed engine). Seven finished subsystems and a 74-case corpus, none of it ever driven by an audio device.

What remains open in #1895 and #2466 cannot be answered offline — whether a monitor round trip is playable, whether an input latency figure is right, how a plugin behaves under a real host at a real block size. All of it needs the engine reachable from a running app, and this is the first thing that makes that possible.

## The delegation, and why it is not a shortcut

`AudioEngine` is 80 pure virtuals. Roughly 25 are not engine questions at all — plugin scanning and exclusion lists, groove templates, `MagdaApi&`, `PluginWindowManager*`, the sampler media list, the tempo-ripple command. Two more, `getAudioBridge()` and `getMidiBridge()`, hand back types built around `te::Edit` and cannot be answered honestly by a non-TE engine.

Narrowing that interface first would have made the first slice a refactor with nothing audible at the end of it. So this owns a fork and delegates that half to it, with the reasoning written on the class rather than left implicit. #2554 is where the delegation is cut and the sync layer becomes compiler passes.

One rule the class states and keeps: **the fork never renders.** Its transport is never started, so the only thing that will fill an output buffer is this class. Both engines holding the device at once is the one failure that would sound like an engine bug rather than a wiring mistake.

## Verification

Through the CLI, both ways:

```
$ magda_cli boot
MAGDA engine booted headless

$ MAGDA_NATIVE_ENGINE=1 magda_cli boot
[engine] MAGDA native engine selected (#2551)
MAGDA engine booted headless
```

`make test` — 3859 passed, 0 failed. `make test-build` and `make cli` clean.

## Next in this slice

The plan compiled from `TrackManager`/`ClipManager`, the device callback driving `EngineSession::process`, devices bound through the `EngineDeviceFactory` that already exists, and transport/tempo/loop moving off the forward. That is the increment that makes it audible.

The forwarding surface was generated from the two headers rather than hand-written, so it cannot drift from the interface by a typo.


magda/daw/engine/TracktionEngineWrapperInit.cpp
```diff
@@ -33,6 +34,12 @@ TracktionEngineWrapper::~TracktionEngineWrapper() {
 }
 
 std::unique_ptr<AudioEngine> createDefaultAudioEngine(AudioEngineOptions options) {
+    // The native engine holds a fork of its own for the half of the interface
+    // that is not an engine question, so this is a choice between two engines
+    // rather than a choice about whether the fork is built (#2551).
+    if (NativeAudioEngine::requested(options))
+        return std::make_unique<NativeAudioEngine>();
```
lucaromagnoli: [P2] Preserve the headless option when selecting native

The CLI calls createDefaultAudioEngine({.headless = true}), but this branch constructs NativeAudioEngine without passing that option, and its owned TracktionEngineWrapper retains forceHeadless_ = false. With MAGDA_NATIVE_ENGINE=1 and MAGDA_HEADLESS unset, CLI initialization therefore opens audio/MIDI devices, constructs GUI/timer services, and can start plugin scanning instead of following the headless path. Please pass the option through and apply it to the owned wrapper before initialize().

magda/daw/engine/MagdaAudioEngine.hpp
```diff
@@ -0,0 +1,152 @@
...
+    bool isPlaying() const override;
+    bool isRecording() const override;
+    double getSessionPlayheadPosition() const override;
+    ClipId getSessionPlayheadClipId() const override;
+    std::unordered_map<ClipId, double> getActiveClipPlayheadPositions() const override;
+    SessionClipPlayState getSessionClipPlayState(ClipId clipId) const override;
+    void stopSessionTrack(TrackId trackId) override;
+    bool isSessionTrackStopPending(TrackId trackId) const override;
+    double getAudioThreadTransportSeconds() const override;
+    void deactivateAllSessionClips() override;
```
lucaromagnoli, line 86: [P2] Forward the non-pure recording methods too

This forwarding surface omits armSessionSlotRecording, beginArmedSessionSlotRecordings, isSessionSlotRecordArmed, isSessionSlotRecording, and getRecordingPreviews. These have default implementations in AudioEngine, so compilation succeeds while calls silently return no-op/false/empty results. With the native engine selected, SessionView's empty-slot record handler calls armSessionSlotRecording and then immediately returns because isSessionSlotRecordArmed is always false; arrangement recording also loses its live preview because TrackContentPanel receives an empty preview map. Please override and forward these methods to preserve the existing recording behavior.

magda/daw/engine/TracktionEngineWrapperInit.cpp
lucaromagnoli: Fixed in 07fb45be. `createDefaultAudioEngine` now passes `options` through, and the constructor applies `setForceHeadless` to the owned wrapper before `initialize()`.

Worth recording how it got past my own check: I verified the CLI with `MAGDA_HEADLESS=1` set, and that variable makes `isHeadlessRuntime()` true whatever the option said — so the test agreed with the bug. Re-verified with it unset, where both engines now boot headless identically.

magda/daw/engine/MagdaAudioEngine.hpp
lucaromagnoli: Fixed in 07fb45be, and there were seven rather than five.

The forwarding surface is generated from the two headers by matching pure virtuals, which is exactly why these were missed: they carry default bodies instead, so inheriting them compiles perfectly and then answers no-op/false/empty for the rest of the run.

Your five on `AudioEngine`, plus `onPunchRegionChanged` and `onPunchEnabledChanged` on `AudioEngineListener`, which have the same shape and the same consequence. All seven are now hand-written outside the generated block, with a note on the class saying why they are not in it.

Issue #2151, opened 2026-08-22
SIMD: one sanitiseBlock() for the 20 compiled FX devices, FloatVectorOperations for the peak reductions


lucaromagnoli:
Part of #2143. Performance lane: no ranges work lands in these files.

The compiled FX devices share a byte-identical per-sample output sanitise loop, and the block peak reductions are scalar. `daw/audio` has zero uses of `juce::FloatVectorOperations` today.

## Sanitise loop (20 files, verified identical)
`out[i] = std::isfinite(sample) ? juce::jlimit(-16.0f, 16.0f, sample) : 0.0f;` in `plugins/compiled/`: MagdaClipper (291), Chorus (407), RingMod (402), Pitch (337), Grit (323), Delay (413), Bitcrusher (299), Phaser (320), Flanger (384), Filter (446), Mod (410), Compressor (468), Reverb (369), GrainDelay (384), FreqShift (297), Eq (374), Dimension (343), Saturator (338), GateExpander (304), Multiband (472), Utility.

The per-sample `isfinite` ternary is a branch libc++ will not fold into a vector compare. Replace with one shared `sanitiseBlock(float* out, int n) noexcept` built on `FloatVectorOperations::clip(out, out, -16.0f, 16.0f, n)` plus one NaN check per block (`findMinAndMax` and a single `isfinite` on the pair, or a vector compare), and call it from all 20 devices.

## Peak reductions
- `MagdaClipperCompiledPlugin.cpp:273`, `MagdaCompressorCompiledPlugin.cpp:447, 451, 469`: `std::max(fabs(x))` -> `FloatVectorOperations::findMinAndMax`
- `plugin_manager/PluginManagerModifiers.cpp:514, 536, 567`: same; for the gain-scaled peak, `copyWithMultiply` then `findMinAndMax`
- `MagdaEqCompiledPlugin.cpp:348, 360, 378`: mono downmix sum + scale -> `FloatVectorOperations::add` then `multiply`; also hoist the `if (!bandEnabled[band]) continue` at 365 out of the sample loop into a pre-compacted active-band array (evaluated numSamples x 8 times today)
- `FollowerSourceTapPlugin.cpp:43`: `getSample()` per sample in the downmix -> hoist `getArrayOfReadPointers()` + `add`/`multiply`
- `SidechainPlugin.cpp:143`: `sidesOnly` is loop-invariant; split into two loops so the non-sides path becomes `multiply` over a gain ramp buffer

[...]

PR #2565, opened 2026-09-10
Move the block reductions onto FloatVectorOperations

lucaromagnoli:
Part of #2151. The scalar peak and downmix reductions in the audio path become
vector ops. `daw/audio` had no `juce::FloatVectorOperations` at all before this.

## peakMagnitude()

New, in `daw/audio/BlockMath.hpp`. Replaces the `peak = max(peak, fabs(x))`
loop in the clipper, the compressor and the three follower-push sites.

That form genuinely does not vectorise, because reassociating a float reduction
needs `-ffast-math` and the release build does not set it. At release `-O3` the
loop compiles to `ldr s1, [x0], #4` on arm64: one sample at a time. Covered in
`tests/dsp/test_block_math.cpp` against the loop it replaces, including a lone
spike in the tail past the last whole vector.

## The rest

- the gain-scaled follower peak is `rawPeak * |gain|` rather than a second pass,
  since scaling by a constant is monotone in magnitude
- the follower downmix reads channel pointers instead of `getSample()`, and sums
  before scaling so the arithmetic order matches the loop it replaces
- the EQ pre/post taps use `add` / `multiply`, and the per-band enable test is
  hoisted out of the sample loop into a compacted active-band list. It ran
  `numSamples x 8` times per block to answer a question constant for the block
- the sidechain gain is written to a fixed 256-sample stack chunk, so applying it
  is a vector multiply per channel rather than a strided store, and the
  loop-invariant `sidesOnly` branch leaves the sample path

Every one of these is bit-identical by construction, not by approximation.

## What this does not do, and why

The `sanitiseBlock()` rewrite the issue leads with. Two problems with it:

**The 20 copies are gone.** They were already collapsed into one
`MagdaCompiledEffect::sanitise()` since the audit was taken in August.

**The premise does not hold.** The issue says the per-sample `isfinite` ternary
is a branch libc++ will not fold into a vector compare. It folds it on both
shipping targets, at the release `-O3`:

| | arm64 | x86-64 SSE2 |
|---|---|---|
| clamp | `fminnm.4s` / `fmaxnm.4s` | `minps` / `maxps` |
| isfinite | `bic.4s` + `cmgt.4s` | `pand` + `pcmpgtd` |
| select | `and.16b` | `andps` |

Branchless, four lanes wide, one pass. `FloatVectorOperations::clip` would need a
second traversal for the NaN pass and would change behaviour: it maps an infinity
to +/-16, where the current form maps it to 0. The issue's own bar is
bit-identical output, which that would not meet.

## Verification

- `make debug` clean
- Catch2: 3887 passed, 13 skipped, 0 failed
- `magda_juce_tests`: all passed

One note on that last line. The first run of this branch hung: the sidechain gain
ramp was sized in `prepare()`, and in `Master sidechain ducks audio from a source
MIDI note` it had no block size, so the chunk length was 0 and `done += chunk`
never advanced. The fixed stack chunk in this PR removes the member, the
`prepare()` dependency and that failure mode together. The suite now runs to
completion.

### DIFF
--- a/magda/daw/audio/BlockMath.hpp
+++ b/magda/daw/audio/BlockMath.hpp
@@ -1,0 +1,23 @@
+#pragma once
+
+#include <juce_audio_basics/juce_audio_basics.h>
+
+namespace magda {
+
+/** @brief Highest absolute sample in a block (#2151).
+ *
+ *  The scalar `peak = max(peak, fabs(x))` form does not vectorise: reassociating
+ *  a float reduction needs -ffast-math, which the release build does not set.
+ *  juce::AudioBuffer::getMagnitude() is this for a buffer; this is the same
+ *  reduction for a bare pointer, which is what a sidechain port and the follower
+ *  scratch are.
+ */
+inline float peakMagnitude(const float* samples, int numSamples) noexcept {
+    if (samples == nullptr || numSamples <= 0)
+        return 0.0f;
+
+    const auto range = juce::FloatVectorOperations::findMinAndMax(samples, numSamples);
+    return juce::jmax(range.getStart(), -range.getStart(), range.getEnd(), -range.getEnd());
+}
+
+}  // namespace magda
--- a/magda/daw/audio/BlockMath.hpp
+++ b/magda/daw/audio/BlockMath.hpp
@@ -2,22 +2,44 @@
 
 #include <juce_audio_basics/juce_audio_basics.h>
 
+#include <algorithm>
+#include <cmath>
+
 namespace magda {
 
-/** @brief Highest absolute sample in a block (#2151).
+/** @brief Highest absolute sample in a block, ignoring NaN (#2151).
  *
  *  The scalar `peak = max(peak, fabs(x))` form does not vectorise: reassociating
  *  a float reduction needs -ffast-math, which the release build does not set.
- *  juce::AudioBuffer::getMagnitude() is this for a buffer; this is the same
- *  reduction for a bare pointer, which is what a sidechain port and the follower
- *  scratch are.
+ *  It also ignores NaN, because the compare against it fails, while still
+ *  counting an infinity. Both are kept here.
+ *
+ *  findMinAndMax cannot be handed a NaN: its SIMD min/max differ by
+ *  architecture, and on the lane that met the NaN the running extreme is
+ *  dropped, so a real peak can go missing without the result being NaN at all.
+ *  The block is screened first, which is an integer reduction over a compare and
+ *  vectorises where the float one will not.
  */
 inline float peakMagnitude(const float* samples, int numSamples) noexcept {
     if (samples == nullptr || numSamples <= 0)
         return 0.0f;
 
-    const auto range = juce::FloatVectorOperations::findMinAndMax(samples, numSamples);
-    return juce::jmax(range.getStart(), -range.getStart(), range.getEnd(), -range.getEnd());
+    int nanCount = 0;
+    for (int i = 0; i < numSamples; ++i)
+        nanCount += static_cast<int>(samples[i] != samples[i]);
+
+    if (nanCount == 0) {
+        const auto range = juce::FloatVectorOperations::findMinAndMax(samples, numSamples);
+        return juce::jmax(range.getStart(), -range.getStart(), range.getEnd(), -range.getEnd());
+    }
+
+    // A NaN in the block is an upstream bug, so this path is not the one worth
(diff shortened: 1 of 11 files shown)


magda/daw/audio/BlockMath.hpp
```diff
@@ -0,0 +1,23 @@
...
+ *  a float reduction needs -ffast-math, which the release build does not set.
+ *  juce::AudioBuffer::getMagnitude() is this for a buffer; this is the same
+ *  reduction for a bare pointer, which is what a sidechain port and the follower
+ *  scratch are.
+ */
+inline float peakMagnitude(const float* samples, int numSamples) noexcept {
+    if (samples == nullptr || numSamples <= 0)
+        return 0.0f;
+
+    const auto range = juce::FloatVectorOperations::findMinAndMax(samples, numSamples);
```
lucaromagnoli: **P2 — Preserve NaN-skipping reduction semantics**

The scalar loop this replaces starts at zero and therefore ignores NaN samples. `findMinAndMax` seeds its range from the input, and its SIMD min/max instructions handle NaNs differently by architecture and lane position. On arm64, for example, a NaN can poison a SIMD lane, either returning NaN or suppressing a later valid peak in that lane. That makes follower detection and compressor metering diverge from the stated oracle. Please add NaN-before-valid-peak cases and explicitly ignore NaNs while continuing to count infinities.

magda/daw/audio/plugins/compiled/MagdaClipperCompiledPlugin.cpp
```diff
@@ -55,11 +55,9 @@ void MagdaClipperCompiledPlugin::beforeCompute(DeviceProcessContext& context, in
...
 
     float peak = 0.0f;
-    for (int channel = 0; channel < channels; ++channel) {
-        const float* samples = context.audio->getReadPointer(channel, context.startSample);
-        for (int i = 0; i < context.numSamples; ++i)
-            peak = std::max(peak, std::fabs(samples[i]));
-    }
+    for (int channel = 0; channel < channels; ++channel)
+        peak = std::max(
+            peak, context.audio->getMagnitude(channel, context.startSample, context.numSamples));
```
lucaromagnoli: **P2 — Keep NaNs out of the persistent clipper meter**

`AudioBuffer::getMagnitude` delegates to the same NaN-sensitive `findMinAndMax` reduction. A NaN here is stored in `inputPeakDb_`; the curve view then applies `pow` and feeds it into `smoothedInputAmp_`, which remains NaN even after later finite blocks. The old scalar loop ignored the bad sample. Please route this through the corrected NaN-safe peak reduction, or reject a NaN before publishing the meter value.

magda/daw/audio/BlockMath.hpp
lucaromagnoli: Correct, and fixed in eff541db7.

Both halves of the old loop's semantics are back. The block is screened for NaN first, which is an integer reduction over a self-compare and vectorises where the float one will not (`cmpunordps` + `psubd` on x86-64, `fcmge`/`fcmlt` on arm64). A clean block takes findMinAndMax; one holding a NaN falls back to the scalar loop this replaced, which walks past it and reports the loudest real sample. Infinities still count, on both paths.

Worth recording why the first version was wrong, since it is subtler than a poisoned lane: on x86 `_mm_max_ps` returns its second operand when either side is NaN, so the NaN does not propagate, it *overwrites* the running maximum in that lane. An earlier peak in that lane is silently forgotten and the result comes back finite and too low. Checking the result for NaN would not have caught it.

Tests now put a NaN at every position across six block lengths and compare against the loop, plus explicit before/after-the-peak and infinity cases.

magda/daw/audio/plugins/compiled/MagdaClipperCompiledPlugin.cpp
lucaromagnoli: Correct, and fixed in eff541db7. It reads through peakMagnitude() now, which ignores NaN and keeps infinities, so the meter cannot latch on a bad sample.

The stickiness you describe is the reason this one mattered more than the reduction itself: inputPeakDb_ is not recomputed from scratch by a later block, it is stored, so a single NaN would have sat on the transfer curve's dot until the device was rebuilt.

Same reduction is behind juce::AudioBuffer::getMagnitude(), so anywhere else that publishes a stored meter from it has the same exposure. Nothing else in this PR does, but it is worth knowing when the next one reaches for it.

Issue #2152, opened 2026-08-22
SIMD: remove per-sample accessors, shifts and invariant branches from the hot loops (TrackMeasurer, Limiter, engine clip path)


lucaromagnoli:
Part of #2143. Performance lane: no ranges work lands in these files.

Per-sample loops that are slow today for reasons unrelated to SIMD width: bounds-checked accessors inside the inner loop, a 12-element shift per input sample, loop-invariant branches inside the sample loop, a `std::cos` per sample. Fix the shape; keep the loops raw.

## daw/audio
- `analysis/TrackMeasurer.hpp:420` (`oversamplePeak`): the 4x true-peak FIR shifts `delay[t] = delay[t-1]` for 12 taps on every input sample before the 48 MACs, on every metered block. Use a circular index into the history (or a contiguous history + `addWithMultiply` per phase). The FIR stays a raw loop.
- `analysis/TrackMeasurer.hpp:122`: `if (capture)` is loop-invariant but sits inside the K-weighting/M-S/correlation loop; hoist it so the five accumulators stay in registers.
- `analysis/AudioTapBuffer.hpp:42, 55`: masked ring index per sample, called from every `applyToBuffer`; split at the wrap into at most two `memcpy`s (capacity is a power of two).
- `plugins/compiled/MagdaLimiterCompiledPlugin.cpp:79`: `getSample()`/`setSample()` in the innermost loop and `% lineLength` per sample; hoist `getArrayOfWritePointers()` once, make the line length a power of two and mask.
- `CompService.cpp:131`: crossfade gain via `getSample()`/`setSample()`; build a gain ramp buffer, then `FloatVectorOperations::multiply` per channel.
- `plugins/MagdaSamplerPlugin.cpp:181, 197`: `addSample()` per sample per channel; hoist write pointers (ADSR and interpolation stay as is).
- `AudioThumbnailManager.cpp:425, 437`, `media_db/MediaDbIndexer.cpp:162`: per-column min/max and multi-to-mono downmix -> `findMinAndMax` / `addWithMultiply` (message thread, but O(total samples) per repaint / per index).

## magda/engine
- `exec/PlanExecutor.cpp:931, 945`: detect-mono downmix and block peak through `AudioBlock::getSample(c, s)` per element; hoist channel pointers, `copy` + `addWithMultiply`, `findMinAndMax` per channel.
[...]

Issue #2143, opened 2026-08-22
Epic: declarative loop style (std::ranges) with a SIMD-first exception list


lucaromagnoli:
Move MAGDA's loop-heavy code to a declarative `std::ranges` / `std::views` style, and in the same pass make the per-sample loops SIMD-friendly. Where a functional form would compromise auto-vectorisation, performance wins: those loops stay raw and get `juce::FloatVectorOperations` instead.

Audit (2026-08-22, six read-throughs of loop bodies, counts are verified sites): https://claude.ai/code/artifact/5343cf0b-f329-40b3-b632-1276639e5be1 (private report).

## Vocabulary
C++20 ranges and views, plus the C++23 subset libc++ actually ships: `ranges::to`, `views::zip`, `views::chunk_by`, `views::repeat`, `ranges::fold_left`, `ranges::contains`. Not available on macOS: `views::enumerate`, `chunk`, `slide`, `stride`, `adjacent`, `join_with`, `std::generator`. C++23 needs no toolchain change (AppleClang 15 takes `-std=c++2b`, GCC 13 `-std=c++23`, MSVC `/std:c++latest`).

## Ground rules
- Per-sample / per-block numeric loops stay imperative. `views::filter` is a data-dependent branch, `zip`/`transform` over several arrays hands the vectoriser iterator pairs it must prove non-aliasing, `fold_left` has no horizontal-max idiom, and envelope/biquad recursions are serial by construction.
- `ranges::to` does not compile for `juce::Array` / `StringArray` (they have `add`, not `push_back`/`insert`); use the `toJuce<C>()` helper or leave those loops.
- `juce::ValueTree` is not a `std::ranges::range`; use the `children(vt)` helper.
- By-value getters (`ClipManager::getClipsOnTrack`, `getSessionClips`, `AutomationManager::getLanesForTrack`) must be bound to a local before piping.
- `filter_view` caches `begin()`: build and consume, never store in a member.
- Keep empty-range guards around `ranges::min/max/minmax`.
- Loops with an early `break` on a sorted scan keep the loop (`filter` does not short-circuit).
- The listener notify loops (`TrackManager.cpp:3204-3351`, `SelectionManager.cpp:988ff`) stay as index loops on purpose: listeners add/remove during the callback.
- MIDI-thread scans (`BindingRegistry.cpp:113/137`, `ControllerRegistry.cpp:107/119`) may use `any_of` but stay allocation-free.

## Order of work
Declarative lane:
[...]

PR #2569, opened 2026-09-10
Take the per-sample accessors and shifts out of the hot loops

lucaromagnoli:
Closes #2152. Last issue in the C++23 and declarative epic (#2143); no ranges work lands in these files, which is what the issue asked for.

Thirteen loops that were slow for reasons unrelated to SIMD width: a bounds-checked accessor inside the inner loop, a twelve-tap shift per input sample, a loop-invariant branch inside the sample loop, a cosine per sample, and a gain envelope rebuilt once per channel. The loops that have to stay serial stayed serial: the limiter's gain follower, the K-weighted recursion, the biquads and the cubic interpolator are all still raw.

## daw/audio

| site | was | is |
|---|---|---|
| `TrackMeasurer::oversamplePeak` | 12 taps shifted per input sample, before the 48 MACs, on every metered block | a mirrored circular window: nothing moves, and each phase reads 12 contiguous samples |
| `TrackMeasurer::process` | sample peak folded into the K-weighted loop; `if (capture)` inside it | `peakMagnitude` per channel; the capture is one downmix pass through the scratch |
| `AudioTapBuffer` | masked index per sample, both directions | split at the wrap into at most two `memcpy`s |
| `MagdaLimiterCompiledPlugin` | `getSample`/`setSample` and `% lineLength` per sample | hoisted write pointers, power-of-two line, mask |
| `CompService` | crossfade gain through `getSample`/`setSample`, per channel | one ramp per section, `FloatVectorOperations::multiply` per channel |
| `MagdaSamplerPlugin` | `addSample` and `getReadPointer` per sample per channel | pointers hoisted, mono fan-out no longer re-interpolates |
| `AudioThumbnailManager`, `MediaDbIndexer` | scalar per-column min/max, scalar downmix | `blockMinMax`, `addWithMultiply` |

## magda/engine

| site | was | is |
|---|---|---|
| `PlanExecutor::renderModSource` | detect-mono downmix and block peak through `AudioBlock::getSample` per element | `clear` + `addWithMultiply`, `peakMagnitude` per channel |
| `SampleRing` | per-sample masked atomic store; per-element zero-pad branch on read | two contiguous runs each way, three precomputed segment lengths; the downmix goes through a fixed stack chunk, so the tap still has no maximum block size |
| `ClipVoice::applyFade` | `getChannelPointer` per sample, envelope recomputed per channel | one ramp per chunk, one `multiply` per channel |
| `FadeCurves` de-click | `std::cos` per sample **per channel**, twice over (start and stop) | one shared raised-cosine ramp, `addWithMultiply` per channel, one helper for both directions |
| `ClipStretcher` | clamp and channel lookup per sample on the way in; `getSample`/`setSample` history shift | pointers hoisted once per call, `clear` for the tail, one overlapping move for the history |
| `ModFollower` | a per-sample one-pole over a value that is constant across the block | closed form |

### The follower's closed form

`advanceFollower` is handed one number for the whole block, so a run of one-pole steps is a geometric decay of the gap: `envelope + coeff^n * (envelope - input)`. Neither branch can change part way through a block, which is what makes it exact rather than an approximation. In attack the step closes a fixed fraction of what is left and never arrives, so the envelope stays under the input; in release it stays over it. The clamp only bites where the source peaked above one, and clamping every step gives the same answer as clamping the run, because the iterate is monotone.

## BlockMath moved to daw/core

The engine includes `core/` and not `daw/audio/`, and three of these sites want the same reduction, so `BlockMath.hpp` moved to `magda/daw/core/` and is allowlisted for device packs on the same terms as `RangesHelpers`: header-only over JUCE, no host state behind it.

It gained `blockMinMax`, and `peakMagnitude` now delegates to it. The NaN screen is the point: `FloatVectorOperations::findMinAndMax` cannot be handed a NaN, because on the lane that meets one the running extreme is dropped, so a real peak goes missing and the result is not NaN either. The block is screened with an integer reduction over a self-compare, which vectorises where the float reduction will not. Eight new cases in `tests/dsp/test_block_math.cpp` pin it against the loop it replaces.

## Deliberately not converted

- **The cubic interpolator's `at()` lambda is not peeled.** The issue asks for the interior range to be peeled so the join test goes away in the middle. The four taps per output sample are eight perfectly predictable branches, and the peel's boundary arithmetic depends on the sign of the step, which is a rendered-output path. Not worth the risk for branches the predictor already gets right.
- **The de-click keeps `std::cos`, once per sample rather than once per sample per channel.** A two-tap recursive cosine drifts, and what it would drift away from is the ramp arriving at zero, which is the click the ramp exists to remove. Sharing the ramp across channels was the actual defect and it costs nothing.

## Verification

- `make debug` clean, Catch2 3908 cases / 1.93M assertions pass, JUCE suite `All tests PASSED`.
- Null-diff corpus rows compared against a local build of the parent commit: **byte-identical on every deterministic row**, `project.sidechain` and `project.fmchain` included. Only `project.retrospect` moves, and it moves run to run on the same binary because it hosts a third-party VST3.
- `make configure` re-run, so the device-pack source validation actually ran against the new allowlist entry rather than a stale cache.

Epic #2143 closes with this.


magda/engine/param/ModFollower.cpp
```diff
@@ -189,16 +190,26 @@ float advanceFollower(FollowerState& state, const FollowerSettings& settings,
...
     const float input = std::max(state.sourcePeak, 0.0f);
 
-    for (int i = 0; i < numSamples; ++i) {
+    if (numSamples > 0) {
         if (input > state.envelope) {
-            state.envelope = (attack * (state.envelope - input)) + input;
+            // Attack, and for the whole block: a step closes a fixed fraction of
+            // what is left and never arrives, so the envelope cannot cross the
+            // input and change branch part way through.
+            state.envelope = decayed(state.envelope, input, attack, numSamples);
```
lucaromagnoli: **P2 — Preserve attack state when the closed form rounds to its target**

The closed form can round exactly to the input even though the scalar recurrence remains slightly below it. At 48 kHz with a 1 ms attack and 512-sample blocks, this path reaches exactly 1.0; the next sustained block enters the hold/release branch and exhausts a 5 ms hold. The old recurrence remains at 0x1.ffffe8p-1, keeps taking the attack branch, and refreshes the hold. When the source then drops, the new version releases immediately instead of holding for 5 ms.

Please preserve the attack-side state/invariant across blocks and add a sustained-input regression case.

— Codex

Issue #2566, opened 2026-09-10
External plugins on the native engine's live path


lucaromagnoli:
`EngineRuntimeFactory::createDevice` (`magda/daw/engine/host/EngineRuntimeFactory.cpp`) resolves a
Device op through `createEngineDevice`, which builds internal SDK devices and nothing else. Every
external plugin in a project is therefore named in the log and bound to nothing: the track renders
without it.

The pieces already exist and are what #2243 built. What is missing is the host side of them.

## What it needs

- `ExternalPluginServices` — a `juce::AudioPluginFormatManager*` and a `const
  juce::KnownPluginList*`. Both live inside the fork's `te::Engine::getPluginManager()`, and
  `AudioEngine` exposes neither. An accessor on `TracktionEngineWrapper`, handed to the host by
  `MagdaAudioEngine`, is the short version.
- `createEngineExternalDeviceAsync` rather than the blocking form. A publish runs on the message
  thread, and a project of plugins loaded synchronously is a frozen UI for as long as the slowest
  one takes.
- The completion half: `PluginAssignments` for the identity boundary, a `CurrentDeviceLookup`, and
  `applyRestoredParameters` onto the model when an instance comes back with state
  (`docs/architecture/engine-device-factory.md` has the contract).
- The two-pass compile the corpus does (`tests/NullDiffNativeLeg.cpp`): resolution is entitled to
  correct an external device's instrument/effect role, and a plan compiled from the project's own
  guess routes MIDI to the wrong places.

## Why it is not in #2551

Slice 1's audible increment is a MIDI clip driving an instrument, and the async load, the identity
boundary and the state writeback are each their own question. Splitting them out kept that slice to
a size somebody could read.

Blocks #2551's "an external plugin loads on a track and is audible" verification.

Zloyparen:
Hi luca,maybe u could help me to add steinberg vst sdk  support for my web api based daw built with electron,u know i want to separate vst host,so to prevent crashes caused by vst host.So u just make template me to connect my index.html to vst hosting,i got bored of testing vst host,it didnot make any sound when i tested it  

[my daw - Copy.html](https://github.com/user-attachments/files/32058917/my.daw.-.Copy.html) It is my project,still working on it

lucaromagnoli:
Done: PR #2571 merged into dev/0.20.0.

PR #2571, opened 2026-09-10
External plugins on the native engine's live path

lucaromagnoli:
Closes #2566 (close by hand — this merges to `dev/0.20.0`, not `main`).

## What was wrong

`EngineRuntimeFactory::createDevice` resolved a plan's `Device` op through the internal
and compiled-Faust catalogs alone. Every external plugin in a project was therefore named
in the log (`nothing to build "..." with`) and bound to nothing: the track rendered without
it. This was the one item on #2551's verification list that had not been started, and it is
what limited hands-on testing to projects built from internal devices.

## What this does

`magda/daw/engine/host/ExternalPluginLoader` is the host side of what #2243 built.

- A publish asks for a device; the loader starts `createEngineExternalDeviceAsync` and
  answers null. The executor already renders an unbound op as a pass-through, and
  `RuntimeStateStore` re-asks the factory rather than remembering a null, so the next
  publish picks the instance up. A project of ten plugins does not freeze the window.
- Completion writes `resolvedDevice` (real bus widths, MIDI capability, role) and
  `restoredParameters` into the model on the message thread, then publishes a plan compiled
  from it. That is the corpus's two-pass compile (`tests/NullDiffNativeLeg.cpp`) spread over
  as many message-loop turns as the plugins take.
- `CurrentDeviceLookup` reads the model when the load finishes, through the same
  `devicesIn` walk the compiler uses — a new per-track overload, since `TrackManager` hands
  out mutable tracks but not a mutable project.

The identity boundary is the loader's own bookkeeping beside the adapter's:

- A slot re-registered for the same plugin keeps its assignment. Every publish registers
  every device, and minting a new assignment each time would expire the load still running
  — repeatedly, so a plugin would never finish loading in a session anybody was using.
- A slot that changed plugin gets a new assignment, and the generation on it makes the old
  completion a no-op rather than something that marks the slot spent and blocks the plugin
  the slot now names.
- A plugin the scan has not found yet is asked about again at the next publish: nothing was
  spent finding that out, and a scan finishing is what changes the answer. Anything that got
  as far as instantiating is left spent, so a plugin that fails to open does not cost seconds
  once per edit.

Two smaller things travel with it:

- The fork's `AudioPluginFormatManager` and `KnownPluginList` reach the host through
  `MagdaAudioEngine`, which keeps the wrapper as itself: `AudioEngine` exposes neither, and
  this is a dependency #2554 deletes rather than one worth growing the interface for.
- The trace's plan line now says how many plugins are still opening, so a plan with unbound
  device ops can be told from one whose plugins are never coming.

## Tested

- Five new cases in `tests/engine/test_engine_external_device.cpp`, over a real
  `AudioPluginFormat` seam. Each of the four behaviours above was verified negatively —
  reverting the guard and watching the case fail: the generation check, `ensureAssignment`
  vs `replaceAssignment`, the not-installed retry, and the factory's routing.
- `magda_tests`: 3902 passed, 13 skipped, 0 failed.
- `magda_juce_tests`: all passed.
- Null-diff corpus: 65 ok, 9 FAIL — the same nine as an unmodified `dev/0.20.0`, checked by
  stashing this branch and rebuilding. Eight are the audio cases the previous session
  recorded as inherited; the ninth, `session.launch.sustained`, is new on dev since then and
  is filed separately.

Not hands-on tested with a real VST/AU yet — that needs a run on Luca's machine with
`MAGDA_AUDIO_ENGINE` on the native engine and a project carrying a plugin.

### DIFF
--- a/docs/architecture/engine-device-factory.md
+++ b/docs/architecture/engine-device-factory.md
@@ -100,3 +100,14 @@ had a preset applied, or been deleted — so:
 
 `isInstalledExternalPlugin` only says a scanned plugin exists to try loading
 — it does not promise instantiation will succeed.
+
+### Who drives the asynchronous path
+
+`magda/daw/engine/host/ExternalPluginLoader.hpp` (#2566). It owns the
+`PluginAssignments` a live session registers against, starts one load per slot,
+and answers `EngineRuntimeFactory::createDevice` with null until the instance
+arrives — which is why the store re-asks rather than remembering a null
+(`RuntimeStateStore.hpp`). Completion writes `resolvedDevice` and
+`restoredParameters` into the model on the message thread and publishes a plan
+compiled from it: the corpus's two-pass compile, spread over as many message
+loop turns as the plugins take.
--- a/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
+++ b/magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp
@@ -16,21 +16,18 @@ namespace magda::daw::audio::engine_adapter {
 
 namespace {
 
-/// One descent, written once, over whichever constness the caller has.
+/// One track's descent, written once, over whichever constness the caller has.
 ///
 /// A flat stage's elements are devices rather than a tree, so it is walked
 /// directly; flat does not mean padless, and a device carrying pads is
 /// descended into the same way a tree device is.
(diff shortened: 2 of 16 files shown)


lucaromagnoli:
Hands-on: Surge XT loads and is audible on the native engine. That closes #2551's "an external plugin loads on a track and is audible" row, which was the last one outstanding.

CI green on all eight checks.

Issue #2579, opened 2026-09-10
One switch: selecting an engine should mean the other one is not running


lucaromagnoli:
Selecting magda does not turn Tracktion off. It leaves it half-alive: no
transport, but a live graph on the audio device, rendering. The engine choice
is a switch with nothing behind it.

`MagdaAudioEngine.hpp` claims otherwise, under a heading "What the Tracktion
engine is not allowed to do":

> Render. Its transport is never started and its edit never gets a playback
> context, so the only thing filling an output buffer is this class.

That is a comment, not a mechanism, and it is false.

## Why it is false

`MagdaAudioEngine::initialize()` calls `tracktion_->initialize()` for the ~25
interface methods that are not engine questions — plugin scanning, exclusion
lists, groove templates, `PluginWindowManager`, `getAudioBridge`. That function
does not separate services from rendering. It runs end to end:

```
[Init] createEditAndBridges()...
Edit loaded in: 1 ms
Playback context allocated for live MIDI monitoring
AudioBridge initialized
```

`TracktionEngineWrapperInit.cpp:365` calls `ensureContextAllocated()`. Nothing
frees it, nothing takes TE off the device, and `EngineHost` adds a second
callback to the same `AudioDeviceManager` — JUCE sums them.

So the switch has nowhere to land: asking for the fork's services and asking
for the fork's renderer are the same call.

## What it accounts for

- **A live note is audible while magda renders nothing for it.** The MIDI trace
  shows magda receiving no note and its meter tap reading zero, yet the synth
  sounds — through TE's copy of the instrument, reached via
  `TrackManager::previewNote` -> `AudioEngine::previewNoteOnTrack` ->
  `audioTrack->injectLiveMidiMessage`. The QWERTY keyboard takes the parallel
  route through `AudioBridge::getQwertyMidiDevice()->keyboardState`.
- **Every external plugin is loaded twice.** `AudioBridge::syncAll()` has no
  engine gate, so the fork instantiates every VST/AU beside the one
  `ExternalPluginLoader` opens. Twice the RAM and load time, two editor windows
  behind one slot.
- **Any work assuming magda is the only renderer is reasoning from a false
  premise.** PR #2577's `setMeteringFedElsewhere` did exactly that, and killed
  the live-note meters that TE's taps had been feeding.

## The switch

Split `TracktionEngineWrapper::initialize()` in two:

- `initialiseServices()` — the `te::Engine`, plugin formats, scanning,
[...]

Issue #2582, opened 2026-09-10
Plugin programs and preset files on the native engine


lucaromagnoli:
The plugin preset menu is empty under the native engine. `DevicePresetMenu` reads
`getPluginNumPrograms`, `getPluginCurrentProgram`, `getPluginProgramName`,
`setPluginCurrentProgram`, `loadPluginPresetFile` and `savePluginPresetFile` on `AudioBridge`
(`DevicePresetMenu.cpp:449-609`), and each resolves a `te::ExternalPlugin`.

#2270 listed program changes and preset loading as "the same shape and not covered" and closed
without them. #2573 covers a chunk-carrying preset applied through the model; this is the
plugin's own program list and its preset files, which never go through the model at all.

## What to build

- Program count, names, current program and select, plus preset file load and save, as
  operations on `DeviceControlPlane`, serialised against rendering like capture (#2268) and
  keyed by `engine::DeviceKey`.
- A program change or a preset file load changes the instance's chunk, so it ends in the same
  capture path as the save (the capture-on-save issue), or the model is left describing a patch
  the plugin no longer has.
- `DevicePresetMenu` asks the rendering engine.

## Verification

Under magda: a VST3 with factory programs lists them, selecting one is audible, saving the
project keeps it. Under Tracktion: unchanged.

Issue #2581, opened 2026-09-10
Capture a native instance's plugin state on save


lucaromagnoli:
A project saved while the native engine holds an external plugin writes the chunk the plugin was
loaded with, not the chunk the instance holds now.

`EngineExternalDevice::captureState()` exists (`EngineExternalDevice.cpp:609`) behind
`LocalDeviceControlPlane::captureState` (`DeviceControl.cpp:53`), and nothing calls it. The save
hook only knows `AudioBridge::captureAllPluginStates()`
(`TracktionEngineWrapperInit.cpp:415-418`), which reads the fork's parallel instances. #2244
closed with the API half; the host half was never needed while the fork's copy stood in for the
native one. After #2579 there is no copy, so a save on magda loses whatever the instance changed
since load.

## What to build

- A host path from the save hook to the control plane, one capture per bound `DeviceKey`,
  assignment checked between the read and the commit as #2270 states, then
  `applyCapturedPluginState()` into the model.
- `captureAllPluginStates` and the other capture sites (`TrackCommands.cpp:327, 889`,
  `PadCommands.cpp:24`, `DevicePresetMenu.cpp:193`) ask the rendering engine.
- Together with #2573, which is the same seam in the apply direction.

## Verification

Load a project with an external plugin on magda, change something only the chunk carries, save,
reload on either engine: the change is there. `test_external_plugin_state_restore_juce.cpp`
stays green.

lucaromagnoli:
Merged to dev/0.20.0 as ab83f2cf (#2586). `Closes #2581` does not fire against dev, so closing by hand.

Issue #2580, opened 2026-09-10
External plugin editors on the native engine


lucaromagnoli:
Under the native engine an external plugin's editor window never opens onto the instance that
renders. `PluginWindowManager` drives `te::ExternalPlugin::windowState`
(`magda/daw/engine/PluginWindowManager.cpp:39-77`), and the native instance is private to
`EngineExternalDevice` (`EngineExternalDevice.hpp:164`) with no accessor by design.

Today it is masked: the fork holds a parallel `te::ExternalPlugin` for every slot, and the window
that opens is onto that copy, which is not the one you hear. After #2579 the copy is gone and no
window opens at all.

#2270 named this ("the editor-window path moves to the same endpoint") and closed with state
capture only. #1893 does not mention editors. `DeviceControl.hpp:20-29` still reserves the place.

## What to build

- Open, close and is-open on `DeviceControlPlane`, keyed by `engine::DeviceKey` like capture,
  serialised against rendering the way capture is (#2268), and fallible: a device that has gone
  mid-call answers rather than dangles.
- One window per slot, owned by whoever owns the instance. Closed on rebuild (#2575), on unload,
  on teardown (#2576), and before the session it renders through goes away.
- The slot's toggle (`DeviceSlotComponent.cpp:350, 1630`, `MiniChainRow.cpp:62`) and
  `DeviceApiLive::openDeviceEditor` (`device_api_live.cpp:353`) ask the rendering engine, not
  `AudioBridge` by name. Under Tracktion the answer is the existing `PluginWindowManager`.

In-process only. Out-of-process is #1899 and is why the endpoint is asynchronous.

## Verification

Under magda: open an external plugin's editor, move a control, hear it. Close the project with
the window open and nothing dangles. Under Tracktion: unchanged.

lucaromagnoli:
Merged to dev/0.20.0 as 1df132bd (#2594).

`DeviceControlPlane` grows show, hide, toggle and query keyed by DeviceKey, on the control executor; `EngineExternalDevice` owns the window as the only object allowed to reach the instance, so a device retired, evicted or torn down takes its window with it. The five call sites ask the engine that is rendering rather than `AudioBridge` by name.

The same pass fixed the device power button, which published values for an edit the compiler treats as structural.

Out-of-process editors stay #1899.

PR #2584, opened 2026-09-10
Split the fork so only the chosen engine renders

lucaromagnoli:
Fixes #2579, and #2578 with it.

Selecting magda left Tracktion half-alive: no transport, but an Edit with a playback context, an
AudioBridge mirroring every device into it, and a second copy of every external plugin. The
engine choice was a switch with nothing behind it.

## The switch

`TracktionEngineWrapper::initialize()` is now `initialiseServices()` followed by
`initialisePlayback()`. `MagdaAudioEngine` calls the first only, so under magda there is no
Edit, no playback context and no AudioBridge, and every external plugin loads once. The
MidiBridge and the project save hooks are services and moved with them. Under Tracktion,
`initialize()` runs both and nothing changes.

## What that took with it, rehomed

- **Meters.** The three metering rings, the master peak and the MIDI activity light live in a
  `TrackMeters` each engine owns and feeds, reached through `AudioEngine::meters()`. The
  "fed elsewhere" gate from #2577 is gone.
- **The transport model.** Tempo, time signature, the tempo map, loop, metronome, count-in,
  trigger state, edit length and the MagdaApi are answered by `MagdaAudioEngine` and
  `EngineHost`, not by an Edit that no longer exists. `getAudioBridge()` returns null in its own
  body.

## Live notes reach the engine (was #2578)

- `EngineRuntimeFactory::createMidiInput` binds the plan's MIDI input op through a
  `LiveMidiInput` over the session's feed, reading the track's audition source and the device
  sources its route names.
- `LiveMidiQueue` carries messages from the MIDI and message threads into the callback
  (farbot fifo, many producers, one consumer); `LiveMidiSources` assigns the ids.
- `CompileOptions::auditionMidi` gives every MIDI-reading track an input op, so a piano-roll
  preview on an idle track has something to reach. The app sets it; the corpus never does.
- The piano roll, chord panel, drum pads, QWERTY keyboard, hardware MIDI and the remote inject
  all arrive through a `LiveMidiSink` on `MidiBridge`, which `MagdaAudioEngine` implements.
  The QWERTY keyboard no longer needs an AudioBridge.

Events land at the start of the block they are pulled in, up to one buffer of jitter. Sub-block
placement belongs with #2553's monitor round trip.

## Tests

- Compiler: the option off, on, on a track with no MIDI consumer, on with monitor In.
- Host: a preview sounds on a track monitoring nothing; a device's live MIDI sounds on the track
  routed to it and on no other; an idle track hears nothing from a device stream.
- The wrapper after `initialiseServices()` alone: no Edit, no AudioBridge, a MidiBridge.
- `MagdaAudioEngine` answers its transport model from no Edit, and the fork under it has none.

Verified negatively: with `createMidiInput` returning null the live-note cases fail; with
`initialisePlayback()` called from `MagdaAudioEngine::initialize()` the no-Edit assertions
fail.

`magda_tests` 3914 passed, 13 skipped. `magda_juce_tests` all passed. The app target links.

A second wrapper in one process is now survivable: the wrapper restores the save hooks it
overwrote and shuts the `ControllerRouter` down only when the router listens to its own
MidiBridge. The new JUCE tests rely on that.

## Not driven by hand

`docs/smoke-tests/engine-switch.md` is the checklist, written for this. Worth a pass under both
engines before merging: a piano-roll key on an idle instrument track, the QWERTY keyboard on a
monitoring track, a hardware keyboard, and one VST3 showing one plugin process rather than two.

## Still dark under magda, each tracked

Editors #2580, capture on save #2581, programs and preset files #2582, chunk presets #2573,
controller and OSC writes #2554 and #2317, slot meters #2570, the remote transport API #2554,
export and bounce #2555, session #2552, recording and input #2553.

https://claude.ai/code/session_012q17kwk7LKPovRoPtrBWjX


magda/daw/engine/MagdaAudioEngine.cpp
```diff
@@ -86,32 +99,39 @@ bool MagdaAudioEngine::initialize() {
...
+    api_->setMidiBridge(midi);
 
-    int numerator = 4;
-    int denominator = 4;
-    tracktion_->getTimeSignature(numerator, denominator);
-    host_->setTimeSignature(numerator, denominator);
-    publishLoop();
+    if (midi != nullptr) {
+        midi->setMeters(&meters_);
+        midi->setLiveSink(this);
```
lucaromagnoli, line 118: [P1] Disconnect the native callbacks during destruction. This installs this as MidiBridge's live sink, but MagdaAudioEngine's destructor is defaulted and the application destroys it with a plain daw_engine_.reset(). host_ can therefore be destroyed while its audio callback is still registered and while MidiBridge can still call pushMidi through a dangling MagdaAudioEngine pointer. Make destruction invoke the idempotent shutdown sequence: clear the sink, detach the host callback, then shut down Tracktion. The new test calls shutdown explicitly, so it misses the production path.

magda/engine/plan/PlanCompiler.cpp
```diff
@@ -1377,15 +1392,14 @@ void Compiler::emitTrack(const TrackInfo& track) {
...
-            plan_.ops[static_cast<std::size_t>(op)].liveness = LivenessDomain::Live;
-            midiSources.push_back(PortRef{op, 0});
+        case RouteKind::External:
+            midiSources.push_back(emitLiveMidiInput(track.id));
             break;
-        }
         case RouteKind::None:
+            // Nothing is routed here, so this is the op a preview reaches
+            // (CompileOptions::auditionMidi, #2579).
+            if (options_.auditionMidi && readsMidi)
```
lucaromagnoli: [P1] Preserve audition input when a track has an internal MIDI route. This only emits the destination track's audition source for RouteKind::None. RouteKind::Track adds only the source track's MIDI, while piano-roll, pad, and API previews are queued under the destination track's audition source, so preview is silent whenever its input is track:N. Add audition independently for every MIDI-consuming track; let the route switch add only the additional routed source.

magda/daw/engine/host/EngineRuntimeFactory.cpp
```diff
@@ -151,6 +186,68 @@ std::unique_ptr<engine::EngineMidiSource> EngineRuntimeFactory::createSessionMid
...
+}
+
+void EngineRuntimeFactory::refreshMidiRoutes(const std::vector<TrackInfo>& tracks) {
+    midiRoutes_.clear();
+
+    // receivesLiveMidiInput rather than monitorsInput: the fork routes a track
+    // in Auto too (MidiInputRouter::updateMidiInputRouting).
+    for (const auto& track : tracks)
+        midiRoutes_.emplace(track.id,
+                            MidiRoute{track.midiInputDevice, track.receivesLiveMidiInput()});
```
lucaromagnoli: [P1] Do not make unarmed Auto monitoring permanently audible. receivesLiveMidiInput intentionally includes unarmed Auto tracks for UI activity, whereas TrackInfo::monitorsInput defines whether input reaches the audible chain: armed or monitor In. Since TrackMidiInput directly appends every source in this table, this makes every unarmed Auto track hear its selected keyboard, unlike both the existing compiler behavior and Tracktion's automatic monitor mode.

magda/daw/engine/host/EngineRuntimeFactory.cpp
```diff
@@ -151,6 +186,68 @@ std::unique_ptr<engine::EngineMidiSource> EngineRuntimeFactory::createSessionMid
...
+    return std::make_unique<TrackMidiInput>(*liveInputs_, sources_->auditionSourceFor(trackId),
+                                            routeTableFor(trackId));
+}
+
+void EngineRuntimeFactory::MidiRouteTable::set(const std::vector<engine::LiveMidiSourceId>& ids) {
+    const auto n = std::min(static_cast<int>(ids.size()), kMaxSources);
+    for (auto i = 0; i < n; ++i)
+        sources[static_cast<std::size_t>(i)].store(ids[static_cast<std::size_t>(i)],
+                                                   std::memory_order_relaxed);
+    count.store(n, std::memory_order_release);
```
lucaromagnoli, line 240: [P2] Route changes can leave synthesizer notes hanging. If source A delivered a note-on and monitoring is disabled or changed before its note-off, replacing this table prevents that note-off from reaching the instrument. This mutation does not republish topology, so the executor's MIDI-panic mechanism never runs. Removing a route needs to panic the affected track or retain the old source until its held notes are released.

magda/daw/engine/host/LiveMidiSources.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    return sources;
+}
+
+int LiveMidiSources::resolveRoute(const juce::String& midiInputDevice) {
+    if (midiInputDevice.isEmpty() || midiInputDevice == "all" ||
+        midiInputDevice.startsWith("track:"))
+        return kNoSource;
+
+    // The list registerAvailableDevices() took, not a fresh enumeration: this
```
lucaromagnoli, line 65: [P2] Newly connected MIDI devices cannot be resolved reliably. available_ is refreshed only by registerAvailableDevices during structural publishPlan, while monitor/input selection changes go through publishValues and only refresh the route tables. Selecting a newly connected device therefore falls through to a source keyed by its hashed Tracktion ID, while incoming events are queued under its JUCE identifier; those sources do not match until an unrelated structural edit republishes the plan. Refresh the device snapshot and route tables on device-list changes or before resolving value-only route changes.

lucaromagnoli (commented):
Review complete. Five correctness findings are attached inline. The focused 2579, Engine Host Publish, and Magda Audio Engine tests pass, but they do not exercise these paths.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -345,6 +409,20 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
         engine::PlanValues values;
         report("values", resolveValues(*livePlan_, tracks, *master, values));
         report("values", session_->publishValues(std::move(values)).messages);
+
+        // A monitor or route change is a track property, so it arrives here
+        // rather than as a plan, and the input the store already holds reads
+        // the table this rewrites.
+        factory_.refreshMidiRoutes(tracks);
```
lucaromagnoli, line 426: [P1] Recompile track-to-track MIDI routes when their input or monitor state changes. TrackManager::setTrackMidiInput and setTrackInputMonitor emit only trackPropertyChanged, which this host turns into publishValues. refreshMidiRoutes can update device-source tables, but it deliberately returns no sources for track:N because that edge is compiled into RenderPlan. Switching track:1 to track:2, switching external to track:N, or toggling monitoring therefore leaves the old compiled edge active until an unrelated structural edit republishes the plan.

magda/daw/engine/host/EngineRuntimeFactory.cpp
```diff
@@ -19,6 +20,44 @@ juce::String deviceIdentityOf(const DeviceInfo& device) {
...
+        const auto count = routed_->count.load(std::memory_order_acquire);
+        for (auto i = 0; i < count; ++i)
+            feed_.appendEvents(
+                routed_->sources[static_cast<std::size_t>(i)].load(std::memory_order_relaxed), out,
+                engine::kMaxMidiBytesPerPort);
+
+        // A source that has gone will never send the note-off for what it is
+        // still holding, and a route change publishes no topology, so this is
+        // the only panic the instrument gets (#2418).
+        dropped_ = routed_->dropped.exchange(false, std::memory_order_acq_rel);
```
lucaromagnoli, line 47: [P1] This panic still does not reach external instruments. PlanExecutor carries raisedAllNotesOff into DeviceBlock::midiInAllNotesOff, and EngineMagdaDevice consumes it, but EngineExternalDevice::process only copies block.midiIn into its JUCE MidiBuffer and never reads midiInAllNotesOff. The unit test proves that this source raises the flag, not that a VST/AU instrument releases its voices, so external synths can still sustain indefinitely after a route is removed.

magda/daw/engine/host/LiveMidiSources.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    return auditions_.emplace(trackId, next_++).first->second;
+}
+
+std::vector<int> LiveMidiSources::deviceSources() const {
+    const juce::ScopedLock held(lock_);
+
+    std::vector<int> sources;
+    sources.reserve(devices_.size());
+    for (const auto& [deviceId, source] : devices_)
```
lucaromagnoli: [P2] Return currently available device sources here, not every device ever registered. registerAvailableDevices refreshes available_, but devices_ is never pruned and this method walks devices_. For an all route, unplugging a keyboard therefore leaves the same source ID in the route table, MidiRouteTable::set sees no removal, and no panic is raised for a held note whose note-off can no longer arrive. The new hot-plug test covers connection only. QWERTY likely needs a separate persistent/virtual-source set so it remains part of all without keeping disconnected physical sources active.

lucaromagnoli (commented):
Second-pass review after the live-MIDI fixes and the dev/0.20.0 merge. The focused compiler, Engine Host Publish, and Magda Audio Engine suites pass. Four remaining runtime findings are attached inline.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -41,12 +48,46 @@ constexpr int kChannels = 2;
 /// 30 fps, which is what the fork's own metering timer runs at.
 constexpr int kMeterIntervalMs = 33;
 
+/// Live MIDI sources the callback has room for. One per enabled input plus one
+/// per track ever auditioned; a project past this loses the sources beyond it
+/// rather than allocating for them on the audio thread.
+constexpr int kMaxLiveMidiSources = 64;
```
lucaromagnoli, line 54: [P2] This fixed table is smaller than the source-ID space feeding it. LiveMidiSources assigns from one monotonically increasing counter and never reuses IDs; physical devices, QWERTY, and one audition source per MIDI-reading track all consume it. Once an ID exceeds 64, collectLiveMidi silently drops every event for that source in normal builds. A 64-track instrument project can hit this immediately, and repeated device hot-plug or track creation can exhaust it in a much smaller live project. Reuse/prune inactive IDs or introduce a bounded slot allocator whose capacity is enforced before routing.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -342,9 +407,32 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
         if (session_ == nullptr || livePlan_ == nullptr || master == nullptr)
             return;
 
+        // An input route or a monitor switch arrives here, being a track
+        // property, and the route tables below cannot carry either: a
+        // "track:N" edge is compiled into the plan, and what a monitored
+        // audio input is at all is an op (#2579).
+        if (inputRoutingOf(tracks) != routing_) {
+            publishPlan();
```
lucaromagnoli, line 415: [P1] Panic retained instruments when a compiled MIDI edge changes. This now republishes the correct topology, but switching `track:1` to `track:2` (or disabling that internal route) can still leave the destination synth holding notes from the old source. Internal routes deliberately resolve to an empty `MidiRouteTable`, so the table removes no source and `TrackMidiInput::raisedAllNotesOff()` stays false; plan crossfading also ignores MIDI edge changes, while `RuntimeStateStore` retains the destination device by key. The new plan therefore drops the eventual note-off without sending a panic to the retained instrument. Please carry a one-shot panic into the first block of the new plan for retained devices whose MIDI input edge changed, and cover note-on → route publish → panic at the actual device rather than only comparing routing snapshots.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -1530,7 +1606,9 @@ void PlanExecutor::renderOp(OpId id, const OpValue& published, const BlockInfo&
                                     // nothing is withheld and nothing would come
                                     // back (#2418).
                                     .midiInAllNotesOff =
```
lucaromagnoli, line 1654: [P1] Consume the reroute latch even when another panic condition is already true. Because this is the final operand of a short-circuiting `||`, `std::exchange` is skipped when `!block.continuous` or `midiInPanic(...)` is true. The flag then survives the block that already delivered the panic and fires again on a later continuous block, which can kill notes chased or reasserted after the first panic. Spend `reroutedMidi_[i]` before constructing `midiInAllNotesOff` (or otherwise evaluate it unconditionally), and cover a reroute whose first block is discontinuous followed by a continuous block.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -322,6 +372,32 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+    if (previous != nullptr && previous != this && previous->plan_ != nullptr) {
+        const auto was = midiBehindDevices(*previous->plan_);
+        const auto now = midiBehindDevices(plan);
+
+        for (std::size_t i = 0; i < numOps; ++i) {
+            const auto& op = plan.ops[i];
+            if (op.kind != OpKind::Device)
+                continue;
+
+            const auto before = was.find(op.key);
```
lucaromagnoli: [P1] Match devices across plans by `DeviceKey`, which is the identity the runtime store actually retains. `RuntimeStateStore::realise` keys instances by `op.key.deviceKey()`, while this lookup uses the full `OpKey`. `moveChainElement` can move the same device between tracks or rack chains without changing its device ID, so the retained instrument has a different `OpKey`; `before` is not found and the code skips the panic even though its MIDI source changed. Key `midiBehindDevices` by `DeviceKey` and add a case where the same device ID moves from one track MIDI ancestry to another.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -322,6 +376,32 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
         }
     }
 
+    // A device the store keeps across the swap whose MIDI is now coming from
```
lucaromagnoli, line 383: [P1] Preserve an owed reroute panic across another publish while the device cannot run. A device in a muted or soloed-out rack returns at `value.silent` before `reroutedMidi_` is spent, which correctly leaves the panic pending in this executor. But if any structural edit publishes again before the rack becomes active, `prepare` compares the already-rerouted plan with the new plan, initializes this vector to zero, and drops that pending state; `RuntimeStateStore` still retains the instrument and its held voices. Unmuting can therefore resume the stale note indefinitely. Either deliver the one-shot panic to a bound device even while its output is silenced, or carry pending reroute state forward by `DeviceKey`. Please cover route change -> silent block -> unchanged plan publish -> unsilence.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -322,6 +376,45 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+        const auto was = midiBehindDevices(*previous->plan_);
+        const auto now = midiBehindDevices(plan);
+
+        // A debt the plan being replaced never got to pay. A device in a muted
+        // rack is not processed at all, so its panic is still owed however
+        // many publishes land before the rack comes back, and the store is
+        // holding the instrument that holds the notes all the while.
+        std::set<DeviceKey> unpaid;
+        for (std::size_t i = 0; i < previous->plan_->ops.size(); ++i)
+            if (previous->reroutedMidi_[i].load(std::memory_order_relaxed) != 0)
```
lucaromagnoli: [P1] Atomically claim this debt instead of copying it. `previous` is still the live executor while `prepare` runs, so this load can observe `1` and then the audio thread can exchange the same entry to zero and deliver that panic before the swap. The new executor is nevertheless initialized with the copied debt and delivers a second panic, potentially killing notes reasserted after the first. Transfer ownership with an exchange/CAS to zero (for example through a const claim method over mutable atomic state) so exactly one executor can spend it. The new test never lets the previous executor render between reading and publishing, so it cannot expose this race.

magda/engine/exec/EngineSession.cpp
```diff
@@ -96,6 +96,13 @@ EngineSession::Result EngineSession::publish(std::shared_ptr<const RenderPlan> p
     // goes live.
     values_.nonRealtimeReplace(prepared->values);
 
+    // Panics the epoch being replaced never got to deliver, because a device
+    // in a silenced rack is not processed at all. Here rather than at prepare:
+    // until this line the publish could still be refused, and a debt taken
+    // over by a plan that never renders is a note nothing releases (#2418).
+    if (live_ != nullptr)
+        prepared->executor.takeUnpaidReroutesFrom(live_->executor);
```
lucaromagnoli: [P1] Do not strip the live epoch of its panic before the realtime pointer swap. This call is merely adjacent to `published_.nonRealtimeReplace`; farbot can still let the audio thread acquire and render the old epoch before that later CAS succeeds. Since the new values were published at line 97, an otherwise identical republish that unmutes the rack can make the old executor process the retained instrument after its debt was claimed but before the new executor is live, leaking the stale held note for one or more callbacks. The new test explicitly renders the old executor after the claim and expects no panic, which demonstrates this gap rather than closing it. The pending debt needs shared single-consumer state across both epochs, or a handoff performed inside the synchronization boundary that prevents either epoch from rendering without owning it.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -322,6 +377,42 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
...
+            // the epoch being replaced is still rendering, and whichever of
+            // the two reaches the device first is the one that should spend
+            // it. Raised only, never cleared here: a device already owed one
+            // is still owed it.
+            //
+            // Read from the bindings rather than from the per-op table, which
+            // is not resolved until further down.
+            const auto owed = bindings.deviceMidiPanic.find(key);
+            if (lost && owed != bindings.deviceMidiPanic.end() && owed->second != nullptr)
+                owed->second->store(1, std::memory_order_relaxed);
```
lucaromagnoli: [P1] Do not raise the shared panic while the replacement is only being prepared. prepare() runs before appliesValues() can refuse the publish and well before published_ is swapped. The live executor can therefore exchange this flag to zero, then process another note-on through its still-current old route before the replacement goes live; that new held note has no remaining panic and its note-off follows the new route or nowhere. A candidate rejected by appliesValues() also panics the live instrument despite no routing change being published. Keeping the debt beside the retained device fixes its lifetime, but the transition still needs to be committed only once publication is guaranteed, with an epoch/generation scheme so the old route cannot consume the new route's debt early. Add a test that lets the old executor consume the panic, sends it another note-on before the swap, and then renders the new executor.

PR #2586, opened 2026-09-10
Capture a native instance's plugin state on save

lucaromagnoli:
Closes #2581.

A project saved while the native engine held an external plugin wrote the chunk the plugin was loaded with rather than the one the instance holds now. The save hook only knew `AudioBridge::captureAllPluginStates()`, which reads the fork's parallel instances, and `EngineExternalDevice::captureState()` had no caller.

## Asked of whichever engine is rendering

`AudioEngine` grows `captureAllPluginStates()` and `capturePluginStateAt()`. The fork answers them from the bridge as before. `MagdaAudioEngine` asks both in turn: the fork first, since a MAGDA device's state is captured nowhere else and its synced plugin is still where that is read from, then the instances it renders through over the top.

The four capture sites the issue named go through the engine instead of reaching for the bridge: `TrackCommands.cpp:327` and `:889`, `PadCommands.cpp:24`, `DevicePresetMenu.cpp:193`. The save hook in `TracktionEngineWrapperInit.cpp` asks the engine and keeps the warp markers on the bridge, those being the fork's own clip state.

The `FaustUI.cpp` capture calls stay on the bridge. They are paired with bridge-only `refreshDeviceParameters` on Faust devices, which the native path has no chunk to read for.

## The host half

`EngineHost` walks the publish's external keys through `LocalDeviceControlPlane`, checks the assignment between the read and the commit as #2270 states, and writes each snapshot back into the model with `applyCapturedPluginState()`. A key nothing is bound to is logged with its reason rather than committed as an empty state over a patch, so a plugin that is not installed on this machine keeps whatever the project was loaded with.

A single-slot capture asks `EngineRuntimeFactory::isExternalKey()` first. `capturePluginStatesUnder` walks every device in a chain, and without that guard a chain of compiled devices would report each one as a state that could not be read.

`SessionDevices` is held weakly by the plane, so a capture that outlives its project answers "the runtime is gone" rather than reaching a session that has been destroyed. `RuntimeStateStore` hands out a device as a shared lease, so a capture can hold what it is reading across a publish that drops it.

## drain()

Both entry points return with the reads done, because a save writes the file the moment they do. `ControlExecutor::drain()` is what makes that true: the message thread runs its own queue in place rather than returning to the loop, and the serial thread waits for its worker. A drain from inside running work does nothing, which is the nesting `run()` already refuses for the same reason. The message-thread queue is the executor's own now, so each post takes one item and a post whose item a drain already took finds nothing.

## Tests

- `test_control_executor_juce.cpp`, new: a drain runs what is queued, work nobody drains still runs once on the loop, a nested drain is not a second transaction, and a drain answers in order. In the JUCE target because `magda_tests` has no message loop.
- `test_control_executor.cpp`: the same two contracts for the serial thread.
- `test_engine_host_publish_juce.cpp`: `externalKeys()` names only the model's external plugins, and `isExternalKey()` refuses a compiled device and a key the model does not carry.

Catch2 3918 passed / 0 failed locally, full JUCE suite green, including `test_external_plugin_state_restore_juce.cpp`.

Branched off `dev/0.20.0` at f344b8a19, so it does not include #2584. Asking both engines in turn is correct on either side of that merge.

Hardware verification still wants a real VST: load a project with an external plugin on magda, change something only the chunk carries, save, reload on either engine.

### DIFF
--- a/magda/daw/engine/host/EngineHost.cpp
+++ b/magda/daw/engine/host/EngineHost.cpp
@@ -4,8 +4,8 @@
 
 #include <algorithm>
 #include <atomic>
-#include <optional>
 #include <ranges>
+#include <set>
 
 #include "../../audio/plugin_manager/ExternalPluginState.hpp"
 #include "../../audio/plugins/engine/ControlExecutor.hpp"
@@ -13,6 +13,7 @@
 #include "../../audio/plugins/engine/EngineDeviceFactory.hpp"
 #include "../../audio/plugins/engine/EngineExternalDevice.hpp"
 #include "../../core/AutomationManager.hpp"
+#include "../../core/ChainWalk.hpp"
 #include "../../core/ClipManager.hpp"
 #include "../../core/TrackManager.hpp"
 #include "../../project/ProjectManager.hpp"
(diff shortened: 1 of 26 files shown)


magda/daw/engine/host/EngineHost.cpp
```diff
@@ -516,6 +597,68 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
...
+    void captureExternalPluginStateAt(const ChainNodePath& devicePath) {
+        if (session_ == nullptr)
+            return;
+
+        const auto key = keyOfDeviceAt(devicePath);
+
+        // A chain is mostly devices this build holds, and those are asked of
+        // the fork. Reaching for one here would find no external plugin bound
+        // and report that as a state that could not be read.
+        if (!key.has_value() || !factory_.isExternalKey(*key))
```
lucaromagnoli: **P1 — Capture the native pad plugins when this path names a Drum Grid**

The single-slot callers deliberately pass the Drum Grid path when snapshotting pad edits (`PadCommands`) and when removing a grid/rack/chain, because that device owns the pad subtree. Here the exact key is internal, so `isExternalKey()` returns false and none of the native external plugins inside its pads are captured. `MagdaAudioEngine` has already run the fork capture first, leaving the model with chunks from the fork’s parallel pad instances; undoing the edit/removal can therefore restore a stale patch instead of what the native engine was rendering.

Please expand a Drum Grid path into its external pad-device keys before draining, or otherwise preserve the subtree-capture behavior of the previous call.

lucaromagnoli: Fixed in 2305a4cbf.

Confirmed the whole chain before changing anything: `PluginManager::capturePluginState` recurses through `captureDrumGridPads`, which handles `te::ExternalPlugin` on a pad explicitly, and `collectTrackDevices` walks with `chain_walk::Pads::Enter`, so a pad plugin is its own `DeviceKey` on the native side. The grid path resolving to an internal key did drop all of them.

`keyOfDeviceAt()` is now `keysOfDeviceAt()`: it returns the device at the path plus everything on its pads, collected with `chain_walk::forEachDevice` so the subtree is not knowledge this file keeps a second copy of, then inverted off the same `devicesIn` walk as before. Every external key among them is requested before the one `drain()`, so the call still returns with the reads done.

The header contract said "the one at devicePath" and now says pads come with it. Added `testPadPluginIsItsOwnKey` to pin the premise: a VST3 on a grid pad is the only key `externalKeys()` names, and the grid is not one.

The save path was already correct, since `captureExternalPluginStates()` walks `externalKeys()`, which enters pads.

Issue #2573, opened 2026-09-10
A preset applied to a live plugin never reaches the native engine's instance


lucaromagnoli:
Apply a preset to a plugin while the native engine renders it, and only the parameter array
changes. Anything the chunk carries that a parameter cannot — a sampler's loaded samples and key
mappings, the current program, every value the plugin never exposed — stays on the old patch.

## Why

`TrackManager::applyDevicePreset()` writes `live->pluginState` and then pushes the chunk into the
**Tracktion** instance:

```cpp
if (audioEngine_)
    if (auto* bridge = audioEngine_->getAudioBridge())
        if (auto plugin = bridge->getPlugin(devicePath))
            ...
                applyExternalPluginChunk(plugin.get(), live->pluginState);
```

The native engine's instance hears nothing. It is handed its state exactly once, at load, inside
`adaptExternalPluginInstance` (`applySavedPluginState`). After that the only thing reaching it is
the plan's value layer, which writes the parameter array before each `process()` — so a
parameter-only preset happens to work and a chunk-carrying preset silently does not.

Same for anything else that rewrites `pluginState` or `vst3Preset` on a device that is already
bound: a DAWproject import onto a live slot, an undo of a preset load.

## What it needs

`DeviceControlPlane` (#2270) is the endpoint this belongs on, and it is half built: it has
`captureState` and no apply direction. So:

- an apply/restore operation on the plane, serialized against rendering on the same
  `ControlExecutor` the capture uses — the "reading state and rendering must not overlap" rule
  (#2268) applies identically in this direction
- the identity boundary on the way in, the way the capture has it on the way out: a chunk applied
  to the assignment it was not meant for is one plugin's patch on another
- a host path from `applyDevicePreset` to whichever engine is rendering, rather than to
  `AudioBridge` by name

## Scope

Not #2566, which is the load path. This is the state slice, beside #2244 (writing a native-engine
instance's chunk back out so a project round-trips between engines) — same seam, opposite
direction, and worth doing together.

Found reviewing #2571.

lucaromagnoli:
Merged to dev/0.20.0 as 3d4255ba (#2591). `Closes #2573` does not fire against dev, so closing by hand.

PR #2591, opened 2026-09-11
Apply a preset to the plugin the native engine is rendering

lucaromagnoli:
Closes #2573.

A preset written into the model reached the parameter array by way of the plan and reached
nothing else. An external instance is handed its chunk once, when it is built, so anything only
the chunk carries -- a sampler's loaded samples and key mappings, the current program, every
value the plugin never exposed -- stayed on the patch the project was loaded with.
`applyDevicePreset()` pushed it into the Tracktion instance by name, which is not the one making
sound under magda. A parameter-only preset happened to work; a chunk-carrying one silently did
not.

## The operation

`DeviceControlPlane::applyState()`, on the same `ControlExecutor` as the capture, so a write and
a read cannot be inside one plugin at once (#2268). Same weak registry, same lease, same
assignment boundary between the ask and the answer as #2581.

It applies and then reads the plugin back in the same turn, answering with that snapshot rather
than with a yes. A chunk moves parameters the array underneath it never mentioned, so a model
left holding the preset's own array would be published over the patch at the next block. The
answer is the same thing a save commits, through the same `commitCapturedState`.

A plugin that throws partway through taking a patch is a named failure, not a committed
snapshot: what it holds then is neither patch and no read would describe it.

## The suspension

`applySavedPluginState()` never suspended the instance, only the capture did. At a load the
plugin is not rendering and it cost nothing, but applied to a plugin that is, it is what keeps a
block from running through one halfway between two patches. Scoped, because that path has early
returns and a plugin left suspended renders silence for the rest of the session.

## The seam

`AudioEngine::applyPluginStateAt()`. The fork answers it with the chunk-and-repopulate logic that
used to sit inline in `applyDevicePreset`, moved across unchanged; `MagdaAudioEngine` asks both
engines in turn the way the capture does. `applyDevicePreset` no longer names an engine.

## Tests

Five cases in `test_engine_external_device.cpp`, on the existing plane harness: the preset
reaches the plugin, the answer carries what the plugin holds afterwards over a deliberately stale
array, a plugin that throws is a failure with a reason, an unbound key says so and names the
slot, and the throwing path does not leave the plugin suspended.

Catch2 3923 passed / 0 failed locally, full JUCE suite green.

Hardware verification still wants a real VST: load a project on magda, apply a preset that
carries a chunk, confirm it is audible, save and reload on either engine.

### DIFF
--- a/magda/daw/audio/plugins/engine/DeviceControl.cpp
+++ b/magda/daw/audio/plugins/engine/DeviceControl.cpp
@@ -110,6 +110,56 @@ bool LocalDeviceControlPlane::captureState(magda::engine::DeviceKey key,
     });
 }
 
+bool LocalDeviceControlPlane::applyState(magda::engine::DeviceKey key, magda::DeviceInfo saved,
+                                         CaptureCallback completed) {
+    if (!completed || executor() == nullptr)
+        return false;
+
+    // Same threading and lifetime rules as captureState().
+    return executor()->run([devices = devices_, key, saved = std::move(saved),
+                            completed](ExecutionState state) mutable {
+        if (state == ExecutionState::Cancelled) {
+            completed(CaptureOutcome::failed("the control plane closed before this apply ran"));
+            return;
+        }
+
+        const auto registry = devices.lock();
+        if (!registry) {
+            completed(CaptureOutcome::failed("the runtime that owned " + describeKey(key) +
+                                             " is gone, so nothing was applied"));
+            return;
+        }
+
+        const auto device = registry->find(key);
+        if (device == nullptr) {
+            completed(CaptureOutcome::failed("no plugin is bound for " + describeKey(key) +
+                                             ", so there was nothing to apply it to"));
+            return;
+        }
+
+        if (device->applyState(saved) == magda::SavedStateOutcome::Failed) {
+            // setStateInformation() threw partway, so the plugin holds
+            // neither the old nor the new state. Reading it back would record
+            // that as the state somebody asked for.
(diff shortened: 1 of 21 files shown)


magda/daw/audio/plugins/engine/DeviceControl.cpp
```diff
@@ -110,6 +110,56 @@ bool LocalDeviceControlPlane::captureState(magda::engine::DeviceKey key,
...
+        const auto device = registry->find(key);
+        if (device == nullptr) {
+            completed(CaptureOutcome::failed("no plugin is bound for " + describeKey(key) +
+                                             ", so there was nothing to apply it to"));
+            return;
+        }
+
+        if (device->applyState(saved) == magda::SavedStateOutcome::Failed) {
+            // setStateInformation() threw partway, so the plugin holds
+            // neither the old nor the new state. Reading it back would record
```
lucaromagnoli, line 142: **P1 — Check the assignment before writing into the plugin**

The assignment token is captured only by the completion in `EngineHost::requestApply()`, so this write has already happened by the time `commitCapturedState()` asks whether the slot still holds the assignment that requested it. If the queued operation runs after that key has been reassigned, `registry->find(key)` returns the replacement instance and this applies the old plugin’s opaque chunk to the new plugin. Rejecting the later model commit cannot undo that mutation (and handing an unrelated chunk to a plugin may itself fail badly).

Please carry the `AssignmentRequest` into the queued operation and validate it immediately before resolving/writing the device, or otherwise tie the lease to the assignment being applied. A test that reassigns the key while the apply is queued should cover this direction.

Issue #2592, opened 2026-09-11
Input routing is split across values and compiled topology


lucaromagnoli:
A track's input routing is decided in two places that publish independently, so what is audible can
be the resolved answer, the compiled answer, or a mixture that neither half intended.

## The split, as it stands after #2584

- **Compiled into the plan.** A `track:N` route is an edge; a monitored audio input is an op that
  either exists or does not; whether a track reads MIDI at all decides its merge. All of it comes
  from `activeMidiInputRoute` / `activeAudioInputRoute`, which gate on `TrackInfo::monitorsInput()`.
- **Resolved into a table.** `EngineRuntimeFactory::MidiRouteTable` carries which device sources a
  track hears, rewritten on a values publish so the input the store already holds follows a monitor
  or device change without a republish.
- **Snapshotted separately again.** `LiveMidiSources::available_` is the device list a route
  resolves against, refreshed on hot-plug and at a structural publish.

The three disagree by construction. #2584 closed four defects that were all the same shape:

- a `track:N` change rewrote a table that deliberately holds no sources for it, leaving the old
  compiled edge live (fixed by escalating a values publish to a plan publish when
  `inputRoutingOf(tracks)` differs — a comparison the host performs because nothing else does);
- unarmed `Auto` tracks were audible through the table while the compiler gated them out;
- an "all" route resolved against every device ever registered rather than the ones connected;
- a device selected between publishes resolved to a source nothing pushed under.

Each was fixed where it surfaced. The escalation is the tell: the host now recompares the whole
model against the plan it last compiled, because a values publish cannot say what it changed.

## What to build instead

- One coherent routing snapshot per epoch. An epoch carries the topology and the values it was
  resolved with, rather than publishing them on separate paths and reconciling afterwards.
- Eligibility resolved once, before live MIDI enters the device chain: monitoring mode, arm,
  selection and device availability answered in one place, so "is this audible" has a single
  reading rather than one per layer.
- A route change expressed as an identity the realtime side observes, not as a table mutated
  underneath a bound input.

## Why not in #2584

That PR is the engine switch: until it, nothing in `magda/engine` had ever been driven by an audio
device. The four defects above are fixed and covered; this is the structure underneath them, and it
touches the host/engine boundary rather than the fork split.

Named in review of #2584. Related: #2553 (monitoring and input under the native engine), #2590
(live MIDI source ids), #2579.

lucaromagnoli:
[...]

Issue #2576, opened 2026-09-10
A project teardown is inferred from an empty model, not declared


lucaromagnoli:
Opening a project tears the previous one down and restores the next one, and
nothing announces the gap between the two. Two subsystems already depend on
being able to see it, and both infer it.

## What exists

`ProjectManagerListener` has `projectOpened()` and `projectClosed()`. Neither
fires in the window:

- `projectOpened()` fires **after** the restore (`ProjectManager.cpp:497`), by
  which time the new project is already in place.
- `projectClosed()` fires only on the explicit close path
  (`ProjectManager.cpp:733`). A load never fires it.

So the only signal between one project's runtime ending and the next one's
model arriving is `TrackManager::clearAllTracks()`'s `notifyTracksChanged()`,
which says "tracks changed" and happens to be firing over an empty model.

## Who leans on it

- **AudioBridge.** `ProjectSerializer` deliberately calls `clearAllTracks()`
  *outside* the restore batch so that notification fires immediately and
  `syncAll()` runs against an empty track set. The comment there
  (`ProjectSerializer.cpp:807`) says what breaks if it is coalesced: track and
  device IDs reset to 1 every project, so the paths collide and a plugin
  becomes a dead husk with no editor.
- **The native engine**, since #2572. `EngineHost::tracksChanged()` asks
  `modelHoldsNoDevices()` and calls `EngineRuntimeFactory::forgetBuiltDevices()`
  when the answer is yes.

Both are reading a notification for something it does not say, and both would
go quietly wrong if somebody coalesced that clear into a batch — which is a
one-line change that looks like a performance win.

## What to do instead

Declare it. Roughly: `ProjectManagerListener::projectTeardown()`, fired by the
load path before the restore and by `newProject`/`closeProject`, so a listener
hears "the previous project's runtime is over" rather than deducing it.

Then:

- `EngineHost` listens to that instead of `tracksChanged()`, and
  `modelHoldsNoDevices()` in `EngineProject` goes away.
- `ProjectSerializer`'s clear-outside-the-batch arrangement stops being
  load-bearing, and the comment explaining it becomes history rather than a
  warning.

Found reviewing PR #2575.

lucaromagnoli:
Landed in #2583.

Issue #2570, opened 2026-09-10
Per-slot device meters read nothing on the native engine


lucaromagnoli:
Every meter in the app is dead when `MAGDA_AUDIO_ENGINE=magda` (#2551). Found on the first
hands-on pass.

`EngineRuntimeFactory` never overrides `RuntimeStateFactory::createMeter`, so it returns the base
`nullptr` and no `LevelTap` is ever built. The plan emits a Meter op at every track, every device
slot and the master; nothing reads any of them.

## The wiring, which is shorter than it looks

The mixer already reads through one place: `MixerView` pops `AudioBridge::getMeteringBuffer()` per
track (`MeteringBuffer::popLevels`). Write into that and the UI needs no changes at all.

- `EngineRuntimeFactory::createMeter(const OpKey&)` returns a `LevelTap` for `OpRole::TrackMeter`
  (post-fader, pre-mute -- what a channel strip shows) and declines the rest, which is the ordinary
  answer for a meter nobody reads. Keep the raw pointers keyed by TrackId; the store owns them and
  they outlive plan swaps by design.
- `EngineHost` reads them on its timer -- `LevelTap::read()` is destructive and off the audio
  thread, exactly the shape a meter wants -- and hands each track's peak to a sink.
- `MagdaAudioEngine` supplies that sink, pushing into the fork's `MeteringBuffer`, since it is the
  only side that holds both.

`OpRole::DeviceMeter` feeds the chain UI's per-slot meters and can follow the same route through
`DeviceMeteringManager`; `LiveInputMeter` belongs with monitoring (#1895).

## The master is its own question

`MixerView` reads the master from `AudioBridge::getMasterPeakL/R()` rather than from the metering
buffer, and those are fed by the TE graph. Either the bridge grows a setter or the master strip
reads the buffer like every other strip.

## Why it was not in #2551

Slice 1 is the audible increment -- a MIDI clip driving an instrument through a real device. A
meter is a display, and wiring one needs a route into UI plumbing that is still entirely
fork-owned. Listed here rather than left as a surprise, which is what it was.

---

## Where this stands after #2584

The track rings and the master peak are done. `EngineRuntimeFactory::createMeter` answers
`OpRole::TrackMeter` with a `LevelTap`, `EngineHost` reads them on its timer, and
`MagdaAudioEngine` feeds a `TrackMeters` both engines own through `AudioEngine::meters()` — so the
[...]

PR #2594, opened 2026-09-11
Open a plugin's editor onto the instance that renders it

lucaromagnoli:
Fixes #2580.

Under magda no plugin window opened at all. `PluginWindowManager` drives
`te::ExternalPlugin::windowState`, and the window that used to open was onto the fork's parallel
copy of the plugin — not the one you hear. #2579 removed that copy, so the click went nowhere.

## The endpoint

`DeviceControlPlane` grows `editorWindow(key, action, completed)` beside `captureState`, with
`EditorAction` of Show, Hide, Toggle or Query. Keyed by `DeviceKey`, answered on the control
executor, fallible: a device that has gone answers with a reason rather than dangling.

The executor is the message thread's, which is where a window may be opened at all, and it is
what serialises an editor being built against a state being read — the second half of why the
plane exists (#2268).

## The window

`EngineExternalDevice` owns it, being the only object allowed to reach the instance: the header
has no accessor by design and this does not add one. A `DocumentWindow` whose close button tells
the device rather than deleting itself, so the device's `unique_ptr` is both the owner and the
answer to "is it open".

Lifetime falls out of that ownership rather than needing a second mechanism:

- a slot whose plugin changed retires the old instance, which `releaseDeleted` destroys at the end
  of that same publish (#2575);
- a project torn down forgets every device it held (#2576);
- the session going away destroys the store, and the store owns the instances.

Each of those runs on the message thread, which the destructor asserts when a window is open.

## The callers

`AudioEngine` grows show, hide, toggle and is-open. The fork answers from `AudioBridge` exactly as
before; `MagdaAudioEngine` asks the host, which maps the `ChainNodePath` to one key — the device
the path names, without the pads a capture also takes, since a pad's editor is its own.

The five sites that reached for `AudioBridge` by name now ask the engine that is rendering:
`DeviceSlotComponent` (the UI button, the double-click, the light its timer draws),
`MiniChainRow`, and `DeviceApiLive::openDeviceEditor`. The slot's timer no longer returns early
when there is no bridge, which is why its light was dead under magda; the bridge is still what the
per-slot meter reads, and that stays #2570.

## And the power button, found in the same pass

A device's power button wrote `device.bypassed` and published values. Bypass is structural — the
compiler takes a bypassed device out of the plan — so the plan kept rendering it and the button did
nothing under magda.

`publishValues` now compiles and compares `planFingerprint` against the live plan whenever the edit
was a property rather than a parameter, and publishes the plan it just compiled when the shape
moved. The fingerprint is the exact question; a list of which properties are structural would be
one more thing to keep in step with the compiler (#2592). It replaces the routing signature from
#2584, which was that list in miniature.

Confirmed by hand: the power button works.

## Tests

- A plugin with no editor of its own opens no window, and answers "not showing" rather than
  failing — nothing went wrong, there is nothing to open. A hide on a window that was never open
  is not a failure either.
- An editor asked of a key with no device bound says which key, the way a capture does.
- A fader move leaves the plan's fingerprint alone; bypassing a device, taking MIDI from another
  track while monitoring, and naming a monitored audio input each move it.

`magda_tests` 3931 passed, 13 skipped. `magda_juce_tests` all passed.

## Not driven by hand

The issue's own check: under magda, open an external plugin's editor, move a control, hear it;
close the project with the window open and nothing dangles. Under Tracktion, unchanged. Worth a
pass with a VST3 before merging — the corpus has no windows in it, and CI has no display.

Out-of-process editors stay #1899, which is why the endpoint is asynchronous.

### DIFF
--- a/magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
+++ b/magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
@@ -1,8 +1,11 @@
 #include "plugins/engine/EngineExternalDevice.hpp"
 
+#include <juce_gui_basics/juce_gui_basics.h>
+
 #include <algorithm>
 #include <atomic>
 #include <cmath>
+#include <functional>
 #include <utility>
 
 #include "core/ParameterUtils.hpp"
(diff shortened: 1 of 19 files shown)


magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
```diff
@@ -78,6 +81,44 @@ void clearNonFinite(juce::AudioBuffer<float>& audio, int numSamples) {
...
+    /// The editor goes before the window it sits in, and while the plugin is
+    /// still there to be told it has gone.
+    ~EditorWindow() override {
+        clearContentComponent();
+    }
+
+    void closeButtonPressed() override {
+        // Not delete-this: the owner holds this by unique_ptr.
+        if (closed_)
+            closed_();
```
lucaromagnoli, line 115: [P1] Defer window destruction until the close callback has returned

`closed_()` invokes `[this] { hideEditor(); }`, and `hideEditor()` immediately resets `editor_`. That synchronously deletes this `EditorWindow` while `closeButtonPressed()` is still on the stack. We already fixed this exact ownership pattern in `PluginEditorWindow::closeButtonPressed()` by posting the close, because the immediate reset caused heap corruption. That implementation also deliberately uses a JUCE title bar: with a native macOS title bar, the OS may continue close processing after this callback returns, so deleting the peer here introduces a second race. Please keep the window alive through the callback (for example, defer via a `Component::SafePointer`) and avoid the native title bar.

Issue #2437, opened 2026-09-06
Translate a 4OSC patch to a Poly Synth patch for the native engine


lucaromagnoli:
Follows the decision recorded on #2298: 4OSC is never ported to the native engine.

`te::FourOscPlugin` is Tracktion Engine's own synth rather than MAGDA code, so it lives and dies with the fork. 0.20 ships the dual-engine switch, which means a user can move a project onto the native leg while 4OSC is still in the browser on the TE leg, and on the native leg the device does not exist at all. The answer for those projects is migration, not a port.

## What it has to do

Read a saved 4OSC patch and write a Poly Synth patch that sounds as close to it as the two instruments allow: four oscillators (wave, level, coarse, fine), the filter's type, cutoff, resonance and envelope amount, both ADSRs, voice mode, glide, bend range and the velocity amounts. `MagdaPolySynthCompiledPlugin`'s host slots are the target and are already frozen.

Unison is Poly Synth's known gap against 4OSC. A patch that uses it needs to say so rather than translate silently into something thinner.

## When it runs

A project carrying 4OSC that is switched to the native engine, and the same project loaded after v1 when the fork and 4OSC are gone. The first of those is a v0 user-facing path, not a v1 migration, which is what moves this earlier than the cutover.

## Verification

The seven legacy projects that `test_real_project_survey.cpp` reports as `no native device for 4osc` stop reporting it and render. They do not become null-diff cases: the two legs would be playing different instruments, so there is nothing to null against. What can be asserted is that the translated patch loads, sounds, and puts every 4OSC control somewhere a reader can point at.

lucaromagnoli:
Done in #2604.

PR #2604, opened 2026-09-11
Translate 4OSC patches to Poly Synth and offer the project migration

lucaromagnoli:
Closes #2437.

4OSC is Tracktion Engine's synth and is never ported, so a project carrying one
has nothing to play on the MAGDA engine. This translates the patch into a Poly
Synth patch and offers the whole thing as a migration when a Tracktion project
is opened on the native leg.

## Translation

`FourOscTranslation` reads a saved 4OSC state and writes Poly Synth host slots:
four oscillators (wave, level, coarse, fine), filter type, cutoff, resonance and
envelope amount, both ADSRs, voice mode, glide, bend range and the velocity
amounts. 4OSC's built-in chorus, reverb, delay and distortion become a rack of
MAGDA devices behind the synth, since Poly Synth has no effects of its own.

A parameter the saved state never wrote takes 4OSC's own default rather than
zero, and the state is addressed by parameter index, which is what 4OSC writes.

Unison is the gap: Poly Synth has no equivalent, so a patch that uses it is
named in the prompt rather than translated into something thinner.

## The prompt

Opening a project that was last written by Tracktion offers to reopen it as a
copy on the MAGDA engine. It fires whether or not there is a 4OSC in it, because
the two engines do not render every device the same way, and it folds the 4OSC
translation in when there is one. The copy is saved beside the original as
"<name> (magda engine)" and the original is never written to.

`ProjectInfo::savedWithEngine` records which engine last wrote the project, so a
project already through the migration is not asked again. An empty word is a
project saved before the field existed, which is Tracktion by definition. A
"don't ask again" tickbox turns the prompt off for good.

## Under the MAGDA engine

- The 4OSC custom UI does not open: every control writes through a
  `te::FourOscPlugin` that the native engine never builds, so the slot shows the
  device and nothing to turn.
- The plugin browser hides any internal device with no native factory, rather
  than offering one that arrives silent.

## Tests

`tests/core/test_four_osc_translation.cpp`: the slot mapping, the defaults, the
effects rack, the gap reporting, the prompt text and the name of the copy.

### DIFF
--- a/magda/daw/audio/FourOscMigration.cpp
+++ b/magda/daw/audio/FourOscMigration.cpp
@@ -114,10 +114,15 @@ juce::File convertedProjectFileFor(const juce::File& project) {
     if (!projects.isDirectory())
         return {};
 
-    const auto name = project.getFileNameWithoutExtension() + " (Poly Synth)";
-    const auto folder = projects.getChildFile(name).getNonexistentSibling();
+    // Unwrapped, and deliberately: saveProjectAs() builds the folder itself,
+    // and only when the file is not already sitting in one of its own name.
+    // Handing it the wrapped path meant it created nothing and wrote into a
+    // directory that was not there.
+    const auto folder =
+        projects.getChildFile(project.getFileNameWithoutExtension() + " (Poly Synth)")
+            .getNonexistentSibling();
 
-    return folder.getChildFile(folder.getFileName() + project.getFileExtension());
+    return projects.getChildFile(folder.getFileName() + project.getFileExtension());
 }
 
 juce::String describeConversion(const std::vector<FourOscCandidate>& candidates) {
--- a/magda/daw/audio/FourOscMigration.cpp
+++ b/magda/daw/audio/FourOscMigration.cpp
@@ -2,7 +2,9 @@
 
 #include <set>
 
+#include "../core/AutomationManager.hpp"
 #include "../core/ChainWalk.hpp"
+#include "../core/DeviceParamMigrations.hpp"
 #include "../core/TrackInfo.hpp"
 #include "../core/TrackManager.hpp"
 
@@ -42,16 +44,23 @@ std::vector<juce::String> distinctGaps(const std::vector<FourOscCandidate>& cand
  *
  * The rack of effects goes in directly after the synth it came from, so it
  * reaches the same signal 4OSC's own effects did.
+ *
+ * @p replaced collects the path of every device converted, which is what the
+ * links and lanes still addressing 4OSC's parameters are found by.
  */
-int convertIn(std::vector<ChainElement>& elements, TrackManager& tracks) {
+int convertIn(std::vector<ChainElement>& elements, const ChainNodePath& parentPath,
+              TrackManager& tracks, std::set<ChainNodePath>& replaced) {
     auto converted = 0;
 
     for (auto index = 0; index < static_cast<int>(elements.size()); ++index) {
         auto& element = elements[static_cast<std::size_t>(index)];
 
         if (isRack(element)) {
-            for (auto& chain : getRack(element).chains)
-                converted += convertIn(chain.elements, tracks);
+            auto& rack = getRack(element);
+            const auto rackPath = chain_walk::rackIn(parentPath, rack.id);
+            for (auto& chain : rack.chains)
+                converted +=
+                    convertIn(chain.elements, rackPath.withChain(chain.id), tracks, replaced);
 
             continue;
         }
@@ -60,10 +69,13 @@ int convertIn(std::vector<ChainElement>& elements, TrackManager& tracks) {
             continue;
 
         auto& device = getDevice(element);
+        const auto devicePath = chain_walk::deviceIn(parentPath, device.id);
 
         if (device.pads)
             for (auto& pad : device.pads->chains)
-                converted += convertIn(pad.elements, tracks);
+                converted += convertIn(
+                    pad.elements, ChainNodePath::padChain(parentPath.trackId, device.id, pad.id),
+                    tracks, replaced);
 
         if (!isFourOscDevice(device))
             continue;
@@ -71,6 +83,7 @@ int convertIn(std::vector<ChainElement>& elements, TrackManager& tracks) {
         auto translated = translateFourOsc(device, [&tracks] { return tracks.allocateDeviceId(); });
         device = std::move(translated.device);
         ++converted;
+        replaced.insert(devicePath);
 
         if (translated.effects == nullptr)
             continue;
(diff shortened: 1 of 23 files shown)


magda/daw/audio/FourOscMigration.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+        if (device.pads)
+            for (auto& pad : device.pads->chains)
+                converted += convertIn(pad.elements, tracks);
+
+        if (!isFourOscDevice(device))
+            continue;
+
+        auto translated = translateFourOsc(device, [&tracks] { return tracks.allocateDeviceId(); });
+        device = std::move(translated.device);
```
lucaromagnoli, line 87: **P1 — Migrate or remove parameter targets before preserving this device ID**

Automation, modulation links, macros, and controller bindings address plugin parameters by device path plus paramIndex. Reusing the ID while replacing the parameter schema silently changes their meaning: 4OSC index 0 is Tune 1, while Poly Synth index 0 is Osc 1 Wave; index 1 changes from Fine Tune 1 to Osc 1 Level, and so on. Please explicitly map compatible targets and remove or report incompatible ones rather than leaving them attached to unrelated controls.

— Codex

magda/daw/ui/dialogs/FourOscConversionPrompt.cpp
```diff
@@ -0,0 +1,108 @@
...
+};
+
+/// Convert, then save as a project of its own. The one that was opened is
+/// never written to, so a 4OSC project stays openable on the Tracktion
+/// engine.
+void convertAndSave() {
+    auto& projects = ProjectManager::getInstance();
+    const auto destination = audio::convertedProjectFileFor(projects.getCurrentProjectFile());
+
+    audio::convertFourOscDevices(TrackManager::getInstance());
```
lucaromagnoli, line 36: **P2 — Mark the in-session conversion dirty before attempting Save As**

convertFourOscDevices mutates the track model directly and only sends device-change notifications. If there is no destination or Save As fails, the conversion remains in memory as the warning says, but the project can still be considered clean and close without an unsaved-changes prompt. Please mark it dirty immediately after a successful conversion; a successful save will clear the flag normally.

— Codex

magda/daw/ui/dialogs/FourOscConversionPrompt.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    // An unsaved project has nowhere to sit beside, so it keeps whatever the
+    // conversion did and the user names it at their next save.
+    if (destination == juce::File{})
+        return;
+
+    if (!projects.saveProjectAs(destination))
```
lucaromagnoli: **P1 — Preserve the source project media when creating the converted copy**

This calls the normal Save As path, which invokes migrateMediaFiles and physically moves recordings, renders, and imports into the new project. The original .mgd is not rewritten, so it continues to reference the old paths and can reopen with missing media. Please use copy semantics for this migration flow so the original project and its media tree are genuinely left intact.

— Codex

magda/daw/audio/FourOscTranslation.cpp
```diff
@@ -0,0 +1,495 @@
...
+        setSlot(device, Chorus::kDepthSlot,
+                clampToSlot(device, Chorus::kDepthSlot, value("chorusDepth", 0.0f) / 20.0f));
+        setSlot(device, Chorus::kWidthSlot,
+                clampToSlot(device, Chorus::kWidthSlot, value("chorusWidth", 0.0f)));
+        setSlot(device, Chorus::kMixSlot,
+                clampToSlot(device, Chorus::kMixSlot, value("chorusMix", 0.5f)));
+        chain.elements.push_back(ChainElement{std::move(device)});
+    }
+
+    if (on("delayOn")) {
```
lucaromagnoli, line 526: **P2 — Translate the 4OSC delay time and sync behavior**

The replacement copies feedback, crossfeed, and mix but never reads the delay property. 4OSC computes delay seconds from that value and the current tempo, with the default representing one beat; this device therefore stays at Magda Delay's unsynchronised 250 ms default. Please enable sync and map the beat value to a division, or otherwise preserve the original timing.

— Codex

magda/daw/audio/FourOscTranslation.cpp
```diff
@@ -0,0 +1,583 @@
...
+        using Delay = compiled::MagdaDelayCompiledPlugin;
+        auto device = compiledDevice<Delay>(nextId());
+        // 4OSC holds its delay in beats and divides by the tempo at render
+        // time (FourOscPlugin.cpp), so the synced division is the same number
+        // and the free-time slot would be the wrong home for it.
+        setSlot(device, Delay::kSyncSlot, 1.0f);
+        setSlot(device, Delay::kDivisionSlot,
+                clampToSlot(device, Delay::kDivisionSlot,
+                            nearestDelayDivision(floatPropertyOr(props, "delay", 1.0f))));
```
lucaromagnoli: **P1 — Store the delay division choice index, not its Faust multiplier**

`kDivisionSlot` is a discrete menu parameter whose saved value is the choice index. `nearestDelayDivision()` instead returns the Faust beat multiplier. For example, the legacy default of one beat is written as `1.0`, which selects menu index 1 (`1/16T`) rather than the quarter-note entry at index 8. Every translated synced delay can therefore use the wrong timing.

Please find and store the nearest menu index (using the harvested division values), and add a test asserting that one beat selects the entry whose Faust value is `1.0`.

magda/daw/audio/FourOscMigration.cpp
```diff
@@ -0,0 +1,200 @@
...
+    });
+
+    if (replaced.empty())
+        return converted;
+
+    // Poly Synth's parameters are different controls in different units at the
+    // same indices, and the path a link names still resolves. A macro left
+    // pointing at 4OSC's Tune 1 would drive Poly Synth's Osc 1 Wave.
+    tracks.forEachTrackIncludingMaster([&replaced](TrackInfo& track) {
+        device_param_migrations::dropParamLinksInTrack(track, replaced);
```
lucaromagnoli, line 230: **P1 — Remove controller bindings and concrete aliases for replaced parameter paths**

This cleanup only visits macro/mod links and automation lanes. Project controller bindings in `BindingRegistry` and concrete path-bound aliases in `AliasRegistry` still retain the reused device path and parameter index. Because that path remains valid after replacement, those references resolve to unrelated Poly Synth parameters and are saved into the converted project.

Please remove or deliberately remap every explicit plugin-parameter reference to a replaced path, including project controller bindings and concrete user aliases.

magda/daw/audio/FourOscTranslation.cpp
```diff
@@ -0,0 +1,583 @@
...
+            return 0;
+    }
+}
+
+/// A Poly Synth at its defaults, which is what everything 4OSC does not
+/// describe is left at.
+DeviceInfo polySynthDevice(const DeviceInfo& fourOsc) {
+    const PolySynth metadata;
+
+    DeviceInfo device;
```
lucaromagnoli: **P1 — Preserve host-owned device state when replacing 4OSC**

Constructing a fresh `DeviceInfo` preserves the id, name, and bypass state, but resets generic host-owned instance state such as `gainValue`/`gainDb`, device macro values and names, and modulators. A non-unity device gain immediately changes the converted project sound, while macro/modulator configuration is silently discarded.

Please copy the generic host-owned fields from the source device, then selectively remove only links that target the replaced 4OSC plugin parameters.

Issue #2600, opened 2026-09-11
A device parameter's value text comes from the fork's processor, so the native engine has none


lucaromagnoli:
Every parameter cell, automation lane and curve tooltip formats its value through
`ParameterInfo::displayText`, and the one implementation of that provider goes
`TrackManager::getAudioEngine()` -> `getAudioBridge()` -> `getDeviceProcessor(path)` ->
`DeviceProcessor::formatParameterValue()` (`DeviceParameterDisplayTextProvider.cpp:14-31`).
That is the fork's chain. A device the native engine holds and the fork does not answers nothing,
and the formatter falls back to dispatching on `unit` and the parameter's range.

For an external plugin that is the whole answer: the range is normalised (the fork wraps every
external parameter in 0..1), `unit` is empty, so a cutoff reads `0.50` instead of `500 Hz` and a
filter type reads `0.33` instead of `Bandpass`. The plugin knows the string —
`juce::AudioProcessorParameter::getText()` — and nothing asks it.

Only visible now that #2595 puts an external plugin's parameters into the model at all; before
that the grid drew placeholders and there was nothing to format.

## Where it goes missing

- `ExternalPluginProcessor::getParameterInfo` sets the provider for the fork's devices
  (`ExternalPluginProcessor.cpp:89`), and `NativeDeviceProcessors.cpp:95` does the same for the
  fork's internal ones. `describeHostParameters()` sets none, because there is no processor behind
  a native-engine device to point one at.
- The fork chose a live provider over a sampled `valueTable` on purpose ("no dangling pointers, no
  stale sampled tables, exact values"), so the fix is another implementation of the same seam, not
  a table captured at load.

## What to build

A provider that reaches the rendering engine's device by `ChainNodePath` / `engine::DeviceKey` and
asks it to format a normalised value — answered by the fork's processor under Tracktion and by the
engine's own device under magda. `EngineExternalDevice` owns the instance and is the only thing
allowed to call `getText()` on it, the same rule `captureState()` and `showEditor()` follow, so the
call belongs on the control plane beside them (#2270).

This is the parameter-text half of the seam #2585 describes for analyzer and custom-UI telemetry,
and it wants the same lookup. Internal SDK devices need the same answer once they render without a
fork mirror.

## Verification

Under `MAGDA_AUDIO_ENGINE=magda`, an external plugin's parameter cell shows the plugin's own string
and it tracks the plugin's editor; an automation lane over it labels its axis the same way. Under
[...]

Issue #2601, opened 2026-09-11
Translate a hosted plugin's normalized parameters into its real units


lucaromagnoli:
A VST3/AU parameter is a number between 0 and 1. The plugin knows it means 500 Hz, or `Bandpass`,
or -6 dB, and MAGDA has everything needed to say so — and then keeps none of it on the device.

After #2595 an external plugin's parameters reach the model, and every one of them arrives bare:
`minValue`/`maxValue` and `teMinValue`/`teMaxValue` all 0..1, `unit` empty, `scale` Linear. That is
deliberate and matches the fork, which refuses to adopt `param->getLabel()` while the range is
normalized (`ParameterInfoBuilder.cpp:29-32`) because "Hz" over 0..1 makes the formatter produce
garbage. So the cell reads `0.50`, the automation lane's axis is labelled 0 to 1, the macro's
range means nothing, and MCP reports a cutoff as a fraction.

## What already exists

- **Detection.** `ParameterDetector::detect()` is a deterministic heuristic pass and
  `detectWithAI()` the fallback for what it cannot resolve; both take a `ParameterScanInput`
  (name, label, range, state count, sampled display texts) and answer with a unit, a scale, a real
  range, bipolarity and choices. `ParameterConfigDialog` drives them from its "Detect" and
  "AI Detect" buttons.
- **A scan that does not need the fork.** `TracktionEngineWrapper::scanPluginParameters()` opens
  its own throwaway instance from the format manager and samples `parameter->getText()`, walking
  discrete states where there are any and falling back to five sample points otherwise. Only the
  internal branch needs an Edit.
- **Storage.** `PluginParameterConfigStore` keeps unit, scale, range, choices and value table per
  plugin, keyed by `uniqueId`, one XML file each.
- **The conversion.** `ParameterUtils` already carries a parameter in two domains at once: the
  display one (`minValue`/`maxValue`/`scale`) and the one the live parameter stores
  (`teMinValue`/`teMaxValue`), with `modelToNormalizedValue()` / `normalizedToModelValue()`
  between them.

Everything is there. Nothing joins it up for a hosted plugin.

## Where it comes apart

**1. The overlay never survives a rebuild.** Nothing applies a stored config when a device is
built. `PluginParameterConfigStore::applyToDevice` is called from `ParameterConfigDialog` on save
(through `refreshLiveDevices`), from the MCP write path (`device_api_live.cpp`), from
`MiniChainRow::resolveParams` onto the live model, and from
[...]

PR #2605, opened 2026-09-11
Ask the rendering engine for a parameter's value text

lucaromagnoli:
Closes #2600.

A hosted plugin's parameters are normalised, so there is nothing for a formatter
to work from: after #2599 the grid shows them, and a cutoff reads `0.50` while a
filter type reads `0.33`. The plugin knows the string — `getText()` — and
nothing asked it.

## The seam was pointing at the fork

`ParameterInfo::displayText` is already the right seam, and the fork chose a live
provider over a sampled table on purpose ("no dangling pointers, no stale sampled
tables, exact values"). Its one implementation went `getAudioBridge()` →
`getDeviceProcessor(path)` → `DeviceProcessor::formatParameterValue()`.

That is the fork's chain, and `MagdaAudioEngine::getAudioBridge()` returns null by
design — an Edit would bring a bridge and a second copy of every plugin (#2579).
So under magda **no** device had value text, internal ones included, not just
hosted plugins.

`AudioEngine::formatDeviceParameter(path, paramIndex, normalised)` is now the
seam: `TracktionEngineWrapper` answers through the processor as before,
`MagdaAudioEngine` asks the host first and falls back to the fork for the
internal devices it still syncs. Empty from either means "format from the
parameter's range", which `ParameterUtils::formatValue` already does — an engine
that cannot answer degrades rather than breaks.

## The one operation that skips the control plane

`EngineExternalDevice::parameterText()` does not go through `DeviceControlPlane`,
and the header says why at the declaration. A value's text is read inside a
paint — once per cell per repaint, hundreds of times over when a lane labels its
axis (`DeviceProcessor::formatParameterValue` already carries a comment about
Vital beach-balling the UI here) — and the plane queues everything it is given
rather than running it inline, deliberately (`ControlExecutor.hpp`). No queued
round trip can serve a paint.

It is also the only device operation that earns the exemption: it neither
suspends the plugin nor changes anything about it, which is why the fork calls
the same JUCE method unlocked from the message thread today. And the message
thread it runs on *is* the control executor's
(`MessageThreadControlExecutor`), so it overlaps no operation that does suspend.

## Also moved

`installDeviceParameterDisplayTextProviderFactory()` moves from the AudioBridge's
constructor to `TracktionEngineWrapper::initialiseServices()`, which both engines
start. The provider no longer touches the bridge, and left where it was the
registration never happened under magda — the factory would have returned null
and nothing would have been attached.

## Tests

Four cases in `tests/engine/test_engine_external_device.cpp`; three fail with the
implementation gutted, and the fourth is a guard with a positive control so it
fails too. The stub parameter gained a `getText` returning `"<n> frobs"` — a
string no formatter working from the range could produce, which is how the test
tells the plugin's own text from the host's arithmetic.

Covered: the text is the plugin's own; the wrapper pair and an out-of-range slot
answer empty while a real slot answers; the value is clamped before the plugin
sees it; and `describeHostParameters` attaches a provider carrying the right
device id and slot, with none on the wrapper pair.

`make test` (3946 passed, 13 skipped) and `make test-juce` green, corpus
included. `make lint-changed` reports nothing on the new code.

## Note for #2601

Attaching a provider also aligns the native path with the fork on something
#2601 has to settle: `isDisplayMappedInternalValue()` gates on
`displayText == nullptr`, so without one a configured display range would have
made the native engine store a real value where the fork stores a normalised
one. Same number on disk under both engines now, which is one less thing for
#2601 to migrate.

### DIFF
--- a/magda/daw/audio/plugin_manager/ExternalPluginState.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginState.cpp
@@ -179,6 +179,12 @@ HostParameters describeHostParameters(const juce::AudioPluginInstance& instance,
         info.stableId = hostParameterId(*parameter);
         info.defaultValue = parameter->getDefaultValue();
         info.currentValue = parameter->getValue();
+
+        // Asked for the plugin's own string rather than sampled into a table,
+        // which is the rule the fork set for the same reason: a table is a
+        // guess between its samples, and a hosted plugin's range is normalised
+        // so there is nothing else to format from (#2600).
+        info.displayText = makeParameterDisplayTextProvider({}, device.id, index);
         described.parameters.push_back(std::move(info));
     }
 
--- a/magda/daw/audio/plugin_manager/ExternalPluginState.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginState.cpp
@@ -180,11 +180,6 @@ HostParameters describeHostParameters(const juce::AudioPluginInstance& instance,
         info.defaultValue = parameter->getDefaultValue();
         info.currentValue = parameter->getValue();
 
-        // Asked for the plugin's own string rather than sampled into a table,
-        // which is the rule the fork set for the same reason: a table is a
-        // guess between its samples, and a hosted plugin's range is normalised
-        // so there is nothing else to format from (#2600).
-        info.displayText = makeParameterDisplayTextProvider({}, device.id, index);
         described.parameters.push_back(std::move(info));
     }
 
--- a/magda/daw/audio/AudioBridge.cpp
+++ b/magda/daw/audio/AudioBridge.cpp
@@ -13,7 +13,6 @@
 #include "../engine/PluginWindowManager.hpp"
 #include "../profiling/PerformanceProfiler.hpp"
 #include "AudioThumbnailManager.hpp"
-#include "DeviceParameterDisplayTextProvider.hpp"
 #include "Vst3Preset.hpp"
 #include "modifiers/ADSRDebugLog.hpp"
 #include "plugin_manager/ExternalPluginState.hpp"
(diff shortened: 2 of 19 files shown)


magda/daw/audio/plugin_manager/ExternalPluginState.cpp
```diff
@@ -179,6 +179,12 @@ HostParameters describeHostParameters(const juce::AudioPluginInstance& instance,
         info.stableId = hostParameterId(*parameter);
         info.defaultValue = parameter->getDefaultValue();
         info.currentValue = parameter->getValue();
+
+        // Asked for the plugin's own string rather than sampled into a table,
+        // which is the rule the fork set for the same reason: a table is a
+        // guess between its samples, and a hosted plugin's range is normalised
+        // so there is nothing else to format from (#2600).
+        info.displayText = makeParameterDisplayTextProvider({}, device.id, index);
```
lucaromagnoli: This provider is created without a `ChainNodePath`, so formatting later falls back to `TrackManager::findDevicePath(deviceId)`. That lookup searches only the main FX tree, while device IDs are section-scoped and may overlap between FX and post-FX. A native-engine external plugin in post-FX will therefore either fall back to generic formatting or query an unrelated main-FX plugin with the same ID and display that plugin’s text. The automation backstop cannot repair this because it only replaces a null provider. Please preserve or recover the section-aware path when creating/resolving this provider.

lucaromagnoli: Right, and it was reachable: the two id counters are independent, so an FX device 1 and a post-FX device 1 coexist routinely.

Fixed by making the id an identity rather than by recovering the path afterwards. `findDevicePath` gains an overload taking a `ChainSegment` — Fx descends racks and pads as before, the two flat sections are one level deep — and the unqualified spelling still means the FX tree for its existing callers.

The provider is now attached where that identity is known. `describeHostParameters` only has a `DeviceInfo`, which does not say which section holds it, so the attach moved to `EngineHost::applyLoadedDevice`, which has the `DeviceKey`. `attachParameterTextProviders` stores the resolved path in each provider, so the id fallback never runs at format time; an unresolvable path leaves the parameters with no provider, which formats from the range — wrong-looking rather than wrong.

`applyLoadedDevice`'s property notification carried the same FX-only guard for the same reason, so it uses the resolved path too and a post-FX slot is now told its device changed.

New tests in `tests/devices/test_device_parameter_text.cpp` pin it, including the colliding-id case: two devices numbered 1, one per section, each resolving to its own path. Both section tests fail if the segment is ignored.

759b03be

PR #2606, opened 2026-09-11
Keep a hosted plugin's detected units, and agree on what its value means

lucaromagnoli:
First slice of #2601. Not `Closes` — the issue's item 4 and the draw-time
appliers are still open, listed at the bottom.

MAGDA already detects a hosted plugin's real units: `ParameterDetector::detect`
plus an AI pass, driven from the Configure Parameters dialog, fed by a scan that
opens its own instance and samples `getText()`, stored per plugin in
`PluginParameterConfigStore`. Nothing kept the result.

## The units survive a rebuild

Both engines rebuild an external plugin's parameter array wholesale on load —
`populateParametersFromEngine` clears both buckets, `describeHostParameters`
returns fresh records — and nothing re-applied the stored config, so a detected
unit lasted until the next device rebuild and no longer. `applyToDevice` was only
ever called from the dialog's save, the MCP write path, and two draw-time UI
sites.

`EngineHost::applyLoadedDevice` applies it now, beside the text providers from
#2605 and for the same reason: that is where a published device is assembled and
where its identity is known.

`PluginParameterConfigStore::applyToDevice` gains an overload addressed by the
device. The id a config is filed under is the device's own — `uniqueId` where it
has one, `pluginId` otherwise — and every caller derived that by hand, most of
them reading `uniqueId` alone and so missing older devices entirely.

## They land on the parameter they were saved for

A `PluginParameterConfigEntry` is stored by its position in the device's
parameter list. The dialog's scan built a *different* list: it walked the
plugin's raw array and dropped unnamed automatable parameters, which the model
keeps. One such parameter shifted every later override onto the wrong control,
silently.

The scan now walks the host list `describeHostParameters()` builds, so the two
agree by construction rather than by both happening to filter the same way. It
also inherits the fork's naming, so a dialog row and a grid cell read the same.

## The value keeps one meaning

A detected range puts the model and the plugin in different units.
`applySavedPluginState()` and `applyRestoredParameters()` were the one pair
converting through the display range unconditionally, while the plan, the UI,
automation, baking and controller writes all go through
`modelToNormalizedValue()` / `normalizedToModelValue()`, which pick the domain by
`isDisplayMappedInternalValue()`. With a range configured, a load wrote a
frequency into a field everything else reads as normalised — a bug that only
bites once units persist, which is what the first half of this PR does.

Both directions use that pair now. This corrects what I originally wrote into
#2601: dropping the `displayText` gate from the predicate would have changed
which domain automation and controllers use for *internal* devices too, since
`NativeDeviceProcessors` attaches providers as well. The predicate stays; the
restore was the thing out of step. (Commented on the issue.)

## Tests

- The round trip with a configured range: a model holding 0.7 against a
  20 Hz–20 kHz log range reaches the plugin as 0.7 and comes back as 0.7. Both
  assertions fail with the old conversion restored — it hands the plugin 0.0 and
  writes 20 back.
- The device-addressed `applyToDevice`, including a device with no `uniqueId`
  falling back to its `pluginId`.

The scan alignment is structural rather than asserted: it now shares
`describeHostParameters()`'s list, and the unnamed-parameter retention that used
to differ is already pinned by "Repeated and missing parameter names are the
fork's". Testing the scan directly needs a real installed plugin.

`make test` (3953 passed, 13 skipped) and `make test-juce` green, corpus
included. `make lint-changed` reports nothing on the new code.

## Left for #2601

- The fork loses the config on rebuild the same way; only the native path is
  fixed here.
- `MiniChainRow` and `DeviceSlotParameterPaging` still apply the config at draw
  time. Now redundant on the native path, but removing them touches UI I cannot
  drive to verify.
- Item 4: the scan's index space is aligned, but `stableId` is still not the
  stored identity — a config written before a plugin update still addresses by
  position.

### DIFF
--- a/magda/daw/audio/plugin_manager/ExternalPluginState.cpp
+++ b/magda/daw/audio/plugin_manager/ExternalPluginState.cpp
@@ -4,7 +4,6 @@
 #include <map>
 
 #include "../Vst3Preset.hpp"
-#include "core/ParameterUtils.hpp"
 
 namespace magda {
 
@@ -88,6 +87,30 @@ void applyVst3Records(DeviceInfo& device, const Vst3PresetRead& read) {
         device.vst3ClassId = vst3::classIdFromPreset(read.preset);
 
     device.vst3Preset = juce::Base64::toBase64(read.preset.getData(), read.preset.getSize());
+}
+
+/// Where @p info's stored value sits in [0, 1], as the plugin takes it.
+///
+/// A hosted plugin's parameters are stored in the plugin's own normalised
+/// domain whatever display range has been detected over them, and
+/// teMinValue/teMaxValue is the record of that domain. Deliberately not
+/// ParameterUtils::modelToNormalizedValue(), which tells a display-mapped
+/// internal device from a hosted plugin by whether a display-text provider is
+/// attached: no project serialises that provider, so the same device would
+/// answer one way on the load that opens a project and another once the host
+/// had attached one (#2601).
+float normalisedFrom(const ParameterInfo& info) {
+    const auto teSpan = info.teMaxValue - info.teMinValue;
+    const auto normalised =
+        teSpan > 0.0f ? (info.currentValue - info.teMinValue) / teSpan : info.currentValue;
+
+    return std::clamp(normalised, 0.0f, 1.0f);
+}
+
+/// The inverse: what the model stores for a plugin reporting @p normalised.
+float modelValueFrom(float normalised, const ParameterInfo& info) {
+    const auto teSpan = info.teMaxValue - info.teMinValue;
+    return teSpan > 0.0f ? info.teMinValue + normalised * teSpan : normalised;
 }
 
 /// The length the fork asks a plugin for its parameter names at.
@@ -227,14 +250,7 @@ SavedStateOutcome applySavedPluginState(juce::AudioPluginInstance& instance,
         if (info == nullptr)
             continue;
 
-        // Through the pair that knows which domain the model stores a value in,
-        // rather than assuming the display one. They differ once a plugin has a
-        // configured display range: the model keeps the plugin's own normalised
-        // number and the range is what the UI draws it against (#2601).
-        parameter->setValue(std::clamp(
-            ParameterUtils::modelToNormalizedValue(ParameterModelValue{info->currentValue}, *info)
-                .value,
-            0.0f, 1.0f));
+        parameter->setValue(normalisedFrom(*info));
     }
 
     // The portable preset is asked first, because a project only carries one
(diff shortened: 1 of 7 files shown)


magda/daw/engine/TracktionEngineWrapperPlugins.cpp
```diff
@@ -715,22 +716,27 @@ std::vector<ScannedPluginParameter> TracktionEngineWrapper::scanPluginParameters
...
-        if (info.name.isEmpty())
-            continue;
+        info.name = model.name;
         info.defaultValue = parameter->getDefaultValue();
         const auto rawLabel = parameter->getLabel().trim();
         info.unit = rawLabel.length() <= 6 && !rawLabel.contains("[") && !rawLabel.contains("(")
                         ? (rawLabel.isEmpty() ? juce::String("%") : rawLabel)
                         : juce::String("%");
-        info.scanInput.paramIndex = i;
+        info.scanInput.paramIndex = model.paramIndex;
```
lucaromagnoli: `model.paramIndex` is the host/plan slot and starts at 2 because it includes the dry/wet wrapper pair. `ParameterConfigDialog::applyDetectionResults()` consumes `DetectedParameterInfo::paramIndex` directly as an index into its wrapper-free `parameters_` vector. This shifts every external detection result by two rows and drops the final two as out of range. Please use the zero-based position in `described.parameters` for `scanInput.paramIndex` (while continuing to use `model.paramIndex` only to retrieve the live host parameter).

magda/daw/audio/plugin_manager/ExternalPluginState.cpp
```diff
@@ -227,8 +227,14 @@ SavedStateOutcome applySavedPluginState(juce::AudioPluginInstance& instance,
...
             continue;
 
-        parameter->setValue(
-            std::clamp(ParameterUtils::realToNormalized(info->currentValue, *info), 0.0f, 1.0f));
+        // Through the pair that knows which domain the model stores a value in,
+        // rather than assuming the display one. They differ once a plugin has a
+        // configured display range: the model keeps the plugin's own normalised
+        // number and the range is what the UI draws it against (#2601).
+        parameter->setValue(std::clamp(
+            ParameterUtils::modelToNormalizedValue(ParameterModelValue{info->currentValue}, *info)
```
lucaromagnoli: This still mis-restores a configured parameter from a freshly deserialized project. Serialization preserves `minValue`/`maxValue`, `teMinValue`/`teMaxValue`, and the normalized `currentValue`, but deliberately drops `displayText`. With a detected 20–20000 Hz range and saved value 0.7, `modelToNormalizedValue()` therefore sees `displayText == nullptr`, classifies the parameter as display-mapped internal, and converts 0.7 as though it were Hz (approximately zero). The new test masks this load case by manually attaching a display provider before adaptation. Please make the restore distinguish hosted parameters without relying on the non-serialized provider, or attach the provider before applying the saved baseline.

magda/daw/engine/TracktionEngineWrapperPlugins.cpp
lucaromagnoli: Confirmed and fixed. `applyDetectionResults()` indexes `parameters_` with `r.paramIndex`, so the host slot shifted every result by two and dropped the last two.

`scanInput.paramIndex` is now the position in the list being built (`result.size()`), and `model.paramIndex` is used only to fetch the live parameter from `hostParameterOrder`.

The internal branch had the same defect against its own loop counter — it `continue`s on a parameter with no name, after which `i` is no longer the position — so that one is the list position now too.

e643a95f

magda/daw/audio/plugin_manager/ExternalPluginState.cpp
lucaromagnoli: Right on both counts, including that the test masked it — it attached a provider before adaptation, so it exercised the one case that already worked.

The root of it is that `modelToNormalizedValue()` distinguishes a display-mapped internal device from a hosted plugin by a field no project carries. Rather than make the restore reconstruct that answer, `ExternalPluginState` now states the rule it owns: a hosted plugin's value is stored in the plugin's own normalised domain whatever range has been detected over it, and `teMinValue`/`teMaxValue` is the record of that domain. `normalisedFrom()` / `modelValueFrom()` convert through it, so the answer is the same on the load that opens a project and an hour later.

That keeps the predicate untouched for the internal devices that depend on it — which is also why I did not take my own earlier suggestion on #2601 to drop its `displayText` gate.

The test now runs both shapes, with and without a provider, and the no-provider one fails against the version you reviewed: it hands the plugin 0.0 and writes 20 back.

e643a95f

PR #2607, opened 2026-09-11
Address a plugin's parameter config by the parameter's own id

lucaromagnoli:
Item 3 of #2601, the second part. `Refs`, not `Closes` — the remaining items are
listed at the bottom.

A `PluginParameterConfigEntry` was stored by its position in the device's
parameter list. A plugin that gains a parameter in an update renumbers
everything after it, and every later override then describes the control that
used to be at that position — a detected `Hz` range landing on a resonance, a
visible-parameter selection pointing at something else. Silently.

## What changed

Entries carry the parameter's own id: `ParameterInfo::stableId`, which for a
hosted plugin is the `paramID` it declares and which #2599 already records.
`applyToDevice` matches on that, and pushes the rebuilt visible / mini-mixer /
AI selections at the position the match found rather than at the one the file
recorded.

**A miss drops the entry** rather than falling back to the position. That was
the first thing my own test caught: a parameter the plugin no longer has would
otherwise hand its override to whatever moved into its slot, which is exactly
the failure this is meant to remove.

**Position is still the answer when nothing can be matched** — a file written
before ids were stored, or a plugin whose parameters declare none — so every
config on disk keeps working unchanged and gains an id on its next save. The
`id` attribute is only written when non-empty, so the format stays readable by
older builds.

The dialog writes ids as well: `ScannedPluginParameter` reports each parameter's
id alongside its name, taken from `describeHostParameters()` on the external
branch and from the parameter's `paramID` on the internal one, and carried
through `MockParameterInfo` into the file the dialog saves.

## Tests

Four cases in `tests/devices/test_plugin_parameter_config_store.cpp`, two of
which fail with position-only matching restored:

- A plugin that inserts a parameter at the front: the override follows `Mode` to
  its new position, and the parameter that took the old one is untouched.
- A plugin that drops a parameter: nothing inherits its override, and the
  parameter that moved into its position keeps its own.
- A config written before ids still applies by position.
- A plugin whose parameters declare no ids still applies by position.

`make test` (3958 passed, 13 skipped) and `make test-juce` green. `make
lint-changed` reports nothing on the new code.

## Left on #2601

- The fork still loses the stored config on rebuild; only the native path
  re-applies it.
- `MiniChainRow` and `DeviceSlotParameterPaging` still apply it at draw time —
  redundant on the native path, but removing them needs a hands-on check.
- `scanPluginParameters` opens its own instance rather than asking whichever
  engine holds the plugin.

### DIFF
--- a/magda/daw/core/PluginParameterConfigStore.cpp
+++ b/magda/daw/core/PluginParameterConfigStore.cpp
@@ -79,6 +79,31 @@ ParameterScale scaleFromString(const juce::String& name) {
     return ParameterScale::Linear;
 }
 
+std::vector<int> entryPositions(const PluginParameterConfig& config,
+                                const std::vector<juce::String>& currentIds) {
+    std::unordered_map<std::string, int> byId;
+    for (int at = 0; at < static_cast<int>(currentIds.size()); ++at)
+        if (const auto& id = currentIds[static_cast<size_t>(at)]; id.isNotEmpty())
+            byId.emplace(id.toStdString(), at);
+
+    const auto matchable = !byId.empty();
+
+    std::vector<int> positions;
+    positions.reserve(config.entries.size());
+
+    for (const auto& entry : config.entries) {
+        if (entry.id.isEmpty() || !matchable) {
+            positions.push_back(entry.index);
+            continue;
+        }
+
+        const auto found = byId.find(entry.id.toStdString());
+        positions.push_back(found != byId.end() ? found->second : -1);
+    }
+
+    return positions;
+}
+
 juce::File configFileFor(const juce::String& uniqueId) {
     return paths::pluginConfigsDir().getChildFile(uniqueId.replaceCharacters(":/\\,; ", "______") +
                                                   ".xml");
@@ -247,31 +272,16 @@ bool applyToDevice(const juce::String& uniqueId, DeviceInfo& device) {
     // to be re-saved.)
     const auto count = static_cast<int>(device.parameters.size());
 
-    std::unordered_map<std::string, int> byId;
-    for (int at = 0; at < count; ++at)
-        if (const auto& id = device.parameters[static_cast<size_t>(at)].stableId; id.isNotEmpty())
-            byId.emplace(id.toStdString(), at);
+    std::vector<juce::String> currentIds;
+    currentIds.reserve(device.parameters.size());
+    for (const auto& parameter : device.parameters)
+        currentIds.push_back(parameter.stableId);
 
-    // The id where there is one to match on, so a plugin that gained or lost a
-    // parameter since the config was written still finds the ones it kept. A
-    // miss is then a parameter the plugin no longer has, and is dropped rather
-    // than falling back: the position it used to hold belongs to something else
-    // now, and that something else has an entry of its own.
-    //
-    // The position is the answer only when nothing can be matched -- a file
-    // written before ids were stored, or a plugin whose parameters declare
-    // none.
-    const auto matchable = !byId.empty();
-    const auto positionOf = [&byId, matchable](const PluginParameterConfigEntry& entry) {
-        if (entry.id.isEmpty() || !matchable)
-            return entry.index;
+    const auto positions = entryPositions(*config, currentIds);
 
-        const auto found = byId.find(entry.id.toStdString());
-        return found != byId.end() ? found->second : -1;
-    };
-
-    for (const auto& entry : config->entries) {
-        const auto index = positionOf(entry);
+    for (size_t at = 0; at < config->entries.size(); ++at) {
+        const auto& entry = config->entries[at];
+        const auto index = positions[at];
         if (index < 0 || index >= count)
             continue;
         if (entry.visible)
--- a/magda/daw/core/PluginParameterConfigStore.cpp
+++ b/magda/daw/core/PluginParameterConfigStore.cpp
@@ -1,6 +1,7 @@
 #include "PluginParameterConfigStore.hpp"
 
 #include <algorithm>
+#include <unordered_map>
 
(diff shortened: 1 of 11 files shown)


magda/daw/core/PluginParameterConfigStore.cpp
```diff
@@ -241,17 +246,42 @@ bool applyToDevice(const juce::String& uniqueId, DeviceInfo& device) {
     // indices 0/1 were dry/wet; those resolve to the wrong slots once and need
     // to be re-saved.)
     const auto count = static_cast<int>(device.parameters.size());
+
+    std::unordered_map<std::string, int> byId;
+    for (int at = 0; at < count; ++at)
+        if (const auto& id = device.parameters[static_cast<size_t>(at)].stableId; id.isNotEmpty())
```
lucaromagnoli: [P1] This map is keyed by `ParameterInfo::stableId`, but hosted VST3/AU instances do not currently populate that with the format parameter ID. `hostParameterId()` in `ExternalPluginState.cpp` only casts to `AudioProcessorParameterWithID`; JUCE `VST3Parameter` and `AUInstanceParameter` instead derive from `HostedAudioProcessorParameter` and expose the stable ID through `getParameterID()`. The cast therefore falls back to `getParameterIndex()`, so after an insertion the stored `"2"` still matches whichever parameter is now at index 2—the exact corruption this change is intended to prevent. Please source the ID from `HostedAudioProcessorParameter::getParameterID()` and cover a hosted-parameter implementation in the test.

magda/daw/ui/dialogs/ParameterConfigDialog.cpp
```diff
@@ -1252,6 +1254,7 @@ void ParameterConfigDialog::saveParameterConfiguration() {
         const auto& p = parameters_[i];
         magda::PluginParameterConfigEntry entry;
         entry.index = static_cast<int>(i);
+        entry.id = p.stableId;
```
lucaromagnoli, line 1257: [P1] `loadParameterConfiguration()` still overlays saved data using `entry.index` (`parameters_[entry.index]`). After a plugin inserts or removes a parameter, the old customization is displayed on the wrong row; this save then writes the scanned `stableId` from that wrong row, permanently rebinding the old configuration to another control. Please make dialog loading resolve `entry.id` against the scanned row IDs, with the same legacy positional fallback as `applyToDevice()`.

magda/daw/core/PluginParameterConfigStore.cpp
```diff
@@ -210,6 +214,7 @@ PluginParameterConfig fromDevice(const DeviceInfo& device) {
         const auto& info = device.parameters[i];
         PluginParameterConfigEntry entry;
         entry.index = static_cast<int>(i);
+        entry.id = info.stableId;
```
lucaromagnoli, line 242: [P2] `DeviceApiLive::setDeviceParameterConfig()` starts from this ID-bearing config but then replaces each current entry with the saved entry at `entry.index`. For a legacy file, that replacement erases the ID set here, so an API save never migrates the file. After a parameter reorder, it also copies IDs into the wrong positions, drops the newly inserted parameter, and can duplicate another even for a prompt-only update. Please merge saved fields by identity while retaining the current entry index and ID.

magda/daw/core/PluginParameterConfigStore.cpp
lucaromagnoli: Correct, and it made the whole change a no-op on real plugins. `VST3Parameter` in `juce_VST3PluginFormatImpl.h` overrides `getParameterID()` returning the Steinberg `ParamID`, and neither it nor the AU equivalent derives from `AudioProcessorParameterWithID` — that is the class a JUCE *plugin* uses for its own parameters. So every hosted parameter fell through to `String(getParameterIndex())`, which looks like an identity and matches whatever moved into the slot: exactly the corruption the change exists to prevent.

Taken from `HostedAudioProcessorParameter::getParameterID()` now, which `AudioProcessorParameterWithID` also implements, so one cast covers both. A parameter that declares nothing gets an empty id rather than its index, so `entryPositions()` can tell "this is its id" from "it has none" instead of being handed a number that looks like one.

The test stub was an `AudioProcessorParameterWithID`, so it could not have caught this. There is a `HostedStubParameter` beside it now, shaped the way the formats' own parameters are, and the new case fails against the old cast.

This does diverge from the fork, which casts the same narrow way in `ExternalPlugin::buildParameterList` — noted at `hostParameterId()`. Nothing crosses between them: `stableId` is not serialised into projects, and TE's `parameterID` is its own automation id.

58561c6c

magda/daw/ui/dialogs/ParameterConfigDialog.cpp
lucaromagnoli: Agreed, and the cascade you describe is the bad part: displaying against the wrong row is recoverable, writing that row's id over the entry is not.

The root of it was three readers of one file each matching entries their own way. There is one rule now — `PluginParameterConfigStore::entryPositions()`, id first with the legacy positional fallback and a miss dropped — and `applyToDevice()`, this dialog's load and the API's merge all ask it.

58561c6c

magda/daw/core/PluginParameterConfigStore.cpp
lucaromagnoli: Right on all three counts. The wholesale `config.entries[entry.index] = entry` put back the position the file was written at and dropped the id the fresh `fromDevice` entry carried, so an API save could never migrate a legacy file — and after a reorder it copied into the wrong positions besides.

It merges field by field onto the entry the parameter has now, at the position `entryPositions()` resolves, so the current index and id survive and a legacy file gains its ids on the first save through the API.

58561c6c

PR #2616, opened 2026-09-12
Revert the parameter-config rebuild handling (#2608, #2611)

lucaromagnoli:
Reverts #2608 and #2611. A device slot comes up with no parameter controls.

#2608 applied a plugin's saved parameter config wherever the parameter array is rebuilt. `applyToDevice` clears `visibleParameters` and rebuilds it from the entries ticked visible — and a config with none ticked therefore wipes the list. The only config file on the machine I tested against has 774 parameters and zero ticked, so every rebuild wiped that device's saved selection.

The UI path #2611 then removed had guarded against exactly that:

```cpp
if (!tempDevice.visibleParameters.empty()) { ...write back... }
```

I argued in #2608 that the guard was unnecessary because `setDeviceVisibleParameters` has a single caller fed from the config store. That was wrong: `TrackSerializer` deserialises `visibleParameters` straight from the project file, so a project's own saved selection was being overwritten.

#2611 is reverted with it — it only removed the redundancy #2608 introduced, and removing it also took the slot rebuild out of `DeviceSlotComponent::setNodePath`, where the call was wrapped around `updateParameterPagination()` / `updateParameterSlots()`.

What the pair was worth on real data: of the 774 entries in that config, one carries a unit. That is what surviving a rebuild was buying.

To re-land: keep the selection when the config has nothing to say about it, and cover the device slot, which no test drives.

`make test` (4007 passed, 13 skipped) green on the reverts.

### DIFF
--- a/magda/daw/audio/processors/base/DeviceProcessor.cpp
+++ b/magda/daw/audio/processors/base/DeviceProcessor.cpp
@@ -2,8 +2,6 @@
 
 #include <utility>
 
-#include "core/PluginParameterConfigStore.hpp"
-
 namespace magda {
 
 DeviceProcessor::DeviceProcessor(DeviceId deviceId, te::Plugin::Ptr plugin)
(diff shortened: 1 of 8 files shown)


lucaromagnoli:
Closing unmerged. The missing controls are a native-engine notification gap that predates both PRs: EngineHost sends devicePropertyChanged on load completion and neither DeviceSlotComponent nor TrackChainContent handles it, so the slot keeps the empty parameter copy it was created with. The fork sends trackDevicesChanged, which does refresh. Reverting #2608 and #2611 does not touch that and did not help.

Issue #2612, opened 2026-09-12
A monitor switch is still a topology change for everything but a MIDI device route


lucaromagnoli:
Switching a track's input monitor on or off recompiles the plan, because the compiler decides
whether the input exists at all. For a MIDI device route that no longer matters: the routing
snapshot decides what the op reads, and the op is emitted either way. For everything else the
monitor switch is still a topology change.

## What still compiles the gate

`activeAudioInputRoute` / `activeMidiInputRoute` (`PlanCompiler.cpp:381-391`) answer `None` for a
track that is not monitoring, and four things follow from that answer:

- **An external audio input.** `RouteKind::External` emits the `AudioInput` op and the
  `LiveInputMeter` beside it (`PlanCompiler.cpp:1325-1340`). No monitor, no ops.
- **A `track:` audio route.** The source track's routed output is pushed into the destination's
  input mix, and the ordering edge that makes the source compile first
  (`PlanCompiler.cpp:474`, `1308-1318`).
- **A `track:` MIDI route.** The same edge on the MIDI side, plus `midiSourceTracks_`
  (`PlanCompiler.cpp:1605`): the switch decides whether the *source* track compiles MIDI ops at
  all, so one track's monitor button changes another track's op set.
- **A named MIDI device on a track whose chain consumes no MIDI.** `readsMidi` is false, so the
  live input op exists only while `route.kind == External` (`PlanCompiler.cpp:1383`).

`testShapeChangesAreAPlanChange` asserts two of these today: "So is taking MIDI from another track"
and "So is a monitored audio input".

## Why it is worth closing

A recompile is a prepare and a swap for what is a button on a track header, and the escalation in
`publishValues()` exists to notice it: the host compiles the whole model and compares fingerprints
on every property edit because a values publish cannot say what it changed. Every gate moved out
of the plan is one less thing that comparison has to catch.

It is also the half of #2592 that the routing snapshot did not reach. Audibility is now answered
once for a device-routed input and still twice for these.

## What to build

**The ops exist unconditionally, and a value says whether they contribute.** `OpValue::silent`
already means exactly this -- "the op contributes nothing this block", set today by a rack chain
muted or soloed out -- and it is published with values, resolved off the same reading of the model
the routing snapshot is. A monitor flip then publishes values, not a plan.

- The external audio input op and its meter are emitted for every clip-carrying track that names a
  device, and `silent` decides whether the track hears it. The meter stays audible-independent, so
[...]

Issue #2553, opened 2026-09-09
Cutover slice 3: recording, monitoring and input under the native engine


lucaromagnoli:
Slice 3 of #1897. Depends on slice 1, and is what #1895 has been waiting for.

Recording's engine half is finished across #1895's seven slices — live input, the capture queue and writer thread, arrangement audio and MIDI takes, the pass reported without a poll, slot recording and the arrangement capture, and a take that survives a plan swap. Roughly sixty native cases assert it. None of it has ever been driven by an audio device.

## Scope

- Arming, and the three monitor modes reaching the plan's input ops.
- The device callback's input reaching `LiveInputBlock`, with the host resolving which hardware channels and which MIDI source a track names.
- Takes becoming clips: `TakeRecorder` and `MidiTakeRecorder` through `RecordingFeed`, materialised into `ClipManager` when a pass closes.
- The recording overlay off `RecordTap` (#2463), and `getRecordingPreviews` answered from it.
- Slot recording and the session-to-arrangement capture (#2464), which slice 2 deliberately left here.
- Count-in, punch, and loop-record passes.
- Input latency: what the driver reports reaching the take's alignment, with the sign and application point the native tests already encode.

## Verification

`docs/smoke-tests/recording.md` is the checklist, and it was written for this. It is the hands-on pass #1895 budgets, and its conventions table is what a human confirms against a loopback cable once — the offline half is already asserted natively and does not want re-asserting through a second engine (#2466).

The two questions only this slice can answer: whether the input latency figure is right, and whether the monitor round trip is playable.

lucaromagnoli:
PR #2720 is merged into dev/0.20.0. Arrangement MIDI recording and live previews are connected and were confirmed in the application; the PR records callback-level regression coverage and Debug/Release builds.

Keeping this issue open for its remaining scope:

- Arrangement audio input delivery, recording, and previews.
- Session slot recording and Session-to-Arrangement capture.
- Count-in and hardware latency calibration/verification.

The next implementation slice is Session MIDI slot recording: connect the existing app-facing slot-recording methods to native quantized capture and atomically publish the finished MIDI into the selected slot. Plan: `docs/issues/native-session-midi-recording.md` on `feat/2553-session-midi-recording`. Audio recording and hardware-I/O ownership (#2588) remain separate work.

PR #2622, opened 2026-09-12
Make a hardware input's monitor switch a value, not a plan change

lucaromagnoli:
Part of #2612 — the hardware **audio** input half. #2612 stays open for the rest.

## What it does

A track that names an audio input device compiles its `AudioInput` op and the
`LiveInputMeter` beside it whether or not the track is monitoring. The input
reaches the chain through a gate (`OpKind::Gain`, `OpRole::LiveInputGate`) whose
`OpValue::silent` is resolved from `monitorsInput()`. Flipping the switch changes one
boolean in the values table: the host publishes values, prepares nothing, and every
instance in the chain keeps its state and its tail.

The meter sits **ahead** of the gate, so a track that records without monitoring still
meters what it records.

## What it deliberately leaves alone

- **A route from another track** (`audioInputDevice = "track:1"`, and the MIDI
  equivalent). Those are ordering dependencies as well as connections. Resolving one
  unconditionally makes a dormant route close the same cycles a live one closes: where a
  track sidechains from the track it takes input from, the breaker drops the sidechain.
  That is already what happens the moment the track monitors — measured, both plans
  identical — but the state where it worked would disappear. It needs an explicit
  feedback policy, not a side effect. Still gated in `activeAudioInputRoute`.
- **Live MIDI.** Unchanged, so no new unbound-binding reports and no corpus churn.
- **The executor's unbound-live-input reports.** Reverted from the first pass. With the
  change confined to audio, no fixture names an audio input and nothing new is reported.
  Distinguishing offline from live binding belongs in its own change.

## Cost

Measured in a Debug build, 512 samples at 44.1 kHz (budget 11 610 µs/block):

| project | µs/block |
|---|---|
| bare track, no input named | 2.10 |
| same track naming an unmonitored hardware input | 4.67 |
| **delta per such track** | **2.57 (0.022% of budget)** |

Four extra ops — input, meter, gate, one mix-edge delay. Release is lower, and only
tracks pointed at hardware pay it. Removing it entirely wants a values-time prune of the
ops nothing audible reads; that pays off for mute, solo and bypass too, and is filed
separately.

## Tests

`[2612]`: the plan's fingerprint is identical across Off / Auto / Auto+armed / In for a
track naming a hardware input; the gate's silence follows `monitorsInput()`; the input
meter is compiled in every position; the input reaches the chain through the gate and
nowhere else. `Auto input monitoring only counts while the track is armed` now asserts
the audio half as a value and the MIDI half as a shape. In the host publish suite,
naming an audio input is a shape change and monitoring it is not.

`make test` — 4035 cases, 1,754,797 assertions, 0 failures, 13 skipped.
`make test-juce` — all passed.

## Before merging

The native engine is not the default, so none of this is audible unless the app is
launched with `MAGDA_AUDIO_ENGINE=magda`. Under that, worth checking by hand: monitoring
off — input meter moves, recording captures, nothing heard through the track; monitoring
on — heard, and toggling during playback keeps plugin state and tails; and both armed and
unarmed.

One thing that check will expose, which this PR does not change: `monitorsInput()` is
`recordArmed || inputMonitor == In`, so **arming a track makes its input audible even with
the monitor switch set to Off**. The gate preserves that rule exactly. Whether it is the
intended control is a separate question.

### DIFF
--- a/magda/engine/exec/PlanExecutor.cpp
+++ b/magda/engine/exec/PlanExecutor.cpp
@@ -271,6 +271,7 @@ int PlanExecutor::activeCrossfades() const {
 void PlanExecutor::reset() {
     plan_ = nullptr;
     planFingerprint_ = 0;
+    reportedUnboundInputs_.clear();
     latencySamples_ = 0;
     carriedDelayLines_ = 0;
     carriedCrossfades_ = 0;
@@ -325,6 +326,11 @@ bool hearsLiveInput(const RenderPlan& plan, const PlanValues* values, TrackId tr
             return i < values->ops.size() ? !values->ops[i].silent : true;
     }
     return true;
+}
+
+std::string describeOp(const RenderPlan& plan, std::size_t index) {
+    return "op " + std::to_string(index) + " (" + toString(plan.ops[index].kind) + " " +
+           toString(plan.ops[index].key) + "): ";
 }
 
 }  // namespace
@@ -565,10 +571,7 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
 
     detectMono_.assign(static_cast<std::size_t>(std::max(context.maxBlockSize, 0)), 0.0f);
 
-    const auto describe = [&plan](std::size_t index) {
-        return "op " + std::to_string(index) + " (" + toString(plan.ops[index].kind) + " " +
-               toString(plan.ops[index].key) + "): ";
-    };
+    const auto describe = [&plan](std::size_t index) { return describeOp(plan, index); };
 
     const auto findAudioSource = [](const auto& map, TrackId trackId) -> EngineAudioSource* {
         const auto found = map.find(trackId);
@@ -591,15 +594,13 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
                                        std::to_string(trackId) + ", it renders silence");
                 break;
 
-            // Reported only where the track is asking to hear the input. Its op
-            // is compiled whether or not the track is monitoring, so that the
-            // meter on it keeps reading (#2612); an input nobody is listening
-            // to is not a binding the host failed to make.
+            // Not reported here: an input op is compiled whether or not the
+            // track is monitoring, so what makes an unbound one worth saying is
+            // the values, and those arrive again without a prepare (#2612).
+            // reportUnboundInputs below answers for the values this plan is
+            // published with, and publishValues asks again for every set after.
             case OpKind::AudioInput:
                 audioSourceForOp_[i] = findAudioSource(bindings.audioInputs, trackId);
-                if (audioSourceForOp_[i] == nullptr && hearsLiveInput(plan, values, trackId))
-                    messages.push_back(describe(i) + "no live audio input bound for track " +
-                                       std::to_string(trackId) + ", it renders silence");
                 break;
 
             // Reported only by a host that has a launcher at all, which is what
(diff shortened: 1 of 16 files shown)


magda/engine/plan/PlanCompiler.cpp
```diff
@@ -1328,17 +1353,17 @@ void Compiler::emitTrack(const TrackInfo& track) {
             const auto op = addOp(OpKind::AudioInput, key, {}, {SignalKind::Audio});
```
lucaromagnoli, line 1353: [P2] This unconditional op introduces a new unbound-input diagnostic for every configured but unmonitored track. The app's `EngineRuntimeFactory` deliberately does not override `createAudioInput()` yet (`EngineRuntimeFactory.hpp:127`), and `PlanExecutor::prepare()` reports every missing `AudioInput` binding as `no live audio input bound ...`. Before this change, Off/Auto-unarmed tracks emitted no op and therefore no warning; now opening or structurally republishing a project logs one even though the silent gate means the user is not asking to hear that input. The same happens to any factory/offline path that intentionally provides no live inputs. The PR body says this slice avoids new unbound-binding reports, but the tests only use a custom factory or inspect fingerprints, so they miss the real host path. Please either make an unavailable input behind a silent gate non-reporting, or land the host binding/availability distinction needed by the always-present op.

lucaromagnoli: Confirmed and fixed in 5366a19f — you were right that the tests missed it: both the custom factory and the fingerprint checks step around `prepare`, so nothing exercised a host that binds nothing.

Took your first option. `PlanExecutor::prepare` now takes the `PlanValues` the plan will render with instead of just the `ParamTable` inside them (the same argument, one level up — every caller already had the values to hand), and reports an unbound `AudioInput` only where the gate over it is audible:

```cpp
case OpKind::AudioInput:
    audioSourceForOp_[i] = findAudioSource(bindings.audioInputs, trackId);
    if (audioSourceForOp_[i] == nullptr && hearsLiveInput(plan, values, trackId))
        messages.push_back(...);
```

An input nobody is listening to is not a binding the host failed to make. A monitoring track with nothing bound still reports, which is what the report is for — and that case already reported before this PR, so the app's existing to-do notice for #2553 stays visible rather than being swallowed.

The new test publishes through `EngineSession` with a factory that overrides nothing, which is what `EngineRuntimeFactory` is for live audio today: no report for Off or Auto-unarmed, still reported for In. Verified negatively — dropping the `hearsLiveInput` guard fails it.

I left the `MidiInput` report alone. It is unreachable from this PR (live MIDI is still monitor-gated here), and the offline/live distinction it needs is #2628.

magda/engine/exec/PlanExecutor.cpp
```diff
@@ -569,9 +591,13 @@ std::vector<std::string> PlanExecutor::prepare(const RenderPlan& plan, const Pla
                                        std::to_string(trackId) + ", it renders silence");
                 break;
 
+            // Reported only where the track is asking to hear the input. Its op
+            // is compiled whether or not the track is monitoring, so that the
+            // meter on it keeps reading (#2612); an input nobody is listening
+            // to is not a binding the host failed to make.
             case OpKind::AudioInput:
```
lucaromagnoli, line 602: [P2] This loses the unbound-input report when monitoring is enabled after an idle publish. A track opened with monitor Off has a silent gate, so this `prepare()` emits no message. Turning the switch to In now keeps the same plan and goes through `EngineHost::publishValues()` → `EngineSession::publishValues()`, which only swaps the values table and never calls `prepare()` or checks bindings. The gate becomes audible, the app's still-unbound `AudioInput` renders silence, and no diagnostic is emitted. Before this PR the switch changed plan shape and the new plan's prepare reported the missing input. Please cover the Off → In values-only transition in the unbound-factory test and arrange for that transition to report an unbound audible input (or retain a persistent diagnostic readable by the host).

lucaromagnoli: Right — my fix moved the hole rather than closing it. Fixed in 443fcd46.

The report was tied to a moment (prepare) while the condition it reports on — audible, and nothing bound — can change afterwards, which is precisely what this PR makes possible. So the question moved out of the binding walk into `PlanExecutor::reportUnboundInputs(const PlanValues*)`, which answers for a given set of values. `prepare` asks it for the values the plan is published with; `EngineSession::publishValues` asks it for every set after, on the fast path, before the swap.

Each op is reported once per prepared plan: a switch is flipped once, while values arrive for every fader move. The set lives on the executor beside the bindings it is about, publishing-thread only — the audio thread reads none of it — and `reset()` clears it, so a new plan reports again.

`ParallelPlanExecutor` forwards it, so both executors answer the same.

The test now covers the transition you named: publish with monitor Off through the unbound factory (no report), then `publishValues` resolved for In on the same plan — reported. Plus a second identical values publish, which must not repeat it. Verified negatively both ways: dropping the `publishValues` call fails the transition assertion, and dropping the guard in `prepare` fails the Off/Auto assertions.

`make test` 4036 cases green, `make test-juce` green.

Issue #2638, opened 2026-09-12
Key a device's parameter selections on slots, not array positions


lucaromagnoli:
Part of #2629, and a prerequisite for #2635.

`DeviceInfo::visibleParameters`, `miniMixerParameters` and `aiSoundDesignerParameters` are read two different ways today.

- As positions in `DeviceInfo::parameters`: `PluginParameterConfigStore::applyToDevice` pushes indices bounded by `device.parameters.size()` and indexes that array with them, and `StandardDeviceLayout`, `DeviceSlotAutomationControls`, `MiniChainRow` and the remote DTO projection all read them that way.
- As parameter slots: `DeviceParamMigrations::migrateIndices` maps them through `migratedParamIndex`, the same function that remaps `ParameterInfo::paramIndex`.

The two agree for an internal device, whose array is dense from zero, and differ for a hosted plugin, whose array starts at slot 2 past the wrapper pair. That is the shape #2601 already ran into.

## The change

The three lists hold slots. The config store resolves its entries to slots, which it can do without a format change: an entry is matched by stable id against the current list and the slot read off the parameter it lands on. Readers translate slot to array position where they need one.

`AddressedParameters` then reads the lists directly rather than translating through `parameters[position].paramIndex`.

## Why it blocks the cut

#2635 leaves only the addressed slots in `DeviceInfo::parameters`. A position into that array no longer names a plugin parameter, and the translation the addressed-set query does today becomes circular: the addressed set is derived from the lists, and the lists would need the array the addressed set determines.

## Project format

A project writes the lists under new keys holding slots. An older project's keys are read as positions and translated through its saved parameter array, which still carries every slot at load time.

Issue #2635, opened 2026-09-12
Mirror only the addressed parameters of a hosted plugin


lucaromagnoli:
Part of #2629. The cut itself, once the addressed-set query, the sparse delivery, the filtered listener and the instance-backed UI are in place.

- `EngineDeviceFactory` reads an addressed slot's value off the instance at bind and leaves the rest alone (`magda/daw/audio/plugins/engine/EngineDeviceFactory.cpp:128`).
- The table carries only addressed slots for a hosted device.
- An unaddressed slot is never seeded, never written and never reported. The chunk carries it.
- Addressing a slot recompiles, which drawing a lane or linking a macro already does.

A parameter the plugin moves in its own editor that nothing addresses does not reach the model. It is in the chunk on save, and nothing else reads it.

Acceptance is a measurement rather than green tests alone: table entries and per-block parameter writes for a project hosting Serum 2, before and after.

Issue #2634, opened 2026-09-12
Read a hosted plugin's parameter metadata off the instance


lucaromagnoli:
Part of #2629.

Everything that shows a hosted plugin's parameters reads `DeviceInfo::parameters`, so the array cannot shrink until those readers take the instance's described list instead.

- The device slot, `StandardDeviceLayout` (`magda/daw/ui/components/chain/layout/StandardDeviceLayout.cpp:10`), `DeviceSlotAutomationControls` and the parameter config dialog enumerate the described list: name, stable id, config.
- A value shown for an unaddressed slot is read off the instance on the message thread when it is shown.
- `devices.listParameters` reports the same list. `devices.setParameter` on an unaddressed slot addresses it.
- The alias and AI paths, gain staging, and preset apply read through the same projection.

No parameter disappears from anything a user can see. What changes is where its name and its value come from.

Test: Catch2 over the projection, plus the remote API tests for list and set.

Issue #2629, opened 2026-09-12
Mirror only the parameters something addresses for a hosted plugin


lucaromagnoli:
Release precondition for the native engine: it must be at least on par with the incumbent for hosted plugins, and the current parameter model is where it is not.

## Today

A hosted plugin's `DeviceInfo::parameters` mirrors every automatable parameter the instance reports (541 for Serum 2, 2623 in the instance). Every one of them is seeded on load, written into the table, delivered to the device every block, re-reported into the model when the plugin moves it, and saved to the project beside the chunk that already carries all of it. The incumbent does the same through `te::AutomatableParameter`, one object and listener per parameter, and is heavy for the same reason.

That mirror is where the two engines disagreed about a value's meaning (#2623), where a configured range poisoned the audio path (#2624), where the device kept its own copy (#2625), and where an edit made in the plugin's editor was reverted (fixed on `fix/2613-model-owns-device-parameters`). Each fix narrows the mirror's reach; this removes the mirror.

## Proposal

The chunk is the plugin's state. The model mirrors a value only for a slot something addresses.

**Addressed** means a slot that has any of: an automation lane, a macro link, a modifier link, a MIDI learn or controller binding, a place in the device UI (`visibleParameters`, mini mixer, AI list). Everything else exists as metadata only (name, stable id, config), read off the instance on demand.

- **Table.** A hosted device's window in the param table is sparse: only addressed slots get keys. `DeviceParams` delivers `(slot, value)` pairs; `EngineExternalDevice` writes a slot when its value moved (the rule already in place). Addressing a slot (drawing a lane, linking a macro) recompiles, which those edits already do.
- **Seeding.** An addressed slot's value is read off the instance at bind and kept current through the plugin-edit path. Unaddressed slots are never seeded and never reported; the chunk carries them on save.
- **Driven flag.** Each delivered pair says whether the host is driving the slot this block (lane, macro, modifier). The device does not report a driven slot's plugin-side changes, so a plugin's own internal modulation, which VST3 reports as output parameter changes, cannot overwrite the model's base.
- **UI and MCP.** The device slot and `devices_listParameters` enumerate from the instance's described list, not the model's array; values for unaddressed slots are read from the instance on the message thread when shown. `devices_setParameter` on an unaddressed slot addresses it.
- **Project format.** A hosted device saves chunk, config, and the addressed slots' values. An older project's full array loads as addressed slots only; the chunk was already authoritative (#2610), so nothing is lost.
[...]

PR #2639, opened 2026-09-12
Read a hosted plugin's parameter list off its instance, and key its selections on slots

lucaromagnoli:
Two slices of #2629: the projection that can answer from a live plugin (#2634, first half), and the addressing change the cut needs before it (#2638).

## Read a hosted plugin's parameter list off its instance (#2634)

`EngineExternalDevice::describeParameters` reports what the instance has, at the values it holds now, and `AudioEngine::describeDeviceParameters` reaches it on the message thread the way `formatDeviceParameter` already does: a query that suspends nothing.

One projection, `deviceParameterList`, is what readers needing the whole list go through. It applies the saved config and the live text providers to the described list, exactly as the model's own array gets them. `devices.listParameters`, `devices.setParameter`, `devices.setParameterConfig` and the automation menu now ask the instance.

Internal devices, and paths the engine holds no instance for, answer from the model, so the fork's readers are unchanged.

The parameter config dialog needed nothing: it already scans a throwaway instance through `scanPluginParameters`.

The device slot grid and the mini chain row take the list too: the slot adopts it when it learns its path, since the path is what reaches the plugin, and again whenever the device it draws changes. That covers the case a hosted plugin is in before anyone configures it, where the grid shows every parameter the plugin has.

## Key a device's parameter selections on slots (#2638)

`visibleParameters`, `miniMixerParameters` and `aiSoundDesignerParameters` were read two ways: as positions in `DeviceInfo::parameters` by the config store, the grid layout, the mini chain row and the remote DTOs, and as slots by `DeviceParamMigrations::migrateIndices`, which maps them through the same function that remaps `paramIndex`. The two agree only for a device whose array is dense from zero, which a hosted plugin's is not: its own parameters start at slot 2, past the wrapper pair.

They are slots now.

- The config store resolves an entry by stable id and stores the slot it lands on, so saved files need no format change.
- A project writes `visibleParameterSlots`, `miniMixerParameterSlots` and `aiSoundDesignerParameterSlots`. An older project's keys are read as positions and translated through its saved parameter array, which still carries every slot at load.
- The grid layout, the mini chain row, the DTO projection and `AddressedParameters` follow.

This has to land before #2635. Once the array holds only addressed slots, a position into it names nothing, and the addressed-set query would be circular: it reads the lists, and the lists would need the array it determines.

### A bug this fixes

The device slot's automation refresh compared an array position against a lane's slot whenever a device had no visible selection, so a lane on a hosted plugin drove the wrong cell or none.

## Tests

4035 passed, 13 skipped, 0 failed. New cases cover a selection round trip as slots, an older project's positions translating on load, and the addressed set reading slots directly. One expectation changed with the convention: a config following its parameter through a plugin renumber now names the slot, which the renumber does not move, rather than the position it moved to. The app target and the JUCE test target build.

### DIFF
--- a/magda/daw/api/remote_handlers.cpp
+++ b/magda/daw/api/remote_handlers.cpp
@@ -628,8 +628,7 @@ HandlerResult devicesSetParameterConfig(MagdaApi& api, const juce::var& input,
                                    "their parameters already accept agent writes");
 
     // The wire indices are the ones devices.listParameters reports, which are
-    // paramIndex when set; the customization lists key on position. Translate
-    // so the two operations speak the same addresses.
+    // slots; a config entry is addressed by its position in the described list.
     const auto parameters = deviceParameterList(*device, *path);
     std::unordered_map<int, int> positionByWireIndex;
     for (size_t i = 0; i < parameters.size(); ++i) {
--- a/magda/daw/api/remote_handlers.cpp
+++ b/magda/daw/api/remote_handlers.cpp
@@ -7,6 +7,7 @@
 #include <unordered_set>
 #include <utility>
 
+#include "../audio/DeviceParameterList.hpp"
 #include "../core/AutomationCommands.hpp"
 #include "../core/AutomationInfo.hpp"
 #include "../core/AutomationTypes.hpp"
@@ -560,7 +561,7 @@ HandlerResult devicesListParameters(MagdaApi& api, const juce::var& input, const
         return HandlerResult::fail(ErrorCode::NotFound, "no device at devicePath");
 
     std::vector<juce::var> items;
-    for (const auto& parameter : makeDeviceParameterDtos(*device))
+    for (const auto& parameter : makeDeviceParameterDtos(*device, *path))
         items.push_back(toJson(parameter));
     return HandlerResult::ok(toJsonArray(items));
 }
@@ -574,7 +575,7 @@ HandlerResult devicesSetParameter(MagdaApi& api, const juce::var& input, const R
         return HandlerResult::fail(ErrorCode::NotFound, "no device at devicePath");
 
     const auto parameterIndex = readInt(input, "parameterIndex", -1);
-    const auto parameters = makeDeviceParameterDtos(*device);
+    const auto parameters = makeDeviceParameterDtos(*device, *path);
     const auto named = std::ranges::find(parameters, parameterIndex, &DeviceParameterDto::index);
     if (named == parameters.end())
         return notFound("parameter", parameterIndex);
@@ -605,7 +606,7 @@ HandlerResult devicesSetParameter(MagdaApi& api, const juce::var& input, const R
     // the caller sees what the model now holds rather than what was sent.
     const auto* updated = api.devices().getDevice(*path);
     if (updated != nullptr) {
-        for (const auto& parameter : makeDeviceParameterDtos(*updated)) {
+        for (const auto& parameter : makeDeviceParameterDtos(*updated, *path)) {
             if (parameter.index == parameterIndex)
                 return HandlerResult::ok(toJson(parameter));
         }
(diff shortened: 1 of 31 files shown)


magda/daw/core/PluginParameterConfigStore.cpp
```diff
@@ -299,14 +298,15 @@ bool applyToDevice(const juce::String& uniqueId, DeviceInfo& device) {
...
+
         if (entry.visible)
-            device.visibleParameters.push_back(index);
+            device.visibleParameters.push_back(parameter.paramIndex);
         if (entry.miniMixer)
-            device.miniMixerParameters.push_back(index);
+            device.miniMixerParameters.push_back(parameter.paramIndex);
         if (entry.aiAgent)
-            device.aiSoundDesignerParameters.push_back(index);
+            device.aiSoundDesignerParameters.push_back(parameter.paramIndex);
```
lucaromagnoli, line 307: **P1 — Migrate the AI opt-in consumers to parameter slots too**

This now stores parameter slots, but `DeviceApiLive::setDeviceParameter()` still checks `aiSoundDesignerParameters` against the parameter's array position, and `GenericSoundDesignAgent::generatePatch()` filters its parameter list by array position too. For a normal hosted array starting at slot 2, opting into the first parameter produces `{2}`: writes to that selected parameter are rejected (position 0), while the third parameter (slot 4, position 2) can be permitted and exposed to the agent despite not being selected. Please update both consumers to compare against `paramIndex` and add a regression test with a hosted array starting at slot 2.

— Codex

magda/daw/audio/plugins/engine/EngineExternalDevice.cpp
```diff
@@ -700,6 +700,21 @@ magda::SavedStateOutcome EngineExternalDevice::applyState(const magda::DeviceInf
...
 }
 
+magda::HostParameters EngineExternalDevice::describeParameters() const {
+    auto described = magda::describeHostParameters(*instance_, magda::DeviceInfo{});
+
+    // The wrapper pair is the host's own, so its live values are these rather
+    // than anything the instance holds.
+    for (auto& parameter : described.wrapperParameters) {
+        if (parameter.wrapperRole == magda::WrapperRole::DryGain)
+            parameter.currentValue = dryGain_;
```
lucaromagnoli: **P2 — Synchronize the wrapper-value reads across threads**

`describeParameters()` is called on the message thread without suspending processing, but `dryGain_` and `wetGain_` are plain floats written by `writeParameters()` on the audio thread. Reading them here (and the wet value below) concurrently introduces a C++ data race whenever the live parameter list is queried during rendering. Please publish atomic snapshots for these reads, or take the wrapper values from the message-thread-owned model instead.

— Codex

magda/daw/api/remote_handlers.cpp
```diff
@@ -605,7 +606,7 @@ HandlerResult devicesSetParameter(MagdaApi& api, const juce::var& input, const R
     // the caller sees what the model now holds rather than what was sent.
     const auto* updated = api.devices().getDevice(*path);
     if (updated != nullptr) {
-        for (const auto& parameter : makeDeviceParameterDtos(*updated)) {
+        for (const auto& parameter : makeDeviceParameterDtos(*updated, *path)) {
```
lucaromagnoli, line 609: **P2 — Acknowledge the updated model value, not the stale live value**

This projection now reads the hosted instance through `deviceParameterList()`, but `setDeviceParameter()` only updates the model synchronously. On the native engine, `EngineHost::deviceParameterChanged()` calls `wantValues()`, which queues `triggerAsyncUpdate()`; the instance receives the new value later during rendering. This immediate readback therefore returns the old plugin value in a successful `devices.setParameter` response, contradicting the acknowledgement's model-value contract. Please project the updated model value for the response (using the described metadata if needed), or acknowledge after confirmed application. Add a regression test where the model write succeeds while the instance still holds its previous value.

— Codex

Issue #2590, opened 2026-09-11
Live MIDI source ids are never released, and the callback holds 64


lucaromagnoli:
Live MIDI ids come from one monotonically increasing counter (`LiveMidiSources::next_`) and are
never released. `EngineHost`'s callback has room for 64 of them:

```cpp
constexpr int kMaxLiveMidiSources = 64;
```

`collectLiveMidi` drops every event whose source is past that, counting it in `droppedEvents_`,
which is printed only when the MIDI trace is on. From the app it is silence with no reason given.

## What consumes the space

- One id per MIDI input the machine has, taken at every device scan and at every hot-plug.
- One per virtual device: the QWERTY keyboard.
- **One per MIDI-reading track**, not per track that was ever auditioned:
  `EngineRuntimeFactory::createMidiInput` asks `auditionSourceFor(trackId)` for every live input
  op the plan carries, and #2579 gives one to every track that reads MIDI.

So a 60-track instrument project exhausts it on its first publish, and a long session of
hot-plugging or track creation exhausts it in a much smaller one. Deleting the tracks does not
give the ids back.

## Why a bigger constant is not the fix

The counter is unbounded, so any constant is a project size. Releasing an id when its track or
device goes is the real answer, and it needs the reuse to be safe against the queue: an event
already queued under an id must not be delivered to whatever takes that id next. A generation
beside the id, or a release deferred past one drain of `LiveMidiQueue`, is what makes that
sound.

`LiveMidiSources`' own header states the invariant the current design leans on -- "Ids come from
a single counter and are never reused, so two id spaces that must not overlap cannot" -- so
reuse is a change to that contract rather than a patch under it.

## Found

Reviewing #2584 (`EngineHost.cpp:54`). Not introduced there: the cap and the counter both
predate it, and #2579 only widened what consumes the space by giving every MIDI-reading track an
audition source.

lucaromagnoli:
Done in #2640 (merged to `dev/0.20.0` as f9969416): a live MIDI id is the model's for as long as the model names it, reuse waits two drains of `LiveMidiQueue`, device ids are left alone, and the room is 256 rather than 64. Closing by hand — the PR's `Closes` line never fires against a `dev/` base.

Issue #2632, opened 2026-09-12
Drain a plugin's own edits through one host-owned pass


lucaromagnoli:
Part of #2629.

`EngineExternalDevice::PluginListener` stores an edit into a per-slot atomic and then queues a `MessageManager::callAsync` per device (`magda/daw/audio/plugins/engine/EngineExternalDevice.cpp:128`). That is one drain and one allocation per device per burst of edits, from whatever thread the plugin chose.

- The device marks itself dirty and triggers the host's existing `AsyncUpdater`, which is allocation free from any thread.
- The host keeps a map from device key to a weak handle on the device's edit buffer, registered where `ExternalPluginLoader` wires the sink today (`magda/daw/engine/host/ExternalPluginLoader.cpp:177`). A dead handle is skipped.
- The drain walks dirty devices on the message thread, reads each pending slot, and writes it through `TrackManager::setDeviceParameterValueFromPlugin`.
- The loop closes itself. The model write reseeds the table, the device compares against what the table delivered last block and sees no movement, and the host's own writes stay masked by the `writing_` flag.
- The drain is the subscriber seam. The model write is its first subscriber; MIDI learn and a journal attach there, not to the plugin.

Test in magda_juce_tests, since the path needs a message thread: `setValueNotifyingHost` on the stub plugin, then a drain, asserting the model moved and the plugin's value did not bounce. The Catch2 test covering the write rule stays as it is.

PR #2640, opened 2026-09-12
Give a live MIDI source id back when its track goes

lucaromagnoli:
Closes #2590.

Live MIDI ids came from one monotonically increasing counter and were never
released, while the callback has room for a fixed number of them. One is taken for
every track that *reads* MIDI rather than for every track anyone previewed (#2579), so
a session of building and deleting tracks walks the counter past the room. Everything
beyond it is dropped in `collectLiveMidi` and counted where only the MIDI trace prints
it — from the app, silence with no reason given.

## Change

- **An id is the model's for as long as the model names it.** `LiveMidiRouting::resolve`
  already walks the whole track list, so that is where the auditions of tracks that have
  gone are handed back.
- **Reuse waits on the callback.** A released id records what `LiveMidiQueue` had drained
  when it went, and is handed out again only two drains later — an event queued under it,
  or in flight when the queue went empty, is consumed before anything else can be that id.
  The registry watches a counter the callback increments after each drain; a host with no
  callback to watch (a headless tool, most tests) reuses nothing.
- **Device ids are left alone**, and the commit says why: they are keyed by the device's
  identifier, so unplugging one and plugging it back in is the id it had, and retiring the
  id of a device a project still routes to would change it under that route — which the
  routing snapshot reads as a source lost and answers with all-notes-off. They also do not
  grow on a rescan, which is what made them look like a leak.
- **The room is now a project size.** With the live count bounded by the project instead of
  the session, 64 was the wrong number, not the wrong idea: it is 256, and a callback
  clears only the buffers the last one wrote rather than all of them, so a project's worth
  of room costs nothing while it is quiet. The drop counter and its trace line stay.

The header's old invariant — "ids come from a single counter and are never reused" — is
replaced by the one that now holds, rather than patched under.

## Tests

`[2590]`: a track that goes gives its id back, and it is not handed out before the
callback has drained past it; a host with no counter to watch reuses nothing; a track that
stays keeps the id it had across a resolve that retires its neighbours.

Verified negatively: without the `retainAuditions` call, two of the three fail.

`make test` — 4053 cases, 1,917,911 assertions, 0 failures, 13 skipped.
`make test-juce` — all passed.

## Not clashing with #2629

Touches `LiveMidiSources`, `LiveMidiRouting` and the live-MIDI region of `EngineHost.cpp`.
The parameter-mirror work claims `DeviceInfo::parameters`, `ParamTable`/`DeviceParams`,
`EngineExternalDevice`, the device slot UI, the MCP parameter APIs and the hosted-device
project format; the only file in common is `EngineHost.cpp`, and #2632's drain lands in a
different part of it.

### DIFF
--- a/magda/daw/engine/host/EngineHost.cpp
+++ b/magda/daw/engine/host/EngineHost.cpp
@@ -52,16 +52,10 @@ constexpr int kChannels = 2;
 /// 30 fps, which is what the fork's own metering timer runs at.
 constexpr int kMeterIntervalMs = 33;
 
-/// Live MIDI sources the callback has room for: one per enabled input, one per
-/// virtual device, one per track that reads MIDI. A project past this loses the
-/// sources beyond it rather than allocating for them on the audio thread.
-///
-/// A project size rather than a session's history, which is what makes a
-/// constant defensible here: an id goes back when its track does (#2590), so
-/// this bounds how much of one project can be heard at once, not how long the
-/// app has been open. The room costs kMaxMidiBytesPerPort each, so it is a
-/// megabyte of mostly empty buffers.
-constexpr int kMaxLiveMidiSources = 256;
+/// Room for the sources a project can be heard through at once, which the
+/// registry hands out and takes back (LiveMidiSources.hpp). A source with no
+/// slot is dropped here rather than allocated for on the audio thread.
+constexpr int kLiveMidiSlots = LiveMidiSources::kSlots;
 
 /// What one queued event costs a MidiBuffer: a sample position and a length in
 /// front of its three bytes.
@@ -248,9 +242,6 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
                       applyLoadedDevice(key, resolved, restored);
                   }) {
         factory_.loadExternalsWith(loader_);
-
-        // What the registry waits on before handing an id out a second time.
-        sources_.observeDrains(drains_);
 
         // A tap read late loses nothing, since the peak is held until
         // something takes it, so this is how smooth a meter looks.
@@ -736,12 +727,12 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
     /// Room for every source the callback may see. Off the device, from
     /// rebuild(), because it allocates.
     void prepareLiveMidi() {
-        midiBySource_.resize(kMaxLiveMidiSources);
-        for (auto& buffer : midiBySource_)
+        midiBySlot_.resize(kLiveMidiSlots);
+        for (auto& buffer : midiBySlot_)
             buffer.ensureSize(static_cast<std::size_t>(engine::kMaxMidiBytesPerPort));
 
-        streams_.reserve(kMaxLiveMidiSources);
-        written_.reserve(kMaxLiveMidiSources);
+        streams_.reserve(kLiveMidiSlots);
+        written_.reserve(kLiveMidiSlots);
     }
 
     /**
@@ -754,20 +745,24 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
     std::span<const engine::LiveMidiStream> collectLiveMidi() {
         // Only what last block wrote, so the room for a whole project's worth
         // of sources costs a callback nothing while they are quiet.
-        for (const auto index : written_)
-            midiBySource_[index].clear();
+        // Only what last block wrote, so the room for a whole project's worth
+        // of sources costs a callback nothing while they are quiet.
+        for (const auto slot : written_)
+            midiBySlot_[static_cast<std::size_t>(slot)].clear();
 
         written_.clear();
         streams_.clear();
 
         queue_.drain([this](const LiveMidiQueue::Event& event) {
-            const auto index = static_cast<std::size_t>(event.source - 1);
-            if (event.source < 1 || index >= midiBySource_.size()) {
+            // The event names both, and the registry is asked whether they
+            // still agree: a track deleted while this was queued has left its
+            // slot to somebody else, and its last note is not theirs (#2590).
+            if (event.slot < 0 || sources_.ownerOfSlot(event.slot) != event.source) {
                 droppedEvents_.fetch_add(1, std::memory_order_relaxed);
                 return;
             }
 
-            auto& buffer = midiBySource_[index];
+            auto& buffer = midiBySlot_[static_cast<std::size_t>(event.slot)];
             if (static_cast<int>(buffer.data.size()) + kQueuedEventBytes >
                 engine::kMaxMidiBytesPerPort) {
                 droppedEvents_.fetch_add(1, std::memory_order_relaxed);
@@ -775,23 +770,17 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
             }
 
             if (buffer.isEmpty())
-                written_.push_back(index);
+                written_.push_back(event.slot);
             buffer.addEvent(event.bytes, event.size, 0);
         });
 
-        // In source order, which is the order this handed over when it walked
-        // every buffer: arrival order would make one block's streams differ
-        // from another's for the same notes.
+        // In slot order, so one block's streams read like another's for the
+        // same notes rather than in whatever order they were played.
         std::ranges::sort(written_);
 
-        for (const auto index : written_)
-            streams_.push_back(
-                {static_cast<engine::LiveMidiSourceId>(index + 1), &midiBySource_[index]});
-
-        // Counted after the drain, and read by whoever is waiting to reuse an
-        // id: what was queued under it has been consumed by the time this
-        // moves (#2590).
-        drains_.fetch_add(1, std::memory_order_relaxed);
+        for (const auto slot : written_)
+            streams_.push_back({static_cast<engine::LiveMidiSourceId>(sources_.ownerOfSlot(slot)),
+                                &midiBySlot_[static_cast<std::size_t>(slot)]});
 
         return streams_;
     }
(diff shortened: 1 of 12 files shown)


magda/daw/engine/host/LiveMidiSources.cpp
```diff
@@ -22,7 +22,28 @@ int LiveMidiSources::sourceFor(const juce::String& deviceId) {
...
+int LiveMidiSources::takeSource() {
+    // A device is keyed by its identifier, so unplugging one and plugging it
+    // back in is the id it had. What grows without the model growing is the
+    // auditions, and this is where what they give back comes from.
+    if (drains_ != nullptr && !released_.empty()) {
+        // Two callbacks, not one: a push that was in flight when the drain
+        // emptied the queue lands behind it, and a callback is milliseconds.
+        const auto drained = drains_->load(std::memory_order_relaxed);
+        const auto ready = std::ranges::find_if(
+            released_, [drained](const Released& entry) { return entry.drains + 2 <= drained; });
```
lucaromagnoli: **[P1] Account for in-flight producers before reusing an ID**

Two completed drains do not establish that a producer has finished publishing an event. `EngineHost::audition()` obtains the source ID and then calls `queue_.push()` outside the registry lock. A producer can pause between those operations while the publishing thread deletes the track and the audio callback drains twice. If another track takes the released ID before the producer resumes, its stale event is delivered to the new track. Please use generation-tagged events or synchronization that accounts for in-flight producers rather than a fixed number of callbacks.

magda/daw/engine/host/LiveMidiSources.cpp
```diff
@@ -22,7 +22,28 @@ int LiveMidiSources::sourceFor(const juce::String& deviceId) {
...
+
+        if (ready != released_.end()) {
+            const auto source = ready->source;
+            released_.erase(ready);
+            return source;
+        }
+    }
+
+    return next_++;
```
lucaromagnoli: **[P2] Recover out-of-range IDs after callbacks resume**

When audio callbacks are stopped, repeated track creation/deletion never makes a released entry eligible, so this fallback still advances `next_` beyond the callback's 256-source limit even with a small live project. Once audio resumes and the retired slots become reusable, surviving tracks keep their already-assigned out-of-range IDs, so their MIDI remains silently dropped by `collectLiveMidi()`. Please provide a recovery path for these assignments, or decouple logical IDs from the bounded callback slots, so editing while the device is stopped cannot leave tracks permanently silent.

lucaromagnoli (commented):
Two findings from the review. Validation: the existing [live-routing] tests passed (8 cases, 72 assertions); they do not cover the scenarios below.

magda/daw/engine/host/LiveMidiSources.cpp
lucaromagnoli: Right, and the same root cause as your P2: I had made the id *be* the slot. Rewritten in 500fceba.

Ids are back to coming from the counter and never being reused — the invariant the header always stated. What goes back when a track dies is the *buffer* it was bound to, and the two are now separate things. An event carries both, and `collectLiveMidi` asks the registry whether they still agree:

```cpp
if (event.slot < 0 || sources_.ownerOfSlot(event.slot) != event.source) { dropped; return; }
```

That answers the in-flight producer directly, and without a time bound: a thread stopped between `auditionSourceFor` and `push` can resume after any number of drains, and its event is refused because the slot no longer answers to the id it resolved. `ownerOfSlot` is an atomic read per slot — the audio thread takes no lock — written under the registry's lock, disowned before the slot is handed on, so the callback sees either the old id, the sentinel, or the new one and never a slot two ids think they hold.

The drain counter and the two-callback rule are gone; nothing waits on the consumer any more.

magda/daw/engine/host/LiveMidiSources.cpp
lucaromagnoli: Fixed by the same rewrite (500fceba) — this finding is what pushed it: you asked to decouple the logical ids from the bounded callback slots, and that is what it now does.

`next_` still climbs without bound while the device is stopped, and that is now harmless: ids are not slot indices. Slots come from a free pool and go back the moment `retainAuditions` drops a track, with no drain to wait for, so editing while callbacks are stopped recycles them immediately and nobody is left holding an id the callback cannot place.

A project genuinely larger than the room gets an id with no slot (`kNoSlot`). It is still the model's id, so the route resolves and the snapshot is correct; what it pushes is dropped and counted where a source past the end always was. `A project past the room gets ids with nowhere to put them` covers that, and `A slot answers to the id that holds it now` covers the handover.

The room also moved from `EngineHost`'s constant into `LiveMidiSources::kSlots`, so the pool and the buffers cannot disagree about how many there are.

lucaromagnoli (commented):
Follow-up review of 500fceba. Separating IDs from slots fixes the original delayed-producer scenario and stopped-callback churn, but the two cases below remain. The updated [live-routing] tests pass (9 cases, 81 assertions).

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -742,32 +743,44 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
...
-        for (std::size_t i = 0; i < midiBySource_.size(); ++i)
-            if (!midiBySource_[i].isEmpty())
-                streams_.push_back(
-                    {static_cast<engine::LiveMidiSourceId>(i + 1), &midiBySource_[i]});
+        // In slot order, so one block's streams read like another's for the
+        // same notes rather than in whatever order they were played.
+        std::ranges::sort(written_);
+
+        for (const auto slot : written_)
+            streams_.push_back({static_cast<engine::LiveMidiSourceId>(sources_.ownerOfSlot(slot)),
```
lucaromagnoli: **[P1] Preserve the event's source identity through buffer delivery**

The ownership check during `queue_.drain()` and this second ownership read are not one snapshot. The callback can accept and buffer an event from source A, then the publishing thread can delete A and assign its slot to B before this loop runs. This code labels A's buffered bytes with B's source ID, so a subsequently acquired routing snapshot can deliver A's note to B. A slot can also accept events from both owners during a single drain. Please keep the accepted source identity alongside each callback buffer and handle ownership changes without mixing or relabelling the buffered events; alternatively hold slot ownership stable for the entire callback. Checking the live registry again here does not preserve the identity of the bytes already collected.

magda/daw/engine/host/LiveMidiSources.cpp
```diff
@@ -16,13 +16,73 @@ void LiveMidiSources::registerAvailableDevices(juce::Array<juce::MidiDeviceInfo>
...
+}
+
+int LiveMidiSources::take() {
+    const auto source = next_++;
+
+    // A project bigger than the room gets an id with nowhere to put it: what
+    // it pushes is dropped and counted, the way a source past the end always
+    // was. The id is still the model's, so nothing else answers to it.
+    if (freeSlots_.empty())
+        return source;
```
lucaromagnoli, line 46: **[P2] Assign a slot to waiting sources when capacity becomes available**

An ID returned here is cached in `auditions_` (or `devices_`), but neither subsequent lookups nor `release()` ever binds that ID to a slot. For example, allocate 257 audition sources, then retain only the 257th track: there are now 256 free slots, but its `slotFor()` remains `kNoSlot` indefinitely and every note is dropped. This can also happen without an oversized final project because `LiveMidiRouting::resolve()` allocates the new tracks before retiring absent ones: replacing tracks 1–200 with 201–400 leaves 144 current tracks slotless despite 200 slots becoming free at the end of the resolve. Please reclaim absent tracks before allocation and provide recovery for live sources that previously could not get a slot.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Confirmed — two reads of something the publishing thread moves, and the second one named the bytes. Fixed in 0bdc18e3.

There is now one read per slot per block. The first id through a slot owns it for the rest of the callback, that recorded id is what names the stream, and any later event naming the same slot with a different source is refused and counted:

```cpp
const auto owner = first ? sources.ownerOfSlot(event.slot) : ownerOfSlot_[slot];
if (owner != event.source) { drop(); return; }
```

So a slot changing hands mid-callback keeps its first owner's notes and drops the second's — one buffer cannot be two tracks', and the failure to avoid is the first track's notes sounding on the second. The registry is never consulted again after the bytes are accepted.

I moved the collection out of `EngineHost` into `LiveMidiCollector` to do it. That half had no test seam at all — both of your findings were in code nothing could reach — and it now has five: a note reaching the source it was played under, a note from a track that has gone reaching nobody, a slot that changes hands keeping one owner's notes, a source with no slot dropped and counted, and a buffer carrying nothing from the block before.

Worth saying plainly: a single-threaded test cannot stage the interleaving itself. What the tests pin is the rule that replaced the second read; what removes the window is structural.

magda/daw/engine/host/LiveMidiSources.cpp
lucaromagnoli: Both halves confirmed and fixed in 0bdc18e3.

**Recovery.** `bindWaiting()` binds the ids that went without, oldest first, whenever room comes back — called at the end of `retainAuditions`, which is once per publish, so a track deletion is what recovers the tracks that were waiting on it. Your 257-track example is a test: allocate `kSlots + 4` tracks, cut down to the last one, and it goes from `kNoSlot` to a bound slot that answers to it.

**Ordering.** `LiveMidiRouting::resolve` now gives the room back before it asks for any — the live set is computed and `retainAuditions` called at the top, before the walk that allocates. Your 1–200 → 201–400 case is the other test, at three quarters of the room so the doubled demand would not fit: every surviving track has a slot afterwards.

Both are in `[2590]` in test_live_midi_routing.cpp. `make test` 4061 cases green, `make test-juce` green.

lucaromagnoli (commented):
Re-reviewed 0bdc18e3. Both findings from the previous pass are addressed: the collector retains the accepted source identity for each buffer throughout the block, and retired tracks release their slots before new allocations while waiting sources are bound when capacity returns. I found no new actionable correctness issues.

Validation: [live-routing],[live-midi-collector] passed (16 cases, 302 assertions); diff whitespace check passed.

Non-blocking coverage note: the test named "A slot that changes hands mid-callback keeps one owner's notes" changes ownership before collect() starts. It covers queued stale events, but does not exercise ownership changing after the first event has been accepted during the same collection. A deterministic interleaving test would strengthen coverage of the race fixed here.

Issue #2613, opened 2026-09-12
Custom faceplates on MAGDA's own devices do not follow the model


lucaromagnoli:
Reported from the app: the custom faceplates on MAGDA's own devices do not work. The Poly Synth's
ADSR is the case that started it, but it is not confined to that device.

## One defect located

`PolySynthUI::syncGraphFromParam` has exactly one caller: a slider's own `onValueChange`
(`PolySynthUI.cpp:83`). The model-driven refresh, `updateFromParameters`, calls
`syncFilterCurveFromParam` for every parameter and never calls `syncGraphFromParam`. So both ADSR
graphs are drawn from whatever the last drag left there: open a project, open the faceplate, and
the amp and filter envelopes show defaults whatever the patch holds, on either engine. Dragging
one of those four sliders is the only thing that moves them.

This is the same shape as the step sequencer faceplate fixed in #2610 -- the block that feeds a
faceplate living only on the path a parameter edit takes, never the one that builds it.

## Why the native engine makes it general

`DeviceCustomUIManager::getLivePlugin()` returns a `te::Plugin::Ptr`, which is null when the MAGDA
engine renders: there is no fork plugin. Everything reading through it is therefore dead on that
engine, and it is not a short list --

- `PolySynthUI::setLivePlugin` / `StruckInstrumentUI::setLivePlugin`: the 60 Hz timer that pushes
  the live filter curve never starts (`refreshLivePluginBindings`).
- `update()` reads the sampler's file, loop points, root note and waveform off
  `MagdaSamplerPlugin`, and the Drum Grid's pad names and chains off `DrumGridPlugin`.
- `createMidiUtilityUI` and `bindAnalyzerPlugins` read their telemetry the same way (#2585).
- `readAndPushModMatrix` walks the live plugin's automatable parameters.

Where a UI has a model path as well, the model path is what survives, and it is thin: the
`syncGraphFromParam` gap above is what a thin model path looks like when nothing polls behind it.
The fork hid gaps like it by repainting from the running plugin every frame.

## What it needs

- Establish what "does not work" is per device, since read and write are separate paths: a control
  that does not move the sound is a write that never reaches the engine's device; a control that
  does not show the patch is this class of bug.
- Every faceplate is fed from the model, at build time and on every edit, with no live-plugin read
  in the path. `syncGraphFromParam` from `updateFromParameters` is the first of them.
- What genuinely cannot come from the model -- a sampler's waveform, an analyzer's spectrum, a
[...]

Issue #2585, opened 2026-09-10
Analyzer and custom UI telemetry from the rendering engine's device


lucaromagnoli:
The oscilloscope and the spectrum analyzer show nothing under the magda engine, and so does
every custom device UI that binds to a live plugin the same way.

`OscilloscopeUI` and `SpectrumAnalyzerUI` take a telemetry source resolved from the fork's plugin
instance: `DeviceSlotAnalyzerContextActions.cpp:37-45` goes `getAudioBridge()` ->
`getPlugin(path)` -> `deviceFromPlugin<OscilloscopePlugin>`. The same shape reaches the fork in
`DeviceCustomUIManager::getLivePlugin` (`DeviceCustomUIManager.cpp:648`),
`DeviceSlotInlineUiFactory.cpp:26`, `PolyStepSequencerUI.cpp:988`, `FaustUI.cpp:194, 385` and
the compiled panels' `bindPlugin` (`DeviceSlotComponent.cpp:986`).

Before #2579 the fork still held a parallel instance under magda, but nothing rendered through
it, so the scope drew a flat line and the spectrum an empty one. After #2579 there is no instance
at all. Either way the native device is the one with the signal, and nothing hands its telemetry
to the UI.

#1630 closed with the UIs reading through `DeviceTelemetry` rather than concrete plugin pointers,
which is the right seam: what is missing is the other implementation of it.

## What to build

- A way for a custom UI to reach the rendering engine's device telemetry by `ChainNodePath` (or
  `engine::DeviceKey`), answered by the fork's plugin under Tracktion and by the native device
  under magda. The native side already exposes `DeviceTelemetry` on the SDK device; the control
  plane (#2270) is where the lookup belongs, beside capture.
- The analyzer popouts and the slot UIs above ask that rather than `getAudioBridge()`.
- Lifetime: a UI bound to a device that is rebuilt (#2575) or torn down (#2576) is rebound or
  cleared, never left on a freed instance.

## Verification

Under magda: an oscilloscope on a playing track draws the waveform, a spectrum analyzer draws
the spectrum, the step sequencer's live position moves. Under Tracktion: unchanged.

PR #2656, opened 2026-09-14
Pin that a MAGDA device's parameter edit reaches what renders it

lucaromagnoli:
Towards #2613 (does not close it — see the bottom).

#2613 reports that moving a control on one of MAGDA's own devices changes nothing about the sound under the native engine. **On current dev it does** — most likely fixed by #2626 — and nothing covered it: the path from a slider to a voice had no test between `TrackManager` and the audio, so the claim could only be settled by ear.

## Test

`[2613]`: a Poly Synth plays two notes; Osc 1's level — the only oscillator the patch has up — is written through `TrackManager::setDeviceParameterValue` while the first note is sounding, and the values are published with no plan change. The note that was already sounding falls to a tenth of its level, and so does the one that starts afterwards. An assertion about the rendered signal, not about a number copied between two structures.

It is measured **after** the DSP's smoothing ramp rather than across it. The level runs through `smoo`, so the first block either side of an edit is still on its way between the two values; a peak taken over a window that includes it reads as "nothing happened". I walked into exactly that while investigating and briefly concluded the engine was dropping edits on sounding voices — it isn't, and that artefact is the reason this is worth pinning rather than leaving to a listen.

Verified negatively: without the values publish, both assertions fail.

## What is left on #2613

The **read** half: a faceplate showing the patch rather than whatever the last drag left. `DeviceCustomUIManager::getLivePlugin()` returns a `te::Plugin::Ptr`, which is null when the native engine renders, so the parts of a faceplate fed from a live plugin — the sampler's waveform and loop points, the Drum Grid's pad names, the mod matrix, the 60 Hz filter-curve push — are dead there. That is #2585's seam, and the issue stays open for it.

Two smaller things found on the way, neither touched here:
- The Poly Synth declares an `Output` parameter (slot 43, dB) that nothing consumes — a declared control with no DSP behind it.
- `PolySynthUI::syncGraphFromParam` is called from `updateFromParameters` since #2626, so the ADSR-graph gap named in #2613's body is already closed.

`make test` (4077 passed, 13 skipped) and `make test-juce` (all passed) are green locally.


tests/engine/test_engine_host_publish_juce.cpp
```diff
@@ -1042,6 +1043,114 @@ class EngineHostPublishTest final : public juce::UnitTest {
...
+        const auto devicePath = magda::ChainNodePath::topLevelDevice(trackId, 1);
+        constexpr int kOsc1LevelSlot = 1;
+        trackManager.setDeviceParameterValue(devicePath, kOsc1LevelSlot, -60.0f);
+
+        const auto* edited = trackManager.getDeviceInChainByPath(devicePath);
+        expect(edited != nullptr && edited->findParameterByIndex(kOsc1LevelSlot) != nullptr &&
+                   edited->findParameterByIndex(kOsc1LevelSlot)->currentValue == -60.0f,
+               "The model holds the edit");
+
+        expect(session.publishValues(valuesNow()).published, "The values are published");
```
lucaromagnoli, line 1147: P2: This calls `EngineSession::publishValues()` directly after `setDeviceParameterValue()`, so the test still passes if `TrackManager` stops notifying `EngineHost` or `EngineHost::deviceParameterChanged()` stops requesting a publish—the failure this PR says it pins. Please assert that the host receives a publish request from this edit (as the adjacent macro/lane test does), or drive the edit through a host-backed rendering path.

— Codex

tests/engine/test_engine_host_publish_juce.cpp
```diff
@@ -1042,6 +1043,114 @@ class EngineHostPublishTest final : public juce::UnitTest {
...
+        // DSP, so the first block either side of an edit is still on its way
+        // between the two values.
+        const auto sounding = render(40, 30);
+
+        // And the note that starts after the edit, which is a voice the change
+        // reached before it sounded rather than during.
+        const auto afterwards = render(220, 130);
+
+        expect(sounding < before * 0.1f, "The note that was sounding followed the edit");
+        expect(afterwards < before * 0.1f, "And so did the one that started after it");
```
lucaromagnoli, line 1167: P2: `afterwards` is only checked against an upper bound. If the second note never starts, the peak is zero and this assertion passes, so it does not establish that a newly started voice used the edited value. Please also verify that the second note sounded, for example with a lower-bound peak or captured note-on.

— Codex

tests/engine/test_engine_host_publish_juce.cpp
lucaromagnoli: Fixed in fd34a144.

tests/engine/test_engine_host_publish_juce.cpp
lucaromagnoli: Fixed in fd34a144.

Issue #2659, opened 2026-09-14
Two custom-UI paths still ask the fork: a Faust reload's parameters and the sequencer's drum lanes


lucaromagnoli:
Part of #1882. What #2585 left behind: two custom-UI paths that still reach the fork for
something that is not telemetry, and answer nothing under the native engine.

## A recompiled Faust device's parameters never reach the model

`FaustUI::tryLoad` and its editor-window twin refresh the model's parameter list and capture the
plugin's state through `getAudioBridge()->getPluginManager().refreshDeviceParameters(path)`
(`custom_ui/FaustUI.cpp:194-198, 385-389`). `refreshDeviceParameters` exists only on the fork's
`PluginManager` (`plugin_manager/PluginManagerSync.cpp:2232`) and `getAudioBridge()` is null
under magda, so loading new `.dsp` source into a runtime Faust device leaves the slot showing the
parameters the old source had. A compiled Faust device is not affected -- its slots come off the
catalog (#2609) and never change at runtime.

The refresh exists because Faust's parameter set changes at runtime, which is the one case
#2609's catalog scan cannot answer. Whatever replaces it belongs beside the model-authority work
in #2317.

## The poly step sequencer's drum lanes are discovered through a te::Track

`PolyStepSequencerUI`'s `DrumLanesView` finds its lanes by walking from the live plugin's
`getOwnerTrack()->state` and listening to that ValueTree (`custom_ui/PolyStepSequencerUI.cpp:988`,
`~540`). Under magda there is no `te::Track`, so the drum view has no lanes -- the keys view and
the playing step are unaffected (#2585 gave the faceplate the rendered device).

Which Drum Grid is downstream of a sequencer in the chain is a model question. The walk belongs
on `TrackManager` with a `TrackManagerListener` behind it, the way the rest of the chain UI
already answers it.

## Verification

Under magda: loading new source into a Faust device shows its new parameters in the slot; a poly
step sequencer with a Drum Grid downstream shows the pad lanes in DRUM view. Under Tracktion:
unchanged.

Issue #2658, opened 2026-09-14
The Levels device is te::Plugin only, so its meters read nothing on the native engine


lucaromagnoli:
Part of #1882, and the one device #2299 did not name.

## What is missing

`LevelsPlugin` is `te::Plugin` (`magda/daw/audio/plugins/LevelsPlugin.hpp:27`) and its entry in
`BaseDevicePack.cpp:470` has `createPlugin` with no `createDevice`, so
`canCreateEngineDevice()` is false for it. Under the native engine
`EngineRuntimeFactory::createDevice` returns null, the device is reported in `unbuilt()`, and
the executor renders the slot as a passthrough.

#2299 crossed the ten hand-written devices that were still `te::Plugin`; Levels is not in its
table and nothing else tracks it. Every other analysis device MAGDA ships -- Oscilloscope,
Spectrum Analyzer -- is a `MagdaDevice`, which is what let #2585 give their faceplates a
telemetry surface the rendering engine answers.

## What it costs

The faceplate is dead under magda: loudness, true peak, correlation and PLR all read nothing,
and `LevelsPluginTelemetrySource` is the one telemetry source in the chain UI still bound to a
fork plugin (`custom_ui/PluginTelemetrySources.hpp`). Audio is unaffected -- the device is
transparent -- so this is silent rather than audible.

## What to build

- `LevelsPlugin` as a `MagdaDevice`, with `TrackMeasurer` behind it unchanged and a
  `DeviceTelemetry` surface for its snapshot, the way #2585 gave the two tap devices theirs.
- `createDevice` in its `BaseDevicePack` entry.
- The UI's source moves to a device-backed one beside `DeviceOscilloscopeTelemetry`, and
  `PluginTelemetrySources.hpp` goes with it.

## Verification

Under magda: a Levels device on a playing track shows moving LUFS, true peak and correlation,
and its hold figures reset. Under Tracktion: unchanged.

lucaromagnoli:
Done in #2666 (merged to dev/0.20.0, which is why the PR's Closes line did not fire).

Issue #1630, opened 2026-06-29
Epic: Decouple MAGDA device UIs from live plugin instances


lucaromagnoli:
## Goal

MAGDA-native device UIs should no longer hold concrete live plugin pointers or depend on plugin lifetime. Custom device views should interact through stable controller and telemetry interfaces so plugin rebuilds, project reloads, device removal, rack nesting, and future process isolation are safe.

## Non-goals

- External VST/AU sandboxing
- Out-of-process plugin UI
- Replacing Tracktion ExternalPlugin
- Large DSP rewrites
- Project format changes unless required by existing transient UI/plugin coupling

## Success criteria

- No MAGDA custom UI stores raw FooPlugin* as long-lived state.
- Device controls write through model/controller APIs.
- Live visuals read through explicit snapshot/telemetry providers.
- Removing/rebuilding/reloading devices while UI is open does not crash.
- UI headers mostly stop including concrete audio/plugins/*Plugin.hpp.

## Child issues

- [x] #1631
- [x] #1632
- [x] #1633
- [x] #1634
- [x] #1635
- [x] #1636
- [x] #1637
- [x] #1638

## Suggested milestones

Milestone 1: Foundation
- #1631
- #1632

Milestone 2: Main migration
- #1633
- #1634
- #1635

Milestone 3: Hardening
- #1636
- #1637
- #1638

Estimated total: 4-5 weeks for a clean full boundary; 3 weeks if we accept some transitional adapters and defer edge-case cleanup.

lucaromagnoli:
## Assessment after auditing the actual UI/plugin coupling

Direction is right and the crash scenarios are real. Two things sharpen the scope: the coupling is **concentrated, not systemic**, and the real driver here is enabling **premium MAGDA devices that live in a separate (private) repo** built against open-source `magda-core`.

### Current coupling (facts from the code)
- **Most UIs are already decoupled.** Sampler, PolySynth, FourOsc, DrumGrid and the effects use `std::function` callbacks + `updateFromParameters()` with no stored plugin pointer. That pattern already *is* the target.
- **Only ~5 UIs hold a long-lived `Plugin*`:** `OscilloscopeUI`, `SpectrumAnalyzerUI`, `LevelsUI`, `NimbusUI` (telemetry) and `FaustUI`. So the realistic output of #1631 is "5 UIs to fix," not a 37-row migration table.
[...]

PR #2660, opened 2026-09-14
Read a faceplate's telemetry off the device that rendered it

lucaromagnoli:
Closes #2585.

The oscilloscope and the spectrum analyser drew nothing under the native engine, and so did
every custom UI that binds to a live plugin the same way: they all resolved through
`getAudioBridge()->getPlugin(path)`, which answers null there. #1630 closed with the faceplates
reading through `DeviceTelemetry` rather than concrete plugin pointers, which is the right seam;
what was missing is the other implementation of it.

## Change

- **One question, asked of whichever engine renders.** `AudioEngine::renderedDevice(path)` hands
  back a lease on the `MagdaDevice` the rendering engine holds -- the fork's plugin under
  Tracktion, the session's device under magda. The native side resolves off the same
  `devicePaths_` map the slot meters read (#2648), so a repaint costs a scan of the project's
  devices rather than a walk of its chains, and the device a slot draws and the device it reads
  telemetry off cannot come from two readings of the model. It sits beside
  `formatDeviceParameter` rather than on the control plane, for the reason that one does: it is
  a query, nothing is suspended, and it is read inside a repaint.
- **`telemetry(key)` answers something worth having.** `AnalysisTelemetry.hpp` adds the surfaces
  behind the key -- `OscilloscopeTelemetry`, `SpectrumTelemetry`, `GrainEnvelopeTelemetry` --
  implemented by the two tap devices and Nimbus. A faceplate asks its device for one by key and
  never for a plugin class.
- **A source resolves per read rather than holding a device.** That is what makes a rebuilt
  device (#2575) a rebind and a torn-down one (#2576) an empty trace, with no invalidation to get
  wrong and nothing left on a freed instance.
- **The compiled panels bind a device, not a plugin.** `bindPlugin(te::Plugin*)` becomes
  `bindDevice(shared_ptr<MagdaDevice>)` across all 22 curve views, each holding the handle its
  timer polls.
- **A faceplate built before the plan publishes rebinds.** The slot is built from the model and
  the device arrives on the publish after it, so a slot still waiting for one asks again on its
  timer instead of never.

Converted with it: the analyzer popout, the slot faceplates, the chord / arp / strum / step /
poly-step devices, the Poly Synth and Struck panels, and the mixer strips' mini scope and
spectrum. `DeviceSlotMidiUiBinding` goes -- it was a second copy of the same binding that only
ever worked under the fork.

## Tests

`[device-telemetry]`: a source reads through the device, follows a rebuilt one with no
rebinding, reads nothing once the device has gone, and answers nothing for a device with no such
surface.

`An analyser's telemetry carries the signal the engine rendered`: an oscilloscope after a Poly
Synth in a published plan, rendered for half a second, holds the note in its ring and says the
rate it was fed at. Verified negatively: with the scope's surface reverted, it fails at "The
device answers its telemetry by key".

## Verification

Automated suites pass and the app builds. Hands-on under magda is still to do: a scope and a
spectrum on a playing track, the step sequencer's position moving, a compiled device's curve,
and the mixer strip's mini analyzers.

## Not in scope

Levels is still `te::Plugin` and renders as a passthrough under magda (#2658). A Faust reload's
parameter refresh and the poly sequencer's drum-lane discovery still ask the fork (#2659).


magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -2231,27 +2215,76 @@ void DeviceCustomUIManager::refreshLivePluginBindings() {
     if (polyStepSequencerUI_ != nullptr)
         polyStepSequencerUI_->setDevicePath(devicePath_);
 
+    bindDeviceFaceplates();
+}
+
+bool DeviceCustomUIManager::awaitingRenderedDevice() const {
+    if (boundDevice_ != nullptr)
```
lucaromagnoli: P1: A successful first bind makes `awaitingRenderedDevice()` permanently false. When a sample-rate or block-size change remakes `EngineHost`'s session, `boundDevice_` keeps the retired instance alive and the slot timer no longer calls `refreshInlinePluginBindings()` for chord, arp, strum, sequencer, Poly Synth, or Struck faceplates. Their note/voice displays (and any live-device actions) then read the old instance until the slot is rebuilt. Please compare the current `renderedDevice(devicePath_)` with `boundDevice_` during the timer refresh, or use a re-resolving binding for these UIs.

— Codex

lucaromagnoli: Fixed in 8b455d9f. The predicate latched on "nothing bound yet", so the first successful bind stopped the slot timer asking again; it now compares the device rendering now against `boundDevice_`, so a session remade for a new sample rate is a rebind and a removed device clears the faceplates and drops the lease. It still answers false for a slot with no device faceplate, which is what keeps most slots off the timer.

Not pinned by a test: the predicate needs a constructed `DeviceCustomUIManager` with its UIs, which lives in `magda_daw_app` and neither test target links.

magda/daw/ui/components/chain/slot/DeviceCustomUIManager.cpp
```diff
@@ -2231,27 +2215,80 @@ void DeviceCustomUIManager::refreshLivePluginBindings() {
...
+    // sample rate rebuilds every device without the model moving, and a
+    // faceplate that stopped asking would poll the retired instance -- which
+    // boundDevice_ is holding open -- until the slot was rebuilt.
+    return polls && liveDevice().get() != boundDevice_.get();
+}
+
+void DeviceCustomUIManager::bindDeviceFaceplates() {
+    // Whatever renders the slot now. Held for as long as the raw pointers
+    // below do, so a rebuild cannot leave one reading a freed device (#2585).
+    boundDevice_ = liveDevice();
```
lucaromagnoli, line 2237: P2: `bindDeviceFaceplates()` runs from `setDevicePath()` for every device slot, even when none of the faceplates listed in `needsDeviceRebind()` exists. Those slots still store a strong `boundDevice_`, but their timer never compares or replaces it. After `EngineHost::rebuild()` remakes the session for a sample-rate or block-size change, ordinary internal-device slots such as Sampler or 4OSC keep their retired device (and its sample/buffer memory) alive until the slot is destroyed. Please only hold this lease when a faceplate actually keeps a raw device pointer, or refresh/release it for all slots.

— Codex

Issue #2663, opened 2026-09-14
The analysers forget their settings, and no authored-state edit reaches a device the native engine renders


lucaromagnoli:
Reported after #2660: under magda the oscilloscope and the spectrum analyser draw, but their
Time, Color and FFT settings come back as defaults after the app is restarted.

Two faults, both of the same shape as #2585 -- a path only the fork answers.

## An authored-state edit never reaches the device the native engine renders

`projectAuthoredStateToEngine` (`core/TrackManagerDevices.cpp:1976`) is how a settings edit gets
from the model to the running device. It goes
`getAudioBridge()->getPlugin(devicePath)->restorePluginStateFromValueTree(tree)`, and
`MagdaAudioEngine::getAudioBridge()` is null, so under the native engine the edit lands in the
model and nothing tells the instance. That is every faceplate whose settings are authored state:
the arpeggiator's, both sequencers', the sampler's loop and root note, the convolution's impulse
response. The fork's path ends in `device_->restoreState(tree)`
(`TracktionMagdaDevicePlugin.cpp:424`), which is what `AudioEngine::renderedDevice()` (#2585) can
now reach directly.

## The analysers' display settings are not authored state at all

`OscilloscopeUI` and `SpectrumAnalyzerUI` write Time, Color, FFT order, slope and smoothing
straight to the live device through the telemetry source, and the device persists them in
`flushState`. Under Tracktion a save captures that tree back into the model; under magda nothing
captures a MAGDA device's state, so the settings die with the instance. They belong in the
model's state document like every other faceplate's settings (#2317), written through
`updateDeviceAuthoredState` and projected back by the fix above.

## A faceplate binds before the device exists and never re-reads

`OscilloscopeUI::setTelemetrySource` reads the timebase and colour once, when the source is set.
The slot is built from the model and the engine publishes the device after it, so the read
happens against nothing and the faceplate shows 10 ms and colour 0 for the life of the slot --
even for a project whose device holds the saved values. The source is stable by design now
(#2585), so the re-read has to follow the device it resolves to rather than the source pointer.

## Verification

Under magda: set a scope's Time and Color and a spectrum's FFT, save, reopen -- the settings come
back, and the faceplate shows what the device holds rather than the defaults. An arpeggiator
settings edit changes what the running device does without waiting for a rebuild. Under
Tracktion: unchanged.

lucaromagnoli:
The reported case is the **track header's analysis toggle**, and that changes what the fix has to
[...]

PR #2664, opened 2026-09-14
Make an analyser's settings authored state, and project one onto the device that renders

lucaromagnoli:
Closes #2663.

Reported after #2660: under magda the analysers draw, but Time, Color and FFT come back as
defaults. Two faults behind it, both a path only the fork answers.

## No authored-state edit reached a running device

`projectAuthoredStateToEngine` went through
`getAudioBridge()->getPlugin(path)->restorePluginStateFromValueTree(tree)`, and
`MagdaAudioEngine::getAudioBridge()` is null. Under magda every faceplate whose settings are
authored state -- the arpeggiator's, both sequencers', the sampler's loop and root note, the
convolution's impulse response -- patched the model and left the instance alone.

The fork's path ends in `device_->restoreState(tree)`
(`TracktionMagdaDevicePlugin.cpp:424`), which is what `AudioEngine::renderedDevice()` (#2585)
reaches directly. `projectAuthoredStateToDevice` hands the device the same tree the plugin would
have been given, including the bare typed tree an empty document projects as.

## The analysers' settings were not authored state at all

They were written straight to the live device, which persists them in `flushState`: a Tracktion
save captures that tree back into the model, and nothing under magda does, so they died with the
instance. They now go to the document like every other faceplate's settings (#2317), through
`writeDeviceSettings` -- moved out of `DeviceCustomUIManager` into `DeviceStateCommands`, so the
popout window and the mixer's mini monitors write the same way rather than each growing a copy.

The device still takes the value during a drag and the document takes it when the drag ends: a
`devicePropertyChanged` per frame is a values publish per frame. A wheel or keyboard change ends
no drag, so those commit on the change.

## A faceplate read its settings before the device existed

`setTelemetrySource` read the timebase and colour once, when the source was set. The slot is
built from the model and the engine publishes the device after it, so the read found nothing and
never happened again -- the source is stable for the life of the faceplate since #2585. The read
now follows the device the source resolves to.

## A fresh analyser started at the factory's settings, not the ones last used

The reported case is the **track header's analysis toggle**, which is not a power switch:
`TrackChainContent::togglePostFxAnalysisDevice` deletes the device on the way off and adds a new
one on the way on. Nothing device-scoped survives that -- including the document the fixes above
give it. What survives a delete is the last-used setting, and that path was broken twice.

`createDetachedDevice` passes an empty session key on purpose (the services behind one are a
running engine's), and the user's preferences were carried in that same struct, so every analyser
the native engine built ignored them. `DevicePluginCreationContext` now carries defaults for a
detached creation and the engine's factory fills them from `Config` at build time -- read then
rather than snapshotted at startup, since the faceplate writes them as they change.

And the trace colour was never a last-used setting at all: `Config` carried the timebase and the
transform and no colour, so Color reset on every toggle under either engine. It joins them, and
both faceplates persist it the way they already persist the rest.

## Tests

`[analyser-defaults]`: an analyser built with nothing saved starts from the last-used settings --
both devices, every key -- and one the project has state for still comes back as what it saved.
Verified negatively: without the defaults, four assertions fail.

`[2663]`: an analyser's settings reach its document, and the projection puts that document onto a
running device -- both analysers, every key, through the same call the native engine makes.
Verified negatively: with the restore dropped, six assertions fail.

Not covered, and said out loud rather than left to look tested: which instance the projection
picks. That is one branch on an `AudioEngine`, and standing one up costs the 85 pure virtuals
`tests/project/test_project_serialization.cpp` already writes out by hand.

## Verification

Both suites pass and the app builds. Hands-on under magda is still to do: set a scope's Time and
Color and a spectrum's FFT, save, reopen; and check an arpeggiator settings edit changes what the
running device does.


magda/daw/ui/views/MixerView.cpp
```diff
@@ -25,6 +25,7 @@
 #include "components/chain/custom_ui/DeviceTelemetrySources.hpp"
 #include "core/ChainNodePath.hpp"
 #include "core/Config.hpp"
+#include "core/DeviceStateCommands.hpp"
```
lucaromagnoli, line 28: P2: This adds `DeviceStateCommands.hpp`, but `MixerView::ChannelStrip::refreshMiniAnalyzers()` still only installs telemetry sources; unlike `MasterChannelStrip`, it never assigns `onSettingsEdited` or calls `refreshSettingsFromSource()` after the rendered device arrives. On a normal track's compact analyzer (or its popout), Time/Color/FFT edits still change only the live instance and disappear after save/reopen, and the controls can keep showing defaults after the first publish. Please wire the track-strip paths through `writeDeviceSettings()` and refresh the controls when the device changes.

— Codex

magda/daw/ui/components/mixer/MasterChannelStrip.cpp
```diff
@@ -220,20 +221,38 @@ void MasterChannelStrip::refreshMiniAnalyzers() {
...
     if (miniOscilloscopeUI_) {
         if (miniOscilloscopeTelemetry_ == nullptr) {
             miniOscilloscopeTelemetry_ = std::make_shared<daw::ui::DeviceOscilloscopeTelemetry>(
                 analysisDevice("oscilloscope"));
+            miniOscilloscopeUI_->onSettingsEdited = editSettings("oscilloscope");
         }
         miniOscilloscopeUI_->setTelemetrySource(miniOscilloscopeTelemetry_);
+        // Every refresh, since the source is the same object for the life of
+        // the strip and the device behind it arrives with the plan.
+        miniOscilloscopeUI_->refreshSettingsFromSource();
```
lucaromagnoli: P2: `refreshSettingsFromSource()` only runs when `refreshMiniAnalyzers()` is called. Enabling a master mini analyzer adds its model device and synchronously calls this from `MixerView::trackDevicesChanged()`, while `EngineHost::wantPlan()` publishes the rendered device asynchronously afterward. The stable telemetry source initially reads fallback Time/Color/FFT values, and no post-publish callback or timer calls this method again, so the master controls can keep showing defaults even though the device has its saved or global settings. Please refresh on rendered-device identity change after publication (as the slot faceplates do), rather than only on model notifications.

— Codex

magda/daw/ui/views/MixerView.cpp
lucaromagnoli: Fixed in 01d6378e. You are right that it was never wired: the edit meant to give `ChannelStrip` its `onSettingsEdited` and its re-read silently matched nothing, so only the include landed. Both strips now share one resolution helper and one edit helper, and the track strips write through `writeDeviceSettings` like the master does.

magda/daw/ui/components/mixer/MasterChannelStrip.cpp
lucaromagnoli: Fixed in 01d6378e. Both strips now compare the device their query resolves against the one they last read and re-read on change, driven from `MixerView::timerCallback()` — the same place the slot faceplates do it — rather than only on the model notification that runs before the publish.

Issue #2669, opened 2026-09-14
The Drum Grid faceplate reads DrumGridPlugin, so its pads show nothing on the native engine


lucaromagnoli:
Part of #1882.

Under the MAGDA engine, a Drum Grid's pads can't be used from its faceplate. Loading a sample or dropping a device onto a pad shows nothing afterwards: no pad name, no device slots.

The pad edits themselves are model edits (`TrackManager::setPadDevice`, `addDeviceToPad`, `removePadChain`, and so on, #2207), and the plan compiler expands the pads (`emitPadRack`). What the faceplate shows is still read off Tracktion's live `DrumGridPlugin`, through `getLivePlugin()`. Under magda that is null, so every read comes back empty.

## Where the faceplate asks the fork

- **`DeviceCustomUIManager::createDrumGridUI`**
  - `getDrumGrid()` feeds `updatePadFromChain`, so pad names, mute, solo, level and pan never refresh after an edit.
  - `padChain.getPluginSlots` builds a pad's device slots from `DrumGridPlugin::getChainForNote` and `getPluginDeviceId`. Under magda it returns no slots.
  - Per-device slot meters come from `DrumGridPlugin::consumeChainPluginPeak`.
  - `onAnalyzePadRoleRequested` reads the pad's plugins through `getPadPlugin`.
  - Note preview and the refresh after a pad swap also go through `getDrumGrid()`.
- **`DrumGridUI`:** pad trigger flashes (`consumePadTrigger`), chain info (`getChainByIndex`) and the collapsed-detail flag (`uiDetailCollapsed` on the plugin's state) are all read off `drumGridPlugin_`.
- **`DeviceSlotDrumGridBridge`:** `appendAvailableDevices` and `appendDeviceParamNames` walk `getDrumGridPlugin()->getChains()`.
- **`DrumGridClipContent`:** `findDrumGridForTrack` walks the track's `te` plugin list.

## Also

A Drum Grid with no pads yet has null `DeviceInfo::pads`, so the compiler plans it as an ordinary device op. The native factory has nothing to build it with, and the log reports `no device bound for D1` and `nothing to build "Drum Grid" with`.

## What to build

- Pad state (names, the devices on each pad, switches, faders) read from `DeviceInfo::pads` in the model.
- Pad device faceplates and meters bound to the rendered pad devices, the way #2585 binds the other faceplates.
- Pad triggers, note preview and pad-role analysis each answered by the engine that renders the pad.
[...]

PR #2670, opened 2026-09-14
Find the drum lanes in the model and offer the Drum Grid on the native engine (#2659)

lucaromagnoli:
Part of #2659: the drum-lanes half. The Faust reload half is not in this PR.

## Drum lanes from the model

`PolyStepSequencerUI`'s `DrumLanesView` found its lanes by walking the live sequencer's owner `te::Track` for a downstream `DrumGridPlugin`, and by listening to that track's ValueTree. Under the native engine there is no `te::Track`, so the DRUM view had no lanes.

- **Query:** `TrackManager::findPadDeviceDownstreamOf(path)` returns the first pad device after the sequencer, in chain order, racks included.
- **Lanes:** built from that device's pads in the model (`DeviceInfo::pads`, each chain's name and `lowNote`). Without one, the view falls back to the General MIDI drum lanes, as before.
- **Refresh:** the view now listens to `TrackManagerListener` (`tracksChanged`, `trackDevicesChanged`) instead of ValueTrees. The Tracktion includes are gone from `PolyStepSequencerUI.cpp`.

## Drum Grid in the device browser on the native engine

Under the MAGDA engine, the browser listed only internal devices the native factory builds. A Drum Grid is never built: the plan compiler expands its pads. The new `engineRendersDevice` returns true for both kinds, and the browser uses it.

The Drum Grid faceplate itself still reads `DrumGridPlugin`, so on the native engine its pads show nothing. That is tracked in #2669.

## Tests

- Catch2: the pad device after a sequencer inside a rack, a grid before the sequencer, and the last grid on the track. The engine plays a Drum Grid it never builds.
- Locally, after merging `dev/0.20.0`: Catch2 4101 passed, 13 skipped. Before the merge, the full JUCE binary passed.

Not tried in the app: the drum view is part of the app only, and on the native engine pads can't be added from the Drum Grid faceplate until #2669.

https://claude.ai/code/session_01X7ucu5dVjSFik1zMuiGG1R


magda/daw/core/TrackManagerPads.cpp
```diff
@@ -52,6 +53,26 @@ const RackInfo* TrackManager::getPads(const ChainNodePath& gridPath) const {
...
+    const auto* track = getTrack(devicePath.trackId);
+    if (track == nullptr)
+        return nullptr;
+
+    const DeviceInfo* found = nullptr;
+    auto passed = false;
+    chain_walk::forEachDevice(track->chain.fxChainElements,
+                              ChainNodePath::trackLevel(devicePath.trackId), chain_walk::Pads::Skip,
+                              [&](const DeviceInfo& device, const ChainNodePath& path) {
+                                  if (passed && isPadRackDevice(device.pluginId)) {
```
lucaromagnoli: **P2 — Follow signal routing when finding downstream pads**

The depth-first walk treats later parallel rack chains as downstream once `passed` becomes true. For example, with a sequencer in chain A and a Drum Grid in chain B of the same rack, this returns B’s grid even though B receives the rack input, not A’s generated MIDI (`Compiler::emitRack()` starts each chain from the same incoming signal). The drum view then shows pads that the sequencer cannot trigger and can overlook a genuinely downstream grid after the rack. Please follow the sequencer’s signal path, skipping sibling parallel branches, and cover this layout in the query tests.

— Codex

magda/daw/ui/components/chain/custom_ui/PolyStepSequencerUI.cpp
```diff
@@ -838,32 +769,12 @@ class PolyStepSequencerUI::DrumLanesView : public PolyStepSequencerUI::PatternVi
...
-        if (child.hasType(te::IDs::PLUGIN) || child.hasType(DRUM_CHAIN_TYPE))
-            triggerLaneRefresh();
-    }
-    void valueTreeChildRemoved(juce::ValueTree&, juce::ValueTree& child, int) override {
-        if (child.hasType(te::IDs::PLUGIN) || child.hasType(DRUM_CHAIN_TYPE))
-            triggerLaneRefresh();
-    }
-    void valueTreeChildOrderChanged(juce::ValueTree& parent, int, int) override {
-        if (parent == trackState_)
+    void trackDevicesChanged(magda::TrackId trackId) override {
```
lucaromagnoli, line 777: **P2 — Refresh lane labels on property changes**

`TrackManager::setChainName()` updates the chain name and emits `trackPropertyChanged`, but this listener handles only `tracksChanged` and `trackDevicesChanged`. Renaming a pad chain through that API therefore leaves its drum-lane label stale until a pattern edit, device change, or reopening refreshes the view. The timer does not repair this when the pattern is unchanged, and `TrackChainContent::trackPropertyChanged()` only updates header widgets. Please also refresh on property changes for the owning track, with coverage for a pad-chain rename.

— Codex

Issue #2552, opened 2026-09-09
Cutover slice 2: the session launcher under the native engine


lucaromagnoli:
Slice 2 of #1897. Depends on slice 1.

The launcher's engine half is finished — #1894's seven slices closed it, and the corpus nulls `session.launch`, `session.launch.midi` and `session.mode.holds.slot`. What it has never had is an app driving it.

## Scope

- Slot launches and scene launches from the UI reaching `LaunchRequestQueue`, at the beat the host resolved the clip's `LaunchQuantize` to. Quantization stays host-side, which is where it already lives (it has no occurrence in `magda/engine/`).
- The launcher's published state answering the interface's session half: `getSessionClipPlayState`, `getSessionPlayheadPosition`, `getSessionPlayheadClipId`, `getActiveClipPlayheadPositions`, `isSessionTrackStopPending`, `deactivateAllSessionClips`. Off `LaunchTap` and the run edges rather than off a poll.
- `stopSessionTrack` and the quantized stop, including the blinking empty-slot affordance the interface documents.
- Follow actions (#2304) reaching the app.
- Per-track session/arrangement switching, which is the section hold a launched handle takes (#2302).

## Not in this slice

Recording into a slot, and the session-to-arrangement capture. Both are built (#2464) and both belong with the rest of recording in slice 3.

## Verification

- Launching a clip, a scene and a stop behaves as it does on the fork, at each quantization including None.
- Clip playheads and play states in the session view track what is sounding, with no drift against the audio.
- A follow action fires and the UI follows it.
- #2457 and #2458 are re-asked by ear on this engine rather than by corpus.

lucaromagnoli:
Closing as completed after the combined work in #2673 (app-to-native launcher bridge), #2713 (transport restart), and #2719 (stop acknowledgement, live loop edits, and playhead consistency).

The native app now drives slot/scene launches and quantized stops, exposes published launch state/playheads, follows engine follow actions, and switches per-track Session/Arrangement ownership. Regression coverage includes 25 launcher cases / 152 assertions and the real UI timer ordering test (13 assertions). Debug and Release builds pass. The user confirmed the Session fixes in the running application.

This closes the launcher integration slice, not the independent shared-engine trim de-click question in #2457. The claimed trim shift in #2458 was disproved by its follow-up measurements; #2549 retains coverage of the actual sustained-signal residual. We are not claiming that every quantization setting or those separate trim cases received a new listening test in #2719.

Issue #2458, opened 2026-09-06
A trimmed session slot's launch lands 23 samples apart between the engines


lucaromagnoli:
Found while answering review on #2456, on a corpus case written to ask a different question.

## What was measured

A session slot 16 beats long, launched on the render's first sample, reading a steady-level source from a trim point 0.375 s into it (`sourceAnchorSamples`). Everything else is `session.launch`, which nulls at -133 dB.

```
session.launch.trim   exact   peak=-113.5   rms=-139.2   shift=-23.000   FAIL
```

The residual is quiet, and the shift is the finding: the comparator's best alignment is **23 samples**, so the two engines resolve the same trim point to positions 23 samples apart. Both render the material correctly either side of it.

Not the launch de-click, which is #2457: the two agree there, and a de-click difference would be full-scale rather than -113 dB.

## Why it is not in the corpus yet

The case exists in working form but was left out of #2456, which is about the de-click and would have been muddied by a failure with a different cause. It wants adding here with whatever this turns out to be — a declared shift with its mechanism if the difference is a known one, or a fix if it is not.

`placement.trims` covers the same trim on an **arrangement** clip and nulls, so whatever moves is on the launcher's path rather than in how a trim is compiled.

## Where to start

The fork sets a slot clip's trim through `AudioClipBase::setOffset` in **seconds** (`ClipSynchronizer::syncSessionClipToSlot` passes `engineOffsetSeconds`), while the engine carries `sourceAnchorSamples`. 23 samples at 44.1 kHz is 0.52 ms, which is the shape of a seconds/samples rounding rather than a wrong trim.

lucaromagnoli:
Measured it. **The trim is not involved, and the 23 samples are not a shift.** PR #2549 adds the case that does find something.

## What the 23 was

The comparator's best-alignment search on material that has nothing to align. The residual is one to two ULP per sample scattered over the whole render — the engine emits a held `0.5` bit for bit, the fork wobbles in the last bit. Straight from the artefacts this issue was filed on, still on disk from Sep 6:

```
sample 0: native= 0.50000000  incumbent= 0.50000012
sample 3: native= 0.50000000  incumbent= 0.49999985
sample 6: native= 0.50000000  incumbent= 0.50000012
```

15799 single-sample regions above -140 dB, peak -113.5 dB at sample 159861. A real 23-sample slip of a full-scale steps signal would sit near 0 dB at every flip edge, not at -113 dB.

## The runs

[...]

Issue #2457, opened 2026-09-06
Neither engine de-clicks a launched clip trimmed into its material


lucaromagnoli:
Found while answering review on #2456, and measured rather than argued.

## What both engines do

A slot clip trimmed into its source starts on a full-scale sample: silence, then a step. Rendered through both engines, launched from the render's first sample, with a steady-level source read from a trim point:

```
native.wav    first samples: 0.5 0.5 0.5 0.5 ...
incumbent.wav first samples: 0.5 0.5 0.5 0.5 ...
```

Neither takes the step out. They agree, so the corpus has nothing to report — which is why this needs an issue rather than a case.

## Why they agree

Each asks the same question in its own words, and it is not the question a trim raises.

`ClipVoice` (`magda/engine/clip/ClipVoice.cpp`):

```cpp
const auto beginsAtItsOwnStart = windowStart <= event.span.seconds.start + (0.5 / sampleRate_);
```

That is "does the voice begin where the clip begins", not "does it begin where the file begins". A trimmed clip satisfies it.

The fork reached the same place from the other direction: its launch de-click used to fire on every launch, and Conceptual-Machines/tracktion_engine#35 stopped it firing where a run begins rather than is joined. #36 says so in the comment, having measured this first.

## The question

A clip's own attack is not a discontinuity and should not be de-clicked — that is #2444, and it is settled. A trim point is a different thing: the material either side of it is real, and the step is an artefact of where the user cut. Arguably both engines should de-click there, in which case both want the same change and the corpus stays quiet either way.

## Not to be confused with

A trimmed launched slot also renders about **23 samples apart** between the two engines, at a peak of -113.5 dB. That is an alignment difference in where the trim point resolves, not a de-click, and it wants its own issue and its own corpus case.

PR #2673, opened 2026-09-14
Drive the session launcher from the app (#2552)

lucaromagnoli:
Slice 2 of #1897. The launcher's engine half has been finished and null-diffed since #1894 --
`EngineSession` publishes a handle per slot, applies what `LaunchRequestQueue` carries and writes
each block's result to a `LaunchTap`. What it had never had was an app driving it, so a click on a
session slot did nothing.

`SlotLauncher` (`magda/daw/engine/host/`) is the other end: a click becomes a request at the beat
the clip's `LaunchQuantize` resolves to, and what a slot is doing is read back off the tap the
block that advanced it wrote -- play states, playheads and the stop affordance alike, never polled
off a handle the audio thread owns.

## The sync point

Quantization stays host-side, which is where it already lived. That needs the timeline beat and
the monotonic count as **one** reading: a launch is queued in monotonic beats and the boundary it
waits for is a timeline one. `TransportClock::syncPoint()` publishes the pair under a seqlock --
two doubles are too wide to pack into one word the way `LaunchTap` does, and a pair read across a
loop wrap is a whole pass out, which would fire a queued launch immediately instead of at the bar.
`SyncPoint::monotonicAt(beat)` is the conversion; `TimeDomains.hpp` says why this two-faced instant
is legitimate where `BlockInstant`'s is not (same unit, one offset, no map between them).

## A scene is one gesture

Not a loop of launches: every slot reaches the audio thread together, with the followers joining
the leader's run so a relaunched scene stays in phase. `launchSessionScene` is a **pure** virtual
on `AudioEngine` -- the fork keeps its own per-clip semantics rather than inheriting a no-op
(#2551's gotcha) -- and `SessionLaunchService` asks the engine instead of looping itself.

## Off #2551's unwired list

`getSessionClipPlayState`, `getSessionPlayheadPosition`, `getSessionPlayheadClipId`,
`getActiveClipPlayheadPositions`, `isSessionTrackStopPending`, `stopSessionTrack`,
`deactivateAllSessionClips`, `processSessionStateEvents`. Eight `reportUnwired` lines deleted; the
recording half stays named for #2553.

## What comes with it

- **Follow actions (#2304)**: the engine ends one run and begins another inside the audio thread
  and tells nobody, so `processStateEvents` takes any slot the taps say is sounding as what its
  track is playing. The same sweep settles a one-shot that reached its end.
- **Section holds (#2302)**: a stop releases the section rather than only silencing the slot, so
  the arrangement under the track is audible again. The test asserts `holdsSection` off the tap,
  because the model's `TrackPlaybackMode` would look right either way.
- **Playheads** wrap on the slot's pass length -- `placement.lengthBeats`, which is what the
  engine re-triggers the handle on (`ClipSnapshot.hpp`), so the drawing cannot drift from the audio.
- A launch against a slot with nothing published -- an empty MIDI clip compiles to no material and
  so gets no handle -- says so once instead of being dropped on arrival in silence.

Recording into a slot and the session-to-arrangement capture stay with the rest of recording in
#2553, as the issue scopes them.

## Verification

`tests/engine/test_slot_launcher_juce.cpp` runs the whole path rather than the translation: a real
`EngineSession` with a published project, blocks rendered, state read off the taps.

- A launch with nothing to wait for sounds on the next block, and the session takes the track.
- A quantized launch is Queued half a beat in and Playing past the bar line.
- A scene starts its slots together, and a track whose slot in that scene is empty stops.
- A stop releases the section and gives the arrangement back.
- The playhead runs inside one pass and wraps rather than running on.

Negatively verified: making `dueBeat` always immediate fails the quantized case alone, and
swapping `backToArrangement` for a plain `stop` fails the section-hold case alone.

- `make test-juce`: green.
- `make test`: 4116 cases passed, 13 skipped.
- Null Diff Corpus: **65 ok**, with the 9 `underCalibration` cases inherited from dev.
- `.pre-commit-hooks/clang_tidy.py` over the changed `magda/daw` translation units: clean.

Hands-on checks the issue asks for and this does not cover: each quantization by ear, a follow
action watched in the UI, and #2457 / #2458 re-asked on this engine.


magda/daw/engine/host/SlotLauncher.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+    auto* session = host_.launchSession();
+    if (session == nullptr)
+        return;
+
+    const auto due = dueBeat(*clip);
+
+    {
+        engine::LaunchRequestQueue::Gesture gesture(session->launchRequests());
+        gesture.stop(keyOf(*clip), due);
```
lucaromagnoli, line 286: P1: `stopTrack()` sends a plain `stop` request. `LaunchHandle::stop()` ends the run but leaves `holdsSection_` true; only `backToArrangement` releases it. After an empty slot in a scene or the track-stop control stops the clip, the model eventually switches to Arrangement while the session still holds that section, so the arrangement underneath stays silent indefinitely. The slot-stop path above uses `backToArrangement` for this reason. Please make the track stop release the section at its quantized due beat, and assert `holdsSection` is false in the empty-scene test.

— Codex

magda/daw/engine/host/SlotLauncher.cpp
```diff
@@ -0,0 +1,553 @@
...
+        if (track == nullptr)
+            continue;
+
+        const auto sounding =
+            state == SessionClipPlayState::Playing || state == SessionClipPlayState::Queued;
+
+        if (sounding && track->activeSessionClipId != clip.id) {
+            track->activeSessionClipId = clip.id;
+            playheadClip_ = clip.id;
+            stopping_.erase(clip.trackId);
```
lucaromagnoli, line 432: P2: A quantized `stopTrack()` clears `activeSessionClipId` and inserts this track in `stopping_` while its tap still reports Playing until the due beat. On the very next `processStateEvents()` tick, this branch treats that still-playing clip as a follow action, restores `activeSessionClipId`, and erases `stopping_`. The pending-stop affordance therefore disappears long before the queued stop lands. Keep a track in `stopping_` until its tap actually stops, and do not re-adopt its clip as a new launch during that interval.

— Codex

lucaromagnoli:
## What this does not fix, found by hand after the PR was opened

Two clips could sound on one track. The engine renders every slot whose handle is playing
(`SessionPlayback.hpp:forEachSlot`), so one clip per track is the grid's rule and not the
launcher's, and nothing asked for the handover. Fixed in 4be65df2 with a test that fails without
it: `launch` and `launchScene` stop the track's other sounding slots inside the launching gesture,
at the beat the new one starts.

The rest of what a session slot sounds like is **not** fixed here, and is upstream of the launcher.
Merging this makes a click launch a clip; it does not make the session grid right:

- **#2674** -- there is no tempo detection at all, on either engine. `runBpmDetection` has returned
  `0.0` unconditionally since it was disabled for crashing, and nothing calls `cacheBPM`. A file
  whose header carries a tempo is read now (`loopInfoFrom`, landing separately); anything else has
  no tempo.
- **#2676** -- a session clip is put into beat mode whether or not a tempo is known, so it claims
  BEAT while `loopLengthBeats()` answers zero and the engine plays it at file rate rather than
  filling the pass.
- **#2675** -- a loop's musical length is a function of the tempo it was measured at, so correcting
  a wrong detection moves it off the grid.

Two more reported by ear and not yet measured: a launched loop's first transient sounds soft, and
SoundTouch sounds glitchy on a stretched session loop even with the tempo entered by hand.

## #2552 stays open

Its verification is hands-on and is not met: each quantization by ear, a follow action watched in
the UI, and #2457 / #2458 re-asked on this engine. The issues above make those hard to judge right
now, because a session clip sounds wrong for reasons that are not the launcher's.

Issue #2674, opened 2026-09-14
Index unscanned files for BPM, and fill the dspBpm stub


lucaromagnoli:
Part of #1882. Rewritten after reading what is actually there: the first version of this issue said
there was no tempo detection anywhere, and that was wrong.

## What exists

`media_db/AudioFeatures.cpp::extractFeatures` derives a BPM from three tiers, cheapest first:

1. `parseBpmFromPath` -- the filename token. Implemented, and the reason a well-named loop never
   pays for analysis.
2. `metadataBpm` -- the ACID chunk, off the reader's metadata. Implemented.
3. `dspBpm` -- **`(void)path; return std::nullopt;`**

The indexer then applies policy: `shape == "one-shot"` clears the BPM, which is right -- nothing
musical should be read off a one-shot's length.

A clip picks the result up at import. `ClipManager::createAudioClipBeats:394` reads
`getUserMetadataForFile`, which is COALESCE(user override, scanner-detected), and seeds
`interpBpm` from it. So for an **indexed** file this works today, on both engines.

The ONNX work in `media_db` -- CLAP audio and text encoders, the zero-shot tagger -- is embeddings
and tagging. Nothing ONNX touches BPM.

## The two real gaps

**1. An unscanned file has no row.** None of the tiers ever run for it, so a loop dragged in from
outside a scanned library arrives with no tempo however it is named. `ClipManager` already has the
on-demand path -- `isFileIndexed(path)` then `MediaDbIndexer::indexFile(path, ForceAll)` -- but only
in the save-to-library flow, never at import. This is the case found by ear in #2552, and it is the
cheap half: an unindexed file gets indexed, and tiers 1 and 2 answer.

When it runs is decided: **on the BEAT button, not on import.** Pressing BEAT on a clip with no
interpretation is what triggers the work. Nothing is scanned on drop, there is no preference to
invent about when to scan, and the press is the signal that this file's tempo is what the user cares
about. An answer seeds `interpBpm` and beat mode follows; no answer falls back to the project tempo,
so the button still does what pressing it means -- *play this at my tempo* -- with the source BPM
and beat-count fields there to correct it.

**2. The DSP tier is empty.** For a file that says nothing in its name and carries no acid chunk,
there is no answer anywhere in the product. `AudioThumbnailManager::runBpmDetection` is the second
stub of the same thing, disabled because SoundTouch's `BPMDetect` crashed "inside its worker
thread".

[...]

PR #2678, opened 2026-09-14
Take a source's tempo from the file it was probed from (#2674)

lucaromagnoli:
First piece of #2674.

Nothing ever wrote `Source::detectedBpm`. A clip's tempo came from Tracktion's `loopInfo` through
`ClipSynchronizer`, so under an engine with no Edit an imported loop had no tempo at all -- it
could not be in beat mode, and it looped at whatever length its seconds happened to be. That is
one of the two halves of what #2552 turned up by ear.

`SourcePool::probe` already opens the reader whose header carries the answer, so the parse #2038
delivered and never wired -- `engine::loopInfoFrom`, over the acid chunk JUCE already decodes --
runs there.

A hook rather than a direct call: the parse lives beside the engine's reader and `magda_daw` does
not see that header. `EngineProject` installs it, because that side sees both, and
`MagdaAudioEngine::initialize` is the only caller -- so **the fork is untouched** and
`Source::detectedBpm` behaves there exactly as it did.

A gap only, like every other seeding rule in this model: a tempo a project saved, or one the user
typed, outranks what the header claims.

## What this does not do

Files that say nothing still have no tempo. There is no detection on either engine --
`AudioThumbnailManager::runBpmDetection` has returned `0.0` unconditionally since it was disabled
for crashing, and nothing calls `cacheBPM`. That is the rest of #2674: the `BPMDetect` crash first,
then loop-length inference, then inter-onset autocorrelation over the transients the engine already
detects.

## Verification

`tests/audio/test_source_loop_info.cpp` gains three cases over a real WAV carrying a real acid
chunk, so the round trip is the one a loop library's file takes -- JUCE writes the chunk, the reader
parses it back into metadata, the parse reads that:

- with no probe installed, a probed source has its rate and duration and no tempo, which is what
  every import got before this
- installed, a file written at 140 bpm arrives as `detectedBpm == 140`
- a tempo the project already carries is not overwritten by a re-probe

Suites, run on the merge base:

- `make test`: 4121 cases, 4108 passed, 13 skipped, 0 failed.
- `make test-juce`: green.
- Null Diff Corpus: **65 ok**, with the 9 `underCalibration` cases inherited from dev.


lucaromagnoli:
Closing without merging: it is a second route to a fact the media DB already carries.

`AudioFeatures::extractFeatures` has had the tiered source all along -- `parseBpmFromPath`, then
`metadataBpm` over the same acid chunk this PR parses -- and a clip already reads the result at
import through `getUserMetadataForFile` (`ClipManager.cpp:394`), which is COALESCE(user override,
scanner-detected). So the metadata tier exists; what this PR added was a parallel path writing
`Source::detectedBpm` while the DB path writes `event->interpBpm`. One fact, two routes, and the
two would drift.

What is actually missing is narrower, and #2674 now says so:

- a file that was never scanned has no row, so none of the existing tiers apply to it -- which is
  the case that turned up by ear
- the DSP tier is a stub (`dspBpm` returns nullopt, as does `AudioThumbnailManager::runBpmDetection`)

`engine::loopInfoFrom` stays where it is. It is a richer parse than `metadataBpm` -- beat count,
signature and the one-shot flag, not just a tempo -- and #2038 delivered it for seeding an event's
interpretation. If the DB's metadata tier is ever widened, that is what it should be widened to.

PR #2742, opened 2026-09-19
Record hardware audio into native Arrangement clips

lucaromagnoli:
Part of #2553.

An armed audio track on the native engine now records its selected hardware input into an editable Arrangement clip, with a growing waveform preview while recording.

- `EngineHost` owns audio takes beside MIDI takes. They share one close path: disarm, rerouting, track deletion, transport or device stop, project replacement, and tempo, signature or loop changes.
- The device-reported input latency is removed from the head of each take.
- `ClipManager::createRecordedAudioClip` installs the file, placement and loop takes before its single notification.
- Design notes: `docs/issues/native-arrangement-audio-recording.md`. Smoke steps: `docs/smoke-tests/recording.md`.

## Known issues (not fixed in this PR)

- Audio and MIDI recording previews share per-track maps in `EngineHost`. A MIDI take on the same track (default input "all") overwrites the audio waveform.
- `rebuild()` resumes recording before `addAudioCallback`. The generation bump restarts audio takes, which can split one performance into two clips.
- `createRecordedAudioClip` pins the live project tempo, not the tempo the take was recorded at. Changing the tempo mid-recording cuts off the end of the clip.


magda/daw/core/ClipManager.cpp
```diff
@@ -440,6 +440,54 @@ ClipId ClipManager::createAudioClip(TrackId trackId, double startTime, double le
...
+
+    AudioEvent event;
+    event.sourceId = SourcePool::getInstance().acquire(recording.filePath);
+    event.speedRatio = 1.0;
+    event.seedInterpretationFromSource();
+    auto& active = clip.audio().addEvent(event);
+    clip.syncSingleEventToClipBounds();
+    active.loopStartSamples = 0;
+    active.setLoopExtent(RegionExtent::WholeSource);
+    active.adoptBpm(projectBpm, Provenance::User);
```
lucaromagnoli, line 480: This pins the live project tempo, but `recording.lengthBeats` was measured at the tempo the take was recorded at. On a mid-take tempo change (120 → 140), `TimelineController` writes 140 to the model before `setTempo` → `stopRecording()` builds this clip. It gets bpm 140 with beats counted at 120, and `WholeSource` then fits the region to `L*60/140` s, which cuts off the end. Pass the take's recording tempo in `RecordedAudioClipData` (or derive it from the file length) instead of reading `currentProjectTempoOrDefault()`.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -2924,7 +3111,7 @@ void EngineHost::setTempo(double bpm) {
         return;
 
     const auto resumeRecording = impl_->arrangementRecording_;
-    impl_->stopMidiRecording();
+    impl_->stopRecording();
```
lucaromagnoli, line 3114: `stopRecording()` / `resumeRecording()` now closes and reopens audio takes too. Audio arriving between `stopTake` and the next rolling block of the new take lands in neither file, so nudging tempo mid-take leaves two clips with a gap. `TakeRecorder` already reads the take end from the tempo map, so audio takes could stay open here and only MIDI be cut.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -2946,13 +3133,13 @@ void EngineHost::setTimeSignature(int numerator, int denominator) {
...
```
lucaromagnoli, line 3136: Same as `setTempo`: a signature change splits the audio take and drops the audio in between.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -2967,11 +3154,11 @@ void EngineHost::setLoop(bool enabled, double startBeat, double endBeat) {
...
```
lucaromagnoli, line 3157: Same gap on loop toggles. `TakeRecorder` already splits passes at loop wraps inside one take, so the audio take doesn't need restarting.

lucaromagnoli (commented):
Post-merge review notes. Of the known issues listed in the description, the shared preview maps and the rebuild ordering are already fixed at this head.

PR #2752, opened 2026-09-19
Compensate native audio takes for round-trip latency

### LINKED ISSUE #2751: Compensate native audio takes for round-trip device latency
Part of #2553. Independent of #2746 and #2747.

## Problem

`EngineHost::audioDeviceAboutToStart` currently gives audio takes only `getInputLatencyInSamples()`. A transient played from the Arrangement reaches a physical loopback after the interface output latency and input latency, so dropping only the input side leaves the recorded event late by the output latency.

The Tracktion path defines its automatic record adjustment as input plus output latency (`DeviceManager::getRecordAdjustmentSamples()`), which is also what the physical loopback check in `docs/smoke-tests/recording.md` expects.

## Scope

- Use the active interface input + output latency as the native audio take correction.
- Carry that correction consistently through Arrangement and Session audio recording.
- Treat a changed correction after an interface restart as a route change, preserving the existing take before continuing with the new value.
- Rename host state away from input-only terminology so the sign and ownership stay clear.
- Extend the callback fixture with distinct input and output latency values and assert the materialized file/timeline alignment.

## Acceptance

- A fake interface with separate input and output latency drops their sum from the recorded head.
- Arrangement and Session takes use the same correction.
- A restart with changed latency closes the old take and continues without applying two different corrections to one file.
- Existing positive/negative `TakeRecorder` correction tests remain green.
- The recording smoke-test documentation states which automatic figure native uses.

## Non-goals

- `AudioIOService` ownership or channel policy (#2746/#2747).
- A user-entered manual recording offset; that remains a separate calibration/UI slice after audio-I/O ownership settles.
- Monitor-path latency or anticipative rendering.

lucaromagnoli:
## Summary

- use the active interface input plus output latency as the native audio take correction
- carry the same correction through Arrangement and Session recording
- preserve the old take and start a separately corrected file when an interface restart changes the recording route
- document the automatic round-trip correction and strengthen callback coverage for file and timeline alignment

## Validation

- Engine Host Audio Input: all nine scenarios passed
- TakeRecorder latency coverage: 141 assertions in two focused cases passed
- magda_juce_tests target built
- magda_daw_app target built
- pre-commit checks passed

Closes #2751


magda/daw/engine/host/EngineHost.cpp
```diff
@@ -887,11 +887,11 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
                  sessionScene ? engine::RecordTarget::slot : engine::RecordTarget::arrangement,
              .scene = sessionScene.value_or(-1),
              .maxPeaks = kRecordingPreviewPeaks},
-            [this, channels = std::move(channels), latency, sessionScene,
+            [this, channels = std::move(channels), adjustment, sessionScene,
              trackId = track.id](engine::RecordTap& tap) {
                 engine::TakeRecorderSettings settings;
                 settings.channels = channels;
-                settings.latencySamples = latency;
+                settings.latencySamples = adjustment;
```
lucaromagnoli, line 943: **The tail of every take is now cut by the full round trip.** `TakeRecorder` stops capturing at the exact stop/punch-out block with no post-roll (`TakeRecorder.cpp:67`), so the last `adjustment` samples of the performance are never written. Tracktion keeps recording `adjustDurationAtStart` past the stop for this reason. This PR grows the loss from input latency to input+output.

- Session slots: `createRecordedAudioClip` sets `loopLengthBeats` from the shortened length, so a 4-bar take loops at 4 bars minus the round trip and drifts against the scene every cycle. A MIDI slot take from the same launch keeps its full length.
- Punch-out: the clip ends early, `ResolveOverlaps` trims the old clip only up to there, and the old take plays through the last round trip of the punched region.

The new assertions at lines 479 and 626 (`4 * kBlockSize - kRecordingAdjustment`) lock this in.

lucaromagnoli (commented):
Review of the round-trip correction. The main issue is the tail truncation (comment on `EngineHost.cpp:894`): the PR moves the head correction to input+output but nothing records past the stop, so every take, and every Session loop length, is short by the round trip.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -2330,7 +2332,9 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
         rate_.store(device->getCurrentSampleRate());
         blockSize_.store(device->getCurrentBufferSizeSamples());
         inputChannels_.store(device->getActiveInputChannels().countNumberOfSetBits());
-        inputLatencySamples_.store(device->getInputLatencyInSamples(), std::memory_order_relaxed);
+        recordingAdjustmentSamples_.store(device->getInputLatencyInSamples() +
```
lucaromagnoli: Tracktion also adds the plan's delay-compensation latency (`tracktion_WaveInputDevice.cpp:406`: `rc->adjustSamples += context.getLatencySamples()`). This leaves it out. With a lookahead plugin on master, native output is delayed by `PlanExecutor::latencySamples()` but the take isn't corrected for it, so it lands late. That breaks the "matching Tracktion" claim in the docs and the smoke test's same-sample check.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -2330,7 +2332,9 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
...
+                                              device->getOutputLatencyInSamples(),
```
lucaromagnoli: No bounds on the driver-reported sum. Tracktion clamps its adjustment to ±3 s (`tracktion_WaveInputDevice.cpp:1661`). A virtual or aggregate driver reporting a huge output latency drops most or all of the take, and a value near `INT_MAX` overflows the `int` sum.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -872,11 +872,11 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
         if (session_ == nullptr)
             return;
 
-        const auto latency = inputLatencySamples_.load(std::memory_order_relaxed);
+        const auto adjustment = recordingAdjustmentSamples_.load(std::memory_order_relaxed);
```
lucaromagnoli: Only audio takes get the output-latency correction. MIDI takes from the same Record gesture still use `MidiTakeRecorderSettings.latencySamples == 0`, so audio and MIDI recorded together (e.g. an e-drum's MIDI and audio on two armed tracks) are now off by the output latency, where before they lined up.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -782,7 +782,7 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
     struct ActiveAudioRoute {
         juce::String input;
         std::vector<int> channels;
-        int latencySamples = 0;
+        int recordingAdjustmentSamples = 0;
```
lucaromagnoli: This duplicates `deviceGeneration`: the only write (`audioDeviceAboutToStart`) also bumps `inputGeneration_`, so the generation check already catches every change. The test at line 726 ("the changed correction records into a second clip") passes because of the generation bump, not this field. Remove the field, or restart with unchanged latencies, and it still splits.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -2881,7 +2885,9 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
     HardwareInputMap hardwareInputs_;
     std::set<juce::String> unresolvedInputs_;
     std::atomic<bool> hardwareInputsStale_{true};
-    std::atomic<int> inputLatencySamples_{0};
+    /// Automatic take correction owned by the active interface: its input and
```
lucaromagnoli, line 3095: Where this value ends up, the docs still say input latency: `TakeRecorderSettings::latencySamples` ("The input's declared latency", `TakeRecorder.hpp:109`), `LiveAudioInput`'s constructor and `latencySamples()`, and `native-arrangement-audio-recording.md:41` ("after input latency correction"). A later caller that passes only the input latency brings back #2751.

tests/engine/test_engine_host_audio_input_juce.cpp
```diff
@@ -677,22 +698,46 @@ class EngineHostAudioInputTest final : public juce::UnitTest {
         devices.device->pump();
         devices.device->pump();
 
+        devices.device->inputLatencySamples = kInputLatency;
+        devices.device->outputLatencySamples = kOutputLatency;
         devices.device->pumpAfterFutureLatencyReads(2);
-        devices.device->restart(channels({0, 1, 2, 3}), 96000.0);
+        devices.device->restart(channels({0, 1, 2, 3}));
```
lucaromagnoli: Dropping `96000.0` means the render context no longer changes, so `rebuild()` never runs. The rebuild → stopRecording → resumeRecording path during a take, and the "follow-up update cannot split the replacement take" invariant, are no longer covered. `pumpAfterFutureLatencyReads(2)` is dead too (the second read only came from the rebuild's `addAudioCallback`).

Restore the rate change. If it stays gone, `restart`'s `sampleRate` parameter, the `96000.0` entry in `getAvailableSampleRates` and the comment on `pumpAfterFutureLatencyReads` are dead.

tests/engine/test_engine_host_audio_input_juce.cpp
```diff
@@ -471,13 +476,20 @@ class EngineHostAudioInputTest final : public juce::UnitTest {
...
+                             4 * kBlockSize - kRecordingAdjustment,
+                             "device input and output latency are removed from the head");
                 juce::AudioBuffer<float> captured(2, 1);
                 reader->read(&captured, 0, 1, 0, true, true);
                 expectWithinAbsoluteError(captured.getSample(0, 0), levelOf(2), 0.0001f);
                 expectWithinAbsoluteError(captured.getSample(1, 0), levelOf(3), 0.0001f);
             }
+            expectWithinAbsoluteError(clip->placement.startBeat, 0.0, 0.0000001,
+                                      "correction keeps the take on its record-start beat");
+            expectWithinAbsoluteError(clip->placement.lengthBeats,
```
lucaromagnoli, line 501: `(4 * kBlockSize - kRecordingAdjustment) / 48000.0 * 2.0` hardcodes the fixture rate and a 120 BPM default, and is written out again at line 633. A shared samples-to-beats helper using the device rate and project tempo would stop this breaking for reasons unrelated to latency.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 1c69a260. Positive adjustments now enter a deferred post-roll: the recorder drops the corrected head, continues receiving callbacks after the logical stop/punch-out, and writes the same number of tail samples before materialization. Added coverage for arrangement punch-out, Session-slot continuity, takes shorter than the adjustment, and repeated loop passes.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 1c69a260. New takes snapshot the live PlanExecutor latency via EngineSession::latencySamples(); audio correction is interface input + output + plan latency. Added an EngineSession PDC assertion that observes the executor latency changing from 0 to 16 samples.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 1c69a260. Driver values are promoted to int64_t before summing and the final correction is clamped to +/-3 seconds at the active sample rate. The host test now exercises INT_MAX input and output reports and verifies a bounded three-second correction without overflow.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 1c69a260. MIDI takes now snapshot output-interface latency plus plan latency, while audio snapshots input + output + plan latency, keeping simultaneous MIDI/audio capture aligned through their shared playback delay.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 1c69a260. Removed the correction from ActiveAudioRoute (and kept deviceGeneration as the device identity boundary). The adjustment is now computed and snapshotted only when a take starts.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 1c69a260. Updated TakeRecorder/LiveInput terminology and the recording docs to describe an automatic recording adjustment, including interface round trip and plan latency, rather than input latency alone.

tests/engine/test_engine_host_audio_input_juce.cpp
lucaromagnoli: Fixed in 1c69a260. Restored the restart at 96 kHz and the second future-latency read/pump, so the rebuild -> stop/resume path and replacement-take stability remain covered.

tests/engine/test_engine_host_audio_input_juce.cpp
lucaromagnoli: Fixed in 1c69a260. Replaced the duplicated 48 kHz/120 BPM arithmetic with beatsForSamples(host, device, samples), which derives both the current device rate and tempo-map conversion.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -1404,31 +1490,36 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
...
+                                          .channels = *channels,
+                                          .deviceGeneration =
+                                              inputGeneration_.load(std::memory_order_relaxed),
+                                          .sessionScene = {}};
             if (activeAudio != audioRecordingRoutes_.end() && activeAudio->second == wanted)
                 continue;
 
-            if (activeAudio != audioRecordingRoutes_.end())
+            if (activeAudio != audioRecordingRoutes_.end()) {
                 stopAudioTake(track.id);
```
lucaromagnoli, line 1580: **An input change mid-take now ends Arrangement recording.** The same happens after a device restart at the same rate (generation bump, no rebuild). The old code did `stopAudioTake` then `startAudioTake`. Now it stops and `continue`s, so no replacement starts, even when the stop is immediate (zero adjustment, nothing deferred).

With only one armed audio track, `hasActiveAudioTakes()` is then false, so `recording_` goes false. Every later `reconcileRecording()` returns at `if (!recording_)` (line 1399), and the post-harvest reconcile in `handleAsyncUpdate` is gated on `recording_` too. The transport stays in record, nothing is captured, and `isRecording()` is false.

`deviceRestartSplitsRecording` used to cover this path. With `96000.0` restored it goes through `rebuild()` + `resumeRecording()` instead, so nothing tests it now. Suggest: start the replacement when the stop was immediate, count a closing take whose track is still eligible as recording, and add a same-rate restart / input-change case.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -854,10 +891,11 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
                  sessionScene ? engine::RecordTarget::slot : engine::RecordTarget::arrangement,
              .scene = sessionScene.value_or(-1),
              .maxNotes = kRecordingPreviewNotes},
-            [this, sources = std::move(sources), sessionScene,
+            [this, sources = std::move(sources), adjustment, sessionScene,
              trackId = track.id](engine::RecordTap& tap) mutable {
                 engine::MidiTakeRecorderSettings settings;
                 settings.sources = std::move(sources);
+                settings.latencySamples = adjustment;
```
lucaromagnoli, line 910: **MIDI takes are now shortened by this adjustment.** `MidiTakeRecorder` sets `end_ = arrivals_ + (to - from) - latencySamples` (`MidiTakeRecorder.cpp:366`), and `lengthBeats` comes from `end_` (`:514`). There's no MIDI post-roll, so with output + plan latency here, MIDI takes and Session MIDI loops lose that much from their musical length. Notes arriving in that window after stop are dropped. This is the bug the audio post-roll just fixed, moved to MIDI.

No test runs with a non-zero MIDI adjustment. The MIDI host suite's fake device reports zero output latency and the plan has none.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -1151,27 +1225,33 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
 
         const auto found = sessionSlotTargets_.find(trackId);
         if (found != sessionSlotTargets_.end() && found->second.scene == sceneIndex) {
+            auto deferred = false;
             if (found->second.launched) {
-                if (session_ != nullptr) {
+                stopSessionMaterialTakes(trackId, true, true);
```
lucaromagnoli, line 1291: **Stopping a recording slot now behaves differently depending on the adjustment.** When the audio close is deferred, `backToArrangement` is skipped here (and at 1242 and 1292), and after harvest it's only sent if materialization fails (line 1080). So on success the clip keeps playing. With a zero adjustment, or on a MIDI track, the close is immediate and `backToArrangement` is still sent, as before this PR. The new `SessionClipPlayState::Playing` assertion only holds on the deferred path. Pick one behaviour and apply it to both.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -1927,11 +2029,17 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
         // intentional remove/add during rebuild reaches this handler only
         // after the callback has been installed again.
         if (recording_ && !audioRunning_.load(std::memory_order_acquire))
```
lucaromagnoli: A take already in post-roll when the device stops isn't force-closed. `stopRecording` cleared `recording_`, and `hasActiveAudioTakes()` excludes closing takes, so this condition is false and the take waits for a callback that may never come. It finalizes on the next rebuild, project close or quit. If the device restarts at the same rate, the tail is filled from the new device's input. Include `!closingAudioTakes_.empty()` here.

magda/engine/exec/EngineSession.cpp
```diff
@@ -394,13 +444,19 @@ void EngineSession::process(int numSamples, juce::AudioBuffer<float>& output,
...
+                const auto named =
+                    std::binary_search((*render)->takes.begin(), (*render)->takes.end(), entry.key);
+                if (named || entry.take->capturesPostRoll()) {
                     entry.take->capture(segment.block,
                                         segment.countingIn ||
                                             (entry.take->followsArrangement() &&
                                              transport->punch.recordingRequested && !captureWanted),
                                         transport->loop);
+                    if (entry.take->readyToClose())
+                        takeCompletionPending_.store(true, std::memory_order_release);
```
lucaromagnoli, line 457: `readyToClose()` stays true for any take that stopped on its own while still named: a relocation mid-record, a punch-out, or a slot run ending before the host stops it. That sets the flag on every block, and `audioDeviceIOCallbackWithContext` calls `triggerAsyncUpdate()` from every callback until the host closes the take, while `closeCompletedTakes` finds nothing to do. Flag only takes in `deferredTakes_`, or set `readyToClose_` only once a close was requested.

lucaromagnoli (commented):
Re-review of 1c69a260. Fixed: post-roll tail, plan latency, ±3 s bound, doc wording, test helpers and the 96000 restart. The main new issue is the input-change / same-rate-restart path (comment on line 1505), where recording now stops.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in f902a948. Reconciliation now starts the replacement immediately when the old close is synchronous, and keeps the Record gesture alive while an eligible route is finishing deferred post-roll so harvest can start its replacement. Added both same-rate restart and positive-adjustment input-change regressions.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in f902a948. MIDI takes now use the same correction-sized deferred post-roll as audio, preserving the logical length and accepting tail events before materialization. Added unit coverage plus a host test with 960 samples of output latency that records note events from the tail without shortening the clip.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in f902a948. Kept the established stop behavior: an explicit Session recording-slot stop returns to Arrangement for both immediate and deferred closes. Deferred close state now carries the release-on-success intent through materialization; audio and nonzero-adjustment MIDI tests both assert the stopped/Arrangement result.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in f902a948. The device-stop path now force-closes when either audio or MIDI deferred-close maps are non-empty, even after recording_ was cleared. Added a test that enters audio post-roll, stops the device before another callback, and verifies immediate materialization.

magda/engine/exec/EngineSession.cpp
lucaromagnoli: Fixed in f902a948. readyToClose() is now true only when the host requested closure and the recorder has stopped; self-stopped takes no longer raise completion on every callback. Added an assertion covering self-stop without a close request.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -1113,7 +1222,7 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
             juce::ignoreUnused(unused);
             const auto target = sessionSlotTargets_.find(trackId);
             const auto launched = target != sessionSlotTargets_.end() && target->second.launched;
-            const auto handedOver = stopAudioTake(trackId, createClips);
+            const auto handedOver = stopAudioTake(trackId, createClips, allowPostRoll, launched);
```
lucaromagnoli: **What happens to a launched slot still depends on the adjustment here** (and on line 1216 for MIDI). `launched` used to go into `releaseSessionOnFailure`. After the rename it goes into `releaseSessionOnSuccess`, which flips its meaning on the deferred path. With a positive adjustment, harvest sends `backToArrangement` even when the take is handed over, so the slot stops. With zero adjustment the close is immediate: `if (launched && !handedOver)` releases the slot only on failure, and the materialized clip keeps playing, as it did before this PR.

This happens whenever `stopRecording()` runs while the transport keeps playing: Record off (`stopMidiRecording`), punch-out, `setTempo`, `setTimeSignature` and `setLoop`. Passing `false` here and at 1216 matches the immediate path. Failure is already handled by the harvest's `!handedOver` check.

magda/daw/engine/host/EngineHost.cpp
```diff
@@ -1151,27 +1262,33 @@ struct EngineHost::Impl final : private juce::AudioIODeviceCallback,
...
+                stopSessionMaterialTakes(trackId, true, true);
+                deferred = hasClosingTake(trackId);
+                if (session_ != nullptr && !deferred) {
                     engine::LaunchRequestQueue::Gesture gesture(session_->launchRequests());
                     gesture.backToArrangement({trackId, found->second.scene});
                 }
-                stopSessionMaterialTakes(trackId);
             }
-            sessionSlotTargets_[trackId] = {.scene = sceneIndex};
+            if (!deferred)
```
lucaromagnoli: **Arming another slot on the same track is dropped while the old take is closing.** While slot A is recording, click record on slot B of the same track. `stopSessionMaterialTakes` defers A's close, `deferred` is true, and B's target is never set. With zero adjustment the close is immediate and B arms. After A's post-roll, harvest removes A's target and releases A, so the user has to click B again.

Once the old take is harvested, arm the new scene instead of dropping it: for example, keep the pending scene on the closing entry and apply it during harvest. A test that switches slots mid-recording with a positive adjustment would cover it.

lucaromagnoli (commented):
Re-review of f902a948. All five comments from the last round are fixed. Two things remain, both on paths where closing the take is deferred.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 5d5b0a6e. stopRecording() now passes false for release-on-success on both MIDI and audio, matching the synchronous path: successful Session materialization keeps playing, while failure still returns to Arrangement. Added non-zero-adjustment MIDI and audio Record-off tests that assert successful handover remains in Session.

magda/daw/engine/host/EngineHost.cpp
lucaromagnoli: Fixed in 5d5b0a6e. A deferred close now carries an optional pending Session scene. Switching A -> B stores B on the closing take and installs it as the armed target when A is harvested; subsequent clicks can replace or clear that pending scene. Added a positive-adjustment test that records A, switches to B during post-roll, and verifies A materializes while B remains armed.

Issue #2758, opened 2026-09-20
Plugin state capture moves to the plugin service


lucaromagnoli:
Part of #2554. `captureAllPluginStates`, `capturePluginStateAt` and `applyPluginStateAt` sit on `AudioEngine`, but what they serve is a project save and a project load.

## Scope

- The three move to the service that owns hosted plugin state, which each engine feeds from its own instances (the native engine through `EngineHost`, the fork through its `AudioBridge`).
- The project save hook asks that service instead of asking whichever engine renders.

## Note

The chunk is authoritative over a project's parameter array (#2243, #2244), and both engines already write it through `ExternalPluginState`. This is about who is asked, not about what is written.

## Acceptance

- The three leave `AudioEngine`.
- A project with a hosted plugin saves and reloads its chunk under both engines, which the existing state-restore suites cover.

lucaromagnoli:
Landed in #2763.

Issue #2756, opened 2026-09-20
Plugin discovery leaves AudioEngine


lucaromagnoli:
Part of #2554. The first and largest group of what `MagdaAudioEngine` forwards to the fork: roughly 18 of its 30 delegations are a plugin scan, and a scan is not an engine question.

## Scope

- A plugin service owns the format manager, the known plugin list and the scan: `startPluginScan`, `abortPluginScan`, `isPluginScanRunning`, the scan status and completion callbacks, `detectNewPlugins`, `scanPluginParameters`, `getPluginScanReportFile`, `getKnownPluginTypes`, `getKnownPluginList`, `getPreferredPluginTypes`, `getSystemPluginSearchPaths`, the plugin list change listeners, and the exclusion pair.
- Those methods leave `AudioEngine`. The app asks the service.
- `EngineHost::setPluginServices` takes its formats and list from the service rather than from the fork.

## The list itself

Tracktion's `Engine` owns the `KnownPluginList` the scan fills, and its own plugin hosting reads it. Until the fork goes in #2557 the service wraps that list rather than replacing it, and owns it outright afterwards.

## Acceptance

- No caller reaches a plugin scan through `AudioEngine`, and the scan methods are off the interface.
- Scanning, the scan dialog, the exclusion list and the plugin browser behave as they do now under both engines.

lucaromagnoli:
Landed in #2762.

PR #2763, opened 2026-09-20
Move plugin state capture to PluginService

lucaromagnoli:
## Summary
- move live plugin state capture and application off AudioEngine and onto PluginService
- register Tracktion AudioBridge and the native EngineHost as state providers
- route project saves, copy/remove flows, pads, and presets through the service
- add focused PluginService state-routing coverage

## Testing
- `cmake --build cmake-build-debug --target magda_tests magda_juce_tests -j 8`
- `cmake-build-debug/tests/magda_tests "[plugin][state-service]"`
- `cmake-build-debug/tests/magda_tests "[project][serialization]"`
- `cmake-build-debug/tests/magda_juce_tests_artefacts/Debug/magda_juce_tests "External Plugin State Restore"`

Closes #2758

Stacked on #2762.


magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -2102,10 +2103,9 @@ bool TrackManager::applyDevicePreset(const ChainNodePath& devicePath,
     live->gainValue = std::pow(10.0f, presetDevice.gainDb / 20.0f);
     live->pluginState = stripPresetRuntimePluginState(presetDevice.pluginState);
 
-    // The rendering engine, not the bridge: only the live instance has the
-    // state (#2573). It may rewrite live->parameters.
-    if (audioEngine_ != nullptr)
-        audioEngine_->applyPluginStateAt(devicePath);
+    // The plugin service reaches whichever engine renders the live instance. It may
+    // rewrite live->parameters while applying the chunk (#2573, #2758).
+    PluginService::getInstance().applyPluginStateAt(devicePath);
```
lucaromagnoli: **Test isolation: this now reaches the shared engine when TrackManager has none.**

Dropping the `audioEngine_ != nullptr` guard defeats the isolation `resetJuceProjectState()` provides. `tests/JuceTestStateGuard.hpp:40` nulls TrackManager's engine, but nothing clears `PluginService::stateProviders_` — the shared wrapper registers itself at `TracktionEngineWrapperInit.cpp:496` on the first `getSharedEngine()` and is a function-local static that never unregisters during the run.

So `tests/devices/test_rack_audio.cpp:61` (model-only, no engine, same `magda_tests` binary) calls `applyDevicePreset` → `TracktionEngineWrapper::applyPluginStateAt` → `audioBridge_->getPlugin(path)`. AudioBridge syncs off the TrackManager singleton, so it can find a live 4osc and call `restorePluginStateFromValueTree`, rewriting `live->parameters`. Passes when run alone; changes behaviour once any `*_juce.cpp` test has run first.

magda/daw/core/TrackCommands.cpp
```diff
@@ -322,10 +323,8 @@ void DuplicateTrackCommand::execute() {
 
     // Capture current plugin state so the duplicate gets the source's live settings.
     // Skipped when we're stripping the FX chain anyway — nothing to carry over.
-    if (duplicateDevices_) {
-        if (auto* engine = trackManager.getAudioEngine())
-            engine->captureAllPluginStates();
-    }
+    if (duplicateDevices_)
+        PluginService::getInstance().captureAllPluginStates();
```
lucaromagnoli, line 326: **Same isolation hole on the capture side, and this one destroys data.**

`PluginManager::captureAllPluginStates()` unconditionally overwrites `devInfo->pluginState` for every synced device — `PluginManager.cpp:454` even comments *"Always overwrite pluginState (even if empty) to avoid stale state"*.

With the engine check gone, a model-level test running `DuplicateTrackCommand` after any `*_juce.cpp` test drives the shared AudioBridge and can blank or replace the authored `pluginState` it just set on the TrackManager singleton. Same exposure at `TrackCommands.cpp:883`, `PadCommands.cpp:23` and `DevicePresetMenu.cpp:192`.

magda/daw/engine/MagdaAudioEngine.cpp
```diff
@@ -341,31 +345,22 @@ const AudioBridge* MagdaAudioEngine::getAudioBridge() const {
...
 }
 
 void MagdaAudioEngine::capturePluginStateAt(const ChainNodePath& devicePath) {
-    tracktion_->capturePluginStateAt(devicePath);
-
     if (host_ != nullptr)
         host_->captureExternalPluginStateAt(devicePath);
 }
 
 void MagdaAudioEngine::applyPluginStateAt(const ChainNodePath& devicePath) {
```
lucaromagnoli, line 357: **On the native engine this only reaches external plugins, so internal devices silently keep their old patch.**

`host_->applyExternalPluginStateAt` (`EngineHost.cpp:2961`) iterates `keysOfDeviceAt` and skips anything `!factory_.isExternalKey(key)`.

Load a preset onto an internal device (Poly Synth, a Faust FX) while the MAGDA engine renders: `live->pluginState` updates in the model and the UI refreshes, but the rendering device keeps its old patch until something else republishes.

The sibling write paths `updateDeviceAuthoredState` / `setDeviceAuthoredState` (`TrackManagerDevices.cpp:2034`, `:2076`) go through `projectAuthoredStateToEngine`, which grew the `audioEngine->renderedDevice(devicePath)` fallback for exactly this in #2663. `applyDevicePreset` doesn't. That also makes the new comment at `TrackManagerDevices.cpp:2106` — "the plugin service reaches whichever engine renders the live instance" — false for internal devices.

lucaromagnoli (commented):
Review of the plugin-state move. Twelve inline comments below, plus two that land in files this PR doesn't touch:

---

**`tests/JuceTestStateGuard.hpp:40` — the guard's reset list is now incomplete.**

`resetJuceProjectState()` clears SelectionManager, UndoManager, AutomationManager, ClipManager, TrackManager and the bridges, then `setAudioEngine(nullptr)`. After this PR that list misses the thing that now decides where plugin state goes: the shared engine stays registered as PluginService's provider for every subsequent test in the binary. Either add `PluginService::getInstance().forgetStateProvider(*engine)` (restored on the next `getSharedEngine()`), or keep the call sites asking whichever engine TrackManager points at. This is the root of the two inline findings in `TrackManagerDevices.cpp` and `TrackCommands.cpp`.

**`magda/daw/engine/host/EngineHost.cpp:2942` — "Internal devices are handled by the fork" is now definitively false.**

This PR removes the `tracktion_->capturePluginStateAt(devicePath)` leg from `MagdaAudioEngine::capturePluginStateAt`, so under the native engine nothing downstream of `captureExternalPluginStateAt` handles internal devices at all. The comment documents an invariant a reader will rely on when deciding whether the internal path is covered, and it points at a fork that has no AudioBridge (`MagdaAudioEngine.hpp:76`).

---

The three I'd hold the merge for: the two test-isolation holes (ordering-dependent, pass when run alone), and the static-destruction-order call in `TracktionEngineWrapperInit.cpp:581`. The native-engine internal-device gap in `applyPluginStateAt` is a real user-facing hole but may be better as a follow-up against the epic.

magda/daw/engine/MagdaAudioEngine.cpp
```diff
@@ -341,31 +345,22 @@ const AudioBridge* MagdaAudioEngine::getAudioBridge() const {
 }
 
 /**
- * @brief Both engines, in that order (#2581).
+ * @brief Read the external plugins this engine renders through EngineHost (#2758).
```
lucaromagnoli: **"Internal devices write their authored state through the model already" isn't true for Drum Grid pads** — and this PR's own `PadCommands.cpp` hunk says so:

> A pad plugin's patch reaches `DeviceInfo` only when the Drum Grid's state is captured (`PluginManager::captureDrumGridPads`), and the ordinary sync does not do it.

That capture lives only in the fork's `PluginManager` (`PluginManager.cpp:454`, `:495`) and is never reached under the native engine, where `capturePluginStateAt` goes straight to `host_->captureExternalPluginStateAt` (external keys only).

Repro: edit a pad's sampler after adding it, remove the pad, undo — with the MAGDA engine selected the sampler comes back as it was *added*, not as it sounded. The behaviour is pre-existing (the removed `tracktion_->` leg was already a no-op with no AudioBridge), but the comment now asserts the opposite and hides it. Worth rewording to say what actually holds.

magda/daw/engine/TracktionEngineWrapperInit.cpp
```diff
@@ -576,6 +576,10 @@ bool TracktionEngineWrapper::initialize() {
 void TracktionEngineWrapper::shutdown() {
     DBG("TracktionEngineWrapper::shutdown - starting...");
 
+    // Stop the service reaching the AudioBridge before that bridge is torn down. A
+    // services-only wrapper never registered, and cannot displace another engine here.
+    PluginService::getInstance().forgetStateProvider(*this);
```
lucaromagnoli, line 580: **Static destruction order: this can touch `PluginService` after it has been destroyed.**

In `magda_tests`, `tests/SharedTestEngine.hpp:13` holds `static TracktionEngineWrapper engine;`. Its *construction* completes before `engine.initialize()` runs, and PluginService's Meyers singleton is first constructed **inside** that `initialize()` (via `initializePluginFormats`).

Function-local statics are destroyed in reverse order of construction completion, so at process exit `~PluginService` runs first, then `~TracktionEngineWrapper` → `shutdown()` → `PluginService::getInstance().forgetStateProvider(...)` touches `stateProviders_` on a destroyed object — and then `forgetEngineList()` joins a destroyed `std::thread` member.

#2756 introduced the `forgetEngineList` call; this PR adds a second one and moves it to the first line of `shutdown()`, so the window is wider now.

magda/daw/engine/PluginService.hpp
```diff
@@ -197,6 +230,7 @@ class PluginService {
     std::function<void(bool, int, const juce::StringArray&)> onScanComplete_;
     std::function<void(const juce::String&)> onScanStatus_;
     std::function<std::vector<ScannedPluginParameter>(const juce::String&)> internalScanner_;
+    std::vector<PluginStateProvider*> stateProviders_;
```
lucaromagnoli: **Raw non-owning pointers with no lifetime coupling.**

`PluginStateProvider` has no self-deregistration in its destructor and `PluginService` validates nothing, so any provider that skips its explicit unregister leaves a dangling entry the service will happily dispatch to — a test helper, an engine whose `shutdown()` isn't reached because `initialize()` threw, or one destroyed while another sits on top (so the stale entry isn't even `back()`). The next project save then calls `stateProviders_.back()->captureAllPluginStates()` through freed memory.

Giving `PluginStateProvider` a destructor that calls `forgetStateProvider(*this)` closes this for every provider at once.

magda/daw/engine/PluginService.hpp
```diff
@@ -148,6 +160,27 @@ class PluginService {
...
     void clearList();
 
+    /**
+     * @brief Make @p provider the source of live hosted-plugin state.
+     *
+     * Providers are retained as a stack so a short-lived engine in a test can leave the
+     * application's provider as it found it. A services-only Tracktion wrapper never
+     * registers: it has no AudioBridge and therefore no live instances to offer.
+     */
+    void useStateProvider(PluginStateProvider& provider);
```
lucaromagnoli: **The provider *stack* looks like speculative generality — and it's what produces the stale-fallback leakage above.**

The doc says it exists "so a short-lived engine in a test can leave the application's provider as it found it", but the one test that needs that already implements save/restore itself (`ScopedStateProvider`, `test_plugin_service_state.cpp:31`), and the app never has two providers — the services-only fork never registers.

A single `PluginStateProvider* stateProvider_` (set/clear, with the test's RAII guard restoring the old value) does the same job, and makes an engine going away mean *the service goes quiet* rather than silently falling back to an older engine.

magda/daw/engine/PluginService.cpp
```diff
@@ -160,6 +160,30 @@ void PluginService::forgetEngineList() {
...
+void PluginService::useStateProvider(PluginStateProvider& provider) {
+    std::erase(stateProviders_, &provider);
+    stateProviders_.push_back(&provider);
+}
+
+void PluginService::forgetStateProvider(const PluginStateProvider& provider) {
+    std::erase(stateProviders_, &provider);
+}
+
+void PluginService::captureAllPluginStates() {
```
lucaromagnoli, line 185: Three byte-identical forwarder bodies repeat the `if (!stateProviders_.empty()) stateProviders_.back()->…` shape across `captureAllPluginStates`, `capturePluginStateAt` and `applyPluginStateAt` (`:172-184`).

A private `PluginStateProvider* currentProvider() const { return stateProviders_.empty() ? nullptr : stateProviders_.back(); }` reduces each body to one line and leaves one place to change when provider selection grows a validity check or logging.

tests/devices/test_plugin_service_state.cpp
```diff
@@ -0,0 +1,83 @@
...
+}  // namespace
+
+TEST_CASE("PluginService sends state operations to the current renderer",
+          "[plugin][state-service]") {
+    auto& service = magda::PluginService::getInstance();
+    RecordingStateProvider first;
+    ScopedStateProvider firstRegistration(first);
+
+    const auto firstPath = magda::ChainNodePath::topLevelDevice(2, 7);
+    service.captureAllPluginStates();
```
lucaromagnoli, line 58: The test covers the stack-ordering path only; two behaviours the new code deliberately implements go unasserted:

- the empty-stack no-op — nothing checks that `captureAllPluginStates()` / `capturePluginStateAt()` / `applyPluginStateAt()` are silent with no provider registered (the `if (!stateProviders_.empty())` guards);
- the dedupe — `useStateProvider` on an already-registered provider must not duplicate it. Delete the `std::erase` at `PluginService.cpp:164` and every existing `CHECK` still passes, because one `forgetStateProvider` leaves the stack in a shape the assertions accept.

magda/daw/core/PadCommands.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "../audio/AudioBridge.hpp"
 #include "../engine/AudioEngine.hpp"
```
lucaromagnoli: This include is dead now — the hunk removed the last `AudioEngine` reference in the file (`grep -n AudioEngine` returns only this line; same in `TrackCommands.cpp:10`).

`AudioEngine.hpp` pulls in JUCE + tracktion types, so leaving it keeps both TUs recompiling on every engine-interface change for nothing.

magda/daw/engine/MagdaAudioEngine.hpp
```diff
@@ -144,6 +146,7 @@ class MagdaAudioEngine final : public AudioEngine,
     const DeviceMeters& deviceMeters() const override {
         return deviceMeters_;
     }
+    /** @brief Feed PluginService from the EngineHost instances this engine renders. */
```
lucaromagnoli: This `@brief` duplicates — and contradicts — the one on the definition at `MagdaAudioEngine.cpp:348`:

- here: "Feed PluginService from the EngineHost instances this engine renders"
- there: "Read the external plugins this engine renders through EngineHost (#2758)"

Doxygen picks one arbitrarily and both now have to be kept in sync. Same duplication was added at `TracktionEngineWrapper.hpp:306` against `TracktionEngineWrapperPlugins.cpp:164`. Keeping the block on the definition only would be enough.

magda/daw/core/TrackCommands.cpp
```diff
@@ -881,9 +880,7 @@ void CreateTrackWithDeviceCommand::undo() {
 namespace {
 
 void capturePluginStateAt(const ChainNodePath& devicePath) {
```
lucaromagnoli: With the `getAudioEngine()` null check gone, this helper (and `capturePadPluginStates`, and `PadCommands.cpp:22`) is a single forwarding line over a one-line service call — a reader following `capturePluginStateAt(devicePath_)` from `:995` hops through a local function that adds nothing.

Inlining `PluginService::getInstance().capturePluginStateAt(...)` at the three call sites removes the hop; `PadCommands.cpp`'s explanatory comment moves with it.

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed in 93e700fa. The state provider now follows `TrackManager::setAudioEngine()`, so `setAudioEngine(nullptr)` unregisters the shared wrapper and leaves model-only operations inert. Added a regression that exercises capture/apply before and after releasing the active renderer.

magda/daw/core/TrackCommands.cpp
lucaromagnoli: Fixed in 93e700fa by tying provider registration to `TrackManager`'s active renderer instead of engine initialization. Clearing the engine now disables all three service operations, with regression coverage for capture-all, capture-one, and apply-one.

magda/daw/engine/MagdaAudioEngine.cpp
lucaromagnoli: Fixed in 93e700fa. `applyDevicePreset` now sends internal authored state through `projectAuthoredStateToEngine` and reserves `PluginService` for hosted plugin chunks. Added a regression using the native rendered-device seam to verify the authored settings reach the live device.

magda/daw/engine/MagdaAudioEngine.cpp
lucaromagnoli: Fixed the misleading claim in 93e700fa. The comment now states that Tracktion captures pad child instances and explicitly records that the native child-instance capture path is a pre-existing limitation; the EngineHost comment is similarly scoped to externally hosted state.

magda/daw/engine/TracktionEngineWrapperInit.cpp
lucaromagnoli: Fixed in 93e700fa. `PluginService` is now a process-lifetime singleton, so engine/static destructors can safely call both `forgetStateProvider` and `forgetEngineList` regardless of construction order. The Tracktion lifecycle suite exits cleanly with this ordering.

magda/daw/engine/PluginService.hpp
lucaromagnoli: Fixed in 93e700fa. `PluginStateProvider` now self-unregisters in its destructor, explicit shutdown still unregisters early, and the process-lifetime service makes destructor-time deregistration safe. Added coverage for destroying a current provider without explicit shutdown.

magda/daw/engine/PluginService.hpp
lucaromagnoli: Fixed in 93e700fa. The vector stack is gone; the service holds one current provider and never falls back implicitly. `useStateProvider` returns the replaced provider solely so a scoped test can restore it explicitly.

magda/daw/core/PadCommands.cpp
lucaromagnoli: Fixed in 93e700fa: removed the dead `AudioEngine.hpp` includes from both `PadCommands.cpp` and `TrackCommands.cpp`.

magda/daw/engine/PluginService.cpp
lucaromagnoli: Superseded in 93e700fa by removing provider selection entirely: there is now one `stateProvider_` pointer rather than a stack/back lookup. The three forwarding guards are direct and no longer duplicate selection policy.

magda/daw/engine/MagdaAudioEngine.hpp
lucaromagnoli: Fixed in 93e700fa: removed the duplicate declaration-side Doxygen blocks and kept the implementation-side documentation.

tests/devices/test_plugin_service_state.cpp
lucaromagnoli: Updated in 93e700fa. Tests now cover the no-provider no-op, replacement without implicit fallback, provider self-unregistration, and TrackManager releasing the active renderer. The dedupe behavior disappeared with the provider stack.

magda/daw/core/TrackCommands.cpp
lucaromagnoli: Fixed in 93e700fa: removed the forwarding helpers and inlined the service call at the pad and path-command call sites.

magda/daw/core/TrackManagerDevices.cpp
```diff
@@ -2102,10 +2103,15 @@ bool TrackManager::applyDevicePreset(const ChainNodePath& devicePath,
...
     live->pluginState = stripPresetRuntimePluginState(presetDevice.pluginState);
 
-    // The rendering engine, not the bridge: only the live instance has the
-    // state (#2573). It may rewrite live->parameters.
-    if (audioEngine_ != nullptr)
-        audioEngine_->applyPluginStateAt(devicePath);
+    if (live->format == PluginFormat::Internal) {
+        // Internal authored state is projected onto either the bridge plugin or the
+        // native rendered device. It is not an externally hosted plugin chunk.
+        projectAuthoredStateToEngine(audioEngine_, devicePath, live->pluginState, live->pluginId);
```
lucaromagnoli: **Behaviour change worth a deliberate decision: this resets on an undecodable state where the old path left the plugin alone.**

`projectAuthoredStateToEngine` (`:1987`) does:

```cpp
auto tree = ta::devicePluginTreeFromState(docText);
if (!tree.isValid()) {
    tree = juce::ValueTree(tracktion::engine::IDs::PLUGIN);
    tree.setProperty(tracktion::engine::IDs::type, deviceType, nullptr);
}
plugin->restorePluginStateFromValueTree(tree);
```

The TE path this replaces returned without touching the plugin instead (`TracktionEngineWrapperPlugins.cpp:188`):

```cpp
if (const auto savedState = ta::devicePluginTreeFromState(live->pluginState); savedState.isValid())
    plugin->restorePluginStateFromValueTree(savedState);
return;
```

`devicePluginTreeFromState` returns an invalid tree when the state is empty or won't decode, so an internal-device preset with empty `pluginState` now resets the live plugin to defaults. The parameter re-push below covers `live->parameters`, so the exposure is limited to authored properties that aren't parameters.

The bare-tree reset is right for `updateDeviceAuthoredState`/`setDeviceAuthoredState` (#2663), which is what it was written for. For a preset it's a different question — and the external branch three lines down guards exactly this shape ("A preset with no chunk must not repopulate", `TracktionEngineWrapperPlugins.cpp:197`). Resetting is arguably the correct preset semantics; it's just currently silent. Worth either a line of comment saying it's intended, or the `isValid()` check.

tests/devices/test_plugin_service_state.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (previous_ != nullptr)
+            service.useStateProvider(*previous_);
+    }
+
+    ScopedStateProvider(const ScopedStateProvider&) = delete;
+    ScopedStateProvider& operator=(const ScopedStateProvider&) = delete;
+
+  private:
+    RecordingStateProvider& provider_;
+    magda::PluginStateProvider* previous_ = nullptr;
```
lucaromagnoli: `previous_` is a raw pointer that the destructor re-registers:

```cpp
if (previous_ != nullptr)
    service.useStateProvider(*previous_);
```

If the provider it points at is destroyed before this guard goes out of scope, that revives a dangling pointer into the service — and it defeats the new `~PluginStateProvider` self-clear, which has already run by then.

Not reachable as the file is written (declaration order puts the guards after what they capture), but it's a sharp edge in a helper that exists to make provider swapping safe. Holding the previous provider only while it's still current, or asserting `previous_ == nullptr` in the cases that expect it, would close it.

magda/daw/engine/PluginService.cpp
```diff
@@ -160,6 +167,30 @@ void PluginService::forgetEngineList() {
...
+    return std::exchange(stateProvider_, &provider);
+}
+
+void PluginService::forgetStateProvider(const PluginStateProvider& provider) {
+    if (stateProvider_ == &provider)
+        stateProvider_ = nullptr;
+}
+
+void PluginService::captureAllPluginStates() {
+    if (stateProvider_ != nullptr)
```
lucaromagnoli: Minor, and much less pressing than in the stack version: the three forwarders still repeat the same null-check-then-dispatch shape (`:179-192`). At two lines each that's near the floor, so this is fine to leave — noting it only because a `currentProvider()` helper would still be the single place to hang a validity check or logging if provider selection ever grows one.

lucaromagnoli (commented):
Follow-up pass over `93e700f`. All fourteen findings from the previous review are addressed — build is green and the full suite is 4443 passed / 13 skipped / 2 expected-fail locally.

Two of the fixes are better than what I suggested:

- Moving registration into `TrackManager::setAudioEngine` and dropping the engines' self-registration fixes both test-isolation findings at the source, rather than patching `JuceTestStateGuard`'s reset list. App teardown order holds up: `TrackManager::shutdown()` runs at step [4] in `magda_daw_main.cpp:599` and the engine is destroyed at [6], so the new `dynamic_cast` on `audioEngine_` is always on a live object.
- The internal-device gap now has a real branch plus a test that asserts projection onto the native rendered device.

On the intentional-leak fix for static destruction order: I checked the `ShutdownDetector` / `MessageManager` leak assertions at exit by reverting just the `getInstance()` hunk and re-running the full suite — byte-identical output, so they're pre-existing and not from the leak. `discoveryThread_` is still joined via `forgetEngineList()` on the app path.

Three residual notes below, none blocking.

magda/daw/core/TrackManagerDevices.cpp
lucaromagnoli: Fixed in f1899984. Internal presets now pass `resetWhenUndecodable=false`, so an empty or undecodable authored snapshot leaves the live rendered device untouched, matching the external-plugin chunk path and the previous behavior. I added a regression test covering an empty internal preset against a non-default rendered device.

tests/devices/test_plugin_service_state.cpp
lucaromagnoli: Fixed in f1899984. I removed the saved raw `previous_` pointer and restoration behavior from `ScopedStateProvider`; the helper now asserts that the service starts clean and only unregisters its own provider. Provider replacement/no-fallback remains covered by the separate test.

magda/daw/engine/PluginService.cpp
lucaromagnoli: Fixed in f1899984. Added a private `currentProvider()` selection point and routed all three state-operation forwarders through it.

Issue #71, opened 2026-01-17
Missing media dialog


lucaromagnoli:
Relocate, search, or create placeholder for missing files.

PR #2767, opened 2026-09-20
Add missing media recovery dialog

lucaromagnoli:
## Summary

- detect missing clip, take, sampler, and drum-pad media after project restoration
- offer background folder search, sequential manual relocation, or offline placeholders
- conservatively reject ambiguous search matches and persist successful relinks as dirty project state
- keep detection and directory traversal off the message thread, with cancellable progress and cross-platform stored-path handling

## Testing

- cmake --build cmake-build-debug --target magda_tests magda_daw_app -j 8
- cmake-build-debug/tests/magda_tests [missing-media] — 45 assertions in 3 test cases
- cmake-build-debug/tests/magda_tests [project] — 1,428 assertions in 90 test cases
- full magda_tests — 4,472 cases / 1,780,253 assertions (13 skipped, 4 expected assertion failures)

Closes #71


magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+// message thread, and return only the pure path matches; applying them remains
+// a message-thread ProjectManager operation.
+class MissingMediaSearchWindow : public juce::ThreadWithProgressWindow {
+  public:
+    using Completion =
+        std::function<void(std::vector<ProjectManager::MissingMediaReplacement>, bool)>;
+
+    MissingMediaSearchWindow(std::vector<ProjectManager::MissingMediaFile> missing,
+                             juce::File directory, Completion completion)
+        : ThreadWithProgressWindow(trEllipsis("missing_media.searching"), true, false),
```
lucaromagnoli: **The search progress window has no cancel button, so a large folder locks the app.**

`ThreadWithProgressWindow(title, hasProgressBar, hasCancelButton)` — this passes `false` for `hasCancelButton`. JUCE then builds the alert with `numButtons = 0` and calls `alertWindow->setEscapeKeyCancels(false)` (`juce_ThreadWithProgressWindow.cpp:52-57`), so the modal has no button and ignores Escape. `threadShouldExit()` can therefore never become true from the UI, which makes the `shouldStop` plumbing in `run()`, the `cancelled_` flag, `userPressedCancel`, and `missing_media.search_cancelled` all dead.

Pick a big sample drive (or the home folder) and `RangedDirectoryIterator` recurses the whole tree with the window modal and unkillable until it finishes. `CollectFilesProgressWindow` two classes up passes `true` here.

```suggestion
        : ThreadWithProgressWindow(trEllipsis("missing_media.searching"), true, true),
```

magda/daw/project/ProjectManager.cpp
```diff
@@ -920,6 +947,141 @@ juce::File ProjectManager::getImportedDirectory() const {
...
+
+    const bool changed = relinkMediaPaths([&paths](const juce::String& path) {
+        const auto found = paths.find(path);
+        return found == paths.end() ? juce::String{} : found->second;
+    });
+    if (!changed)
+        return 0;
+
+    markDirty();
+    return static_cast<int>(paths.size());
```
lucaromagnoli: **`relinkMissingMediaFiles` reports the number of paths *attempted*, not repaired.**

`relinkMediaPaths` returns a single bool for "anything changed at all", and on `true` this returns `paths.size()` — every path that passed the pre-filter, whether or not it was actually repointed. A path can be in `paths` and still not get fixed: in `relinkMediaPaths` an audio event with `sourceId == INVALID_SOURCE_ID` sets `touched = true` (so the function returns true) but `continue`s past `pool.relink`, leaving the path missing.

With 5 matches where 2 fail that way, `completeMissingMediaSearch` prints `missing_media.search_result` as "Reconnected 5 file(s). 2 file(s) are still missing" for a project that only had 5. Count the real repairs instead — e.g. snapshot `getMissingMediaFiles()` before and after and diff the path sets.

magda/daw/project/ProjectManager.cpp
```diff
@@ -920,6 +947,141 @@ juce::File ProjectManager::getImportedDirectory() const {
...
+        int bestCount = 0;
+        juce::File best;
+        for (const auto& candidate : candidates) {
+            const int score = commonPathSuffixLength(file.path, candidate.getFullPathName());
+            if (score > bestScore) {
+                bestScore = score;
+                bestCount = 1;
+                best = candidate;
+            } else if (score == bestScore) {
+                ++bestCount;
```
lucaromagnoli, line 1082: **`best` can stay default-constructed while `bestCount == 1`, emitting a proposal with an empty `juce::File`.**

`bestScore` starts at 1 and `best` starts empty. A candidate scoring exactly 1 (filename-only) takes the `else if` branch, which bumps `bestCount` but never assigns `best`. If that is the only candidate and nothing scores higher, `bestCount == 1` passes and line 1036 pushes `{file.path, juce::File{}}`.

Reachable whenever two missing paths share a filename (so the `candidates.size() == 1 && missingCountByName == 1` fast path is skipped) and only one of them has a matching trailing directory. E.g. missing `/old/drums/kick.wav` + `/old/perc/kick.wav`, tree contains only `/new/drums/kick.wav`: the second gets an empty-File proposal that survives into the returned `matches`. Today `relinkMissingMediaFiles` filters it out via `existsAsFile()`, so it is latent — but any caller that trusts `matches` (a preview list, a confirm dialog) gets an invalid file, and the comment on line 1022 says filename-only matches are supposed to be rejected, which this branch defeats.

Track the best candidate explicitly, e.g. start `bestScore = 0` and only accept at the end when `bestScore > 1 && bestCount == 1`.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+                juce::AlertWindow::showAsync(
+                    juce::MessageBoxOptions{}
+                        .withIconType(juce::MessageBoxIconType::WarningIcon)
+                        .withTitle(tr("missing_media.title"))
+                        .withMessage(tr("missing_media.relink_failed"))
+                        .withButton(tr("dialogs.ok"))
+                        .withAssociatedComponent(safeThis.getComponent()),
+                    [retrySafe, missing = std::move(missing), index, projectFile](int) mutable {
+                        if (retrySafe != nullptr)
+                            retrySafe->locateMissingMediaFiles(std::move(missing), index,
```
lucaromagnoli: **The failure branch retries the same index, which loops with no way forward when the failure is permanent.**

`relinkMissingMediaFile` returns false whenever `missingPath` is no longer in `getMissingMediaFiles()` — not only when the chosen file is bad. If the path stopped being missing since `missing` was captured (the drive remounted, or an earlier relink merged the pooled source), *every* file the user picks returns false, so `missing_media.relink_failed` → same index → chooser → false → forever. The only exit is cancelling the chooser, which abandons the whole remaining sequence.

Re-check whether the path is still missing before blaming the user's choice, and give the sequence a per-file Skip so one unfindable file doesn't force an all-or-nothing abort.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+        juce::FileBrowserComponent::openMode | juce::FileBrowserComponent::canSelectFiles,
+        [safeThis, missing = std::move(missing), index,
+         projectFile](const juce::FileChooser& chooser) mutable {
+            if (safeThis == nullptr)
+                return;
+            const auto replacement = chooser.getResult();
+            safeThis->fileChooser_.reset();
+            if (!isStillCurrentProject(projectFile))
+                return;
+            if (!replacement.existsAsFile()) {
```
lucaromagnoli, line 338: **Cancelling the chooser mid-sequence abandons the flow with no summary, after the project has already been marked dirty.**

A user with 20 missing files who relinks 12 and then cancels on the 13th falls into this branch: it jumps straight to `offerFourOscConversion()` with no `finishMissingMediaRecovery` alert. They are never told that 12 were reconnected and 8 are still offline, and never told to save — yet each successful `relinkMissingMediaFile` called `markDirty()`, so the project silently differs from its file and the next close pops an unsaved-changes prompt out of nowhere.

Route a cancel through `finishMissingMediaRecovery` with the `all_relinked` / `still_missing` summary, as the index-exhausted path at line 283 does.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+  private:
+    std::vector<ProjectManager::MissingMediaFile> missing_;
+    juce::File directory_;
+    Completion completion_;
+    std::vector<ProjectManager::MissingMediaReplacement> matches_;
+    bool cancelled_ = false;
+};
+
+bool isStillCurrentProject(const juce::File& projectFile) {
+    const auto& projects = ProjectManager::getInstance();
```
lucaromagnoli: **`isStillCurrentProject` can't tell an imported or unsaved project from a different one.**

The identity check is `getCurrentProjectFile() == projectFile`. `importDawProjectAsync` sets `currentFile_ = juce::File()` while `isProjectOpen_ = true` (ProjectManager.cpp:618-620), and a never-saved project is the same. So after importing a .dawproject with missing media, `projectFile` is `{}` and this returns true for *any* subsequently opened unsaved project — the async continuations (folder chooser, search completion, per-file locate) keep running against a project that is no longer the one that was scanned. Same for the empty `projectFile.getParentDirectory()` handed to both choosers as the initial directory.

`MainWindow` already has the right token for this: `projectOpenGeneration_`, which `projectOpened`/`projectClosed` bump. Thread that through the async steps instead of the file path, and drop the two staleness mechanisms down to one.

magda/daw/project/ProjectManager.cpp
```diff
@@ -920,6 +947,141 @@ juce::File ProjectManager::getImportedDirectory() const {
     return mediaDirectory_.getChildFile(kImportedDir);
 }
 
+std::vector<ProjectManager::MissingMediaFile> ProjectManager::getMissingMediaFiles() const {
+    std::vector<MissingMediaFile> missing;
+    std::map<juce::String, size_t> indexByPath;
+
+    const auto add = [&missing, &indexByPath](const juce::String& path) {
+        if (path.isEmpty() || juce::File(path).existsAsFile())
```
lucaromagnoli: **Every media path is `stat`ed on the message thread, and the whole scan runs up to three times per relink.**

`existsAsFile()` is a blocking syscall. For a project whose media sits on a disconnected SMB/NFS share each call can block for seconds, and this runs for every clip event, every take and every sampler reference — on the message thread, right after project open. That is the exact cost the PR moved the *directory search* off the message thread to avoid.

It also compounds: `relinkMissingMediaFiles` calls `getMissingMediaFiles()` (full model walk + stats + a full `getSamplerMediaReferences()` plugin sweep of the Edit), then `relinkMediaPaths` sweeps the plugins again, and `completeMissingMediaSearch` / `locateMissingMediaFiles` call `getMissingMediaFiles()` once more afterwards. Because `relinkMissingMediaFile` (singular) delegates to the plural form, the sequential locate flow pays all of that per file — O(N) full scans for N missing files.

Hoist the `currentlyMissing` set (or the whole scan) out of the per-file path, and consider caching the result for the duration of one recovery flow.

magda/daw/project/ProjectManager.cpp
```diff
@@ -920,6 +947,141 @@ juce::File ProjectManager::getImportedDirectory() const {
...
+
+std::vector<ProjectManager::MissingMediaReplacement> ProjectManager::searchForMissingMedia(
+    const std::vector<MissingMediaFile>& missing, const juce::File& directory,
+    const std::function<bool()>& shouldStop) {
+    if (!directory.isDirectory() || missing.empty())
+        return {};
+
+    std::map<juce::String, std::vector<juce::File>> candidatesByName;
+    for (const auto& entry :
+         juce::RangedDirectoryIterator(directory, true, "*", juce::File::findFiles)) {
```
lucaromagnoli, line 1045: **The traversal buffers every file in the tree before it looks at what is actually missing.**

`candidatesByName` gets a `juce::File` for *every* file under `directory`, then only the handful of names in `missing` are ever read back. Point it at a 500k-file sample library (or a home folder) and that is a half-million-entry `std::map<juce::String, std::vector<juce::File>>` built to answer a few dozen lookups — and, with no cancel button on the progress window, nothing can stop it.

`missingCountByName` doesn't depend on the traversal. Build the set of wanted lowercase filenames first and skip any entry not in it:

```cpp
std::map<juce::String, int> missingCountByName;
for (const auto& file : missing)
    ++missingCountByName[storedFileName(file.path).toLowerCase()];

std::map<juce::String, std::vector<juce::File>> candidatesByName;
for (const auto& entry : juce::RangedDirectoryIterator(directory, true, "*", juce::File::findFiles)) {
    if (shouldStop && shouldStop())
        return {};
    const auto file = entry.getFile();
    const auto name = file.getFileName().toLowerCase();
    if (missingCountByName.count(name) != 0)
        candidatesByName[name].push_back(file);
}
```

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+                safeThis->locateMissingMediaFiles(std::move(missing), 0, projectFile);
+            } else {
+                daw::ui::offerFourOscConversion();
+            }
+        });
+}
+
+void MainWindow::chooseMissingMediaSearchFolder(
+    std::vector<ProjectManager::MissingMediaFile> missing, juce::File projectFile) {
+    if (fileChooser_ != nullptr || !isStillCurrentProject(projectFile))
```
lucaromagnoli: **A busy `fileChooser_` silently swallows the whole recovery flow *and* the 4OSC prompt.**

This `return` (and the matching one at line 293) is the only exit from the missing-media flow that shows the user nothing and never reaches `daw::ui::offerFourOscConversion()`. Every other exit either re-offers the conversion prompt or is a deliberate project-changed bail-out. Since this PR moved `offerFourOscConversion()` out of `MainWindow::projectOpened` and made it the tail of this state machine, a chooser left open by an in-flight export (MainWindowExport.cpp keeps `fileChooser_` alive across the render) now costs the user both dialogs with no feedback.

At minimum call `daw::ui::offerFourOscConversion()` on these branches so the pre-existing prompt is not lost.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+        std::function<void(std::vector<ProjectManager::MissingMediaReplacement>, bool)>;
+
+    MissingMediaSearchWindow(std::vector<ProjectManager::MissingMediaFile> missing,
+                             juce::File directory, Completion completion)
+        : ThreadWithProgressWindow(trEllipsis("missing_media.searching"), true, false),
+          missing_(std::move(missing)),
+          directory_(std::move(directory)),
+          completion_(std::move(completion)) {}
+
+    void run() override {
```
lucaromagnoli, line 117: **The progress window shows a bar frozen at 0% and an empty message for the whole search.**

`hasProgressBar` is `true`, but `run()` never calls `setProgress()` or `setStatusMessage()`. `ThreadWithProgressWindow` initialises `progress (0.0)` and paints whatever `message` holds, which is empty, so the window is a title and a 0% bar that never moves — indistinguishable from a hang on exactly the long scans this class exists for.

`CollectFilesProgressWindow` sets a status message in its constructor and drives `setProgress` from its callback. A directory walk has no known total, so either `setProgress(-1.0)` (JUCE renders that as indeterminate) or `setStatusMessage(entry.getFile().getFullPathName())` per directory would do.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+    std::vector<ProjectManager::MissingMediaReplacement> matches_;
+    bool cancelled_ = false;
+};
+
+bool isStillCurrentProject(const juce::File& projectFile) {
+    const auto& projects = ProjectManager::getInstance();
+    return projects.hasOpenProject() && projects.getCurrentProjectFile() == projectFile;
+}
+
+juce::String missingPathFileName(juce::String path) {
```
lucaromagnoli: **`missingPathFileName` is a verbatim copy of `storedFileName` in ProjectManager.cpp.**

Both are `path.replaceCharacter('\\', '/')` then `fromLastOccurrenceOf("/", false, false)`, added in this same PR, in two anonymous namespaces. The two must agree — the UI titles the chooser with one and `searchForMissingMedia` keys its candidate map with the other — so a future fix to the Windows-separator handling has to land in both or the locate dialog stops naming what the search is looking for.

Put the one implementation on `ProjectManager` (next to `MissingMediaFile`, which is where its input comes from) and call it from both.

lang/en.json
```diff
@@ -355,6 +355,27 @@
     "alert.cancelled": "Collect was cancelled. Some files may not have been copied.",
     "alert.no_edit": "Open or create a project first."
   },
+  "missing_media": {
+    "title": "Missing Media",
+    "intro": "This project references {0} missing {1} from {2} {3}. Search a folder automatically, locate each file yourself, or keep the offline placeholders.",
```
lucaromagnoli: **The pluralisation scheme here is inconsistent with the three strings right below it, and doesn't survive translation.**

`intro` assembles a sentence from four substitutions — `{0}` a count, `{1}` a `file`/`files` word, `{2}` a count, `{3}` a `reference`/`references` word — which needs six extra keys and a `missing.size() == 1 ? ... : ...` in the caller. Yet `search_result`, `all_relinked` and `still_missing` three lines down use the repo's existing convention, `file(s)` (`edit.render_clip`, `status.new_plugins`, `status.plugins_failed`).

Sentence assembly from translated fragments also breaks outside English: languages with more than two plural forms (Polish, Russian, Arabic) can't express this with two words, and languages that inflect the noun after a numeral get the wrong form. Use `file(s)` / `reference(s)` here too and delete `missing_media.file`, `.files`, `.reference`, `.references`.

magda/daw/ui/windows/MainWindow.cpp
```diff
@@ -519,14 +518,23 @@ void MainWindow::updateWindowTitle() {
 
 void MainWindow::projectOpened(const ProjectInfo&) {
     updateWindowTitle();
-    daw::ui::offerFourOscConversion();
+    const auto generation = ++projectOpenGeneration_;
+    const auto safeThis = juce::Component::SafePointer<MainWindow>(this);
+    // ProjectManager notifies listeners before its onAfterLoad hook. Defer the
```
lucaromagnoli: **This comment states a reason that isn't true, which will mislead the next person who touches the ordering.**

`onAfterLoad` is set in exactly one place (TracktionEngineWrapperInit.cpp:427) and it restores the active view mode, the project bindings, and zoom/scroll — no sampler or drum-pad state. Sampler sample files come back via `ProjectSerializer::commitStaged`, before `notifyProjectOpened()`.

The second half is right and is the reason that matters: `notifyProjectOpened()` runs before `onComplete`, which is what calls `hideLoadingMessage()`, so the alert would otherwise open behind the loading overlay. Say that and drop the sampler claim.

magda/daw/ui/windows/MainWindowMenus.cpp
```diff
@@ -90,8 +91,270 @@ class CollectFilesProgressWindow : public juce::ThreadWithProgressWindow {
...
+                      .replace("{0}", juce::String(static_cast<int>(missing.size())))
+                : tr("missing_media.still_missing")
+                      .replace("{0}", juce::String(static_cast<int>(remaining.size()))));
+        return;
+    }
+    if (fileChooser_ != nullptr)
+        return;
+
+    const auto& item = missing[index];
+    auto initialDirectory = juce::File(item.path).getParentDirectory();
```
lucaromagnoli: **`juce::File(item.path)` trips JUCE's assertion on a project authored on the other platform.**

`storedFileName` and `pathComponents` both normalise `\\` to `/`, so this PR clearly expects stored paths to cross platforms. But the raw path is handed straight to `juce::File` here and in `getMissingMediaFiles`'s `add` (ProjectManager.cpp:955). On macOS/Linux `File::parseAbsolutePath` runs `jassert(! p.containsChar('\\'))` and then `jassertfalse` for a path that doesn't start with `/`, so opening a Windows-saved project in a debug build fires two assertions per stored path — and the resulting File is `cwd/C:\Samples\kick.wav`, so the initial directory for the chooser is garbage rather than falling through to the `projectFile.getParentDirectory()` branch below.

Normalise once (the `pathComponents` helper already does it) before constructing any `juce::File` from a stored path.

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Fixed in 4dce5f79: matching starts at score zero and only accepts a unique candidate with a score above the filename-only level. Added a regression test for the empty File case.

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Fixed in 4dce5f79: relinking now tracks and returns the distinct old paths actually changed instead of the number of proposed mappings.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: the search window now exposes Cancel, uses indeterminate progress, and shows the selected directory as status.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: cancelling a manual chooser now reports both the repaired and remaining counts before continuing to the 4OSC prompt.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: a failed relink now advances automatically when the old path is no longer referenced, otherwise the dialog offers Retry and Skip.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: every asynchronous recovery step now carries projectOpenGeneration_ rather than using the project file as identity.

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Fixed in 4dce5f79: model references are snapshotted on the message thread, filesystem existence checks run in a worker, and the cached missing set is reused through recovery. Relinking no longer rescans missing files.

magda/daw/project/ProjectManager.cpp
lucaromagnoli: Fixed in 4dce5f79: wanted filenames are built first and directory traversal only retains entries whose names are actually missing.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: the search bar is indeterminate and the selected directory is displayed as the status message.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: both busy chooser branches now continue to the pre-existing 4OSC prompt instead of silently ending the flow.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: filename extraction now has one shared ProjectManager implementation used by both search and UI.

lang/en.json
lucaromagnoli: Fixed in 4dce5f79: the intro follows the existing file(s) and reference(s) convention and the fragment keys/caller plural branching were removed.

magda/daw/ui/windows/MainWindow.cpp
lucaromagnoli: Fixed in 4dce5f79: the comment now describes only the real ordering constraint, which is waiting for the loading overlay completion.

magda/daw/ui/windows/MainWindowMenus.cpp
lucaromagnoli: Fixed in 4dce5f79: stored paths are normalized and classified by host syntax before any juce::File construction. Foreign-platform paths remain missing and the chooser falls back safely. Added a regression test.

